PR #2, opened 2017-02-25
ChatUserList passing rank # not object, but pros aren't > 1000 any more

roy7:
It's probably better for ChatUserList to pass a ranking object, so the professional value can be tested for future compatibility. But I couldn't figure out what to change to pass the object instead of the rank. This work around also works.

### DIFF
--- a/src/lib/misc.ts
+++ b/src/lib/misc.ts
@@ -115,14 +115,6 @@ export function rankString(r) { /* {{{ */
     }
     if (r > 900) {
         return interpolate(pgettext("Pro", "%sp"), [(((r - 1000) - 36))]);
-    } else {
-        // ChatUserList.tsx isn't showing professionals correctly, shows 8D instead of 1P.
-        // Probably need to send an opject to rankString() but I can't figure that out. So,
-        // instead we could also test for > 36 when it isn't > 900.
-        //
-        if (r > 36) {
-            return interpolate(pgettext("Pro", "%sp"), [(r - 36)]);
-        }
     }
     if (r < -900) {
         return "?";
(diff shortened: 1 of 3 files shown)


anoek:
Unfortunately it's not quite that simple, normal players can actually have a rank of 36 (or higher!) without being professionals, the only legitimate fix for the missing 'p' on the games and whatnot is to get the `professional` flag to where it needs to be. 

(though you're probably right in that I don't think the >1000 hack is needed anymore, that's carry over from a few years ago. Everything *should* be using the `professional` flag now.)

src/lib/misc.ts
```diff
@@ -115,6 +115,14 @@ export function rankString(r) { /* {{{ */
     }
     if (r > 900) {
         return interpolate(pgettext("Pro", "%sp"), [(((r - 1000) - 36))]);
+    } else {
+        // ChatUserList.tsx isn't showing professionals correctly, shows 8D instead of 1P.
+        // Probably need to send an opject to rankString() but I can't figure that out. So,
+        // instead we could also test for > 36 when it isn't > 900.
+        //
+        if (r > 36) {
+            return translate_1.interpolate(translate_1.pgettext("Pro", "%sp"), [(r - 36)]);
```
anoek: It's just `interpolate` and `pgettext`,  `translate_1` looks like the javascript module that typescript fills in for us at compilation time.

roy7:
The rankString function is checking professional flag correctly, but apparently it is being called with an object on places like the User Profile page, but it is being passed a rank # instead of an object from the CharUserList. I assume it's where the map() is done and the field "rank" but I couldn't figure out how to pass the user objects instead of the ranking. So I came up with the above instead. :(

roy7:
Changed it around to cache professional field from messages. Once professional field is added to messages (and really it should be, like ranking, since it's needed to show proper ranking?) it might start working in the main chat room as-is. If not, further testing will take place. :)

anoek:
Working good!

PR #3, opened 2017-02-25
misc(build): add an npm script for convenience

bobylito:
This works because npm scripts have node_modules/.bin in their
environment as bin.


anoek:
Much nicer, thanks!

Issue #5, opened 2017-02-25
Byo-yomi extra periods presentation is confusing


lemurek:
### Summary

*Byo-yomi presentation is confusing*

Consider a game with time settings of 2 mins base time and 1 extra byo yomi period of 30s, how much time left in game does this player have?
0:05+1x30s

It can mean that either:

1. There are 5 more seconds of base time and extra byo-yomi period
2. Player has entered overtime and his last byo yomi ends in 5 seconds

Even both situations are presented in the same way, breaching that 5 seconds limit in the second option means losing a game on time. 

### Environment

 * Operating system: any
 * Browser: any
 * Browser Version: any

### Steps To Reproduce:

1. Start a game with time settings 2 mins base time + 1 x 30s byo-yomi
2. Observe time presentation in base-time and in over-time

### Additional Information:

![screen shot 2017-02-25 at 12 17 40](https://cloud.githubusercontent.com/assets/12175058/23330980/79e9ad16-fb54-11e6-96bd-750e01bb8689.png)

anoek:
Changes to the clock display are a bit of a sensitive topic amongst players, I'd like some additional weigh in from folks about this proposed change before merging any code.

matburt:
I'm pretty sure, for better or worse, this is how the code has worked since the Nova days. We've had lots of folks comment on it over the years... I'm not sure we can change it without major backlash at this point.

mark5000:
I do think OGS has a troublesome clock. In the current implementation, using the above image as an example, it's impossible to know if the user is on their first overtime period or on the last seconds of main time. There's absolutely no difference. Let's consider other servers:

![tygem timing](https://cloud.githubusercontent.com/assets/13010650/23332660/38fe2bbe-fb32-11e6-9206-29c96f2a1690.png)

On tygem, the main time converts into a visual countdown. There is also a voice cue for "overtime countdown has begun" and "the last overtime" period. It is fundamentally clear what the situation is at all times.

![foxgo timing](https://cloud.githubusercontent.com/assets/13010650/23332887/2f80509a-fb36-11e6-9f4c-d6b9069c68af.png)

On Fox Go, the main time is displayed in minute:second format while overtime is displayed as seconds only (and in bigger numbers). There is also a voice cue for "byo yomi now starts" and "last byo yomi."

![kgs timing](https://cloud.githubusercontent.com/assets/13010650/23332668/56b6226a-fb32-11e6-88e7-d0ef90da7e0d.png)

[...]

PR #6, opened 2017-02-25
Present one less byo period when in overtime

lemurek:
Fix suggestion for https://github.com/online-go/online-go.com/issues/5


anoek:
I think #5 needs to evolve into an update of the interface to make things less confusing, not a reduction in the number of byo-yomi periods displayed

lemurek:
Fair point. This was a rather lazy proposition, but I'm glad the discussion has been started ;)

Issue #7, opened 2017-02-25
Indicate beginning of overtime with an audible cue other than Countdown.


lemurek:
### Summary

Countdown at the end of base time is unnecessary, as there is nothing real penalty for breaching the threshold there. I.e. if you slip by 1 second, you just lose one second, vs. slip in byo = 1 byo period gone. 

Sometimes I find myself playing a rushed move in the end of base time for no reason other than the misleading extra countdown.

### Environment

 * Operating system: any
 * Browser: any
 * Browser Version: any

### Steps To Reproduce:

1. Start a blitz game under byo-yomi or canadian time control
2. Observe the countdown close to the end of base time

### Additional Information:

Suggested fix:
For base-time only time controls all the cues should be present. For an overtime based time settings (byo-yomi and canadian) the cues should happen only close to the end of overtime period(s).

anoek:
I think instead of just disabling this we want to expand on the optionality around voice count down. IE, folks should be able to turn this on or off, change how much of a countdown there is and when, if they want three beeps for the last three seconds or full voice, that sort of thing

anoek:
I'm going to close this for now, I'd like any changes to how the clocks behave to be discussed in a more public setting, such as the forums or something of the sort. Once there's been some discussion and input from a number of people and it's clear that the community agrees that X should be done about clocks, then we can open up an issue and alter code.

lemurek:
So let me explain my way of thinking here. Normally the countdown is a clear signal that you need to hurry, and indeed in other time settings the urgency is welcome.

In all 3 of:
Base time, Simple and Fischer settings you lose a game, so the move is definietly urgent. 

On the opposite when there is an overtime available, making the move at this point in time is actually disadvantegous. 

In Canadian setup is better to wait 1 extra second in result you gain some time as you need to make only 24 moves instead of 25.
In Byo-Yomi it would be actually best to spend whole extra minute on thinking about the move, which is effectively lost if you make the decision, just before the base time is over.

So that's why I was arguing that this countdown doesn't bring value to players. 

It is indeed related to https://github.com/online-go/online-go.com/issues/5. Instead of a countdown there should be a clear hint that the player is entering the overtime at this point, which somehow relates to the concern I raised in the other issue. 

Of course the decision is yours, but I do think that there are valid, objective reasons to remove the countdown after base time limit.

lemurek:
[...]

PR #8, opened 2017-02-25
Display byo cues only in overtime or if no overtime present

lemurek:
Fix proposition for: https://github.com/online-go/online-go.com/issues/7


lemurek:
Tested the logic under all 5 time control modes in manual playouts.

anoek:
As per comment on #7 about wanting to have any changes to the clocks vetted by the community in a more public place, i'm going to close this for now. Feel free to re-open if/when that discussion happens and something is decided upon.

PR #15, opened 2017-02-25
Countdown overflow

lemurek:
Fix proposition for https://github.com/online-go/online-go.com/issues/11

Manually tested on byo-yomi and Fischer settings.

Issue #11, opened 2017-02-25
Blitz 10s time controls result in saying "10" each move


roy7:
### Summary

When playing blitz with 10 second periods, after you make your move, the voice says "10" when resetting the clock. If possible it shouldn't say your time left after you've moved.

### Environment

 * Operating system: Windows 10
 * Browser: Chrome
 * Browser Version:

### Steps To Reproduce:

Start a blitz game, no main time, 5-10 second periods, play and make your move.

lemurek:
I've been working on a fix for the issue, and managed to clean-up the code significantly - to refresh the countdown only on opponents turn. However to my surprise the problem still persisted. I decided to print out the clock updates I'm seeing, and there is some awkwardness in the engine.playerToMove() behaviour. 

![screen shot 2017-02-25 at 19 35 12](https://cloud.githubusercontent.com/assets/12175058/23334180/90c34d02-fb91-11e6-93a0-a8abb0a4ef9d.png)

The print out may be obscure, but let me try to explain. Current player is 500, and opponent is 501, byo-yomi was set to 6 seconds.

1. **sound_to_play: 5**  
says 'Five'
2. **sound_to_play: 4**  
says 'Four', end of turn
3. **sound_to_play: 6**  
silent because it's opponents turn (resets last sound played)
4. **sound_to_play: 6**  
silent nothing because it's opponents turn (notice last sound played reset)
5. **sound_to_play: 6**  
This is the weird part, where for 1 clock update it is 'my' turn again 
6. **sound_to_play: 6**  
silent

I could hack it around, but it seems that fixing it on a countdown level is not the correct solution. Is there a good reason why engine 'thinks' it's my turn for a split second after the move was made?

roy7:
Maybe look at the behavior of active_game vs move events. Once you move, you'll get 2 events back (not always in same order). A move telling you a move was made, and an active_game telling you new game state and whose turn it is and so on. There is also a clock event any time clock info changes for a game, that is also separate from the other two events.

Perhaps when you make the move it resets the clock, so timer changes and new sound needs to go off, before active_game tells you it is your turn. In that case we might need a client side toggle of some sort, set it when you make a move and check that for sound regardless of active_game, then turn sounds back on next time active_game says it is your turn? Or clear the "i made a move" flag as soon as the next active_game arrives?

lemurek:
If it's a correct behaviour I can easily suppress that single event from being said out loud. My concern was more around if the set of events I'm seeing is actually intended.

[...]

PR #18, opened 2017-02-25
Pronounce countdown only when counter decreases

lemurek:
Simple fix for https://github.com/online-go/online-go.com/issues/11

Compared to https://github.com/online-go/online-go.com/pull/15 this approach does not try to proactively detect turn changes. This one simply pronounces countdown only when the observed counter decreases. 

As a result, fix is much simpler, but in case of byo-yomi timer it doesn't pronounce the maximum counter.

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -125,7 +125,7 @@ export abstract class Goban extends EventEmitter {
     private last_phase;
     private last_review_message;
     private last_sent_move;
-    private last_sound_played;
+    private last_countdown_registered;
     private last_sound_played_for_a_stone_placement;
     private last_stone_sound;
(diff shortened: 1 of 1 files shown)


anoek:
Seems like if you're playing a blitz game and have an increment that's less than 10s, then this wouldn't work so well no?

lemurek:
I tested the solution with byo-yomi period of 6 and it worked as expected. This simplistic solution just sacrifices the top counter, so as expected the countdown was 5,4,3,2,1. If we want to retain the top index, then https://github.com/online-go/online-go.com/pull/15 is a better option.

anoek:
Fixed with https://github.com/online-go/online-go.com/commit/4e2ea6e8df07a271304e9c1f9dc9cdc28264840f

PR #20, opened 2017-02-25
Added supervisor setup to Makefile to support developers without supervisor preinstalled

msayson:


msayson:
Currently, developers who don't have the "supervisor" node module installed will receive an error message when running "make" to set up their development environment.

This change adds "supervisor" to the list of npm modules to install, allowing new developers to successfully use the Makefile for set up.

anoek:
Huh, I wonder why https://github.com/online-go/online-go.com/commit/9653b8df8b4f4ab97f1fc216b28c93c11f84ae6f didn't solve this. Happy to merge this, though we may want to switch the `Makefile` from the env setting `node_modules/.bin` calling setup they are now to npm scripts similar to the changes in https://github.com/online-go/online-go.com/commit/3e3c4d69db33d1323181e8ae989117f912b587a2 , which might obviate this patch

PR #24, opened 2017-02-26
Fixed channel "leave" button issue + added little test to make pseudo auto-completion more robust.

Alderi-Tokori:


Alderi-Tokori:
The "leave" is now visible even when there's no new message in the channel. 
A drawback of this quick fix is the "leave" button is also available on channels you haven't joined, but clicking it doesn't cause any unintended behavior.

Alderi-Tokori:
I've modified the code and now the "leave" button only appears for open channels.

Issue #29, opened 2017-02-26
Remove Tracking


hloeffler:
### Summary

*Remove Tracking*
https://github.com/online-go/online-go.com/blob/0b599cc055faa4871f91e61c9a7cd669e8d27770/src/index.html#L223-L270

### Environment

 * Operating system:  All
 * Browser: All
 * Browser Version: All

### Steps To Reproduce:

1. Go to https://online-go.com
2. get tracked
3. See that this is wrong 

### Additional Information:

i know you use ads, but there must be a better way

anoek:
Unfortunately we're contractually obligated to have this

PR #30, opened 2017-02-26
Remove Nielsen Online SiteCensus

hloeffler:
just see diff

this helps: https://github.com/online-go/online-go.com/issues/29

but is only 1 provider


anoek:
As per comment, we can't remove these

hloeffler:
this is sad,
@anoek can you please add a comment inside the source, that explains that?

PR #32, opened 2017-02-26
Ensure graph is drawn on Play page

gemedet:
This PR fixes a minor issue where the SeekGraph isn't displayed due to its container div sometimes having no size when first mounted.

### DIFF
--- a/src/views/Play/Play.tsx
+++ b/src/views/Play/Play.tsx
@@ -73,6 +73,9 @@ export class Play extends React.Component<PlayProperties, any> {
         if (w !== this.seekgraph.width || h !== this.seekgraph.height) {
             this.seekgraph.resize(w, h);
         }
+        if (w === 0 || h === 0) { // Wait for positive size
+            setTimeout(this.resize, 500);
+        }
     }}}
     updateChallenges = (challenges) => {{{
         let live = [];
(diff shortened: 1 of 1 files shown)


anoek:
I had to do a similar thing on the game page for the boards: https://github.com/online-go/online-go.com/blob/devel/src/views/Game/Game.tsx#L383-L392

I found that just a timeout wasn't enough, I had to keep checking - this was more of a problem when the tab was loaded in the background on recent versions of firefox (and i think i saw it on chrome too).

Do we need the setInterval here too, or is this setTimeout enough for all browsers and when the tab is loaded in the background?

gemedet:
I think the setTimeout here might be enough, since it calls its own function, so it effectively acts like a setInterval.

I tested the issue using background tabs on the latest versions of Firefox, Chrome, & Safari, and it seemed to work fine, though my testing wasn't exhaustive by any means. :)

src/views/Play/Play.tsx
```diff
@@ -73,6 +73,9 @@ export class Play extends React.Component<PlayProperties, any> {
         if (w !== this.seekgraph.width || h !== this.seekgraph.height) {
             this.seekgraph.resize(w, h);
         }
+        if (w === 0 || h === 0) { // Wait for positive size
```
gemedet: @anoek Is this condition strict enough? I could change it to `if (Math.min(w, h) <= 0)`, similar to [your check](https://github.com/online-go/online-go.com/blob/067eadacc12844d60733908a49e3eb5e3deceabc/src/views/Game/Game.tsx#L387).

anoek: Honestly I don't *know* for sure if there's ever been a case where I saw `< 0`, but I've been bitten enough times over the years by doing things like checking for zero when what I really want is to ensure the value is positive that I just check for `<= 0` or `> 0`  by default now.

PR #37, opened 2017-02-27
goban size fix for webkit2-based browsers such as midori, ps4 browser

blamarche:
Simple style change causes the goban offsetHeight to report correctly and fixes the issue displayed below. Tested in Chrome and Firefox, goban still works as expected after the change.

sshot


anoek:
Cool, thanks for the fix!

blamarche:
No problem, happy to help!

PR #40, opened 2017-02-27
flex grow fix for goban applied to puzzle styl as well

blamarche:
Didn't realize this style would be duplicated somewhere else. I did a search and didn't find anywhere else this fix was necessary so I think this should do it.


anoek:
bleh, yeah probably shouldn't have duplicate styles floating around like that..

PR #70, opened 2017-03-03
Make autoplay delay configurable

mlopezviedma:
Adds an **Autoplay delay (in seconds)** option on Settings.


anoek:
Nice :) Can we change this to allow for fractional seconds as well? Maybe add `step=0.1` for the number field, and then just swap out those `parseInt` calls for `parseFloat`, that aughta do it?

anoek:
Also probably bound that to min of say 0.1, so we don't have folks setting 0 or negative values and then reporting bugs because autoplay "just jumps to the end" or something :)

mlopezviedma:
That sounds a lot better :) On it.

anoek:
Awesome :)

PR #75, opened 2017-03-03
Improve left navbar styling and delete whitespace

ofekih:
From first image:
2fa1d2f2f5acf7435f836861c32c4f29cd31f085

To second image:
66974c94ffb3b81b1b4fe27b2384ea9f79bee130

Tested in Chrome and Firefox

### DIFF
--- a/src/components/NavBar/NavBar.styl
+++ b/src/components/NavBar/NavBar.styl
@@ -116,7 +116,7 @@ navbar-width= 17rem;
     .sign-in {
         padding-right: 1.0rem;
     }
-        
+
 
     .nav-menu-modal-backdrop {
         display: none;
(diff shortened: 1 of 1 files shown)


src/components/NavBar/NavBar.styl
```diff
@@ -231,6 +231,11 @@ navbar-width= 17rem;
             box-shadow: 2px 0px 2px 0px rgba(0, 0, 0, 0.16);
         }
 
+        ul#items {
+            margin-left: 0.75rem;
+            width: initial;
```
anoek: Does this only override the the `width: 100%` on line 205? If so we should just comment that out. Although, I think this has the side effect of limiting the width of the horizontal rule, which I'm not sure if I like better or not. If we want to keep the horizontal rule at full width, we may want to swap the `margin-left:` with a `padding-left` and omit the width limit.

ofekih: I'll double check and see if that works

anoek:
Feel free to re-open when you're ready.

Issue #63, opened 2017-03-02
Puzzle next page button overlap


RobertT53:
### Summary

The bottom tsumego choice on the puzzle page overlaps the next page button.

### Environment

 * Operating system: 
 * Browser: Windows 10
 * Browser Version: any

### Steps To Reproduce:

1. Go to Puzzles
2. Hover mouse over next page button
3. See that the upper half of the button highlights the bottom tsumego

### Additional Information:

[Clicking above the red line goes to the tsumego](http://i.imgur.com/fkzarN2.png)

PR #77, opened 2017-03-04
fixes margin of page-controls

1forh:
Adds a bit of margin to the top of `.page-controls`.

Fixes issue #63.


1forh:
Before: 

before

---

After: 

after

anoek:
Awesome, thanks!

PR #79, opened 2017-03-04
Open user's rating history graph in a new tab

mlopezviedma:
A **Full View** link is added to a user's profile page below
the rating graph,

screenshot-2017-03-04-08h59m18s

which opens a new tab with a fullscreen view of it.

screenshot-2017-03-04-09h02m16s


anoek:
Nice :)

PR #80, opened 2017-03-04
Fix profanity filter bugs

BarneyStratford:
When a user types into a private chat, the recipient receives the chat in their chat window, and also gets a notification from the browser. If the profanity filter is switched on, then it is applied to the contents of the chat window but not to the notification. This commit applies the filter to the notification as well.
On inspection, there was a bug in the English profanity filter itself. Under certain circumstances it would censor the shortest matching profanity when a longer one also matched. The regex has been altered to correct this.
There were a number of words that are extremely offensive in British English that weren't covered by the regex, so I have added them in.


anoek:
:+1:  Thanks for the PR! Good work.

PR #83, opened 2017-03-05
Use more colors on seek graph

mlopezviedma:
Open challenges are shown as follows:

- Green: Ranked 19x19
- Magenta: Ranked 13x13
- Cyan: Ranked 9x9
- Orange: Unranked
- Grey: Not eligible (as always)

### DIFF
--- a/src/components/SeekGraph/SeekGraph.tsx
+++ b/src/components/SeekGraph/SeekGraph.tsx
@@ -399,10 +399,22 @@ export class SeekGraph extends EventEmitter {
             if (C.live_game) {
                 ctx.fillStyle = "#4140FF";
                 ctx.strokeStyle = "#58E0FF";
-                ctx.fillRect(sx, sy, d, d);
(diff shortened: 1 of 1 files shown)


randomra:
Are those colors randomly chosen? If yes I would use more intuitive colors e.g.:
green - ranked 19x19 (as it was already)
yellow - ranked 13x13
orange/red - ranked 9x9
blue - unranked
grey - not eligible

mlopezviedma:
I played along with colors for a while. For your specific suggestion, there are two problems:

1) I tried yellow and it's barely visible on light theme.
2) Blue was already used for live ongoing games. Although they're not visible right now, they are still inside the code, so I assumed that those may be visible again in the future.

anoek:
Nice :) I suspect the selection of what colors are optimal are going to be highly subjective. PR's welcome, though I imagine in the long term we just need to move various color settings out into preferences, and allow for customization.

PR #92, opened 2017-03-05
Add search commands to chat

crodgers:
/sensei
/google
/lmgtfy

All commands will generate a search link.

All commands respond to an optional flag
-user 
and preface the link with the targeted user's username.

All commands use the body of the command to search the specified site.

e.g.

/google -user crodgers let me google that for you
/google let me google that for you


anoek:
Nice :)

I feel like we're going to need tab completion on these commands.. and maybe /help someday... but for a future PR :)

Issue #64, opened 2017-03-02
Active games list view incorrectly displaying opponent


RobertT53:
### Summary

Active game list view displays whoever is playing black instead of the opponents name.

### Environment

 * Operating system: Windows 10
 * Browser: Any 

### Steps To Reproduce:

1. Go to a users page with enough active games to sort by list.
2. See their name is listed under the "Opponent" column if they are playing black.

### Additional Information:

[Example from forums](https://ogs-forums.s3.amazonaws.com/original/2X/0/0da923b1f094b8713c591ab9f4983c9367904f73.png)

BarneyStratford:
I have a counterexample to the claim that it lists only the black players. https://online-go.com/game/7466601 is in the Active Games list at https://online-go.com/player/179886/, with her playing white and her opponent listed as herself. The game is the one at move 205 with no clock. I am her opponent, so I'll arrange that we don't make any more moves.

BarneyStratford:
I know why this is happening now. In order to work out who is the player's opponent, data.get ("user") is compared with Black and White's users. The problem is that data.get ("user") is the currently logged-in user, who is a third party to the games. Hence, the calculation of who is the opponent goes awry. Working on a fix.

anoek:
Fixed with https://github.com/online-go/online-go.com/pull/102

PR #93, opened 2017-03-05
Fix for bug #64.

BarneyStratford:
The problem was that the machine decides which player to make as the opponent by comparing the relevant player to black and to white, displaying whichever player is not the relevant player. Unfortunately, the "relevant player" was being set as the currently logged-in user rather than the player whose profile was being viewed. This caused the display of the wrong opponent in about half of cases.
To fix this, GameLists and GobanLineSummaries no longer have an opponentStyle property, and instead take an optional player property. If the player is undefined or not in the game, then the GameList will list Black and White players. If the player is in the game, then it will display only an Opponent player, which is the player that is not given in the player property.

### DIFF
--- a/src/views/Game/Chat.styl
+++ b/src/views/Game/Chat.styl
@@ -79,9 +79,7 @@ game-chat-user-list-width=10rem;
 
     .chat-line {
         width: 100%;
-        word-wrap: break-word;
-        word-break: break-all;
-        white-space: normal;
+        overflow-wrap: break-word;
     }
 
     .timestamp {


anoek:
Good find!

I did notice a couple of regressions with this patch though:

1) The 'Clock' field no longer works on the home screen
2) The 'Watch' tab game lists seem to be broken with this patch

BarneyStratford:
> I did notice a couple of regressions with this patch though:
Oops. These regressions are fixed in my latest commit.

Cheers,
barney.

anoek:
The clock seems to be fixed, though the watch list still appears to be broken :(

PR #94, opened 2017-03-06
Edit ellipsis on accept

cassidoo:
This is minor but it was bugging me, ha.  This changes the text when you accept a game to say `Accepting...` instead of `Accepting..`.


anoek:
:+1:

PR #97, opened 2017-03-06
One color Go implementation

mlopezviedma:
When activated in Settings, it replaces the Black stone theme with the White stone theme on the entire site. So it works well as an "honour based" one color Go variant.

### DIFF
--- a/src/components/GobanThemePicker/GobanThemePicker.tsx
+++ b/src/components/GobanThemePicker/GobanThemePicker.tsx
@@ -63,6 +63,9 @@ export class GobanThemePicker extends React.PureComponent<GobanThemePickerProper
 
                 this.selectTheme[k][theme.theme_name] = () => {
                     preferences.set(`goban-theme-${k}`, theme.theme_name);
+                    if (data.get("one-color-go") && k === "white" && theme.theme_name !== "Plain") {
+                        preferences.set(`goban-theme-black`, theme.theme_name);
+                    }
                     let up = {};
                     up[k] = theme.theme_name;
                     this.setState(up);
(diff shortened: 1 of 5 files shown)


arnaudgo:
Hello,

Whereas I'm happy that OGS provides the ability to emulate one color go through customization, the PR title is a bit misleading :) I still hope to see in a future a game type dedicated for that variant because as you mention it: not only it changes it for the whole site so for all games but also it changes it only from one side.

So to make it work, the wo players need to agree to put the full site in one color when maybe both want to keep on playing other games with distinct colors. Well, there's no need for you someone explains it to you, I just put in different words what you already say here.

Arnaud

mlopezviedma:
Arnaud,

Thanks for the remarks, I certainly was very lazy when writing the description.

About the technical aspects:

- A one color go variant handled on the server is tricky, because it's easy for your opponent to hack his/her local client to show the stones normally.
- Having the whole site in one color mode allows you practice by a way of observing games in one color.
- Switching between one color and normal modes is just one click away in settings (and all non-refreshed tabs remain as they are) so you will be able to apply one color mode to one tab only. Being a feature that you won't really use much, I don't think it's troublesome.

That said, I know what I did is a very dirty hack, but in my opinion it suits its purpose just fine, and it is rather simple and elegant way to achieve it (lichess.org does a similar thing for blindfold chess).

anoek:
I think I'd rather see this implemented as a proper game mode as well, perhaps as simple as a flag within the game object that forces the theme to be set to same color stones. While this would require a little bit of api support, i don't think we need to get too crazy with anti-cheating in this mode, it's just a for-funsy mode, client side enforcement should be good enough, which can be enforced on a per game level by looking at a flag. 

While I can appreciate the simplicity of this patch, I don't want to add extra stuff in the settings page if we don't have to, especially since the same effect can be pretty easily accomplished by just setting the stone colors manually (which, if you're OK with playing with a hack for the time being, I think is a fine thing to ask of folks when the alternative fast implementation would be something like this, which would effectively do the same thing and just set those colors the same for you.)

arnaudgo:
It was not a criticism, from the beginning, I think we were sharing the same opinion and again, I'm quite happy to use that feature for playing one color go. I was just thinking it would be a pity to forget about *actually* implementing that go variant :)

I would prefer a flag server-side. As far as players can play ranked one color games, it's a must.
We could even imagine tournaments, ladders of one color go for crazy people, hehe...

Of course one color or not, we are playing go for fun mostly but if we make games ranked, kind of consistency must be respected.
Anyway .... it won't never be something very serious... :)

mlopezviedma:
I see, very fair points. Now I feel even more excited about seeing this fully implemented. :D

Closing.

Issue #99, opened 2017-03-06
both /sensei and /senseis should create search urls


crodgers:
### Summary

OCD led me to leave the 's off of /sensei, but /senseis should also work since that is the common reference name.

### Steps To Reproduce:

1. use the command /senseis <insert search terms here>
2. this won't work, but it should

PR #100, opened 2017-03-06
Add /senseis to the list of commands

crodgers:
same as /sensei, but with an s.

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -794,6 +794,12 @@ function ChatLine(props) {{{
         );
       }
 
+      if (body.substr(0, 9) === '/senseis ') {
+        body = generateChatSearchLine(
+          'http://senseis.xmp.net/?search=', '/senseis ', body
+        );
+      }
+
       if (body.substr(0, 8) === '/google ') {
           body = generateChatSearchLine(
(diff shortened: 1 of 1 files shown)


crodgers:
For issue #99.

src/components/Chat/Chat.tsx
```diff
@@ -794,6 +794,12 @@ function ChatLine(props) {{{
         );
       }
 
+      if (body.substr(0, 9) === '/senseis ') {
```
anoek: Could we merge this if statement with the /sensei one we don't have duplicated cases?

Perhaps something like `if (/^\/senseis?\s/.test(body))` ?

crodgers: I tend to avoid regexp because I am not too familiar with it, but this is a much cleaner solution, It'll be good for me.

BarneyStratford: If you want to learn about regexes, then there are lots of good sites out there. They look quite intimidating, but are really very simple. Eg https://regexone.com

crodgers: @BarneyStratford Thanks for the recommendation. That's a great tutorial.

crodgers:
I'll close this and reopen another pull request with the regec.

PR #102, opened 2017-03-06
Fix for bug #64

BarneyStratford:
The game list now displays the player's opponents correctly.

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -41,43 +41,13 @@ export class GameList extends React.PureComponent<GameListProps, any> {
         };
     }
 
-    sortByClock = () => {
-        if (this.state.sort_order === 'clock') {
-            this.setState({sort_order: '-clock'});
-        } else {
-            this.setState({sort_order: 'clock'});
-        }
-    }
-
-    sortByOpponentClock = () => {
-        if (this.state.sort_order === 'opponent-clock') {
-            this.setState({sort_order: '-opponent-clock'});
-        } else {
-            this.setState({sort_order: 'opponent-clock'});
-        }
-    }
-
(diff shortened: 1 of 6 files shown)


src/components/GameList/GameList.tsx
```diff
@@ -41,48 +41,17 @@ export class GameList extends React.PureComponent<GameListProps, any> {
...
-    }
-
-    sortByMoveNumber = () => {
-        if (this.state.sort_order === 'move-number') {
-            this.setState({sort_order: '-move-number'});
-        } else {
-            this.setState({sort_order: 'move-number'});
-        }
+    sortBy(name: string) {
+        return () => {
```
anoek: I like this style a lot, it's much cleaner, and for this component I think it should be fine. Just to be aware of though, the reason why we have a lot of method assignment `fn = () => {...}` style declarations is because when react is in the `render()` method, if you have any properties that change in a sub-component it forces a re-render (or at the least, some extra work figuring out it doesn't need to re-render) the sub-component. By doing `onClick={this.sortBy("thing")}` we change the value that's assigned to the `onClick` property for each header sub-component every time we are in the `render()` call for our game list. (We're also creating a new closure each pass through render, for each header). In this case it's not very expensive in practice as the lists don't change that much, but for other components it's a good thing to be aware of.

A compromise that I've done elsewhere is to simply make several fields and assigned them with `.bind` in the constructor, ie `this.sortByName = this.sortBy.bind(this, "name")`, and then assign `onClick={this.sortByName}`

Regardless, I'm clicking merge anyways because it's not going to make much of a difference here, and frankly this is prettier :)

anoek:
Nice, thanks! And thanks for the code cleanup!

Issue #95, opened 2017-03-06
Search tool does not accept spaces


juanchodepisa:
### Summary

When using the search tool, it is not possible to enter the space character. The only way to do so is to write everything in a single chunk and then use the cursor (or the arrow keys) to put the space in between the words

### Environment

 * Operating system: Windows 10
 * Browser: Chrome
 * Browser Version: 56.0.2924.87

### Steps To Reproduce:

1. Click on the OGS logo to reveal the search tool
2. Try typing "two words"
3. Fail
4. Try typing "twowords"
5. Move the cursor in between to put the space in "two words"
6. Succeed

PR #103, opened 2017-03-06
Whitespace in the search box

BarneyStratford:
Fix bug #95.

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -205,8 +205,11 @@ export class NavBar extends OGSComponent<{}, any> {
         try {
             let q = ev.target.value || "";
 
-            if (q !== this.state.omnisearch_string) {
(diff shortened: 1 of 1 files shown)


src/components/NavBar/NavBar.tsx
```diff
@@ -203,10 +203,13 @@ export class NavBar extends OGSComponent<{}, any> {
...
     updateOmnisearch = (ev) => {
         try {
-            let q = (ev.target.value && ev.target.value.trim()) || "";
+            let q = ev.target.value || "";
 
-            if (q !== this.state.omnisearch_string) {
+            if (q.trim() !== this.state.omnisearch_string.trim()) {
                 this.abortOmnisearch();
+            } else {
+                this.setState({ omnisearch_string: q });
```
anoek: I don't think we need to set this setState here as we set it down elow in both the `q === ""` and `q !== ""` conditions

anoek:
Awesome thanks!

PR #105, opened 2017-03-07
Page control improvements

### LINKED ISSUE #88: Page number entry box on /observe-games shows "NaN" when trying to clear current value to enter a new one
### Summary

The page number entry textbox (part of "page-controls" div) in /observe-games immediately switches to display "NaN" when trying to remove the current value to be able to enter a new one.

A workaround is to first "select" the current value and then enter the new value - that way the input is never cleared and whatever validation that wants to turn it to "NaN" does not trigger.

### Environment

 * Operating system: Chrome OS
 * Browser: Chrome
 * Browser Version: 56.0.2924.110 (64-bit)

### Steps To Reproduce:

1. /observe-games
2. Click on the input box displaying current page number (should be 1 to start with)
3. Backspace away the 1
4. Observe input immediately swtiching to "NaN" value

mlopezviedma:
- Larger font size, this way buttons are easier to access on mobile and numbers are more legible.
- Fix overlap betweeen Total Pages and Next Page button on /observe-games:
screenshot-2017-03-07-12h44m09s
- Make "**of**" string translatable.
- Fix #88

### DIFF
--- a/src/views/ObserveGames/ObserveGames.tsx
+++ b/src/views/ObserveGames/ObserveGames.tsx
@@ -120,7 +120,11 @@ export class ObserveGames extends React.PureComponent<ObserveGamesProperties, an
         this.setPage(this.state.page + 1);
     }}}
     setPage = (ev_or_page) => {{{
-        let page = parseInt(typeof(ev_or_page) === "number" ? ev_or_page : (ev_or_page.target as any).value) || 1;
+        if ((ev_or_page.target as any).value === "") {
+            this.setState({page: ""});
+            return;
+        }
+        let page = parseInt(typeof(ev_or_page) === "number" ? ev_or_page : (ev_or_page.target as any).value);
         page = Math.max(1, Math.min(Math.ceil(
             (this.state.viewing === "live" ? this.state.live_game_count : this.state.corr_game_count)
                 / this.state.page_size), page));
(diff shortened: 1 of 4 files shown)


anoek:
I think the author of #88 was hitting backspace to clear the value before entering a new one. I'd like to support that work flow a little better so the user can select the value, hit back space, then enter something like '2' to jump to the second page. The question being what to do with the query while it's in an invalid state... I think I'd advocate simply not doing a refresh while that text box is empty

Also, with this patch the next/previous step icons render a little bigger than the text, can we reduce the size of those a little?

mlopezviedma:
There you go.

Next/previous icons were the same size as the text, both before and after that first commit. Seems like its height is more aparent now that is bigger.

src/views/ObserveGames/ObserveGames.tsx
```diff
@@ -120,7 +120,11 @@ export class ObserveGames extends React.PureComponent<ObserveGamesProperties, an
         this.setPage(this.state.page + 1);
     }}}
     setPage = (ev_or_page) => {{{
-        let page = parseInt(typeof(ev_or_page) === "number" ? ev_or_page : (ev_or_page.target as any).value) || 1;
+        if ((ev_or_page.target as any).value === "") {
```
anoek: Need to handle when ev_or_page is a number here (will raise an exception as it is currently authored)

anoek: (this happens when you click the 'next' and 'previous' icons btw)

mlopezviedma: Duh! I didn't even test those buttons afterwards. Sorry for the caveman coding. ^.^

anoek:
Excellent, thanks!

PR #121, opened 2017-03-09
Devel

crodgers:
Replace explicit comparison with a regex for a cleaner solution to implementing both '/sensei' and '/senseis' commands.


anoek:
:+1:

PR #126, opened 2017-03-09
We had some issues with sorting players of undefined ranking. Here is…

BarneyStratford:
… a further place where the ranking may be undefined. I've been looking at the bug that causes the number of spectators to be incorrect, but been unable to reproduce it. This fix might help in that regard as well.

I've also included an alm config file to cause it to follow the project's coding standard more closely.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*.{js,jsx,ts,tsx}]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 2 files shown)


anoek:
:+1:

BarneyStratford:
> In the future please keep file adds like .editorconfig as their own commit and probably own pull request
Will do.

Barney.

PR #129, opened 2017-03-09
Fetch rating on /overview to fix progress bar

### LINKED ISSUE #125: Rank bar on /overview doesn't work
### Summary

[Overview.tsx](https://github.com/online-go/online-go.com/blob/devel/src/views/Overview/Overview.tsx) uses `data.get("config.user")`, but that object doesn't have a `rating` field. As a result, the rank bar always shows 50%.

### Environment

 * Operating system: Any
 * Browser: Any

### Additional Information:

screenshot-2017-03-09-16h49m00s

screenshot-2017-03-09-16h50m37s

mlopezviedma:
Fixes #125

### DIFF
--- a/src/views/Overview/Overview.tsx
+++ b/src/views/Overview/Overview.tsx
@@ -30,6 +30,7 @@ import {Player} from "Player";
 import {PlayerIcon} from "components";
 import online_status from "online_status";
 import data from "data";
+import player_cache from "player_cache";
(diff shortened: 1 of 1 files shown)


anoek:
Totally works but this is better fixed on the server side so we don't have to do another fetch, that rating info was simply missing from the config being sent out. I've added that in and now things work :)

PR #132, opened 2017-03-10
"Learning hub" view

emauton:
This is (the start of) an implementation of
https://forums.online-go.com/t/accessible-teaching/11039

The idea is to use a stripped down "Group" view so that we can drive the
content of the learning hub page (news, teachers) from existing
infrastructure.

We'll need to create an invite-only, private "OGS Teachers" group on the
main site to drive the page from, and presumably a `global-teaching`
channel to link with from `EmbeddedChat`.

Once we have this hashed out, next we'd like to create an interactive
tutorial that we'll link to from this page along with a player's
progress - that'll be more involved though.

There are some problems I know about with this initial stab:

*   `PaginatedTable` may not be the right thing to use for the list of
    teachers, given that we'd like to sort the whole group by online
    status, and keep it refreshed.
*   The "Resources" section is anaemic.

Any advice appreciated.

A screenshot running against the Beta server:

*   http://cian.furiousthinking.org/picture/ogs-learn-1.png

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -300,6 +300,7 @@ export class NavBar extends OGSComponent<{}, any> {
                 {(!this.state.user.anonymous || null) && <Link to="/overview">{_("Home")}</Link>}
                 {user && <Link to="/play">{_("Play")}</Link>}
                 <Link to="/observe-games">{_("Watch")}</Link>
+                <Link to="/learn">{_("Learn")}</Link>
                 <Link to="/chat">{_("Chat")}</Link>
                 <Link to="/puzzles">{_("Puzzles")}</Link>
                 <Link to="/tournaments">{_("Tournaments")}</Link>
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -65,6 +65,7 @@ import {LibraryPlayer} from "LibraryPlayer";
 import {Play} from "Play";
 import {Moderator} from "Moderator";
 import {ObserveGames} from "ObserveGames";
+import {Learn} from "Learn";
 import {Puzzle} from "Puzzle";
 import {PuzzleList} from "PuzzleList";
 import {PuzzleModify} from "PuzzleModify";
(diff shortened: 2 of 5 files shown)


anoek:
Cool, thanks for taking the lead on this :)

Re the paginated table for teachers, we could just show the list of online users in the learning hub room if you wanted? It's been in the back of my mind to highlight group admins in chat, I'd be happy to make that happen so they can show up as "official" teacher folk in the chat room or something like that?

Also if you'd like I can help get a template down for an interactive tutorial. We had one in v4, the functionality was decent enough I thought but the content was sub-par, i'd be happy to drop in what we had though and presume the content will be gutted and replaced with something that doesn't suck.

emauton:
Yeah, just a list of who's in chat + a highlight for teachers (i.e. members of the "OGS Teachers" group) would be good. Ideally no-one needs to actually join a group to come hang out in `global-teaching` but we'd still like to highlight the teachers.

Tutorial template sounds good, then we could sick mark5000 & Kaworu Nagisa on it to make it better, and I could handle the updates? I think we can probably get this "hub" idea into play soonish though.

emauton:
Oh, and by the way, I'm impressed with how easy you made it to get started with the codebase, it's been fun and easy to work with.

mlopezviedma:
Indeed, the code is very neat and self-documented. Thumbs up!

anoek:
FYI @emauton I've ported over the old tutorial which should be useful in seeing at least one way of approaching a series of interactive boards and the like: https://github.com/online-go/online-go.com/tree/devel/src/views/Tutorial

mark5000:
I'd like to get started on the content of the "Learning Hub." @anoek how can a non-programmer put his time to best use in creating the necessary content?

anoek:
Excellent, thanks in advance @mark5000 

For interactive components, internally i've re-used the puzzle system, so what I did to create the initial tutorial was to create several puzzles using the existing puzzle creation system, then I went through and copied out the config from them when creating the tutorial. So, for interactive components, maybe just create a puzzle collection on the beta server that contains the interactive boards you want?

For rich non interactive content, I reckon whatever works best for you (google doc or something?) and then we go from there to web page as appropriate.

anoek:
No pressure, but is there still interest in doing this?

emauton:
Yeah, I think so, but I'm afraid I've been somewhat distracted by Life.​
Sorry!

anoek:
Feel free to re-open this if you decide to pick it back up

PR #137, opened 2017-03-11
Improved text contrast for player name and score (especially dark theme)

ihoover:
### Issue I am trying to fix
In light theme, the black player's name and rank are hard to read. In the dark theme, the White player's name and rank are almost impossible to read. Those are the main points.

### Method
I made a black and white name color in each theme. I made the player-name-container div also  subclass the color (I think that's how you describe it. It has been  a while since I did anything web). The color style attribute for `.[color] .player-name-container` is then themed. This implies removing the color responsibility from the `Player` class.

### Other Minor Tweaks
I removed the transparent border as it was creating an unsightly white line under the player boxes while in the dark theme. Also, I lessened the contrast between the score text and the background while in the dark theme.

### DIFF
--- a/src/components/Player/Player.styl
+++ b/src/components/Player/Player.styl
@@ -16,7 +16,6 @@
  */
 
 .Player {
-    themed color user
     cursor: pointer;
     white-space: nowrap;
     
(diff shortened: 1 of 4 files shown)


anoek:
This is a good step in the right direction I think, though I think we might need to make the white clock a bit more legible still?

ogs minibeast com-game-7172647 iphone 6 plus

ihoover:
Yep. Clock contrast was a big oversight of mine. I think that is fixed with the latest commit.

Issue #140, opened 2017-03-11
Miniature stone rendering


lemurek:
### Summary

Miniature game previews (especially 19x19) render is unnecessarily ugly. Because stone spacing is fixed, in lower resolutions stones seem to be misplaced and misaligned.

I believe it might be a problem as this is usually one of the first screens newcomers see after logging.

### Environment

 * Operating system: Any
 * Browser: Any
 * Browser Version: Any

### Steps To Reproduce:

1. Navigate to 'Watch' section.

### Additional Information:

I believe that this:
![image](https://cloud.githubusercontent.com/assets/12175058/23822721/8622a078-064a-11e7-8ab7-35dfe5c28ba0.png)
Could be easily fixed by scaling stone spacing proportionally to stone sizes, and applying some a little bit of floating points to the transforms.

ihoover:
I honestly can't see anything that looks particularly misaligned in the example screen-shot you give. Is it very subtle or am I just missing something obvious?

lemurek:
Well... let's zoom in a little bit. 

![screen shot 2017-03-11 at 11 11 15](https://cloud.githubusercontent.com/assets/12175058/23822779/a46b1b72-064b-11e7-9d73-0b67eab9de81.png)

I wouldn't say it's a major problem, but since Go is visually pleasing game, we should make sure that the Main preview screen is pleasing to look at. I'm not sure why stones get smaller in low resolution, but space between them have to stay at exactly 2 pixels.

lemurek:
Btw. problem only affects rendered stones. Plain stones are aligned perfectly! :) The distance between stones could be still reduced in that case, but that's a rather subjective matter.

![image](https://cloud.githubusercontent.com/assets/12175058/23823087/69f0e16e-0652-11e7-8149-ad108eecc6df.png)

lemurek:
Ok, so I noticed that line paths are traced at 0.5 pixel to make sure they are as thin as possible, i.e. just 1 pixel wide. At the same time rendered stones are centred at full-pixel number. That inconsistency produces this 1 pixel shift. If it's easy enough I'll try to submit suggested fix in form of pr

PR #144, opened 2017-03-11
Align stone and line centers

lemurek:
Fix proposition for https://github.com/online-go/online-go.com/issues/140

The change acknowledges specific position of stone inside of rendered square. Pre-renders all stones shifted 0.5 pixel to the left, which aligns with the way goban lines are printed. 

Also, since the stone alignment is done correctly, prerender function is able to accept fractional radius parameter. I used that opportunity to make radius scale proportionally to board size instead of leaving 1 pixel of margin at all times.

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -2957,7 +2957,7 @@ export abstract class Goban extends EventEmitter {
     abstract getSelectedThemes();
 
     private computeThemeStoneRadius(metrics) {{{
-        return Math.round(Math.max(1, metrics.mid - 1.0));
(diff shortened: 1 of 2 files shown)


lemurek:
Before: 

screen shot 2017-03-11 at 15 48 14

After: 

screen shot 2017-03-11 at 16 32 42

lemurek:
Before: 

screen shot 2017-03-11 at 15 47 50

After:

screen shot 2017-03-11 at 16 33 20

lemurek:
Before:

screen shot 2017-03-11 at 15 47 21

After:

screen shot 2017-03-11 at 16 34 35

anoek:
Nice :)

PR #145, opened 2017-03-12
Fix reports tracker scrolling

mlopezviedma:


anoek:
Awesome thanks for that

PR #149, opened 2017-03-12
Text contrast

ihoover:
Whoops, looks like I missed something. I realized they way I did it before removed coloration from the chat window. So I added the color back to the .Player style, but then just made it override if inside a plyer-name-container class.


anoek:
Thanks for following up with a fix :)

PR #181, opened 2017-03-26
I believe I've nailed the bug that causes the number of players in a …

BarneyStratford:
…game to be different to the number of players listed. It was caused by the GameChat object subscribing to the wrong chat_proxy. The Game object can initially contain the wrong game_id. The game_id gets corrected at a later stage, but after the GameChat has initialised. I corrected this by making the user count follow the method used by the user list.


anoek:
Ahh, makes sense, good job hunting that one down! And good solution / code cleanup too :)

roy7:
3 cheers for @BarneyStratford !

Issue #189, opened 2017-03-29
Ranking bars missing


BarneyStratford:
### Summary

Ranking bars are not always visible in user profile

### Environment

 * Operating system: Mac OS X 10.11.6
 * Browser: Safari 10.1 and Firefox 52.0.2

### Steps To Reproduce:

1. Go to the page of any user whose rating comes less than half-way between ranks. Mine is currently in this state: https://online-go.com/user/view/351937
2. Whenever the rating comes less than half-way between ranks, the coloured bar has height 0 and so becomes invisible.

### Additional Information:

![ranking](https://cloud.githubusercontent.com/assets/5913995/24450111/80501b8c-1471-11e7-81a3-0c01e7a8bb35.png)
I know what is causing this and will submit a pull request in due course.

anoek:
Confirmed on firefox, chrome renders it correctly though.

anoek:
Fixed by #191

PR #191, opened 2017-03-29
Fix bug #189 that caused rankings to display incorrectly

BarneyStratford:
Each player's profile page displays the rankings of the players using a progress bar style UI element with a number indicating the progress. When the progress is less than 50%, then the number is supposed to be displayed to the right of the progress bar, whereas for more than 50% the number is on top of the progress bar. This ensures that the number doesn't overflow the progress bar element.
In the first case (<50% progress) the bar was invisible. This commit fixes that.

### DIFF
--- a/src/views/User/User.styl
+++ b/src/views/User/User.styl
@@ -131,25 +131,28 @@
         margin-left: -0.2rem;
         min-height: 10em;
     }
-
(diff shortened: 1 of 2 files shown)


anoek:
Excellent, thanks for the fix!

Issue #193, opened 2017-03-31
Slate and shell stones are clipped and too large on mobile devices


anoek:
### Summary

The slate and shell, and other rendered stones, appear to be too big and cropped on mobile devices. 

### Environment

 * Operating system: Android 7.x
 * Browser: Chrome
 * Browser Version: 57

### Steps To Reproduce:

1. Select Slate and Shell as your stone themes
2. View a game

### Additional Information:

This is probably a side effect of https://github.com/online-go/online-go.com/pull/144 as those have been the only changes to the affected theme code. Pinging @lemurek to see if he might have any insight.

![image](https://cloud.githubusercontent.com/assets/168460/24548818/8941f822-15e4-11e7-8346-045dde521006.png)

lemurek:
Hey! I'll take a look. I don't have experience with testing on Android, so any suggestions on how to connect a phone to my local dev environment would be helpful.

anoek:
@lemurek awesome thanks! As for debugging on android, they make it really easy see: https://developers.google.com/web/tools/chrome-devtools/remote-debugging/ 

Basically follow those instructions and you can access the chrome developer console on your desktop's chrome.

As for accessing the development site from your phone, simply go to http://<your local ip>:8080/ where local ip is the ip of the machine running `npm run dev`

lemurek:
That should be sufficient, I'm on it :) Expect some pr today

lemurek:
Hmm... I think I run into the same problem before (I did try to test it on my phone when submitting original change). 

![screen shot 2017-04-01 at 12 06 07](https://cloud.githubusercontent.com/assets/12175058/24578211/ac41f094-16d3-11e7-8117-d612b7b08b9c.png)

When connecting from a phone service sees the request (see 'GET /' and 'GET /game/868' ) but somehow the communication breaks.

Luckily Chrome has a built in mobile emulator, which allows me to recreate the issue on my laptop.

lemurek:
Hm, so I understand the problem now. Not sure about the right solution yet. See this log:

![screen shot 2017-04-01 at 13 04 08](https://cloud.githubusercontent.com/assets/12175058/24578615/b83d75e6-16db-11e7-9b5c-ad61adbe64a1.png)

When scaling is applied stones are pre-rendered with a different radius than the one passed in render phase. 

49/23.9399 = ~2.046
21/9.12 = ~2.302

[...]

PR #194, opened 2017-04-01
Pre-render stones in canvas resolution

lemurek:
Quick fix for https://github.com/online-go/online-go.com/issues/193

Pre-renders stones in canvas resolution.

screen shot 2017-04-01 at 14 46 22
screen shot 2017-04-01 at 14 46 33
screen shot 2017-04-01 at 14 46 44
screen shot 2017-04-01 at 14 46 59
screen shot 2017-04-01 at 14 47 14
screen shot 2017-04-01 at 14 47 26

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -2956,7 +2956,7 @@ export abstract class Goban extends EventEmitter {
     abstract getSelectedThemes();
 
     private computeThemeStoneRadius(metrics) {{{
-        return Math.max(1, this.square_size * 0.48);
+        return Math.max(1, this.square_size * 0.49);
     }}}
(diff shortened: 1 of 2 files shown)


anoek:
mm while the clipping is fixed the stones are fuzzy because you've gotten rid of the device scaling factor :(

lemurek:
My original idea was using the scaling factor. I can revert it. Personally I think now it looks better as the extra manual scaling during rendering looks worse/not better than this. 

With the high res pre-render I could see some artifacts on stone edges, so I'd have to go back to 0.48 multiplier (which is by no means fine) 

I think I'll submit another pull request with the original pre-render size and will let you decide.

anoek:
Closed in favor of https://github.com/online-go/online-go.com/pull/198

PR #195, opened 2017-04-01
Account for device canvas scale when repositioning pre-rendered stone.

lemurek:
Another quick fix proposition for:
https://github.com/online-go/online-go.com/issues/193

This one keeps the stone render at high res and makes necessary adjustment during rendering.

screen shot 2017-04-01 at 20 29 08
screen shot 2017-04-01 at 20 29 21
screen shot 2017-04-01 at 20 29 31
screen shot 2017-04-01 at 20 30 02
screen shot 2017-04-01 at 20 29 41
screen shot 2017-04-01 at 20 29 50

### DIFF
--- a/src/lib/ogs-goban/themes/rendered_stones.ts
+++ b/src/lib/ogs-goban/themes/rendered_stones.ts
@@ -359,24 +359,35 @@ function preRenderStone(radius, seed, options) { /* {{{ */
     return [{"stone": stone[0], "shadow": shadow[0]}];
 } /* }}} */
 function placeRenderedStone(ctx, shadow_ctx, stone, cx, cy, radius) {{{
-    let ss = square_size(radius);
-    let center = stone_center_in_square(radius);
-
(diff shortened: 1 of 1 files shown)


lemurek:
I think I'll need to look into this approach a bit more. There may be a problem with retrieving pre-rendered stones from cache. After refreshing the view for IPhone 6 Plus, this is what I'm seeing:

image

It looks like it defaults to 1-pixel sized stones. Unfortunately I'm not sure if it's the problem with mobile environment simulation or something else. I'll need to debug it

lemurek:
False alarm, the solution is ok. It's just that particular stone radius renders awkwardly. After removing coordinates the slight difference makes the stones 'round' again:

image

I still think the 'fuzzy' stones rendered for the canvas resolution look better ;)

lemurek:
Closing in favor of https://github.com/online-go/online-go.com/pull/198

PR #197, opened 2017-04-02
Render stones in 2x res for scaled devices.

lemurek:
Yet another fix proposition for https://github.com/online-go/online-go.com/issues/193

I've had this extra idea that allows keeping the pre-rendering with affixed centre spot, while reverting the handling of scaled devices to before my change. By simply pre-rendering the stones in 2x resolution we get few pleasant advantages:

  - pre-rendered stone fills almost 100% of the image so radius becomes the valid offset again as rounding becomes negligible
  - manual rounding to full pixels can be avoided and a sub-pixel smoothing can be left to the browser

The result looks pretty good while maintaining sharpness of the image:

screen shot 2017-04-02 at 12 03 52
screen shot 2017-04-02 at 12 03 19
screen shot 2017-04-02 at 12 04 05
screen shot 2017-04-02 at 12 04 18
screen shot 2017-04-02 at 12 04 31
screen shot 2017-04-02 at 12 04 40

### DIFF
--- a/src/lib/ogs-goban/themes/rendered_stones.ts
+++ b/src/lib/ogs-goban/themes/rendered_stones.ts
@@ -218,7 +218,10 @@ function clearAboveColor(ctx, width, height, r, g, b) { /* {{{ */
     ctx.putImageData(image, 0, 0);
 } /* }}} */
 function preRenderStone(radius, seed, options) { /* {{{ */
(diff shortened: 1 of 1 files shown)


lemurek:
Closing in favour of https://github.com/online-go/online-go.com/pull/198

PR #198, opened 2017-04-02
Focus on symmetry for scaled stones rather than center placement.

lemurek:
This fix I'm happy with. Hope you like it as well 😀 

https://github.com/online-go/online-go.com/issues/193

I thought about it for a while before understanding that the problem is in symmetry. For the higher resolution of mobile devices I can ignore the centre alignment (as the pixels are so small) and focus on a simple stone symmetry to make them look nicer. Visually this should revert the look and feel for mobile to before my original change, while keeping centre positioning for non-scaled devices. 

I believe that visually this is the best solution. Here are the examples:

screen shot 2017-04-02 at 14 23 32
screen shot 2017-04-02 at 14 23 42
screen shot 2017-04-02 at 14 23 52
screen shot 2017-04-02 at 14 24 13
screen shot 2017-04-02 at 14 24 30
screen shot 2017-04-02 at 14 24 44

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -2956,7 +2956,7 @@ export abstract class Goban extends EventEmitter {
     abstract getSelectedThemes();
 
     private computeThemeStoneRadius(metrics) {{{
-        return Math.max(1, this.square_size * 0.48);
+        return Math.max(1, this.square_size * 0.488);
     }}}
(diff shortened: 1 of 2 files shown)


lemurek:
And here you can see the difference the 20% decrease of gap sizes makes in high resolution:
screen shot 2017-04-02 at 14 06 37
screen shot 2017-04-02 at 14 05 19

Issue #199, opened 2017-04-03
Broken Puzzle: "TypeError: Cannot read property 'left' of null"


mark5000:
### Summary

A puzzle I designed on beta.online-go.com is giving an error message that reads "TypeError: Cannot read property 'left' of null."

### Environment

 All devices and browsers

### Steps To Reproduce:

1. Go to https://beta.online-go.com/puzzle/9
2. See an error message "TypeError: Cannot read property 'left' of null"

### Additional Information:

This error occurred after I had deleted all branches of the puzzle and attempted to save.

anoek:
Fixed by https://github.com/online-go/online-go.com/pull/200 thanks to @BarneyStratford

PR #200, opened 2017-04-03
Fix bug #199

BarneyStratford:
Puzzle.getBounds can return null. Puzzle.fetchPuzzle failed to account for this. Added a check for the null return value.


anoek:
Awesome

Issue #208, opened 2017-04-06
Ruleset not displayed when accepting game


bajidu:
When accepting a game from the list there is no way of knowing which ruleset will be applied; is it possible to display that along the other information which is shown beforehand?

anoek:
Thanks for the patch @DmitriyKirakosyan !

PR #211, opened 2017-04-08
#208: header <Rules> added to game list page

DmitriyKirakosyan:
New <Rules> column and small refactoring in Play.tsx.


anoek:
Looks good! And good refactor.

The only thing I'd ask is to use the `rulesText` function ( https://github.com/online-go/online-go.com/blob/devel/src/lib/misc.ts#L36 )  to process `C.rules` so the names are translated and such.

Thanks for the PR!

DmitriyKirakosyan:
I updated with using rulesText. Thanks for the remark.

PR #218, opened 2017-04-13
Observed games can now be sorted by move number

john-banks:
I also added a ONLINE_GO_PORT environment variable to the gulpfile
Port 8080 is used quite commonly so now you can override it

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -136,7 +136,7 @@ function background_webpack(done) {
 }
 
 function dev_server(done) {
-    let port = 8080;
+    let port = (process.env.ONLINE_GO_PORT ? process.env.ONLINE_GO_PORT : 8080);
     let express = require('express');
     let body_parser = require('body-parser');
     let http = require('http');
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -124,7 +124,12 @@ export class GameList extends React.PureComponent<GameListProps, any> {
                 case 'move-number' :
                     lst.sort((a, b) => {
                         try {
-                            return a.json.moves.length - b.json.moves.length || a.id - b.id;
+                            if(a.json && b.json){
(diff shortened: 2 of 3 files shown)


src/components/GameList/GameList.tsx
```diff
@@ -124,7 +124,12 @@ export class GameList extends React.PureComponent<GameListProps, any> {
                 case 'move-number' :
                     lst.sort((a, b) => {
                         try {
-                            return a.json.moves.length - b.json.moves.length || a.id - b.id;
+                            if(a.json && b.json){
```
apetresc: The Shippable test failures have to do with the linter complaining about the lack of whitespace, on this line and [line 130](https://github.com/online-go/online-go.com/pull/218/files#diff-4a58b42a1aaf5f8daaffc4de471dc12aR1300).

src/views/ObserveGames/ObserveGames.tsx
```diff
@@ -190,7 +191,7 @@ export class ObserveGames extends React.PureComponent<ObserveGamesProperties, an
                 </div>
             </div>
 
-            <GameList list={this.state.game_list} disableSort={true} emptyMessage={_("No games being played")} />
+            <GameList list={this.state.game_list} disableSort={false} emptyMessage={_("No games being played")} />
```
anoek: The problem with enabling sorting here is that it doesn't actually sort the entire list, only the current page. Support for sorting needs to be added on the server side to support this.

Gulpfile.js
```diff
@@ -136,7 +136,7 @@ function background_webpack(done) {
 }
 
 function dev_server(done) {
-    let port = 8080;
+    let port = (process.env.ONLINE_GO_PORT ? process.env.ONLINE_GO_PORT : 8080);
```
anoek: :+1:

anoek:
Feel free to re-open once things are fixed up.

PR #287, opened 2017-05-24
Fix timecop notification

mlopezviedma:
Notification showed "You have timeout to make your move" instead of e.g. "You have 12 hours to make your move".


anoek:
Oh good, thanks for finding and fixing this!

Issue #271, opened 2017-05-06
Produce a sound every time we change the volume from a game page


arnaudgo:
### Summary

From a game page:

The game page should play a sound everytime we click on the volume button. It helps us gauging it.

### Environment

 * Operating system: n/a
 * Browser: n/a
 * Browser Version: n/a

### Steps To Reproduce:

Feel free to ask if needed

### Additional Information:

n/a

Arnaud

Nazgand:
Preferably the stone placement sound.

PR #288, opened 2017-05-24
Play a stone sound every time we change the volume

mlopezviedma:
This applies to game pages, implements #271


anoek:
Awesome thanks :)

Issue #262, opened 2017-04-28
Game thumbnails list threshold should be settable directly from the games page


arnaudgo:
### Summary

Sometimes we want to see thumbnails but maybe not everytime you load the games page. 
It could be nice to have a drop-down list on the games page that allow to dynamically change it without going to our settings.

### Environment

 * Operating system: any
 * Browser: any
 * Browser Version: any

### Steps To Reproduce:
N/A
### Additional Information:
N/A

anoek:
As per the discussion in #290, I think I'd like to change the scope of this request to decouple the list limit functionality on this page, and instead just have a toggle that switches the view from list to thumbnails.

anoek:
This issue has been marked stale and will be closed soon without further activity. To keep the issue open, please respond to the comment to keep the discussion going.

PR #290, opened 2017-05-25
Set game thumbnails list threshold on Observe page

mlopezviedma:
Implements #262.

### DIFF
--- a/src/views/ObserveGames/ObserveGames.tsx
+++ b/src/views/ObserveGames/ObserveGames.tsx
@@ -41,6 +41,7 @@ export class ObserveGames extends React.PureComponent<ObserveGamesProperties, an
             num_pages: 1,
             page_size: preferences.get("observed-games-page-size"),
             viewing: preferences.get("observed-games-viewing"), /* live / correspondence */
+            game_list_threshold: preferences.get("game-list-threshold"),
             game_list: [],
             live_game_count: 0,
             corr_game_count: 0,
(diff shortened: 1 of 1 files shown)


anoek:
Hmm, not sure I'm a fan of how this affects the balance of the page. I kinda question how used the original feature request would be to begin with too, there hasn't been any comments on it.

mlopezviedma:
You mean the position and the space it takes?

anoek:
Yeah, without some confirmation that this is a widely desired feature, I don't think we should implement this in such a prominent wau (and perhaps not at all)

mlopezviedma:
Yeah, I pretty much agree with you. Let's see what everyone thinks :)

apetresc:
Personally, I like this change, just because it's such a standard UI thing in most web applications with pagination to set the page size directly from the interface its modifying. (Though I haven't actually pulled @mlopezviedma's patch yet to see how this particular implementation looks)

anoek:
The effect here is two pagination sizes though, which I think is confusing.

I was thinking about it more, I think that's my primary complaint. We might consider decoupling this list from that limit, and instead just have a toggle on the watch page that toggles from list to thumbnails, and completely ignore that size setting. Then it can be just a simple single icon instead of a big list, which would fit a lot better I think. What do you guys think?

mlopezviedma:
I like that a lot! Furthermore, list threshold is also applied to profile pages, so changing this preference in other place than **Settings** looks somewhat inconsistent to me.

anoek:
Closing this in favor of waiting for a patch that does the thumbnail/list toggle button instead

Issue #298, opened 2017-05-28
Identity Crisis


BarneyStratford:
### Summary

A friend (user id 179886) has recently changed her username from "wolf guard 🐺" to "Uconn Huskies". She seems to be having an identity crisis as a result, with her username flipping back and forth between the two. I've confirmed that it's not an elaborate practical joke this time...

### Environment

 Many users have reported seeing this across several platforms.

### Steps To Reproduce:

1. English chat.
2. Whenever Uconn Huskies comments, her name becomes correct. It stays like that for a while, then the chat screen flickers as it reverts to her old username.

There's also something weird going on with her private chat.

### Additional Information:

![chat1](https://cloud.githubusercontent.com/assets/5913995/26531868/c8a0bb82-43ea-11e7-8ba1-ede8956390c8.png)
![chat2](https://cloud.githubusercontent.com/assets/5913995/26531870/c9faf3f8-43ea-11e7-9596-06a4bbc6da03.png)
![chat3](https://cloud.githubusercontent.com/assets/5913995/26531871/cc6e62a0-43ea-11e7-9454-f8f29f542391.png)

roy7:
Assuming she didn't change it back I actually see the Uconn name right now on

https://online-go.com/player/179886/

PR #329, opened 2017-06-09
Types

BarneyStratford:
I've started a project to improve the reliability and maintainability of the OGS codebase. The aims are:
    * Separate abstract data from its implementation details (especially the UI)
    * Use the type system to prevent bugs
    * Increase the amount of documentation associated with the code
At the moment this is at a very early stage, and is still very much open for debate and change.


anoek:
So :+1: for typifying this stuff, and while the structure that you've come up with would probably be pretty decent in a new project, it deviates significantly from our current object schemas which I'd like to avoid. The object structures and field names (as unfortunate as some of them are) appear in a lot of places on both the front end and the back end, and so I'd like to keep them consistent.

src/lib/player_cache.ts
```diff
@@ -16,185 +16,192 @@
...
+
 export const player_cache = {
-    update: update,
     lookup: lookup,
+    lookup_by_id: lookup_by_id,
     lookup_by_username: lookup_by_username,
-    nicknames: nicknames,
     fetch: fetch,
-    watch: watch,
-    Listener: Listener,
```
anoek: Was there a reason to remove the Listener functionality?

BarneyStratford:
> Was there a reason to remove the Listener functionality?
Yes, but it’s mostly a matter of style. You see, I’m very much in favour of simplifying everything as much as possible. It leads to fewer places for bugs and less code to maintain. I made a bit of a name for myself when I was at IBM for throwing away whole modules and making the software work better for it. You’ll notice that most of my contributions to date have resulted in less code rather than more.

It seemed to me that the Listener didn’t belong in the player cache. It was used in only one other place (the player icon) and is also duplicated in the data module. Having said that, I can see that there is a good case to be made for having some kind of callback system so that components can update in response to changes in the player cache, and for it to be used to force updates of Player components in general, as usernames, flags, attributes (player.is.*) and rankings can all change in addition to icons.

Perhaps the right way to go is some kind of EventEmitter? Or is that also overkill? Or to create a Listener class somewhere else and refer to it?

See you,
Barney.

anoek:
I'll agree that `Listener` being shoved in there was poor style, especially since apparently it's duplicated code from `data.ts` (these were some of the first files I did during the port.. no excuse for duplicated code, but no surprise that things are a bit ugly.. all this was before my discovery of being able to just make things `EventEmiters`). So with that, I'd be happy to see that transform using a more standard `EventEmitter` or something of the sort, whatever ends up making sense. Regardless, the functionality is important though, it as you noted, it drives the magic feature that updates player's rating, ranks, icons, supporter status, etc throughout the interface wherever a `Player` element is being used. 

It might (not necessarily?) make sense to unify that update stuff with what's being done in `lib/online_status.ts`, where we *do* use a `EventEmitter` instead of some made-at-home `Listener` setup :) Food for thought.

BarneyStratford:
> It might (not necessarily?) make sense to unify that update stuff with what's being done in lib/online_status.ts, where we do use a EventEmitter instead of some made-at-home Listener setup :) Food for thought.
Funny you should mention that - I had had the same thought. :-)

Barney.

BarneyStratford:
OK, seems fair enough. I’ve removed the incompleteness checks in the latest version. I think there might well be some scope for optimising all this a different way, but this project has got quite big enough to be going on with.

See you,
Barney.

src/components/IncidentReportTracker/IncidentReportTracker.tsx
```diff
@@ -73,27 +73,27 @@ export class IncidentReportTracker extends React.PureComponent<IncidentReportTra
             delete this.active_incident_reports[report.id];
         } else {
             report.unclaim = () => {
-                post("moderation/incident/" + report.id, {"id": report.id, "action": "unclaim"})
```
anoek: Missed a %% I think

anoek: Following up.. I'm not sure I'm a fan on this %% change, % style subs have always been a bit more error prone and not as clear as just inline variable substitution, in my opinion

anoek:
So.. this PR is getting massive, which it's awesome that you're doing so much! but at the same time, I'm getting a bit worried about things that breaking along the way. I much prefer PR's that can be a bit more surgical in nature.

BarneyStratford:
I didn’t like doing it that way either. The reason is that in order to be able to infer the type of the data returned from the request, you have to have a fixed string for the URL. If you look in compatibility/index.ts you’ll see that there are a couple of interfaces there that give the types of the data returned, and translations that are to be applied. There’s no way to put in a variable URL there that I can see.

anoek:
Interesting, ok i'll ponder that today while I go through the PR

BarneyStratford:
Yeah, I know what you mean about breakages. I wonder if we could split this up into smaller pieces to make them more digestible. We could make the data cache changes fairly easily as one PR, with the requests changes as another, and pubsub as a third. Then, having got the type machinery in place, the Rank and Player types can be made into a 4th one, followed by more later.

What do you think?

Barney.

anoek:
No need to break up this PR, but maybe we can hold off on introducing other major changes in this PR, and break future ones off? I know you did a lot of core refactoring so that kind of necessitates a big PR like this sometimes.. but yeah, it gets a bit scary :)

So any remaining fixes or minor completions of work related to this PR, keep pushing to this PR, but moving forward if we can break out future work that's reasonably isolated into a subsequent PR that'd be good, that way we can get this one stabalized and merged without having to worry about anything new coming in, and once that's done then we can move on to the next PR's that should be easier to validate and merge :)

anoek:
Yeah beta first, though honestly there's never that much traffic on there and only the big stuff is ever caught. We'll do that, but in reality the buck stops with us when it comes to making sure things are tested enough to not break the live site too badly :)

BarneyStratford:
Might also be an idea to do the update at a pre-planned time so that we can both be available to fix anything that is reported broken. And if it does all go horribly wrong, then there’s always the nuclear option of reverting the merge commit.

anoek:
Sure that'd be great :) We can figure out when that should be once we get things merged.

src/components/Player/AbstractPlayer.ts
```diff
@@ -0,0 +1,153 @@
...
+                    // player to be republished after we've been mounted.
+                    update_state(this.setState.bind(this), player);
+                });
+            };
+        }
+        this.subscribe = new player_cache.Subscription(callback);
+        this.setup(props, (state) => { this.state = state; });
+    }
+
+    componentWillReceiveProps(new_props) {
```
anoek: This is probably where the redraw slowdown  elated to Player's is coming from, this should only doing updates if something changed (React doesn't try to make assumptions for us). 

We also need to implement `componentShouldUpdate` either here or in `Player` to check if we need to redraw based on new / old props and state

src/components/Player/Player.tsx
```diff
@@ -18,405 +18,88 @@
...
-    } catch (e) {
-        console.error(e);
-    }
-});
-
-
-let __html_player_link_id = 0;
-let player_link_data = {};
-let player_link_last_id = 0;
-export function makeHTMLPlayerLink(field_order, player, link_target) { /* {{{ */
```
anoek: Good riddance :)

src/components/SeekGraph/SeekGraph.tsx
```diff
@@ -666,11 +686,17 @@ export class SeekGraph extends EventEmitter {
             }
 
             if (C.live_game) {
-                e.append(makePlayerLink(["name", "rank", "nolink"], {"user_id": 0, "ranking": C.black_rank, "username": C.black_username}));
+                let f = $("<span>");
+                e.append(f);
+                ReactDOM.render((<Player user={{ "user_id": 0, "ranking": C.black_rank, "username": C.black_username }} rank nolink />), f[0]);
```
anoek: 👍

src/data/Rank.ts
```diff
@@ -0,0 +1,168 @@
...
+    rank = validate_rank(rank);
+    return rank.level + " " + rank.type;
+}
+
+export function rank_short_string(rank: Rank): string {
+    rank = validate_rank(rank);
+    return rank.level + rank.type[0].toLowerCase();
+}
+
+export function parse_rank(rank_string: string): Rank | void {
```
anoek: I don't think this will work like you desire when the string is coming in as a translated string, as would currently be the case with your usage of `parse_rank` https://github.com/BarneyStratford/online-go.com/blob/types/src/components/ChallengeModal/ChallengeModal.tsx#L545 https://github.com/BarneyStratford/online-go.com/blob/types/src/components/ChallengeModal/ChallengeModal.tsx#L744-L748

src/data/Rank.ts
```diff
@@ -0,0 +1,168 @@
...
+    let level: number = rank.level;
+    level = Math.floor(level);
+    level = Math.min(level, maximum_level[rank.type]);
+    if (level < 1) {
+        return adjust_type[rank.type](level);
+    }
+    else {
+        return {level: level, type: rank.type};
+    }
+}
```
anoek: I gotta be honest, the rank updates you have here seem like they've complicated things a lot.. it might just take some getting used to, so I'll try, but so far it seems like it's introduced a lot of code and complexity and I'm not sure I feel the gain yet.

src/lib/data.ts
```diff
@@ -62,112 +61,56 @@ export function set(key: string, value: any, dontTriggerListeners?:boolean): any
...
-            remove(key);
-        } catch (e) {
-            console.error(e);
-        }
+        remove(key);
     }
 }
-export function get(key: string, _default?: any): any {
+
+export function get<K extends keyof LocalData>(key: K, default_value?: LocalData[K]): LocalData[K] | undefined {
```
anoek: 👍

src/lib/preferences.ts
```diff
@@ -109,24 +104,17 @@ export function watchSelectedThemes(cb) {
...
-    dont_call_right_away = false;
-    let c = watch("goban-theme-white", call_cb);
-    return {
-        remove: () => {
-            a.remove();
-            b.remove();
-            c.remove();
-        }
-    };
+    subscribe.to(["goban-theme-board", "goban-theme-black"]);
```
anoek: Not sure where the bug is, but the goban-theme watcher functionality seems to be broken currently. You should be able to update the theme and see the effects immediately on any goban on the screen

src/views/Tournament/Tournament.tsx
```diff
@@ -1191,13 +1205,13 @@ export class Tournament extends React.PureComponent<TournamentProperties, any> {
                                     : <span>
                                         <select className="rank-selection" value={tournament.settings.lower_bar} onChange={this.setLowerBar}>
                                             {ranks.map((r, idx) => (
-                                                <option key={idx} value={r.rank}>{r.label}</option>
+                                                <option key={idx} value={rank_long_string(r)}>{rank_long_string(r)}</option>
```
anoek: Tournament creation currently does not work, `{"error":{"min_ranking":["A valid integer is required."],"max_ranking":["A valid integer is required."]}}` - related to these lines perhaps?

BarneyStratford:
It looks like the re-rendering in Chat is the same bug that’s causing https://github.com/online-go/online-go.com/issues/298.

What’s happening is that the player details next to each message contains only partial information about the player, and in particular sets the country to “un” regardless of the player’s actual country. There also seem to be some missing ui_classes there. When the Chat gets updated every 30 seconds, it updates the state of the main Chat view, which re-renders all of its children. This includes the (incorrect) Players beside each message, and the (correct) Players on the right-hand side. They end up fighting each other in the player_cache.

Previously, the player_cache’s Listener wasn’t very widely used, but the new version is much more proactive about pushing changed details to anyone that’s asked for them. That has caused the pre-existing bug to become more visible.

The new player_cache only re-publishes a Player’s details when they have actually changed - as, in this case, they keep doing.

The fix is a one-liner, which is to disable player_cache updates from the incorrect Player. I think I’ll make this a separate pull request against devel. I’m also going to change the new player_cache to increase its resilience in cases where the ui_class is missing altogether.

BarneyStratford:
The more I think about it, the more I think you're right about this PR having got too unfocussed. I really think I should split it up into chunks concerning single aspects of this PR. It'll be a lot less risky that way.

src/lib/player_cache.ts
```diff
@@ -15,186 +15,383 @@
...
+        // If the cached information has changed, then inform everyone who is
+        // subscribed to the player.
+        if (changed && !dont_overwrite) {
+            publisher.publish(cached.id.toString(), cached);
+        }
+
+        // If the player is new to us, then request to be kept informed
+        // of their online status.
+        if (!was_cached) {
+            connect_online(cached.id);
```
anoek: Mm, this probably isn't a good idea to do for every player we ever see, this can quickly grow to be quite a lot of players when we start thinking about things like viewing tournament results (sometimes hundreds of players), big games on the site (again hundreds of players), every person they've seen in any chat (currently connected or from lingering messages), every custom game that gets created on the play screen pops up a player, all that and coupled with the fact that a lot of people never close their browser, someone could very easily start getting online status updates for hundreds and even thousands of players.

It's this same reason why we have that annoying system in `fetch` to only actually do a call back to the server for more information if we actually need fields that we don't have, most of the interfaces (like the tournament interface, game chat records) will send along enough information about a player for display purposes, but not necessarily enough information for display elsewhere.. so by specifying which fields we need we can prevent having to resolve say 200 different player's on a tournament screen all at once or when a big chat record loads for a game, we have everything we need and can display without hundreds of requests back to the server.

PR #359, opened 2017-07-02
Fix the overlapping of the link_matcher and email_matcher regexes in …

BarneyStratford:
…the chat_markup function of PrivateChat.


anoek:
Awesome thanks :)

PR #372, opened 2017-07-12
Fix for #298 , which…

BarneyStratford:
… was caused by conflicting player details being added to the player_cache. I've prevented the incorrect details from being added, allowing the correct and current ones to stick around.


anoek:
Awesome thanks :)

anoek:
FYI if you include a line like

fixes #298 

in your commit, then when the commit is merged, it'll automatically close the issue :)

BarneyStratford:
OK, will do.

PR #374, opened 2017-07-12
Type error fixes

BarneyStratford:
The latest tsx typings show up a type error on the <select> elements. The select element has no "value" attribute; instead, its <option> children have a "selected" attribute.
See https://developer.mozilla.org/en/docs/Web/HTML/Element/select and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option for more info.

### DIFF
--- a/src/components/AutomatchSettings/AutomatchSettings.tsx
+++ b/src/components/AutomatchSettings/AutomatchSettings.tsx
@@ -266,9 +266,9 @@ export class AutomatchSettings extends Modal<AutomatchSettingsProperties, any> {
                                     <ConditionSelect value={settings.handicap.condition} onChange={this.setHandicapCondition} />
                                     {settings.handicap.condition === 'no-preference'
                                         ? <i>{tab === 'blitz' ? _("Default is disabled") : _("Default is enabled")} </i>
-                                        : <select value={settings.handicap.value} onChange={this.setHandicapValue} >
-                                            <option value='enabled'>{_("Handicaps enabled")}</option>
-                                            <option value='disabled'>{_("Handicaps disabled")}</option>
+                                        : <select onChange={this.setHandicapValue} >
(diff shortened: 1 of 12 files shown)


anoek:
Re the usage of `<select value={...}>`, that's a react thing not an html5 thing, https://facebook.github.io/react/docs/forms.html#the-select-tag - AFAIK there's no reason to prefer the much more verbose and error prone method of enumerating out `selected={...}` in all of the options?

BarneyStratford:
Oh, you’re right. I wonder why I was getting a whole string of type errors for it after I updated my npm modules. I updated by throwing away my node_modules and redoing npm install. It was also giving type errors on the Link components for target=“_blank”, but I found docs showing that that was supposed to work and didn’t fiddle with it. Would you mind trying the same update and seeing what you get?

anoek:
Interesting.. yeah if I npm install I see those too - doing a `yarn` to install things works fine though FYI. The main difference I think is npm install installs `tslint 5.5.0`, yarn installs `5.4.3` because that's what we have locked in currently... so yeah that'll have to be addressed at some point, but I can't find anything that would indicate that being an invalid attribute, so not sure what's going on, if it's actually something we need to change, or if it's maybe a bug elsewhere that will magically work itself out in the next few weeks :)

anoek:
I haven't confirmed that it's `tslint 5.5` yet that is the difference btw, just noticed that it did install a higher version with `npm install`, could be some other type defs were updated as well though.

BarneyStratford:
Now that I’m using Yarn, all the errors have magically disappeared. I think we can close the PR.

It does beg the question though, of what build instructions we make available. I like to periodically flatten my build and re-run from scratch to check that it still works. The instructions we’ve got at the moment say to run npm install. Perhaps this need updating? The Makefile uses Yarn.

anoek:
Yeah probably so, yarn is really nice in it's ability to keep everyone locked in together, we should probably just make that what we use officially

PR #376, opened 2017-07-13
Update the build instructions

BarneyStratford:
Update the build instructions to use Yarn for module installation instead of npm. Fix broken image.


anoek:
:+1:

PR #377, opened 2017-07-13
Introduce pubsub mechanism

BarneyStratford:
We have a Listener class in the data cache as well as the player cache. We replace both of these with a general pubsub mechanism that can be used wherever it is required.

Along the way, we removed the dont_trigger_listeners functionality from data.set, which was used almost not at all and seems like a nice pistol to shoot yourself in the foot with. We also removed the src/lib/config module, which contained little functionality and was used in only one other place.


anoek:
In #329 we'd talked about ditching `Listener` in favor of a standard `EventEmitter`, I think I'd still like to try that instead of using a home brew solution, for consistency if nothing else. Would you agree, or do you see there being a particular advantage of continuing to use a home brew solution?

src/lib/preferences.ts
```diff
@@ -69,15 +68,11 @@ for (let k in defaults) {
 
 
 export function get(key: string): any {
-    return data.ensureDefaultAndGet(`preferences.${key}`);
```
anoek: This has protected me more than once, I'd like to preserve the functionality. We can move the check here though and ditch the method in `data` if we want, I don't think we use it elsewhere.

BarneyStratford:
The reason I ended up making my own was that there is relatively little type-checking on the standard EventEmitters, and I couldn’t find an alternative library that did what I was aiming for. The homebrew pubsub module provides very strong guarantees about what types are allowed on which channels, and if you make a mistake that violates those guarantees, then the code won’t even compile, let alone get out to production. TypeScript’s generic types are a really powerful tool if we’re willing and able to use them!

Try making a deliberate mistake with the pubsub module, and see how loudly the type-checker shouts! Now imagine it was an accidental mistake instead and you haven’t noticed. Arguing with the type-checker can be frustrating at times, but the type-checker is always right and the code will be more reliable for it.

A second reason was to simplify code using the module. Client code doesn’t have to keep track of which channels it’s already subscribed to and add or change subscriptions as necessary; it can just say “these are the channels I’m interested in now”. If you really want to subscribe or unsubscribe to individual channels, then you can still do that easily by creating multiple publisher.Subscription objects, as I’ve done in several places.

anoek:
Hmm, I figured we'd just want to do a thin typing wrapper around `EventEmitter`, maybe something like:

```
class TypedEventEmitter<T> {
    private emitter = new EventEmitter();

    addListener<K extends keyof T>(event: K, listener: (arg: T[K]) => any): this {
        this.emitter.addListener(event, listener);
        return this;
    }
    on<K extends keyof T>(event: K, listener: (arg: T[K]) => any): this {
        this.emitter.on(event, listener);
        return this;
    }
    once<K extends keyof T>(event: K, listener: (arg: T[K]) => any): this {
        this.emitter.once(event, listener);
        return this;
    }
    removeListener<K extends keyof T>(event: K, listener: (arg: T[K]) => any): this {
        this.emitter.removeListener(event, listener);
        return this;
    }
    removeAllListeners<K extends keyof T>(event?: K): this {
        this.emitter.removeAllListeners(event);
        return this
    }
    listeners<K extends keyof T>(event: K): ((arg: T[K]) => any)[] {
        return this.emitter.listeners(event);
    }
    emit<K extends keyof T>(event: K, arg: T[K]): boolean {
        return this.emit(event, arg);
    }
}
```
?

There might be a cleaner fancier way of doing that too.. but just toying around, that seems to work pretty good.

```
class XY {
    x: number;
    y: number;
}

interface XYEvents {
    'update': XY;
    'xset': number;
}

class XYEmitter extends TypedEventEmitter<XYEvents> {
}

let xyemitter = new XYEmitter();
let alt = new TypedEventEmitter<XYEvents>();

xyemitter.on('update', (xy) => {
    xy.x = 4;
    xy.z = 4; // compile error, .z is undefined in XY
});

xyemitter.on('xset', (num) => {
    num.x = 4; // error, num is a number not an XY
});

xyemitter.emit('update', new XY());
xyemitter.emit('update', 3); // error, 3 is not an XY

xyemitter.on('invalid', () => { }); // error, no event 'invalid'

```

src/components/PlayerIcon/PlayerIcon.tsx
```diff
@@ -43,7 +43,7 @@ export function getPlayerIconURL(id, size): Promise<string> {{{
 
 export class PlayerIcon extends React.PureComponent<PlayerIconProps, {url}> {
     mounted: boolean = false;
-    listener;
```
anoek: Would we need a `this.subscribe.to()` in `componentWillUnmount` here? Seems like this is a leak that exists with the current listener implementation too.

anoek: sorry not *here* on this line, but in this component, wouldn't let me comment in that method since no changes were there

BarneyStratford:
Yes, I noticed that PlayerIcon thing too. It’s coming in the version 3 types PR that I’m making.

BarneyStratford:
My first attempt to do this was something fairly similar to what you’ve suggested, but it quickly became apparent that to make this work the way I wanted would complicate things. The shim was becoming considerably larger than the current pubsub module, and itself relied on a library that’s 3-4 times the size of the pubsub module with (in my opinion) less functionality. I hope you don’t mind if I try to persuade you that the current pubsub is a good way to go in spite of it being not not invented here. The aim has always been to produce an API that’s resilient against human error and easy to use, and I see that I haven’t made it clear how my proposed change achieves this.

As I alluded to in my earlier message, I wanted to make the subscription work based on what we’re interested in, not on how the things we’re interested in have changed. The EventEmitter.on and .off calls require you to detect changes yourself, and to do this in every location where you might wish to change a set of subscriptions. If you accidentally subscribe to the EventEmitter twice, then you’ll get two callbacks. In contrast, after subscribe.to(things), you’ll be subscribed to the things and only the things. Any subscriptions that are no longer required will go away, and any new ones will be added automatically. Doing it this way is more reliable, simpler, and more recoverable if a bug does ever occur.

An example of a place where this is particularly convenient is in the FriendIndicator component, where the updateFriendCount method currently has to work out which friends are new. In the new system (once online status has been added to the player cache) this can be reduced to  a one-liner - this.subscribe.to(this.friend_list); and we get unsubscribing from any lost friends for free. OK, online_status uses its own pubsub system rather than an EventEmitter, but the principle is the same.

Another benefit of doing it this way is that it allows us to separate the publishing and subscribing parts of the interface if we want to, thereby preventing us from being able to accidentally publish to the wrong publisher. The old Listener classes and watch functions achieved something similar, as there was no way to publish to a Listener outside of the file in which it was defined, and I really wanted to keep that protection. If you look in the new data cache, you’ll see that the Subscription is exported whereas the publisher is not. By contrast, an EventEmitter has its emit, on and off methods on the same object.

Finally, if there ever is a situation in which the EventEmitter’s style is preferable, then you can do that really easily with the new pubsub module.
ee.on = (event, callback) => new publisher.Subscription(callback).to([event])
ee.off approximately = (event, sub) => sub.to() // Use sub instead of the callback
ee.emit = (event, data) => publisher.publish(event, data);
In fact, I’d go as far as to say that you could create an EventEmitter shim over the new pubsub module and it would still be a great deal smaller and simpler than the original EventEmitter.

Because the other PR was getting too big and too unfocussed, I split it up to make the change less daunting, but that’s had the unfortunate consequence of presenting the changes out of their full context. I hope this makes it clearer what I’m aiming for and why it’ll help move the project along to a greater degree than the EventEmitter system can.

You’ll see that I’ve taken your other comments on board fully. The preferences.get function now checks for the presence of the default value, both in the type of its parameter and at runtime (in case you have to coerce the parameter to type any). And although I haven’t put it into a PR yet, the new-style Players will once again have numerical ranks. You were right - the new rank system was a bad idea.

anoek:
So first off let me say thanks for the discussion around this :) Please know that I'm always open to changing the way things work and going different directions, and when I raise questions or make comments like "why not X?", it's me assuming there was a reason to not do X, but it's not apparent to me, so I'd like to better understand. And lets be honest, it's not uncommon for us programmers to get excited about one solution and miss a simpler/better way of doing things, so if we ask each other these questions while we're doing a review and sometimes hit on 'Oh hm.. yeah X makes more sense huh', then there's more of a chance that together we'll hone in on the simpler / better solution and end up with a more manageable project in the end. (Mind you, it is *also* not uncommon for us programmers to be able to respond to that "why not X" problem with a very good reason that the reviewer didn't think of.)

I'm not necessarily arguing against a custom pub sub system we maintain ourselves, but I'm trying to understand the benefit of using this pub sub system everywhere over the widely used EventEmitter pattern that is so prevalent in the JS world.

There are definitely some things I like about your pubsub system, being partially type safe is one, and the `.to([channels..])` system I could see as being very handy in certain circumstances. I also like not having to specify the callback handler in `.to()` as you would with `.off(ev, cb)`, and both like and dislike specifying the callback only once on construction - yay to not having to repeat that 3 times for different events, but bleh to physically separating the event you want to act upon, and what you're going to do when you receive that event.. so we'll call that a wash, +1 for DRY, -1 for adding to the spaghetti code factor :) 

There are also some things that I see as adding friction for other developers working on the code base. The first of which is that people have to learn it as opposed to just knowing it (yeah i know, "big ask", but hey if everything else is equal, we should absolutely prefer commonality so people don't have to keep so much in their brain while they're working on stuff). The second point of friction is that users of any instance of the pubsub system need to be proactive to take advantage of the type system, for instance `new Subscription<"user">((channel, user) => user.xyz).to(["user"]);` will get you a nice compile error, but the easier to type lazy version `new Subscription((channel, user) => user.xyz).to(["user"]);` won't, but it'll still function if it's correct. The third point of friction is the definition of `.to([..])`, while potentially useful that it overwrites all previous subscriptions, that is equally potentially confusing for a user programmer as that's not traditionally how you think of a publish subscribe system, usually subscriptions are additive and only go away when you specifically unsubscribe. And lastly one point of probably minor concern I have with the pubsub system is performance of `.to` when you have a fair amount of subscriptions streaming in, namely a user programmer needs to know to buffer and manage any channel list they may be building before calling `.to([..])` else suffer `O(n^2)` performance.. which in practice probably won't happen a lot, that's one of those things where you test locally with like 50 "things" (whatever you're working with), it doesn't even hit the radar because it's so quick, but then you throw some code into production and now you have 500 things and your browser hangs. That might not be a practical concern, but it made me pause and think about how the friends list stuff works for instance.. I *think* that is currently getting a list of friends in one big blob from the server, so that's probably not a problem here, but there are several places where we stream updates to user lists and the like (game and lobby chat for instance) .. so if this were used in conjunction with those, we'd have to keep it in mind.

So to me the two systems are, for all intents and purposes, functionally equivalent, it comes down to which makes most sense given our use cases. Looking through this and the types branch, it doesn't look like the array capability in `.to([..])` came in that handy in practice (the friends list being the only real place where it did, though I feel like there are probably a lot of ways to achieve that functionality more gracefully than what is on `devel` without specifically building out a custom messaging system that replaces `EventeEmitter` usage everywhere.) . I do like the net effect of being able to just do `.to()` to get rid of any lingering event emitter bindings one has, that's a less error prone way of doing things. On type safety I imagine it can be improved, but as it stands right now just doing something like `TypedEventEmitter` would provide stricter type safety since it would be enforced at the emitter not where subscriptions are made.


All that being said, I haven't myself tried to use `TypedEventEmitter` to replace the `Listener`'s, so maybe that's not the right fit either, I'm just having a hard time justifying replacing one custom thing for another custom thing when it *seems* like they all do the same thing as your standard EventEmitter just with a different way of thinking about things (hence, "Why not a (Typed)EventEmitter?").

anoek:
Closing in favor of pubsub v2

PR #378, opened 2017-07-14
Added profile button to Navigation Bar

briank621:
Requested from:
https://forums.online-go.com/t/putting-profile-button-on-top-menu-bar/12711


anoek:
My gut says this link would be better under the little profile icon... but hard to visualize that being nice with what's currently under there so maybe it's fine on the main bar, though it may get reorganized later if we clean up the profile icon drop down at some point.

briank621:
One idea I have is clicking on the profile picture goes to the profile, while the drop down remains the same.

anoek:
Well, the other functionality there would then be a bit awkward - checking
notifications shouldn't ever change your page, and the quick theme change
stuff would also be out of place

mlopezviedma:
What about adding one icon? Profile pic links to profile, and on the right, a gear/bell icon drops down the notification and theme area.

anoek:
Yeah, the problem is primarily on mobile where there is very limited space up there. We'd have to see how it looks, I'm not opposed to trying though.

PR #385, opened 2017-07-23
Types version 4

BarneyStratford:
Introduce the basic type-checking and translation machinery to the data cache and the backend requests.


anoek:
It occurs to me that the `data` actually depends on the request type too, I suppose we need to account for that. For example, `PUT groups/%%/members` is `{player_id:number, is_admin?:boolean}`, `POST groups/%%/members` is `{player_id: number, delete?:boolean}`, and `GET groups/%%/members` are list filter/sort/limit parameters, and of course the result's vary between them all as well.

Second topic, how would you feel about making the id parameter optional? Maybe something like: 

```typescript
function validate_parameters<K extends keyof URLCommunication>(url:K, id_or_data:number | URLData[K], data:URLData[K]): {id:number, data:URLData[K]} {
    let id:number = 0;

    if (url.indexOf('%%') > 0) {
        if (arguments.length < 2) {
            throw new Error(`Missing ID parameter in request: ${url}`);
        }
        if (typeof(id_or_data) !== 'number') {
            throw new Error(`Invalid ID in request: ${url} ${JSON.stringify(id_or_data)}`);
        }
        id = id_or_data;
    } else {
        if (arguments.length === 2) {
            data = id_or_data;
            id = 0;
        } else if (typeof(id_or_data) !== 'number') {
            throw new Error(`Invalid ID in request: ${url} ${JSON.stringify(id_or_data)}`);
        } else {
            id = id_or_data;
        }
    }

    return {id:id, data:data};
}

export function get<K extends keyof URLCommunication>(url: K, id_or_data?: number | URLData[K], data?: URLData[K]): Promise<URLResult[K]> {
    let res = validate_parameters(url, id_or_data, data);
    return request("GET", url, res.id, res.data);
}
export function post<K extends keyof URLCommunication>(url: K, id_or_data: number | URLData[K], data?: URLData[K]): Promise<URLResult[K]> {
    let res = validate_parameters(url, id_or_data, data);
    return request("POST", url, res.id, res.data);
}
export function put<K extends keyof URLCommunication>(url: K, id_or_data: number | URLData[K], data?: URLData[K]): Promise<URLResult[K]> {
    let res = validate_parameters(url, id_or_data, data);
    return request("PUT", url, res.id, res.data);
}
export function patch<K extends keyof URLCommunication>(url: K, id_or_data: number | URLData[K], data?: URLData[K]): Promise<URLResult[K]> {
    let res = validate_parameters(url, id_or_data, data);
    return request("PATCH", url, res.id, res.data);
}
export function del<K extends keyof URLCommunication>(url: K, id_or_data: number | URLData[K], data?: URLData[K]): Promise<URLResult[K]> {
    let res = validate_parameters(url, id_or_data, data);
    return request("DELETE", url, res.id, res.data);
}
```

BarneyStratford:
You’re absolutely right on both counts. I’m on the case.

barney.

BarneyStratford:
This part of the work should now be complete. I'm not totally sure that it's worth explicitly detecting which parameters to requests.get, post, etc are missing (commit 7fb7b27), as it's much simpler to just pass undefined or 0 as the id when we don't actually need an id. But it's there if you want to use it.

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -143,7 +143,7 @@ export class ChallengeModal extends Modal<ChallengeModalProperties, any> {
             conf: {
                 mode: this.props.mode,
                 username: "",
-                bot_id: data.get("challenge.bot", 0),
```
anoek: I suspect this was an accident given the recent pattern removal of the `get("...", 0)` - or was this on purpose?

anoek:
Great stuff @BarneyStratford :) Had the one possible catch of to address and aside from that I think we're good to go for merge :)

src/components/ChallengeModal/ChallengeModal.tsx
anoek: eh, darn inline diff thing doesn't show what was replaced - this went from `data.get("challenge.bot", 0)` to `data.get("challenge.bot")`, ie the default of `0` was dropped and I suspect it's just because you were running through ripping out all of the request `get` calls and did this one by accident.

PR #388, opened 2017-07-25
Remove underused src/lib/config.ts module.

BarneyStratford:


anoek:
:+1: good riddance :)

PR #393, opened 2017-07-30
Hint button in Puzzle screen

DmitriyKirakosyan:
- Refactoring in Puzzle.tsx
- Add Hint button for showing one of the correct moves.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -22,12 +22,12 @@ let ts_sources = ['src/**/*.ts', 'src/**/*.tsx'];
 gulp.task('default', ['dev-server', "livereload-server", "background_webpack", "build_styl", "watch_styl", "watch_dist_js", "watch_html", "watch_tslint"]);
 gulp.task('watch_dist_js', () => { gulp.watch(['dist/*.js'], livereload.reload); });
 gulp.task('watch_html', () => { gulp.watch(['src/*.html'], livereload.reload); });
-gulp.task('watch_styl', () => { gulp.watch(['src/**/*.styl', 'src/*.styl'], ['build_styl']); }); 
+gulp.task('watch_styl', () => { gulp.watch(['src/**/*.styl', 'src/*.styl'], ['build_styl']); });
 gulp.task('build_styl', build_styl);
 gulp.task('min_styl', min_styl);
 gulp.task('livereload-server', () => { livereload.listen(35701); });
 gulp.task('background_webpack', background_webpack);
-gulp.task('watch_tslint', () => { gulp.watch([ts_sources], ['tslint']); }); 
+gulp.task('watch_tslint', () => { gulp.watch([ts_sources], ['tslint']); });
 gulp.task('dev-server', dev_server);
 gulp.task('tslint', lint);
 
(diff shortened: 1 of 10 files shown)


anoek:
This looks great! And nice work cleaning up / refactoring along the way, :+1: 

I'm wondering if maybe we should show all hint branches instead of a random one, what do you think? Would that be a notable undertaking, or is it simply just iterating through all the branches in showHint and doing the setCustomMark with highlight true?

Only other things that come to mind is maybe the hint button should toggle off the hints if pressed again? And maybe it should be disabled or hidden when the we're already highlighting moves?

DmitriyKirakosyan:
> I'm wondering if maybe we should show all hint branches instead of a random one, what do you think?

I thought this at first too but decided to show one only, random from those which can lead to the correct answer. Do you think it is worth showing all correct branches?

> Only other things that come to mind is maybe the hint button should toggle off the hints if pressed again?
Yes, why not. I will do it.

> And maybe it should be disabled or hidden when the we're already highlighting moves?
Hm.. There are some types of highlighting apart from hint highlight. It could be letters, it could be highlight with the same color (which will not conflict with hint highlight) and some shapes. Shell we disable hinting on squares with letters and shapes?

DmitriyKirakosyan:
I wasn't sure about this change: https://github.com/online-go/online-go.com/pull/393/commits/fb23e86706a5a42e4b66a14441a5f7c02ab98461#diff-055b04377c299d8e62fe406303171a7fR273

Could you verify if this replacement is fine?

anoek:
> I thought this at first too but decided to show one only, random from those which can lead to the correct answer. Do you think it is worth showing all correct branches?

I think it is, I think when you're using hint when there is more than one correct answer, showing them only one might give them the false impression that another correct path is incorrect, so for that reason I think we should show them all. Plus, if you've solved it one way, you might want to use hint to see if there was another way.

> > And maybe it should be disabled or hidden when the we're already highlighting moves?

> Hm.. There are some types of highlighting apart from hint highlight. It could be letters, it could be highlight with the same color (which will not conflict with hint highlight) and some shapes. Shell we disable hinting on squares with letters and shapes?

I wouldn't say disable them, although I suppose since we're using the same color it doesn't really show anything different when they're already highlighted.. for that reason maybe we should use a different color for the hint squares?

What I was referring to though is when `puzzle_player_move_mode` is `fixed`, the places where you can play are already highlighted, and I was thinking that if that's true maybe we shouldn't have the hint button. But now that I'm thinking about it again, that's probably wrong since those marks aren't necessarily the correct path, so having a hint button is still a good idea - but it should probably be a different color so you can tell which of the already highlighted options lead to correct solutions.


> I wasn't sure about this change: fb23e86#diff-055b04377c299d8e62fe406303171a7fR273
>
> Could you verify if this replacement is fine?

Yep that looks good :)

DmitriyKirakosyan:
Thanks, everything you said makes sense. I will do it and update PR soon.

DmitriyKirakosyan:
- [x] - Hint all correct branches
- [x] - Remove hints on button second click
- [x] - Set a different color for hints and `fixed` option

Issue #395, opened 2017-08-02
[beta] Cannot access Moderator Controls


mlopezviedma:
### Summary

At the beta site, clicking on **Moderator Controls** button on a profile page freezes the whole page and the mod controls modal is not visible.

### Environment

 * Operating system: Linux
 * Browser: Opera
 * Browser Version: 46

PR #396, opened 2017-08-02
Fix for #395.

BarneyStratford:
Apparently, abstract React components don't actually define a componentWillMount method.


anoek:
Awesome thanks for the quick fix :)

PR #400, opened 2017-08-03
Pubsub2

BarneyStratford:
Introduce the pubsub module to the player_cache. This allows us to subscribe and unsubscribe to a player in constant time.


src/lib/player_cache.ts
```diff
@@ -182,7 +206,7 @@ export function fetch(player_id: number, required_fields?: Array<string>): Promi
                         console.log("Batch requesting player info for", queue.map(e => e.player_id).join(','));
                     }
 
-                    get("/termination-api/players", {"ids": queue.map(e => e.player_id).join('.')})
+                    get("/termination-api/players/%%", queue.map(e => e.player_id).join(','))
```
anoek: Picked the wrong merge here, should be
```
get("/termination-api/players", {"ids": queue.map(e => e.player_id).join('.')})
```

src/lib/player_cache.ts
```diff
@@ -61,13 +104,7 @@ export function update(player: any, dont_overwrite?: boolean): PlayerCacheEntry
         return;
     }
 
-    let id:number = "user_id" in player ? player.user_id : player.id;
-
-    if (!id) {
```
anoek: Lets keep this check, it helped me not too long ago

src/components/PlayerIcon/PlayerIcon.tsx
```diff
@@ -27,48 +27,38 @@ interface PlayerIconProps {
...
-export function getPlayerIconURL(id, size): Promise<string> {{{
-    return new Promise((resolve, reject) => {
-        player_cache.fetch(id, ["icon"]).then((user) => {
-            resolve(icon_size_url(user.icon, size));
-        })
-        .catch(reject);
-    });
-}}}
+export function getPlayerIconURL(id: number, size: number): Promise<string> {
+    return player_cache.fetch(id, ["icon"]).then(user => icon_size_url(user.icon, size));
```
anoek: This does not do the same thing as the function you replaced: https://github.com/online-go/online-go.com/pull/400/files#diff-4a58653571e8b38e193938055200ebfcL34 and the types don't match up.. not sure why this compiles tbh

BarneyStratford:
It should be all ready to go now.

PR #401, opened 2017-08-03
Batcher

BarneyStratford:
There are a couple of places now where we batch up operations. Refactor these into a new Batcher class.


anoek:
Nice :)

BarneyStratford:
Thanks. Should be good to go now that I've merged it.

PR #403, opened 2017-08-06
Introduce machinery for validation of the contents of the data cache.

BarneyStratford:
This will be used to detect invalid data and prevent it from bringing the whole thing crashing down.


BarneyStratford:
I've thought of a better way to do this.

PR #408, opened 2017-08-10
a new interface - stones + board

seunin:
black
board texture seamless
white-1
white-2
white-3
white-4
white-5


anoek:
These look really nice :)

Currently we don't use images for any of the stones, we procedurally generate them on the fly. That doesn't mean we can't use these, but it does mean some code will need to be written in order to incorporate these. 

Because of this, can you open up an issue with these attachments (instead of a pull request), that way we can track the implementation details? Once that code is written, then we can make a pull request of that.

Also, do you have the originals for these? It could be interesting to use the images prior to the application of the light and shadow layers so that we can rotate those, then apply the light and shadow dynamically, giving more variation to the placed stones.

anoek:
Discussion moved to https://github.com/online-go/online-go.com/issues/411

PR #413, opened 2017-08-11
Typed players

BarneyStratford:
Introduce the new typed players throughout the frontend. This has led to major modifications to the player_cache and online_status modules as well as the Player, PlayerDetails and PlayerIcon components.

There are still a number of places to typify, including chat and games. I'll do those in a future PR to prevent this one from getting too enormous.


BarneyStratford:
This PR is now complete and ready for review and merging.

anoek:
It looks like there were several places that the `is_` to `is.` transition was missed so can't currently be merged. As I mentioned in my first comment on https://github.com/online-go/online-go.com/pull/329 I'd like to avoid changing the object structures like this anyways though. I'm all for typifying, but I am not in favor of changing the structure as that same structure is used all over the front end and both back end systems and it creates a lot of inconsistencies which I don't want.

The second blocking issue, which could certainly be worked around (see `lib/goban.ts` and my comment on https://github.com/online-go/online-go.com/pull/426 ), is that for all intents and purposes `lib/ogs-goban` needs to remain an isolated module since it is used on the backend as well. We can pull code from it, but not vise versa.

BarneyStratford:
I've restored the ogs_goban to its original state and hacked window["user"] to contain the "anonymous" attribute again. I've also fixed the merge.

I've had a look, and can't see any places where is_* attributes are still being used on the Player type. Can you point out where you can see them? There are a few places, mostly in User and ModerateUser, where is_* variables are used, but these are not using the Player type. And if anyone does ever use an is_* by mistake, then the type checker will catch them now.

Apart from the ui_class and the is_* attributes, I believe that the rest of the Player type closely follows the original schema, and I've made many changes to increase its conformance. I've also made the pro/professional, rank/ranking and icon/icon-url naming issues go away.

As for the ui_class itself, I really think that the new way is more reliable in the long term. Giving it a type of "string" doesn't capture what it really is - a set of flags - and doesn't prevent you from saying ui_class.indexOf("typo"). All I've really done is to refactor it into a true set of flags to enable the typechecker to catch mistakes in its usage. There are only a handful of places where the ui_class is referred to on the frontend, and they are all easily converted to use the new style. In fact, the ui_class field is now not used on the frontend at all except in the compatibility code.

Once we have converted the ui_class into flags, then it makes sense to do away with the duplicate flags on the main Player object. The is_*flags don't always exist; the ui_class doesn't always exist; and there's always a risk that a bug might cause them to become inconsistent with each other when they are both present. Much better to have a single source of truth that's always available to query.

If it would make you feel more comfortable, perhaps I could add getters for ui_class and the is_* fields to the Player type that construct their results from the is.* attributes? That might give us the best of both worlds? And once we're confident that no code is using those fields, then we can remove them at a later stage.

anoek:
Feel free to re-open this if you decide to pick this back up

PR #414, opened 2017-08-11
Introduce a debugging module

BarneyStratford:
… allowing us to enable per-module debugging as the need arises. Remove a bunch of assignments to global properties into the debugging module, where they belong.


BarneyStratford:
This PR is intended as a possible solution for something I've wished for on several occasions. It would be nice to have some means of making assertions and logging any that fail in production. I imagine logging to a machine that's separate from the main backend so that if there's a problem and it gets flooded then the main servers carry on quite happily. Would that be possible?

It would also be nice to collect the various debugging-related parts of OGS together in one place rather than having them scattered around.

Any thoughts?

anoek:
I like it

How would you feel about currying the module name parameter so we could just specify it once at the top of the file near the import?

Also we should probably include mappings for `.info`, `.error`, and `.warn` too

PR #426, opened 2017-08-26
Add "always-disable-analysis" preference

basti1302:
This new setting would disable the analysis mode (and conditional moves) in all of your games for you, no matter if the game was created with analysis enabled or not. By default, this setting is off, so everything works as before unless you change the setting for yourself.

This setting would only affect yourself, of course - that is, if the game has been created with analysis enabled, your opponent can still use the analysis feature, but you can't.

See <https://forums.online-go.com/t/option-to-always-disable-analysis-for-yourself/13408> for the motivation behind this setting. There has been some positive feedback (from two users) about the suggestion, not sure if that is enough or what the criteria for adding a new feature are?

Assuming for a moment, that this feature is considered useful, let's talk about the implementation: I'm not familiar with the online-go.com code base at all, this was the first time I did anything with it. So the changes could be all in the wrong place or whatever. If that is so, I'm happy to redo them from scratch or improve on the existing changes.

I encountered one weird thing: It seems GoEngine has an `disable_analysis` attribute, which it takes from the config object that is being passed around. Some places use `go_engine.config.disable_analysis` while others use `go_engine.disable_analysis`. This looks a bit like duplicated state, which is usually a bad thing. Also, I'm not entirely sure what the `config.original_disable_analysis` is for.

### DIFF
--- a/src/lib/goban.ts
+++ b/src/lib/goban.ts
@@ -85,13 +85,14 @@ export class Goban extends OGSGoban {
     }
 
     isAnalysisDisabled(perGameSettingAppliesToNonPlayers = false):boolean {
-        if (preferences.get("always-disable-analysis")) {
-            // Player's settings to always disable analysis overrides the per-game setting.
+        // The player's preference setting to always disable analysis overrides the per-game setting for
+        // their own games.
+        if (preferences.get("always-disable-analysis") && this.isParticipatingPlayer()) {
             return true;
         }
 
-        // If the user hasn't enabled the always-disable-analysis option we check the per-game
-        // setting.
+        // If the user hasn't enabled the always-disable-analysis option (or they do not participate in this game),
+        // we check the per-game setting.
         if (perGameSettingAppliesToNonPlayers) {
             // This is used for the SGF download which is disabled even for users that are not
             // participating in the game (or not signed in)
--- a/src/lib/goban.ts
+++ b/src/lib/goban.ts
@@ -84,9 +84,21 @@ export class Goban extends OGSGoban {
         return preferences.get("show-variation-move-numbers");
     }
 
-    isAnalysisDisabled():boolean {
-        // Player's settings to always disable analysis overrides the per-game setting.
-        return this.engine.config.disable_analysis || preferences.get("always-disable-analysis");
+    isAnalysisDisabled(perGameSettingAppliesToNonPlayers = false):boolean {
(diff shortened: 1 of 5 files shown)


anoek:
This looks good, unfortunately there's an annoying thing we have to dance around a bit before we can merge though. The `GoEngine.ts` source code is used by both the front end and the back end, so we can't import and use `preferences` there.

I *think* what we'll want to do is to add a method to `lib/goban.ts` that does does that `this.engine.config.disable_analysis || preferences.get("always-disable-analysis");` check and then replace the appropriate places elsewhere in the code where we look at that flag.

basti1302:
Wow, so online-go.com has a Node.js backend and shared code between client and server. How cool is that?

I changed the PR according to to your suggestions. I have a question about src/views/Game/Game.tsx, line 2286, though. This is for the SGF download. It is the only place that used this.goban.engine.config.original_disable_analysis before. As far as I can tell, the intention is that you can not download the SGF even when you are not one of the participating players, or, if you are not signed in at all. The analysis functions itself is enabled for non-participating players or people that are not signed in (by overriding the value in the game config object in lib/ogs-goban/GoEngine.ts, lines 170-174). 

With the current PR, this behaviour would be changed so that SGF download is enabled for non-participating players or people who are not signed in.

I wonder what the intention was here. More importantly: Is this a feature that we want to keep as it was before? If so, I'll change the PR once more so that the previous behaviour for SGF download stays unchanged.

basti1302:
Update: To speed up the process a bit, I went ahead and changed the SGF download so that it works exactly as before. SGF download is now disabled for non-players when the per-game setting for disabling analysis is turned on.

basti1302:
Hey Akita,

I can't reproduce the issue with the review button. In all situations I tested it is displayed as disabled and clicking on it has no effect. Can you still reproduce this? If yes, in which exact situation:
* when signed in/not signed in?
* in a game where the user participates (plays black or white) or when being a spectator?
* in a game which has analysis disabled or when the new setting is activated for the signed in user?

In other news, good catch about the preference taking effect when spectating. I changed this with my latest commit.

anoek:
Hm, yeah I can't reproduce it now too, not sure what I saw.

Looks good to me, thanks for the PR!

basti1302:
Thanks for giving feedback throughout the process. It was a pleasure.

PR #432, opened 2017-09-06
Fix the ChatPresenceIndicator.

### LINKED ISSUE #263: 'Player online' indicators in-game do not always work
This happens everytime I open a new game from the notifications list after someone accepts my challenge, but it is not the only time that this problem occurs.
When I open the game this way, both of the indicators show 'Offline', even my indicator. Both of the indicators stay 'Offline' the whole game unless the page is refreshed, even as the opponent alternates between 'Online' and 'Offline' on the friends list.
I would like the in-game indicators to be as reliable as the friends list. If that is too difficult, then, as fail-safes:
1) Make it so that the user's indicator always shows 'Online'.
2) Make it so that when the opponent moves, their online status is updated.

BarneyStratford:
The Game view would attempt to display a ChatPresenceIndicator before the engine is initialised, causing the indicator to receive an undefined user id. We fix this by performing more checks before rendering the indicator.
Fixes #263.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -2210,7 +2210,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
                       }
 
                       <div className="player-icon-clock-row">
-                          {(goban.engine.players[color] || null) &&
+                          {((engine.players[color] && engine.players[color].id) || null) &&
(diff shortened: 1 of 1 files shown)


anoek:
Good catch, thanks :)

Issue #434, opened 2017-09-09
Restrict rank range in Custom Game screen is broken.


IDex:
### Summary

The "restrict rank" setting found in Custom Game setup form is severely misaligned for new accounts (and reportedly for others). E.g. I have a new 7k account who is able to only restrict maximum/minimum rank from 7k to 25k, meaning I can only sandbag with any new account.

### Environment

 * Operating system: Linux 4.9.43-1-MANJARO \#1 SMP PREEMPT Sun Aug 13 20:28:47 UTC 2017 x86_64 GNU/Linux
 * Browser: Firefox 55.0.1 (64-bit)
 * Browser Version: 55.0.1 (64-bit)

### Steps To Reproduce:

1. Go to Custom Game creation on a new, but ranked account (5-10 games played).
2. Click Maximum Ranking dropdown.
3. See that the range of available rankings is wrong. 

### Additional Information:

Little poking around revealed that the issue seems to be related to the data.get('user').rating and data.get('user').ranking not matching data.get('user').ratings.overall. Ratings found in data.get('user').ratings.overall seem to match the real rating while data.get('user').rating is severely behind or possibly not updated at all.

PR #436, opened 2017-09-09
Fixed range of ranks in Restrict Rank, see issue #434

IDex:
This small change fixes the issue in #434. The change doesn't really address why user.ranking is wrong in the first place though.

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -23,7 +23,7 @@ import {_, pgettext, interpolate} from "translate";
 import {post, del} from "requests";
 import {Modal, openModal} from "Modal";
 import {termination_socket} from "sockets";
-import {longRankString, rankString, MaxRank, amateurRanks, allRanks, rankList} from "rank_utils";
+import {longRankString, rankString, MaxRank, amateurRanks, allRanks, rankList, rating_to_rank} from "rank_utils";
 import {errorLogger, errorAlerter, rulesText, dup, ignore} from "misc";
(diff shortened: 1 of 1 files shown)


BarneyStratford:
I'm not sure this is going to work as it stands, as professional players have ranks assigned to them rather than calculated from their rating. You're quite right though that the ranks that the backend returns to us are no longer being updated for amateur players, and hence data.get("user") can have the wrong rank.

This bug is actually already fixed upstream of here in the player_types PR that I've made.

IDex:
Nice catch. Given the bug makes custom games nearly unplayable for any new, non-DDK players it would be nice to see it get fixed asap. I'm wondering if augmenting this quick fix is worth it/wanted, or is your fix going to get merged soon?

BarneyStratford:
At the moment, I'm not sure if or when my PR will be accepted as there are some differences of opinion over it. If it is accepted, then it would be better to leave this part of the code as it is; otherwise, it's better to fix it here for now.

anoek:
Sorry for the delay in reviewing this.

There's a few other considerations beyond just looking at a raw overall.rating -> rank conversion, so the preferred way to get a player's rank is to use the `getUserRating` method from `rank_utils`. In this case, you'd want to work with the `bounded_rank` which deals with appropriately bounding the rank between `25k` and `9d`. This method also deals with all of the pro/non-pro stuff as well.

So in short, do something like `let rank = getUserRating(data.get('user')).bounded_rank` and work from there. 

Thanks for the catch and the patch!

anoek:
Awesome, thanks!

PR #441, opened 2017-09-16
Don't display PlayerDetails for guest players.

BarneyStratford:
We were displaying PlayerDetails for guest players. (This kind of bug could have been prevented by the type system.)


anoek:
Indeed. I need to spend some time devising some better testing strategies too with all this refactoring, looks like this bug was introduced way back in 3eb99c12eb13db151ad6e2e5ce13bf5f42b3147d and just slipped through the cracks, and unfortunately this has become a common occurrence.

BarneyStratford:
Eek! You’re right! What was I thinking?

PR #442, opened 2017-09-18
Introduce testing frameworks.

BarneyStratford:
As requested, I've made a start on adding some testing frameworks to OGS. @anoek, I suspect that you know much more than I do about this stuff. Am I taking this in broadly the right direction? Or do you know of a better way to do it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "license": "AGPL-3.0",
   "private": true,
   "scripts": {
-    "dev": "gulp"
+    "dev": "gulp",
+    "selenium": "node dist/selenium.js"
   },
   "devDependencies": {
     "autoprefixer": "^7.1.1",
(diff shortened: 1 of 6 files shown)


anoek:
*Effectively* testing user interfaces is something that still eludes me. Selenium is certainly a good piece of the puzzle to use I believe, as are having unit and functional tests, probably.

The trick with testing is to be able to have rigid enough tests so as to catch errors, but still keep it fluid enough so it doesn't bog down development and changes. If it starts taking significant amounts of time to update tests when you make some tweaks or add a small feature, then we start to not make those tweaks and changes which isn't good. On the other hand, if the test system is too lax so as to not get in the way, then chances are it won't be catching the bugs you need it to catch and you might as well not even have it. In the past I have found it difficult to find the sweet spot in between for UI code :( But I'm all for trying!

anoek:
Feel free to re-open if you want to pick this back up.

BarneyStratford:
> Feel free to re-open if you want to pick this back up.
Yes, I will do at some point. I’ve not gone away, I’m just not able to do any more for now due to other commitments. I’ll probably be back after Christmas.

See you,
Barney.

PR #445, opened 2017-09-19
Ensure that the player.rating.professional field is treated as a boolean in the PlayerDetails.

BarneyStratford:
Sometimes the backend uses an integer for the player.rating.professional field, and sometimes it uses a boolean. React displays integers literally, but treats false or null as a signal to not render the component. Hence, if the backend sends a zero to indicate that the player is not a professional, then that zero gets displayed literally.


anoek:
Thanks for the fix!

PR #447, opened 2017-09-22
Type enforcement for all usages of the REST API

anoek:
This is somewhat of an experiment to see how typifying all REST interactions might work and if the undertaking seems worth it. If we think it is, it'll turn into a real PR.

This implements the typing envisioned by @BarneyStratford


BarneyStratford:
Sorry, I’ve not had time to have a proper look at it. I’ve been crazy busy after a big job landed on my shoulders with another organisation that I volunteer for. It’s still on the list though.

PR #455, opened 2017-09-30
Fix for exception on initial display of new game. (https://github.com…

GreenAsJade:
…/online-go/online-go.com/issues/454)

I don't know if this is the "right" fix, since it is my first look at the codebase, but it "seems to work" :)

(I didn't find any testcases to run - are there some?)


GreenAsJade:
Hmm - this might be based on an out of date version of devel.  Please ignore.

Issue #454, opened 2017-09-30
Bug: When a new game is created, an exception is thrown trying to display the score that has not been initialised


GreenAsJade:
### Summary

When a new game is created, an exception is thrown trying to display the score that has not been initialised

### Environment

 * Operating system: 
 * Browser:
 * Browser Version:

### Steps To Reproduce:

1. In developer mode, start a game vs a bot (for ease), with pause on exception enabled
2. See the exception"

Game.tsx:2289 Uncaught TypeError: Cannot read property 'black' of undefined
    at Game.tsx:2289
    at Array.map (<anonymous>)
    at Game.frag_players (Game.tsx:2238)
    at Game.render (Game.tsx:1724)
    at react-dom.js:5262
    at measureLifeCyclePerf (react-dom.js:4541)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (react-dom.js:5261)
    at ReactCompositeComponentWrapper._renderValidatedComponent (react-dom.js:5288)
    at ReactCompositeComponentWrapper._updateRenderedComponent (react-dom.js:5212)
    at ReactCompositeComponentWrapper._performComponentUpdate (react-dom.js:5190)

### Additional Information:

I'm about to submit a pull request for the way I have fixed it locally. 

Since it's my first look at the code, it might not be the "right" solution.

PR #456, opened 2017-10-02
Fix exception on game creation - initialise state.score so the initia…

GreenAsJade:
…l score can be displayed.  (#454)

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -451,7 +451,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
 
         // We need an initial score for the first display rendering (which is not set in the constructor).
         // Best to get this from the engine, so we know we have the right structure...
-        this.state.score = this.goban.engine.computeScore(true);
+        this.setState({score: this.goban.engine.computeScore(true)});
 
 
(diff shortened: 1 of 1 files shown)


src/views/Game/Game.tsx
```diff
@@ -449,6 +449,11 @@ export class Game extends React.PureComponent<GameProperties, any> {
             this.goban.setMode("analyze");
         }
 
+        // We need an initial score for the first display rendering (which is not set in the constructor).
+        // Best to get this from the engine, so we know we have the right structure...
+        this.state.score = this.goban.engine.computeScore(true);
```
anoek: Nice find! There's a react-ism we need to account for, which is we can never set the state like this *except* for in the constructor. Everywhere else, we use `this.setState`, specifically in this case it would be `this.setState({score: this.goban.engine.computeScore(true)});`

GreenAsJade:
Hah - I should have known that!

Interestingly, I didn't get a warning message about that until just this morning - not sure whether I wasn't looking properly, or just the sequence of events.

Anyhow - fixed that aspect of things now...

anoek:
Excellent, thanks!

PR #465, opened 2017-10-04
Modal: Increased space b/w the buttons and the edges of the modal.

evoluteur:
Increased the space b/w the buttons and the sides of the modal.

Current & proposed looks:
31157707-9fe7f5d4-a871-11e7-8010-946cf5981503

For comparison, Facebook buttons are placed at the same distance from both edges of the modal:
facebook-buttons

same w/ Bootstrap:
screen shot 2017-10-03 at 8 22 31 pm

### DIFF
--- a/src/components/Modal/Modal.styl
+++ b/src/components/Modal/Modal.styl
@@ -85,7 +85,7 @@
 
     .buttons {
         flex: 0;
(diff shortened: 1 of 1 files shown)


anoek:
Great tweaks @evoluteur , thanks for the PRs!

PR #467, opened 2017-10-06
Pie-chart for win loss data

GreenAsJade:
I think it is nice to be able to see the win-loss numbers in a pie chart.

This PR implements that.

It also does two other things that I think are good:

 - It adjusts the colour scheme of wins and losses so that strong opponents are bright and weaker ones are darker.     It makes the win-loss bars tell a clear story of when you were playing vs stronger and vs weaker.  It also make the pie more intuitive.

 - It narrows the aspect ratio of the ranking graph.   This puts a bit more emphasis on vertical (rank) movement, which I like.

This PR also ~~sets the scene for the scene for~~ implements displaying total ranked games above the pie.


anoek:
Cool, I like it :) The placement might be a bit problematic for all layouts though, in particular on mobile it makes the graph even harder to see and a lot of the text is clipped. 

Maybe we can get @evoluteur to weigh in, he seems to have a good knack for where things can live and look and feel.

For reference here's a current before and after: 

without_pie

vs

with_pie

My current thoughts are to either a) move the pie chart underneath the graph when the screen size is below a certain width, say either 544px or 768px, or b) move the the chart into the right column above the trophies / activity. Another thought would be c) reimagine the Ratings breakdown into something simpler, clearing some space up there for the chart, that breakdown hasn't been very popular I don't think.

`b` or `c` would be a little more complicated than `a` because we'd have to have a separate component as opposed to just drawing everything in the same space, but it shouldn't be *that* bad to do.

Anywho, that's what comes to mind over here.

anoek:
Minor tweak thought too, the chart looks a little flat, would adding a stroke around the edge and between the slices make them pop out and look a little better I wonder?

GreenAsJade:
It was actually quite deliberate to put the pie next to the graph, to alter the graph's aspect ratio, and also have the pie legend serve as the legend for the win loss bars.

But obviously I have to check changes I make on mobile, which I hadn't done. 

I guess I'll have to learn to make different layouts on mobile - that will be good to learn.

GreenAsJade:
@anoek I pushed a new version that gets rid of the problem on mobile by not displaying the pie on the mobile :)

I'm sure a more creative solution might be possible, that puts the pie in a better place, but I'd kind of rather see this out there and be able to access it at least one way, rather than get held up trying to re-layout the other way.

The same about the look, with strokes etc.  I think that the plain look of it will motivate me to improve that, once it's out there, but it seems premature to beautify before hearing everyone's complaints about it :)

It also may be the case that a proper pie widget would be the way to go, if it considered at Good Thing, and needs to look better.

I also put in a solution for the legend text getting crammed up in mobile, by having less ticks at lower screen width.

anoek:
```
The same about the look, with strokes etc. I think that the plain look of it will motivate me to improve that, once it's out there, but it seems premature to beautify before hearing everyone's complaints about it :)
```

I'd like to polish it up before it's released, I *try* not to push things up before they're pretty much finished. If you're concerned about it, maybe take some screenshots and reach out to some folks on the forum or chat?

GreenAsJade:
Ah - that's fair.   One of the great things about OGS is that it always looks great :)

GreenAsJade:
Getting closer?

screenshot 2017-10-12 16 25 30 2

BHydden:
I like it :) and if nothing else it makes a natural looking place to display the total ranked games played haha

src/components/RatingsChart/RatingsChart.tsx
```diff
@@ -216,17 +222,39 @@ export class RatingsChart extends React.PureComponent<RatingsChartProperties, an
...
-        let self = this;
+    setRanges = () => {{{
         let sizes = this.chart_sizes();
-        let width = this.width = sizes.width;
+
+        this.width = sizes.width;
+        this.graph_width = 2.0 * sizes.width / 3.0;
+
+        if (this.width > 768) {  /* it gets too bunched up to show the pie */
+            this.show_pie = true;
```
anoek: Mutable state like this should be stored in `this.state`, and ideally this should be updated as the window resizes.

This applies to any variable `width` kinda stuff too, it should be computed on render or it should be updated and stored in `this.state`.

Doing this, while extending `React.PureComponent`, lets react intelligently decide when we need to update the component.

src/components/RatingsChart/RatingsChart.tsx
```diff
@@ -255,6 +283,15 @@ export class RatingsChart extends React.PureComponent<RatingsChartProperties, an
         this.rating_graph = this.svg.append('g')
             .attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
 
+        /* Win-loss pie chart goes to the right of the rating graph... */
```
anoek: Lets avoid trailing ellipsis in comments in general, they indicate omitted text, doesn't seem applicable.

src/components/RatingsChart/RatingsChart.tsx
```diff
@@ -613,7 +733,7 @@ export class RatingsChart extends React.PureComponent<RatingsChartProperties, an
             }
         }
 
-        /* Plot */
+        /* Plot graph.... */
```
anoek: And if you're going to use ellipsis, it's exactly 3 dots, not 4! :) (Note: I might be feeling a bit pedantic this morning.)

BTW I'm a big overuser of ellipsis myself, but I'm trying to get better. As a result, you get to get better with me!

src/components/RatingsChart/RatingsChart.tsx
```diff
@@ -802,23 +932,26 @@ export class RatingsChart extends React.PureComponent<RatingsChartProperties, an
...
                     ? <div className='loading'>{_("Loading")}</div>
                     : this.state.nodata
                         ? <div className='nodata'>{_("No rated games played yet")}</div>
-                        : <PersistentElement elt={this.chart_div}/>
+                        : <div className='ratings-graph'>
+                            <PersistentElement elt={this.chart_div}/>
+                        </div>
                 }
-                {this.renderWinLossNumbers()}
+                {this.show_pie ? "" : this.renderWinLossNumbersAsText()}
```
anoek: Should be `null` instead of `""` as putting in a `""` will actually create a text node

src/components/RatingsChart/RatingsChart.tsx
GreenAsJade: Does it help that the only time setRanges is called is on intitialisation and resize?

anoek: Well, it *shouldn't* be. There's what I consider to be a bug, which is resizing the window doesn't resize the chart nor recompute the pie chart. Updating the width's and hiding/revealing the pie chart, should all be updated in an `onResize` event ideally. I didn't bring that up though because that bug existed prior to your feature, so it felt unfair to require that be fixed as part of this PR :) 

Beyond that though, just to maintain consistency if nothing else, dynamic variables should always be tracked in `this.state` as that's where other contributors know where to look for stuff. That and we get to take advantage of `PureComponent` a lot like that :)

anoek: Eh, yeah never mind, looking at the code again it's just a funky component, most of that is my fault.

anoek:
Alrighty I'm ready to hit merge. +10 points to you though if you have any ideas as to why the resize code isn't working right :) 

UI update is going out on Tuesday, I'll merge before then unless you note otherwise.

GreenAsJade:
I guess it depends what you mean by "doesn't work".

If I gradually decrease the width of my browser, the chart stays the same width and gets scroll bars, rather than scaling.

This appears to be because in:

   chart_sizes() {{{
        let width = Math.max(chart_min_width, $(this.container).width()  - margin.left - margin.right);
       
$(this.container).width()   is constant under these circumstances.

Which seems to imply css at the higher level deciding that the container doesn't shrink?

I'll play around with it some more when I have a chance - might raise a "it would be nice if the chart scaled instead of got scroll bars" enhancement, to separate it from this.

anoek:
Yep you're absolutely right

Adding

```
.RatingsChart {
    max-width: calc(100vw - 2em);
}
```

seems to work, though only the graph part is resizing, the pie chart isn't changing it's starting x position

GreenAsJade:
Rats.  Can I cash in my 10 points, and you fix it?  :D :D 

(Just kidding, I'll take a look when I get a chance).

GreenAsJade:
> "Adding
> 
> .RatingsChart {
>     max-width: calc(100vw - 2em);
> }
> seems to work, though only the graph part is resizing, the pie chart isn't changing it's starting x position"

This _almost_ works, but the timescale of the win-loss bar and timescale selector doesn't resize either!

I'm not convinced scaling with screen size is the right answer anyhow - I have a feeling a fixed size and scroll is more practical.

But I will press on because either way various things need to all be lined up and working together.

GreenAsJade:
So - I've made the pie chart more resize friendly.

However, the win-loss bars don't appear to support resize properly - in fact they have a bug that reveals itself on resize (IE the shape of the win-loss bars changes dramatically on the first small increment of resizing).

Therefore, I think that this non-resizing PR with Pie Chart should be merged, and "how to make the graph and bars resize properly" should be a separate ticket.

anoek:
Hmm, go ahead and push up your changes that fix everything related to the pie chart, then I can base a fix for the graph bar stuff on that code base

GreenAsJade:
Hi Anoek, 

I think that edcbe00  make the pie chart good to go.

BHydden:
any idea when this is going live? :)

anoek:
Last I checked it wasn't resizing properly? I fixed up the chart stuff, so as soon as the pie stuff is fixed up I reckon we can click merge :)

GreenAsJade:
Oh - I had thought the last commit I did fixed the sizing: at least, for the manipulations I did it seemed OK.

Current status: I think it's OK, can you let me know what isn't right?

anoek:
So if you resize your window a bit, you'll start seeing stuff like this:

screen shot 2017-12-12 at 12 15 09-fullpage
screen shot 2017-12-12 at 12 15 31-fullpage

GreenAsJade:
Hmm - I'll need remind myself what was going on there, obviously.

The first thing I notice is that I said (further up):

> However, the win-loss bars don't appear to support resize properly - in fact they have a bug that reveals itself on resize (IE the shape of the win-loss bars changes dramatically on the first small increment of resizing).
> 
> Therefore, I think that this non-resizing PR with Pie Chart should be merged, and "how to make the graph and bars resize properly" should be a separate ticket.

After that I said that I "turned off resizing".

I don't properly recall whether it was turned off or on when I started all this :)

I'll go take another look, when I have a chance.

GreenAsJade:
I've started taking a look again.

My initial observation is that it doesn't seem like the screenshots you showed can be from the head of my branch, because at the head of my branch, the horizontal dynamic resizing is turned off.   This means that as you make the window narrower, you simply have to scroll to see the pie chart.

I find that behaviour quite acceptable - my suggestion was that we run with that.

The next thing I will do is the big merge of all the recent stuff and see where it gets to.

GreenAsJade:
Okaaaay... continuing the debugging narative ... merging my head into devel head does result in a resizing bug: devel head has resizing turned back on and apparently working, but the pie chart breaks it.

In this instance, I reckon just turning off resizing would be sufficient: arguably being able to scroll to see the graph over a sensible width is better than squishing it up :)

However, I have the feeling @anoek prefers resizing, so I will try to find what broke.

GreenAsJade:
Okaaay again!

I found the issue - interestingly the necessary piece of code was sitting there, commented out!

So now it seems, once again, to work for me :)  

I've resized like mad and can't break it.

(There is still a quirk with the ratings bars themselves, but that's tracked under a separate bug)

Maybe some Pie for Christmas? ;)

BHydden:
Still really wanting this to be implemented <3 I know you're busy @anoek so no rush... just wanted to ping the project after almost a month :)

arnaudgo:
Hello,

Great job ! I think it would be nice to display the time range even if there's some redundancy.

Maybe, the string "same period" is enough or maybe dates only on mousehover are enough.

Arnaud

GreenAsJade:
It's something that could be tweaked considered.   But not worth messing much more till we know whether this one is going in or not...

anoek:
It's going in, just been slammed with life and fires to put out. I want to
do some more testing with this on all the various devices and probably some
tweaking so I just need to sit down with it for a few hours.

GreenAsJade:
Cool!   I'm just conscious that just because it was coded, doesn't mean it makes it :)

And it still might not survive the "test of all devices" ...  I'd rather know what issues there are before polishing it further :)

anoek:
Ok sorry for the excessively long delay on merging this one. Looks good!

BHydden:
Woohoo! Big thanks to both of you for your hard work!

GreenAsJade:
Welcome back, anoek :)

PR #471, opened 2017-10-07
Different layout of Game Creation Card for improved intuition about what does what

GreenAsJade:
In this thread, some time ago, I suggested an alternative layout for the Game Creation Card:

https://forums.online-go.com/t/clearer-layout-of-game-finder-pane/12354

This PR implements that suggestion.

The rationale is explained in the thread.

As noted in the thread, the "computer" button is not a perfect fit for the layout logic, but close enough I think.

### DIFF
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -71,6 +71,7 @@
         justify-content: space-around;
         align-content: stretch
         width: 100%;
+        flex-wrap: wrap;
 
         button {
             font-size: 1.1rem;
@@ -79,7 +80,8 @@
             flex-basis: 13rem;
             height: 4rem;
             margin: 0.9rem;
-            min-width: 160px; // so none shrink less wider than the widest
+            min-width: 155px; // to stop shrinking and expanding differently to others
+            max-width: 155px;
 
             .fa, .ogs-turtle {
                 //font-size: 1.2rem;
@@ -123,6 +125,8 @@
             flex-basis: 13rem;
             height: 3rem;
             margin: 1rem;
+            min-width: 155px; // to stop shrinking and expanding differently to others
+            max-width: 155px;
 
             .fa {
             //font-size: 1.2rem;
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -79,6 +79,7 @@
             flex-basis: 13rem;
             height: 4rem;
             margin: 0.9rem;
+            min-width: 160px; // so none shrink less wider than the widest
 
             .fa, .ogs-turtle {
                 //font-size: 1.2rem;
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -167,9 +167,16 @@
         cursor: help;
     }
 
-    .custom-games-list-header {
-        padding-bottom: 0.5rem;
+    .custom-games-list-header-row {
+        display: table-row;
         font-weight: bold;
+        white-space: nowrap;
+
+        .cell {
+            padding-left: 0.5em;
+            display: table-cell;
+            padding-bottom: 0.5rem;
+        }
     }
 
     .challenge-row {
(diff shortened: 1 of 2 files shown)


anoek:
I think I like this better, one minor tweak though - can you move the settings gear icon to the left a la https://github.com/online-go/online-go.com/pull/463 ? @evoluteur's change there was good too I think :)

anoek:
Hmm, also though, as I'm recalling now, my main reason for having 'Correspondence' on it's own line is because that view on mobile looks messed up when it's aligned with the other buttons:

phone_layout

so I reckon we need to figure out how to make it look good on mobile again before we can merge

GreenAsJade:
Hah - there had to be a reason!

I'll have to think about that!

BHydden:
Out of curiosity, if all the options were stacked vertically, how much room would be needed?

GreenAsJade:
@BHydden FWIW, that's one thing I'd been considering: making the buttons "flatter" and "wider" in the mobile format, and having them all stacked like pancakes :)

BHydden:
In my head it looks good but haven't put it to paper so to speak so thought I'd throw the idea out there

anoek:
Hmm, looks a bit funky on smaller devices still =/

dev beta online-go com-8080-play iphone 5 1

GreenAsJade:
Ah man.  Obviously I didn't go that small :)

Back to the drawing board!

GreenAsJade:
Another attempt, trying to stick with what flex will do naturally.

anoek:
Alrighty I think we're getting closer

https://github.com/online-go/online-go.com/pull/463 still needs to be incorporated. The only other thing I noticed was the width of the correspondence button in spanish wasn't quite wide enough to properly center the text so it looked a little funky, and in some languages the icon is on top of the text for some buttons and on the left for others, so looks kinda wrong. Maybe we could make the buttons a bit wider to fix that? could probably just be a percentage or something?

GreenAsJade:
I did #463 and tweaked the button size up.

It looks OK to me (and I never noticed we had a language selector in the app before!)

screenshot 2017-10-23 18 13 14

GreenAsJade:
Oh - I think we have to accept that the turtle goes above the word "Correspondence".  I tried making it wide enough for the turtle to come down, and it gets silly.

I'm also trying to think of a better (shorter!) word than "correspondence", because we don't play by correspondence, but I can't think of one!

evoluteur:
What do you think of this change on your fork?

    .automatch-row {
        display: flex;
        justify-content: space-around;
        align-content: stretch
        width: 100%;
        flex-wrap: wrap;

        button {
            font-size: 1.1rem;
            font-weight: bold;
            flex-grow: 1;
            flex-basis: 13rem;
            height: 4rem;
            margin: 0.5rem;

            .fa {
                margin-right: 0.2rem;
            }

            .time-per-move {
                font-size: font-size-smaller;
            }
        }
    }

GreenAsJade:
@evoluteur I find it "quite acceptable" :D  

Pros:

- It's simpler css
- It puts the turtle on the same line as the word like the others

Cons:

 - The buttons have less space around them: there is lots more "blue".
 - It flips to vertically stacking the buttons rather sooner, as you reduce the width of the screen.

I would find either workable.   I actually don't mind the turtle above "correspondence" - it's mildly inconsistent, but not jarring.   I think my eyes accept that it's a long word, so it looks different anyhow :D

evoluteur:
I also played w/ different layouts. They all have pros and cons. It may become subjective at this point.

Another possibility would be to make the buttons panel larger than the scatter plot panel. It would give us more space.

GreenAsJade:
Yes - I think that any of the variants that we have, or something close, will be a good step forwards.

GreenAsJade:
It's great to have this implemented: this was actually the initial driver for me to do some coding :)

I'm interested to see that the text in the buttons was moved to "off centre, downwards" - I guess this is a matter of taste, but it feels a little "unbalanced" to me.   I understand that the result is that the "main text" is centred, and the note is below it.   Not entirely sure it's "best", but not fussed, very happy to see the layout adopted.  Just thought I'd mention it in case it warrants tweaking further...

PR #473, opened 2017-10-07
Debounce playing sounds on the volume control slider.

GreenAsJade:
I was motivated to do this when I saw Dwyrin go "Argh! I hate that!"  as he slid the volume control while playing on OGS in one of this videos, and it went "clickclickclickclickclick".

I thought "we can't have Dwyrin hating OGS!!"

Now it just goes "click" when you arrive at the target volume.


anoek:
Hah, yeah that was annoying. Thanks for the fix!

PR #474, opened 2017-10-08
"Fuzzy stone placement"

GreenAsJade:
In [this thread](https://forums.online-go.com/t/fuzzy-stone-placement/14199) there is a request for "fuzzy stone placement".

Here is an implementation, which seems to deliver an appearance much like the Sabaki that was offered as inspiration.

JokesAside had mentioned that it would be better to nudge the stones deep inside preRenderStone, for best graphical effect some how.   I can imagine that there is a graphics-guru reason that makes sense.

The problem is that you need to nudge the stone in the same way each time for each given location on the board, and down in preRenderStone you don't know what board location this stone is going on.

So I just nudged the stones in Goban.ts.   It looks OK to my undescriminating eyes :D 

I'd also be inclined to just make this the way it's always done, rather than a user option, if people think it looks good.  It's kind of subtle, so seems like it ought not to be offensive.

But the option is there to easily turn it off if needed.

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -2342,8 +2342,8 @@ export abstract class Goban extends TypedEventEmitter<Events> {
                         if (!this.stone_offsets[i][j]) {
                             let nudge_space = (NUDGE_PERCENT / 100) * this.theme_stone_radius / NUDGE_SCALE; // note that radius was already scaled
 
-                            x_nudge = nudge_space * (Math.random() > 0.5 ? -1 : 1);//(2 * (Math.random() - 0.5));
-                            y_nudge = nudge_space * (Math.random() > 0.5 ? -1 : 1);//(2 * (Math.random() - 0.5));
+                            x_nudge = nudge_space * (Math.random() > 0.5 ? -1 : 1); //(2 * (Math.random() - 0.5));
+                            y_nudge = nudge_space * (Math.random() > 0.5 ? -1 : 1); //(2 * (Math.random() - 0.5));
 
                             this.stone_offsets[i][j] = {x: x_nudge, y: y_nudge};
                         }
@@ -2354,7 +2354,6 @@ export abstract class Goban extends TypedEventEmitter<Events> {
 
                         nx = nx + Math.round(x_nudge);
                         ny = ny + Math.round(y_nudge);
-                        
                         /* Note that letters and marks on this stone will be displaced by this amount as well.... */
                     }
                     if (color === 1) {
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -36,7 +36,7 @@ const SCORE_ESTIMATION_TOLERANCE = 0.30;
 const AUTOSCORE_TRIALS = 1000;
 const AUTOSCORE_TOLERANCE = 0.30;
 
-const STONE_SPACING_SCALE = 0.976; // Factor by which stone size is reduced to leave a gap when touching
+const STONE_SPACING_SCALE = 0.98; // Factor by which stone size is reduced to leave a gap when touching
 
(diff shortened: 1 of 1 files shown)


GreenAsJade:
FWIW, I put in a separate branch an implementation that moves stones further and results in more touching and a "closer stones" feel.

Unfortunately, this doesn't work properly because the first thing __drawSquare() does is clear the square, which means that stones can't overlap into other stone's squares.    It looked "very hard" to me to get around that.

anoek:
Hah fun stuff.

So first off, this would definitely have to be an option, some people will absolutely love it and some people will absolutely hate it. I don't know what would be default, we'd put it to a poll and feel people out.

For reference for anyone following along, here's the change:

Original:
stones_original

With random placements:
stones_chrome

Some observations:

1) Things look much better on Firefox than on Chrome, the above pic is from Chrome, you can see lots of fuzzy edges and such, on firefox you don't get that:

stones_firefox

so we'd definitely want to figure out how to get chrome to place crisp stones.. this might be something simple like some flag/option we pass, or it might be a little more involved like rounding the offsets.

2) The offsets don't seem to be consistent, see the relative placement between `n9` and `o9` in the above firefox/chrome screenshots

3) I know exactly why you reduced the size of the stones, so we can "erase" them without any further changes to the code. However, people will most certainly complain about this, we used to have small stones like you have now and heard no end of complaint until we made the stones large enough to "touch eachother". Now, how to do random placement and have large stones, that starts to get trickier. But we might need to solve for it if we want to keep people happy.

4) The offset doesn't seem to scale with the board quite right:

image
image
image

as you can see the offset seems greatly more pronounced in the smaller board sizes, and not so pronounced as things get big.

5) There seems to be some clipping problems still at some resolutions, see: 
clipping_problem

6) The placement of the `last move` circle doesn't seem to be offset such that it's centered with the stone, I'm thinking it should be? Although I'm also thinking the other marks *should not* in general be offset.

And finally, since you're into this, another thought and consideration along this same vein is the fact that we are using squares here, traditional gobans are not square, see https://senseis.xmp.net/?EquipmentDimensions - which state that they should have a ratio of `22:23.7`, with white stones being 22.1 wide and black being "slightly larger", which since they are both wider than there is space on the board, necessitates the "imperfect" placement of the stones. 

I'll be honest, I'm not entirely sure the current rendering strategy of erasing squares and only drawing updates will really work well for what you're trying to do. However I'll also note that the reason why we do things the way we do is because 5 years ago we kinda needed to as things like webgl weren't widely available, so we had to do a lot of extra work to be fast enough. If we switched to a webgl renderer, we could probably just redraw the whole board every time, thus making it much easier to do all of this. Some experimentation would certainly need to happen though before we fully committed to that strategy.

GreenAsJade:
All of which goes to show that Jokes_Aside comment that this won't be done quickly was correct!

GreenAsJade:
Oh one important "discovery" - I think I mentioned in the thread - the reason why the stones have to be shrunk a little is not directly to do with erasing, it's to do with avoiding needing to move surrounding already-offset stones when you place a new offset one.

Even Sabaki has this gap.   It can be minimised, but not removed entirely, if you are allowed to placestones overlapping into the next square.   Sabaki has a gap enough such that the max offset only ever displaces one neighbor, so it never has to move more than one stone per axis per placement.

The link to the erasing strategy - we can't overlap stones into the next square.  That means we can only do smaller offsets unless we redo the whole drawing (!).

About the inconsistent rendering you mentioned - I think this is probably just the fact that the offsets are randomised every time the component mounts....

GreenAsJade:
I might be imagining it, but rounding the offsets seems to eliminate the added blurriness.

anoek:
Feel free to re-open this if you decide to pick this back up

arnaudgo:
Pinging @GreenAsJade I'm one of those who would enjoy this :+1: 

Arnaud

Issue #468, opened 2017-10-06
Feature request: hint for the date in the chat backlog


arnaudgo:
### Summary

Hello, in chatrooms, we have time hh:mm for every entry of the backlog but there's no hint for the day.
It would be nice to have that.

### Environment

 * Operating system: any
 * Browser:any
 * Browser Version: any

### Steps To Reproduce:

Ask if needed

### Additional Information:

Ask if needed

arnaudgo:
Hello @anoek and @MartijnTheunissen !

Does it also work for private chatlogs ?

If not, it would be nice to have it working for both.

If it is already working for both private and public chat logs, or for any other reason, this issue can be closed.

Arnaud

BHydden:
bump @anoek this can be closed

anoek:
Thanks :)

PR #475, opened 2017-10-08
Resolve issue #468 ; add date hover text in chat backlog

MartijnTheunissen:
Resolve issue #468. Hover text with the full date was added for timestamp in the chat.

### DIFF
--- a/src/components/Chat/Chat.styl
+++ b/src/components/Chat/Chat.styl
@@ -318,5 +318,11 @@ channel-width=12rem;
         text-overflow: ellipsis;
 
         white-space: normal;
+
+        .date {
+            font-size: 8pt; 
+            font-family: 'Source Code Pro', monospace;
+            color: #666;
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


arnaudgo:
Hello !

I'm happy you submited a MR for fixing the issue I opened :-)
Does it appear while hovering the mouse on the log entry?
I understand that and I feel a bit frustrated because most of the time, I visit OGS from my mobile.
But, I hope many desktop ogs users will enjoy your commit if it's accepted :-)

Arnaud

MartijnTheunissen:
Hi!

You are correct, it appears while hovering a mouse over the entry.
I'm sorry you feel frustrated, what would you have rather seen?

Another format? If so which one? I was afraid it would take up too much space.
Otherwise I could look into a hover effect for mobile users if you would like that.

arnaudgo:
I was thinking of a separation line every time a new starts (with the new date within it).

What do you think?

Arnaud

MartijnTheunissen:
That seems like an elegant solution, I'll have a look at that.

anoek:
Awesome, thanks for working on this! Yeah a date line separator would be awesome. We do something similar in the game chat page: https://github.com/online-go/online-go.com/blob/0f09c5e1d977c40df27809889abbe8f44dbb22af/src/views/Game/Game.tsx#L2980-L2986 if that's of any use to reference.

MartijnTheunissen:
separator

This is how it looks like now :)

anoek:
Awesome, looks great! I'm on the road for a few days, I'll do a full review and get this merged when I arrive, thanks for the PR!

PR #477, opened 2017-10-10
There's currently a slight display bug with the the "You have won

GreenAsJade:
There's currently a slight bug in the "you have won x games against ..." section:

screenshot 2017-10-10 21 09 20 2

This PR fixes that:

screenshot 2017-10-10 21 13 54 2


anoek:
Hah, whoops! Thanks for the fix!

PR #482, opened 2017-10-12
Settings: rearranged page.

evoluteur:
I think it is an improvement but it could be subjective...

Current:
settings-current

Proposed:
settings-new


anoek:
:+1:

PR #492, opened 2017-10-21
Search icon for OmniSearch

GreenAsJade:
... as requested: https://forums.online-go.com/t/how-do-i-search-for-a-player-name/12254

screenshot 2017-10-21 13 13 14 2

### DIFF
--- a/src/components/NavBar/NavBar.styl
+++ b/src/components/NavBar/NavBar.styl
@@ -228,6 +228,27 @@ navbar-width= 17rem;
             }
         }
 
+
+        .search-row {
+            display: flex;
+            align-items: flex-end;
+        }
+
+        .OmniSearch-input {
+            width: (navbar-width - 3rem);
+            margin-right: 1.0rem;
+            margin-top: 1.0rem;
+        }
+
+        .fa-search {
+            themed color shade1;
+            margin-left: 0.8rem;
+            margin-top: 1rem;
+            margin-right: 0.2rem;
+            padding-bottom: 0.35rem;
+
+        }
+
         .fa-check {
             themed color primary;
         }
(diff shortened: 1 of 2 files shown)


evoluteur:
What do you think about changing the word "Search" (not needed anymore now that the icon is there) for "Player name"?

GreenAsJade:
Well - here's thing thing, that I only discovered after working on the code: it's an OminBar.

It doesn't only search for players, it searches for groups and tournaments and ... goodness knows what else!!

evoluteur:
I didn't know, then "Search for players, group, and tournaments" will never fit ;-)

anoek:
I like it :) Should the icon be aligned with the other icons, or is it better unaligned because it's not a link do y'all think?

evoluteur:
I prefer not aligned w/ other icons. 

Personnaly, I like the search icon to be a button on the right of the search box (like Amazon). That way people can click the button to submit the search (even though most people will use the enter key anyway). Here every key stroke triggers the submit so it is not necessary.

It is not obvious that the search applies to different entities. Currently on the site, if I type "ab" I only get players but if I type "abc" I get players and groups. We may want to trigger the search at 2 letters for everything (searching for one letter add to the server work but is not meaningful).

Another nice thing would be to have the little "x" inside the search box to clear it. With the current design, as soon as I type something the menus disapear to show me the search results but I cannot click back to get the menu back, I have to highlight the search string and remove it. Clicking the "x" would be easier.

screen shot 2017-10-21 at 1 54 03 pm

GreenAsJade:
If you change the design of it, you probably should do the other ones as well (on the Profile page).

evoluteur:
Exact. I guess the change in design is an enhancement request. I do not know when I will find the time to do it but I would like to.

For now, adding the icon as you did is already an improvement.

anoek:
Cool, thanks for the improvement!

PR #496, opened 2017-10-21
Minor CSS tweaks

evoluteur:


src/views/docs/GoResources.styl
```diff
@@ -25,16 +25,14 @@
...
 
     dt {
-        text-align: left;
         font-size: 1.3em;
-        margin-top: 1em;
+        margin-bottom: 0.5rem;
     }
     dd {
-        text-align: left;
+        -webkit-margin-start: 1rem;
```
anoek: We should use `margin-inline-start: 1rem;` here instead of the prefixed variant, autoprefixer will take care of adding the webkit/moz prefixes for us :)

Issue #450, opened 2017-09-25
Long usernames obfuscate rank


crodgers:
### Summary

Usernames that are too long hide rank in certain contexts.

### Environment

All

### Steps To Reproduce:

Just see the included image

### Additional Information:

[Example](https://www.dropbox.com/s/3q2jiykshto0xaa/nick%20too%20long%20for%20challenges.png?dl=0)

I imagine this can be fixed by cutting off the overflow for the name container and moving the rank to a separate, juxtaposed, container.

PR #513, opened 2017-11-06
Dedicated column for rank in game list

nclavaud:
Fixes https://github.com/online-go/online-go.com/issues/450

Show the rank in a separate column when listing custom challenges.

### Before
screenshot from 2017-11-06 15-38-52

### After
screenshot from 2017-11-06 15-55-23


anoek:
Awesome, thanks for the patch!

Issue #509, opened 2017-11-03
long words in private chat create a scroll bar instead of splitting to several lines


Fuseques:
### Summary

when writing a very long word in a private chat, it won't split it to several lines but instead put it all in one line and create an ugly horizontal scroll-bar.

### Environment

*chrome

### Steps To Reproduce:

1. open a private message tab
2. write a very long word (something like "lalalalallalalallalalalalalalalalalalalalalalalalalalalalalal")
3. ta-da!

PR #521, opened 2017-11-09
Prevent long words overflow in private chat

nclavaud:
fixes https://github.com/online-go/online-go.com/issues/509

### Before
screenshot from 2017-11-09 21-26-55

### After
screenshot from 2017-11-09 21-27-22


anoek:
Indeed, thanks for the fix!

PR #535, opened 2017-11-30
Play: Moved "Settings" to the left + increased margin.

evoluteur:
Current:
settings-right

Proposed change:
settings-left

### DIFF
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -124,10 +124,9 @@
     .automatch-settings {
         font: 1rem;
         flex: 0;
-        text-align: right;
-        align-content: right;
-        width: 100%
-        margin-right: 0.5em;
+        text-align: left;
+        width: 100%;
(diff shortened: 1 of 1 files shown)


anoek:
Honestly it seems a bit out of place in both places.

What do you think about moving Settings to where the size selector is, and just bumping the size selector down, centered on it's own row right below?

BHydden:
That feels a little too "out of sight" to me... Personally I like its current position.

arnaudgo:
Hello,

I'm not 100% sure to understand what you describe @anoek or maybe yes ? Well, I would prefer see it in the same row of the size selector ... perhaps at the left side.

Arnaud

anoek:
That's true, might fit there and look good in general. On mobile we have less realestate than is what's depicted in those images, but we could overflow to the next line and center there if there wasn't enough room.

evoluteur:
Another try:
screen shot 2017-12-04 at 1 47 43 pm

PR #569, opened 2018-02-24
Add "quick chat" function to in-game chat box

stephancurran:
Add a toggle beside the chat input which displays a list of preset, but
user editable, messages (kept in localStorage) which are sent via game
chat with one click.

### DIFF
--- a/src/lib/ogs-goban/GoEngine.ts
+++ b/src/lib/ogs-goban/GoEngine.ts
@@ -1804,16 +1804,58 @@ export class GoEngine {
         return se.score();
     } /* }}} */
     public getMoveByLocation(x, y) { /* {{{ */
-        let m = this.cur_move;
-        while (m) {
-            if (m.x === x && m.y === y) {
-                return m;
-            } else {
-                m = m.parent;
+        let m = null;
+        let cur_move = this.cur_move;
+        let branch_point = this.cur_move.getBranchPoint();
+        let prev_move = null;
+        while (!m && cur_move !== branch_point.parent) {
+            m = this.findMoveInBranch(x, y, cur_move, prev_move);
+            prev_move = cur_move;
+            cur_move = cur_move.parent;
+        }
+        if (!m) {
+            m = this.findMoveInTrunk(x, y, branch_point);
+        }
+        return m;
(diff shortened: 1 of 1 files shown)


src/views/Game/Game.tsx
```diff
@@ -2612,6 +2677,67 @@ export class GameChat extends React.PureComponent<GameChatProperties, any> {
...
+
+    renderQC = (user) => {{{
+        let quick_chat: JSX.Element = null;
+
+        if (this.state.qc_visible) {
+            if (user.qc_phrases === undefined) {
+                let qc_local = localStorage.getItem("ogs.qc.messages");
+                if (qc_local === null) {
+                    let qc_messages: string[] = [
+                        "Hi",
```
anoek: Can we localize these too? (Should just be a matter of wrapping them like `_("Hi")`

anoek:
Nice, editable and everything :) I'm knee deep in some feature work at the moment, once I wrap it up I'll be going through all the PR's to merge/give feedback. This feature has been asked for for a long time, thanks for taking the time to implement it :)

src/views/Game/Game.tsx
anoek: Ah, yeah the reason why we need to do it explicitly for every string is because the string extractor looks for _("blah") calls and adds "blah" to the list of strings to be translated, otherwise those strings won't make it to the translators to be translated. If the string is already somewhere in our translation dictionary then what you had would work, but those are new strings so we gotta wrap em :)

stephancurran: Okay, gotcha. I'll commit the change now.

stephancurran:
Ummmm....as you can see, I committed and then reverted because it was for a different issue...should I create a separate fork for each issue I look at?

anoek:
Yep separate pull requests would be great, thanks!

stephancurran:
Damnit...I'm making a mess of this...I'm not used to Github at all...please bear with me.

stephancurran:
Okay, sorry for all the faffing about. I'm closing this, and have opened another clean one, plus the other pull request I've been trying to do. I promise to read a Github tutorial before next time :P

Issue #563, opened 2018-02-16
Review Tool: Quick Jump to Boardstate


Tr1bute:
In the current state of the **Review** mode, players are able to **Shift+Click** on a stone to go back to the board state when that stone was played. It would be really useful if you could also **Shift+Click** on an empty intersection to skip **forward** to the board position where a move on that spot was played.

My teacher uses this feature to a great degree, and I can't imagine how long a review would take if this wasn't possible for him. I truly believe that this will improve the teaching community on OGS a lot, if this feature were to be implemented.

stephancurran:
I'm working on a fix for this, but I need some advice on the exact behaviour required. At the moment, shift-clicking a point searches to the end of the current branch and it's sub-branches for the move, and then works backwards to the trunk (looking for a "nearby" match), and checks remaining branches of that trunk move before checking the trunk itself. It doesn't check any other branch moves because there could be any number of matches and no way of knowing which one the the user might require.

Does this seem like suitable behaviour, or should I use a different tack?

Tr1bute:
Great! This sounds right to me.

After playing around a bit on the CGOBAN tool, this is what I found.
In addition to searching the remaining branches of the current trunk, it will prioritize the current branch, then the parent branch, in-case there are moves that overlap each-other.

For example.

1. I'm explaining something in the trunk branch, making a new sub-branch.
2. I remove a stone that was previously played way before, and make up a new sequence that includes this move, but later.
2. (Currently in the sub-branch) I want to move back(or forward) to this place on this branch, so I shift-click on the stone, or the empty intersection where it is.
3. The system should now take me to the move on the sub-branch I'm currently in.
4. In the case where the system does not find a match in the sub-branch for that particular move, there seems to be three different behaviors on CGOBAN.
4.1 If the user clicked on a stone. (Trying to go backwards) It will take the user back to that move, and to the branch where that move was played, still prioritizing the current sub-branch.
4.2 If the user clicked on an empty intersection. (Trying to go forwards) It will not find the move if it is only found in the parent of the current branch, and instead requires that the user switches branches manually first.
4.3 If the user is currently in trunk, and clicked on an empty intersection, and the system did not find the move in the trunk branch, it will search the sub-branches of trunk for that move. If it was found, it will move the user to that board position and branch.

Essentially. 
Going **forward** can take you down to children of the branch, but not up to the parent.
Going **backward** can take you up to the parent, only if the move doesn't exist in the current sub-branch.

[...]

PR #575, opened 2018-03-04
Review Tool: Quick Jump to Boardstate #563

stephancurran:
Shift+clicking seeks a move played on that point. If the current move is
a sub branch, it will be explored before checking the main branching
point, followed by any other subbranches of that point. Then the trunk
after the main branching point is checked followed by the trunk before
it. If the current move is a trunk move, all subbranches of that move
will be searched before returning to the trunk as above.

### DIFF
--- a/src/lib/ogs-goban/GoEngine.ts
+++ b/src/lib/ogs-goban/GoEngine.ts
@@ -1804,16 +1804,58 @@ export class GoEngine {
         return se.score();
     } /* }}} */
     public getMoveByLocation(x, y) { /* {{{ */
-        let m = this.cur_move;
-        while (m) {
-            if (m.x === x && m.y === y) {
-                return m;
-            } else {
-                m = m.parent;
+        let m = null;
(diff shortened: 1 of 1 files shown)


anoek:
Intuitively it seems to me like shift clicking should always jump to an official move, then exploring variations should be easily done using the up/down keys to jump around between variations. It seems like one could get easily lost if shift clicking iterated through variations?

I'm happy to explore the reasoning here though, especially if there's a precedent established by some other editors that people like to use.

stephancurran:
You're probably right. I don't use this tool a lot, to be honest, so I don't know what the ideal behaviour is. What about adding an active branch  as Tr1bute mentioned in #575? First search that, and then go to the trunk, ignoring any sibling branches?

anoek:
Yeah indeed, that makes sense to me

stephancurran:
As this is also from 'unknown repository', I'll close it and do it properly from a branch.

anoek:
Indeed thanks :)

Issue #141, opened 2017-03-11
Can't sort puzzles


BarneyStratford:
### Summary

Bug report from Eidolonia.
On the old site, the Puzzles section was sortable (so e.g. I could arrange the collections by difficulty, or by number of puzzles). The new site doesn't seem to let me do that, at least in Chrome. I found this pretty useful. Could we have it back please?

### Environment

 * Operating system: iOS
 * Browser: Chrome
 * Browser Version: ?

### Steps To Reproduce:

1. Go to https://online-go.com/puzzles
2. Try to sort the puzzles

tnan123:
Also having this problem. On windows 10, I've tried both firefox and chrome with no luck :(

stephancurran:
This should be closed now.

PR #577, opened 2018-03-10
Make PaginatedTable optionally sortable

stephancurran:
Added optional props to PaginatedTable so that columns can be sorted ascending/descending as required. This option is used by PuzzleList in response to #141.


src/components/PaginatedTable/PaginatedTable.tsx
```diff
@@ -124,7 +126,7 @@ export class PaginatedTable extends React.Component<PaginatedTableProperties, an
             query[k] = filter[k];
         }
         //console.log(query);
-        let order_by = this.props.orderBy ? this.props.orderBy.concat(sorting || []) : sorting || [];
+        let order_by = this.state.orderBy.concat(sorting || []);
```
anoek: I'm seeing an error `Cannot read property 'concat' of undefined` places where `orderBy` isn't used, such as on the profile page

stephancurran: Just committed a fix for this.

stephancurran:
Opened as new pull request due to making a mess of this one.

PR #586, opened 2018-03-17
Add "quick chat" feature to in-game chat box

stephancurran:
Add a toggle beside the chat input which shows/hides a list of editable
messages which can be sent with one click.


anoek:
Cool, nice work :)

PR #587, opened 2018-03-17
Make PaginatedTable optionally sortable

stephancurran:
Add optional props to PaginatedTable so that columns can be sorted
ascending/descending. Used by PuzzleList in response to #141.


anoek:
Awesome thanks :)

PR #588, opened 2018-03-18
Review Tool: Quick Jump to Boardstate #563

stephancurran:
Shift+clicking seeks a move played on that point.

Current thinking is that the current active branch should be checked first, and if the move isn't found, look in the trunk. Don't worry if the move appears in some other subbranch. Any suggestions for better behaviour are welcome.


anoek:
Nicely done! And kudos for getting in that deep into the code and making a nice clean patch :)

Tr1bute:
Great work! Thank you so much. 😄

Issue #506, opened 2017-10-29
Feature Request: Games Screen should have a size column


BHydden:
### Summary

Games Screen should have a size column.

stephancurran:
This is a duplicate of #310, no?

PR #594, opened 2018-03-21
Add Size column to game screen #506

stephancurran:
Add a column displaying the board size to game lists in Observe Games, Overview, and Profile.

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -188,7 +188,8 @@ export class GameList extends React.PureComponent<GameListProps, any> {
                             white={game.white}
                             player={this.props.player}
                             gobanref={(goban) => game.goban = goban}
-                            size={game.height}
+                            width={game.width}
+                            height={game.height}
                             />)}
                 </div>
             );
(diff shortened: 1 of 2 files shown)


src/components/GameList/GameList.tsx
```diff
@@ -185,6 +188,7 @@ export class GameList extends React.PureComponent<GameListProps, any> {
                             white={game.white}
                             player={this.props.player}
                             gobanref={(goban) => game.goban = goban}
+                            size={game.height}
```
anoek: We should probably do width x height as we do allow for non square games, also that's what we do everywhere else so in the name of consistency if nothing else :)

stephancurran: :-o  I had no idea you could play non-square games. I did think it was weird that there was height and width, but I didn't think about it too much.

What about putting a single number for square games, and something like "17x12" otherwise?

anoek: Well i think for consistency we should keep with the NxN format regardless since it's what we do everywhere else (also I think it's a good UI practice as your eyes can scan for that easily enough.. having two formats for the same information might be a bit jarring I'd think)

stephancurran:
I've just pushed the above mentioned change. Size is now displayed as NxM. e,g. 19x19 or 17x13.

anoek:
Awesome thanks!

Issue #601, opened 2018-03-25
Be able to filter out challenges that you can't accept


GreenAsJade:
### Summary

Be able to filter out challenges that you can't accept

### Additional Information:

https://forums.online-go.com/t/show-only-can-accept-games-as-default-on-play-page/16259?u=greenasjade

GreenAsJade:
Proposed implementation in PR #602.

PR #602, opened 2018-03-25
Filter challenges #601

GreenAsJade:
(apologies some spurious formatting changes got included in the source code, my editor auto format went nuts)

screenshot 2018-03-25 14 58 18 2

screenshot 2018-03-25 15 02 51 2

### DIFF
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -195,16 +195,12 @@
         cursor: help;
     }
 
+    // Flex to escape from the table
     .custom-games-list-header-row {
-        display: table-row;
+        display: flex;
         font-weight: bold;
         white-space: nowrap;
-
-        .cell {
-            padding-left: 0.5em;
-            display: table-cell;
-            padding-bottom: 0.5rem;
-        }
+        padding-bottom: 0.5 rem;
     }
 
     .challenge-row {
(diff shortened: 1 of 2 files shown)


lemurek:
This PR is showing 4 commits being included. I'm not a regular user of github, so I'm not sure what went wrong, but is it possible that you've included a bit more than you wanted in the PR ? 

You might want to rebase and squash your filter-challenges branch on the HEAD of main repo before submitting, to make it clearer.

GreenAsJade:
I made it a little messy getting my fork up to date.

Maybe I should have done something like your suggestion to clean it up.  There's nothing else in there though.

mlopezviedma:
Great! Although I would put the filter at the top, so you don't have to scroll down to change it.

GreenAsJade:
It's at the bottom so it's mostly not visual clutter.

Since it's usually "unchecked", you don't have to scroll down to see it, because there's usually not much in the list (see first picture).

If you check it then the list grows, and true you need to scroll down to see it again, but I guess if you turned it on, then you wanted to scroll and look at all those anyhow?

I prefer it tidily out the way at the bottom.

anoek:
Nice :) I took the liberty of moving that div outside of the table (looks and works the same, it just prevents  word wrapping when there are challenges with long names, it was getting treated like a table cell and got squished with the accept column)

PR #606, opened 2018-03-28
Update layout of buttons on Game Page to avoid misclicks (and improve consistency)

GreenAsJade:
In this change, buttons that are not for actually manipulating the current game are now green instead of blue.   Blue buttons actually affect the game.

Also, they are moved to the bottom of that general area above the chat window, away from where the game-affecting buttons are (at the top of that area).

Unchanged: it's my turn, I have not placed a stone.  Here is the "Pass" button.  Note that it is blue, because it affects this game.

screenshot 2018-03-28 20 39 42 2

Now I placed a stone.  Also unchanged, here is the "Submit" button:

screenshot 2018-03-28 20 40 02

I submitted the move, then hit the arrow keys in disabled-analyze-mode.   This takes me back one move, so I have "back to game", but no analysis tree.   The "Back to Game" button is green and below the status.

screenshot 2018-03-28 20 40 14 2

I resigned, so now I have "Rematch".  Also green, and now centered, like the others:

screenshot 2018-03-28 20 40 37

At last the one we were all really interested in.  I started a new game with Analysis enabled, and went into Analysis mode:

Note that the buttons are both green, and they are significantly differently placed than the Back To Game button.

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -110,6 +110,19 @@ goban-view-bar-width=400px
             text-align: center;
         }
 
+        .analyze-mode-buttons {
+            display: flex;
+            justify-content: center;
+            margin: 5px;
+            button {
+                // this works in both light and dark theme
+                background: green;
+                border: darkgreen;
+            }
+            button:hover {
+                background: darkgreen;
+            }
+        }
         .pause-controls {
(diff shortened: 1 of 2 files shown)


anoek:
Awesome, good improvement :)

PR #608, opened 2018-03-29
Put the popup above the target element if it doesn't fit below - fixes #607

### LINKED ISSUE #607: Misclick protection on popups (and better general position)
### Summary

It's been reported that the player details popup can arrive in a place where there is an active button under the mouse.  This can result in a misclick on a mouse bounce.

It turns out that this same problem is associated with the popup obscuring the source of the popup (IE the popup goes over the player name link that it is associated with) which is not ideal.

### Steps To Reproduce:

1. Go to the chat page, in a group with lots of people in it
2. Click on a player name in the user list near the bottom
3. See that that popup obscures the link you clicked on, and likely your mouse is pointing direct at a button on the popup, liable for misclick.

### Additional Information:

Reported here: https://forums.online-go.com/t/a-ui-tweak-to-avoid-misclick-correspondence/16285/2?u=greenasjade

GreenAsJade:
As simple as the summary: put popups above a target element if they don't fit below as requested by config.bottom.

Before:

screenshot 2018-03-29 10 46 23

After:

screenshot 2018-03-29 10 25 14

### DIFF
--- a/src/lib/popover.ts
+++ b/src/lib/popover.ts
@@ -93,19 +93,29 @@ export function popover(config: PopoverConfig): PopOver {
     if (config.at) {
         x = config.at.x;
         y = config.at.y;
+
+        x = Math.min(x, bounds.x - minWidth);
+
+        if (y < bounds.y - minHeight) {
+            container.css({minWidth: minWidth, top: y, left: x});
+        } else {
+            container.css({minWidth: minWidth, bottom: $(window).height() - y, left: x});
+        }
     }
     else if (config.below) {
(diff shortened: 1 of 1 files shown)


anoek:
This is great :) Though, current bug is it doesn't work when you are scrolled down on a page. I think what you want is `document.body.scrollHeight` instead of `$(window).height()`

GreenAsJade:
Good catch.  There's something a little mysterious to me about "what is relative to what" - I probably should draw a picture with all the elements on it.

But this works :)

Issue #609, opened 2018-03-30
Feature Request: Esperantujo as user country


GreenAsJade:
### Summary

https://forums.online-go.com/t/add-another-flag-location/16316?u=greenasjade

PR #610, opened 2018-03-30
Let Esperanto speakers pretend they come from Esperantujo (implements #609)

GreenAsJade:
I put Esperantujo in as a way to get the Esperanto flag as a user country.

screenshot 2018-03-30 11 26 36 2


anoek:
Awesome, thanks!

Issue #611, opened 2018-03-30
Minor improvement: sort dropouts to the bottom of display list in tournament


GreenAsJade:
### Summary

It feels more just to put dropouts at the bottom of the ongoing results table :)

### Steps To Reproduce:

1. Go to https://online-go.com/tournament/34930
2. See disqualified people still "above" other people in the list

GreenAsJade:
Implemented by PR #612

PR #612, opened 2018-03-30
Sort players displayed in a tournament table so that dropouts are at the bottom (implements #611)

GreenAsJade:
screenshot 2018-03-30 12 23 57 2


anoek:
Cool, good improvement :)

Issue #615, opened 2018-04-02
Improvement: Less obstructive placement of "waiting for opponent"


GreenAsJade:
### Summary

Quite a regular request is to not have the "Waiting for opponent" modal obscure so much of the "Play" page.

Latest: https://forums.online-go.com/t/waiting-for-opponent-box-is-too-big/16358/1
### Steps To Reproduce:

1. Go to Play page
2. Click "Create" to start a custom live challenge.
3. See that the big modal obscures much of the information on the Play page.

GreenAsJade:
Implemented by PR #616

GreenAsJade:
Implemented by PR #620

PR #616, opened 2018-04-02
Improved placement for Waiting For Opponent (implements #615)

GreenAsJade:
This branch implements the "Waiting for opponent" dialog as a popover instead of a swal, allowing it to be placed directly over the automatch Card, where it obscures only that card.  Sending the message that those controls aren't really intended to be used while waiting for a live match, but not obscuring anything else.

It pulls in improve-popup-location branch because I needed to extend the popover functionality, and wanted to build on that change.

This branch also themes the double-bounce spinner.

screenshot 2018-04-02 21 54 32

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.styl
+++ b/src/components/ChallengeModal/ChallengeModal.styl
@@ -15,6 +15,13 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+.waiting-for-opponent-container {
+    themed background-color card-background-color;
+    themed color text-color;
+
+    text-align: center;
+}
+
 .ChallengeModal {
     width: 50rem;
     max-width: 100%;
@@ -48,8 +55,6 @@
     label {
         user-select: none;
     }
-
-
 
     .left-pane {
         margin-right: 0.5rem;
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -25,6 +25,7 @@ import {Modal, openModal} from "Modal";
 import {termination_socket} from "sockets";
 import {longRankString, rankString, getUserRating, MaxRank, amateurRanks, allRanks, rankList, bounded_rank} from "rank_utils";
 import {errorLogger, errorAlerter, rulesText, dup, ignore} from "misc";
+import {close_all_popovers, popover} from "popover";
 import {PlayerIcon} from "PlayerIcon";
 import {timeControlText, shortShortTimeControl, isLiveGame, TimeControlPicker} from "TimeControl";
 import * as preferences from "preferences";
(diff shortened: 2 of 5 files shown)


anoek:
The core improvement you've made is good, I think we just need to polish it up a bit. Right now I'm not sure how I feel about having a click anywhere on the screen simply cancel that dialog, it's a bit confusing to use I feel since the user clicks something, and instead of doing that thing they clicked on, it changes that view. (It also doesn't cancel the open challenge, it just changes the view).

That being said, the user should not be allowed to do things like accept other live challenges while the open challenge is going on, it should automatically cancel the open challenge before accepting the new challenge if we're going to allow the user to interact with the page while the open challenge is ongoing.

Also, navigation is a little tricky. The automatch system works entirely differently so the user can wander around the site while the automatch is looking for stuff.. unfortunately the open challenge system does not and is tied to watching the seek graph, so leaving the page for a minute will cancel any open 
 live challenge the user has ongoing... so because of that, I'm not sure the correct approach in terms of how to tell the user they are going to be abandoning their open challenge if they leave that page. (Obviously the best approach would be replumb that system, but that's a notable undertaking on the backend...)

GreenAsJade:
Looking at "whether this is modal or not, and why" made my head spin :)    

So I went for the smallest possible change to the existing flow that I could think of.

I felt that this was "no worse" than the existing, in terms of letting the person interact with the page while the challenges is waiting, because you can simply open another tab and do the same thing right now.

And in fact, people are doing that, so that they can see the Play page without the waiting dialog obscuring it.   (The other trick I've seen is giving the challenge a really long "notes" comment, to change the layout of that table, so they can see more).

I agree that "clicking on the page makes the automatch controls come back" is a bit odd - I wasn't delighted by that effect.  

I saw it as another example of "well, this is no different than opening another tab, and it's perfectly clear that the current challenge is still there", so I thought it probably liveable.

I'll have a think about "what could be done about clicking outside the popup".

However, in the light of the fact that the current solution is not really modal anyway, do we have to tackle this:

"The user should not be allowed to do things like accept other live challenges while the open challenge is going on, it should automatically cancel the open challenge before accepting the new challenge if we're going to allow the user to interact with the page while the open challenge is ongoing."

... they are already able to do that.

GreenAsJade:
Hmm - I don't really like how this turned out.   It's rather work-around-ish.

I think it should work like the automatch: control should return to the Play page after hitting "Create Challenge" in the "Create challenge modal".

If there are things you should not be able to do, then the state of the Play page should control that, so it works even if you have two tabs open.

I think the Play page should just do the before unload warning thing to tell them they are abandoning their challenge if they leave.

I might look into this refactor ... let me know if you think it's a bad idea so I don't go too far needlessly :)

Thanks!

anoek:
Closing in favor of #620

Issue #618, opened 2018-04-03
Improvement: delay pop-out of Dock


GreenAsJade:
### Summary

It's rather distracting to have the Dock on the Game page pop out every time the mouse strays onto it.

Actually, once you know what the icons are, you don't really even need it to pop out at all.

Therefore, let's delay its pop-out, so that if you pause on it you can see the items in full, but you don't have to do that to use it.

GreenAsJade:
Implemented by #619

GreenAsJade:
@anoek  FWIW, I'm finding 1s just a little to short to click an icon and get my mouse off the Dock before it starts popping out.

PR #619, opened 2018-04-03
Apply a delay to the pop-out-on-hover of the Dock (implements #618)

GreenAsJade:
As per #618, it is irritating having the Dock pop out when the mouse strays on to it.

This PR delays the opening of the Dock by 2 seconds, enough for someone to click an icon if they know what it is without the Dock opening, and not too long to have to wait if you want to confirm what the items are.


anoek:
Yeah that's better I think, :+1:

PR #620, opened 2018-04-04
Do away with waiting dialog for live custom challenges.

GreenAsJade:
@anoek  How about this approach?

It seems to work, simplistically testing. 

I'm not sure what gotchas there are.

I'm not sure what "active_check()" is doing in ChallengeModal - I left it there.

I can see that there is a case in ChallengeModal after the user cancels:

(this.props.mode === "open" ? "challenges/%%" : "me/challenges/%%"),

where I don't know what the false side is for.

That is to say, I cancel the challenge in Play.tsx the same way the "Remove" button does, which always uses challenges/%% not me/challenges/%%

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -432,7 +432,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                 notification_manager.event_emitter.on("notification", checkForReject);
 
                 if (open_now) {
-                    swal({
+                    /* swal({
                         title: _("Waiting for opponent"),
                         html: '<div class="spinner"><div class="double-bounce1"></div><div class="double-bounce2"></div></div>',
                         confirmButtonClass: "btn-danger",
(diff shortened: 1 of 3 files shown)


anoek:
That feels pretty nice, and I really like that it works across tabs too.

The `me/challenges/%%` is for direct challenges, the `challenges/%%` is for open challenges. So for example, if you challenge someone directly that dialog was still present, that happens on any page. 

Since this dialog is/was used for direct challenges as well, I think we need something to replace that (or just keep the dialog for direct challenges for now). Otherwise, when you're in chat and you challenge someone, there's no real feedback that the challenge has been sent.

FYI playing with this exposed a bug on the backend which I wasn't aware of, basically if you accept someone elses challenge, it will cancel your open challenge but it wont remove it from the live challenge list.. so folks get an error trying to accept or remove said challenge, whoops ;) .. so if you see that, don't worry about it, I'm pretty sure it's a backend fix.

GreenAsJade:
Ah - great to find out what the other use for the modal is.   (I realise I should have just grepped for that :) )

Keeping the dialog in that case sounds good.

Can you clarify what active_check() does - it's called for all open_now challenges.

I'm nervous whether it's logic will be right for this change...

GreenAsJade:
I put back the dialog for direct challenges - seems to work well.

I know that I need to test this more, especially what happens when someone clicks a different challenge while there is a live one pending.   I'll do that.

It will be also good to clear up whether active_check() logic is still OK.  With that sorted, this might be a thing...

GreenAsJade:
I just found the bug you  mentioned :)

It seems like if that bug is fixed, then the whole thing appears to hold up to "normal" kinds of activities that I know of...

anoek:
Nice work :)

PR #621, opened 2018-04-04
Themed colour of double bounce spinner, so it's visible in dark.

GreenAsJade:
I put this into a separate branch, so we can have it even if we don't move forwards with the waiting dialog fix for a while.


anoek:
:+1:

PR #623, opened 2018-04-05
Extended columns for games list

KrilYura:
I noticed that each Friday I want to know which games have the pause on weekends enabled.
Also, time to time I'm interested in some additional information like if the analysis is enabled or which games are ranked.
So I decided it will be useful for a user to be able to configure which columns should be shown on the overview page when games are represented as a table.

Note: at the moment it's more like proof of concept.

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -23,6 +23,7 @@ import {termination_socket} from "sockets";
 import {MiniGoban} from "MiniGoban";
 import {GobanLineSummary} from "GobanLineSummary";
 import * as data from "data";
+import {rulesText} from "misc";
 
 interface GameListProps {
     list: Array<any>;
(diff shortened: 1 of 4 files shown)


anoek:
This is pretty neat :)

I'm wondering if for this feature we want to maybe make a little gear settings popup box thing that is just attached to those tables, so people can click them and adjust what they see right on those pages. I feel like under the settings page it's not immediately clear what you're adjusting.

Counter point, most everything else is under settings to maybe it makes sense to keep it there, with maybe some clarification about what is being adjusted.. not sure.

I feel like maybe a list of checkboxes would be easier to use too, as opposed to a multi select list.

KrilYura:
Somehow I wasn't subscribed to notifications so I missed your answer...
I can create few mocks to consider if you don't have objections to this change.

anoek:
No objections to the core change, my only ask is to make it configurable from the table since I think that's the UI pattern that people are used to in general. Good stuff though I think :)

anoek:
Feel free to re-open this PR if you pick it back up

PR #624, opened 2018-04-06
Made dock delay configurable by preferences

GreenAsJade:
As per popular request.


GreenAsJade:
Yay for the extra "off"... the same thought occured to me this morning!  Sorry I forgot about _ !

Issue #627, opened 2018-04-08
Make it easier to hit the right challenge, without it moving out from under your mouse


GreenAsJade:
### Summary

People are reporting that it's easy to accept the wrong challenge because just as you go to click the button, the list updates and moves the one you wanted away.

GreenAsJade:
Implemented by #628

PR #628, opened 2018-04-08
Easy hit correct challenge (implements #627)

GreenAsJade:
As per the issue, it can be a hassle when the challenge list updates just as you click the accept.

This change freezes the list while they are moving their mouse over it.

I've left logging in, because it is tough to test/debug without it.

### DIFF
--- a/src/views/Play/Play.tsx
+++ b/src/views/Play/Play.tsx
@@ -37,7 +37,7 @@ import {bot_count} from "bots";
 import {SupporterGoals} from "SupporterGoals";
 import {boundedRankString} from "rank_utils";
 
-const CHALLENGE_LIST_FREEZE_PERIOD = 1000; // Freeze challenge list for 1s while they move their mouse on it
+const CHALLENGE_LIST_FREEZE_PERIOD = 1000; // Freeze challenge list for this period while they move their mouse on it
 
 interface PlayProperties {
 }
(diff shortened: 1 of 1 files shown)


anoek:
I like this, though it seems to me like this might make it so we might have stale challenges.. ie if you are about to hit accept, and someone else accepts first, your ui might not update so you're going to get a challenge does not exist error.

Do you think it'd be tough to make it so when a challenge is accepted and the list is frozen, the accept button becomes disabled or something like that?

GreenAsJade:
Hmmm - that's an interesting 'challenge' :D 

I think if it could be done it would be really elegant.   Darn, I was supposed to do some work today ;)

This being said, when I submitted this I had it in mind that it would be easier to get a "challenge does not exist" with this in place, and felt that this would be an acceptable improvement over clicking on the wrong one and having to back out.  I think this is a workable fallback position - maybe  I'd make the dialog less "alarming" - more "matter of fact" along the lines of "Oops, you were too late, someone else accepted that one". 

But ... I'll see what I can do disabling the button while freeze is active.

GreenAsJade:
Ooo - that turned out surprisingly tidy :)

I've left the logs in still.  I find that with CHALLENGE_LIST_FREEZE_PERIOD set to 10s then it's quite to able to see the effects and test.

anoek:
I definitely like where this is heading. My two suggestions are:

1. Right now you have the X second delay between updates, though I think the intent is to prevent the list from updating while the mouse is moving - this doesn't happen currently since you start moving and never reset the timer as the mouse moves, so basically the net effect is you just drop the update frames per second to 1/X FPS instead of locking the list. I think the desired behavior would be achieved by simply canceling and recreating that timeout whenever the mouse moves maybe? We might also want to bump that X delay to more than a second, not sure what the optimum is, but give folks a chance to skim the line they're about to click

2. When the mouse leaves the area, that list should go back to updating pretty quick I'd think (maybe after some small denounce delay to allow for movement slop), ie after a little bit of a debounce (maybe half a second?), cancel that pending timer and unfreeze the list

GreenAsJade:
Hah - you're right, I fixed up the timer.

Playing around, I found that 1 second is pretty right for everything.  1 second resuming updating after mouse leaving the list, and after mouse stops moving, feels OK.

PR #629, opened 2018-04-08
Fix height of win loss bars, don't show incomplete months.

GreenAsJade:
There are two changes in here.

First, the code was not agreeing whether the total height of the win-loss bar area was 60 or 65.  This causes small bars to be compressed, possibly out of view.

Second, it's hard-to-understand at best the fact that the partial months at the beginning and end of the graph are drawn as if the month extends off the graph.   The result of this algorithm is that it can look like one or two of the 4 possible bars are missing, because in fact they are off the graph.

We could either compress the bars into the partial month, or just not draw incomplete months.

I chose the latter here.

### DIFF
--- a/src/components/RatingsChart/RatingsChart.tsx
+++ b/src/components/RatingsChart/RatingsChart.tsx
@@ -786,7 +786,7 @@ export class RatingsChart extends React.Component<RatingsChartProperties, any> {
             let w = this.getUTCMonthWidth(d.ended) * alpha;
 
             // Don't plot bars for incomplete months at each end
-            let x = X(d,0);
+            let x = X(d, 0);
             w = isFinite(w) && x > 0 && (x + this.getUTCMonthWidth(d.ended) < this.graph_width)  ? w : 0;
             //console.log("W", w);
             return w;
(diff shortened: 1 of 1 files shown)


anoek:
Whoops, that's what I get for using hard coded numbers...

Re change two, I get what you're saying, though I'm trying to think if there's another way of representing the partial months. The thing about extending the drawing off the screen is it provides consistency as you scroll around and change your viewing window. Compressing seems like it might give you a weird stretching effect (maybe not that bad though? I haven't tried) and omitting partial months would have a popping effect as a full month came in to view it would seem. Also, if you are viewing only a few days of a month, it would be empty, which doesn't seem right?

GreenAsJade:
Actually, I'm pretty sure that the whole thing is only rendered once, and scrolling around and zooming happens on the rendered chart.   So the only months that are truncated are month when you joined and the current month.

GreenAsJade:
BUT I'm not totally sure I like the look of it better with those two months chopped out.

I was interested to see what you think about that, I should have submitted a screenshot, though I think you actually need to play with it a bit live to get a feel.

I just couldn't think of a better way to represent them, and it definitely generates questions as it is: the information displayed looks "wrong" until you know what is going on.

PR #635, opened 2018-04-13
Add semi-transparent, non-gray shadow to message box

cassidoo:
The gray shadow in the "Loading..." (and other messages) has bugged me, since it's lighter than the thing casting the shadow, and solid. This makes it a semi-transparent, dark shadow.
It's super minor and doesn't affect any functionality, but it's something that I've noticed for a while and I decided to finally do something about it. 😄


anoek:
I always thought it was kinda ugly too, so thanks for the fix!

PR #636, opened 2018-04-17
Reworked mobile style for cancel game button...

### LINKED ISSUE #625: Move Resign button in mobile view width
The resign button is directly below the chat box when in mobile view and I accidentally press it all the time. Thanks to the pop up I usually don’t accidentally resign but it has happened.

It should be moved.

Bloodyaugust:
... to be more difficult to click when trying to chat, and also to improve design language.

Fixes #625.

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -176,7 +176,7 @@ goban-view-bar-width=400px
         flex-grow: 1;
         flex-basis: 95%;
     }
-    
+
     .center-col {
         .play-controls {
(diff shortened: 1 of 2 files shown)


anoek:
:+1: Yeah that's better

anoek:
Hmm except this wasn't just a mobile fix, it looks great on mobile, not so great on desktop

anoek:
OK I updated things so that it uses your big red button for mobile, and keeps the original for the desktop :)

PR #637, opened 2018-04-17
Moved chat resolve to componentDidMount, so that in flight requests…

### LINKED ISSUE #592: Chat rooms not appearing under refresh
### Summary

When I log into OGS and click Chat, I don't see my group chats on left (even if I'm already in the room) until I refresh the browser.

### Environment

Windows, MSIE or Chrome I think behave the same?

### Steps To Reproduce:

Log in, click Chat, I only see: http://puu.sh/zMvgN/77bea9a274.png

I click refresh, then I see: http://puu.sh/zMviP/b51195ad02.png

Bloodyaugust:
…were not canceled by leaving overview (and to better be in line with other components initial data load). Fixes #592.

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -171,7 +171,6 @@ export class Chat extends React.Component<ChatProperties, any> {
             show_say_hi_placeholder: true,
         };
 
-        this.resolve();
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for that fix, it was bugging me a lot!

PR #639, opened 2018-04-17
Added a logout button to the right nav dropdown. Implements #638.

### LINKED ISSUE #638: Add Sign-Out Button to Signed-In Dropdown
### Summary

The design language around signing out is confusing: Sign in is in the top-right, but sign-out is nowhere to be found in the user drop-down that replaces it. Instead, you must open the left-hand menu. This is un-intuitive.

Bloodyaugust:
Closes #638.

### DIFF
--- a/src/components/NavBar/NavBar.styl
+++ b/src/components/NavBar/NavBar.styl
@@ -126,7 +126,6 @@ navbar-width= 17rem;
     .sign-in {
         padding-right: 1.0rem;
     }
-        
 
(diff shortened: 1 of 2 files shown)


anoek:
I'm open to some iterations around this, but just adding a big red button there doesn't make any sense for what else is in that menu I don't think.

anoek:
Feel free to re-open this PR if you pick it back up

PR #642, opened 2018-04-17
Re-worked math that calculates which page of a ladder to send the…

### LINKED ISSUE #562: Ladder sometimes displays wrong initial page
### Summary

When a user clicks on the link to the ladder, the normal behaviour is to arrive on a page of the ladder that has them on it.

However, if the user is near the "edge" of a page, the ladder may display the adjacent page, rather than the one they are on.

Further details : https://forums.online-go.com/t/wrong-page-while-looking-at-ladder-site-9x9/12050/5

### Environment

 * Operating system: 
 * Browser:
 * Browser Version:

### Steps To Reproduce:

1. Go to ...
2. Click ...
3. See that X wrong 

### Additional Information:

*Include any screen shots, links to games/reviews, wonky SGFs, or other information that is useful*

Bloodyaugust:
…player to so that it does not overshoot. Fixes #562.


anoek:
Thanks for the fix!

PR #644, opened 2018-04-19
Fix logic around play buttons in analysis disabled mode (fixes #643)

### LINKED ISSUE #643: Submit button showing when it should not be, on "showPrevious"
### Summary

As [reported in the forum](https://forums.online-go.com/t/submit-move-bug-in-disabled-analysis-mode/16607?u=greenasjade), if you place a move in analysis disabled mode, then click "back", the Submit button appears instead of the Pass button.  And if you click the Submit button you submit the move that you backed out of.

There is a related bug, which is that if you then go forwards you see your stone again, and have the Submit button, but you can no longer click anywhere else.  You have to go back to get rid of your stone, then you can place another.  I'm not raising a separate ticket for this because the fix for the first should fix this anyhow.

GreenAsJade:
There were two underlying problem here.

The first is easy: when you go back from placing a stone speculatively on the board, you should get the pass button back again, so setSubmit(null) needs to be called in showPrevious(), so that we tell the Game to not display the Submit button.

The second is not so easy: when you go forwards again, you would want to be able to Submit that stone.  But - goban does not update move_selected when you go from the LastOfficialMove to the next move past that.  And I'm not sure this would always be correct.   It's correct in disabledAnalysis mode, but maybe not others.   And goban doesn't know about that.  Can of worms.

Therefore, the second part of this fix is that if you go back from a speculative placement, you've lost it.  If you want it back, you have to place it again: you can't go forwards past LastOfficialMove with the showNext() in disabledAnalysis mode.   This seems solid, and not too bad experience.

### DIFF
--- a/src/lib/goban.ts
+++ b/src/lib/goban.ts
@@ -42,6 +42,21 @@ export class Goban extends OGSGoban {
         };
     }
 
+    showNext(dont_update_display?) {
+        /* This is a workaround for a bug (?) in ogs-goban where a showNext to the
+           lastOfficialMove plus 1 does not set move_selected.  But at least in
+           analysis disabled mode, you would expect to be able to submit that move
+           Easiest thing is to just not let them "next" forwards to a previously submitted
+           speculative move that they went back from.  The (?) is because ogs-goban doesn't
+           know about isAnalysisDisabled, so how does it know whether to set move_selected?
+           Meanwhile, sidestep the issue :)
+         */
+        if (this.isAnalysisDisabled() && this.isLastOfficialMove()) {
+            return;
+        }
+        super.showNext(dont_update_display);
+    }
+
     getClockDrift() {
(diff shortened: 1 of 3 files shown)


GreenAsJade:
I put the fix into goban.ts to avoid the exposed previously private member.

Having played around with this some more, and seeing the code in goban itself, I'm not actually sure it's a workaround: maybe it's the actual solution.   It feels solid to me: in disabled analysis mode, it seems fine that after you back-arrow away from a speculative placement, the furthest forwards you can go is LastOfficialMove.   At first I thought ideally you'd go back to your placed stone, but after using it some more in testing, I don't find myself "wishing it did".

anoek:
I ended up fixing this a different way in cb72f1dc4a968ff7548f80192c460 , which just fixes that show_submit boolean flag to resolve the issue

PR #646, opened 2018-04-28
Make it so easy-hit-challenge works when unavailable challenges are f…

GreenAsJade:
…iltered.

Yeah.  Um oops  :%)  .   I had this feeling as I piggy backed on "eligible", that a separate flag might be a better idea...


anoek:
Cool thanks for the fix :) I guess we both need to get better about testing :(

GreenAsJade:
Yeah - I thought I had tested like mad.   Sometimes you just don't think of stuff eh...

PR #650, opened 2018-04-30
Play a dingdingding sound when an automatch starts

GreenAsJade:


GreenAsJade:
screenshot 2018-05-01 09 37 55 2

anoek:
Thanks for this :) I'm "out of the office" for just a bit longer then I'll get this merged. IIRC we'll need to convert the mp3 to a few other formats too (wav and ogg if memory serves me correctly, though I haven't checked browser capabilities in a long time to see if firefox can do mp3's yet or not), I can take care of that though when I'm sitting down again.

GreenAsJade:
Yay :)

I looked into the wav format question, and couldn't really see why this would be the case.  I'd be keen to learn what is the driver for that: as far as I could see all browsers are supposed to support mp3.

PR #654, opened 2018-05-18
Added common links to right navbar with styling - Adds logout to new section

IFOV37:
Addresses #638 and adds more common links to the right navbar.  This could go through some iterations, but I felt it continues the theme of the site while reusing some elements already in place.

I don't want to step on anyone's toes is someone is already working on this.  I am a software dev, but this is my first contribution to an open source project.  My apologies if something was done wrong. :)

### DIFF
--- a/src/components/NavBar/NavBar.styl
+++ b/src/components/NavBar/NavBar.styl
@@ -185,6 +185,36 @@ navbar-width= 17rem;
         }
         .center {
         }
+
+        .quick-links {
+            margin: auto;
+            width: 65%;
+
+            li {
+                list-style-type: none;
+                list-style: none;
+
+                a, span.fakelink {
+                    display: block;
+                    line-height: 2.0rem;
+                    width: 100%;
+                }
+            }
+
+            i {
+                themed color shade1;
+                width: 1.5rem;
+                text-align: center;
+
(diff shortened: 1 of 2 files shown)


IFOV37:
After thinking about this some more, the term "Common Links" would probably be better than "Quick Links."

anoek:
Feel free to continue iterating but at the moment I'm not a fan of just having a just a list of links there, it doesn't really fit with what else is there.

IFOV37:
Do you have any preferences or ideas?  Links, nor buttons really fit, so I'm not sure what might fit there. Shifting the profile drop down over or sliding a "sign off" link in that area where "sign in" was doesn't make sense in terms of usability.  It would be too close to the move notification or the profile/game based notifications, thus mis-clicks causing users to log out.  Hmm.  I need to think more about this.  Might look at some other issues while trying to come up with something that fits.

anoek:
```
Do you have any preferences or ideas?
```

I do not, there's a discussion thread here: https://github.com/online-go/online-go.com/issues/638 but it seems kind of dead, seems like a hard ui/ux problem to solve.

anoek:
Feel free to re-open if work resumes on this

PR #655, opened 2018-05-21
Add a setting to show/hide the tournament indicator

drichard:
## Add a general setting to show/hide the tournament indicator in the nav bar.

I never play tournaments and the tournament countdown in the top right gives me a little anxiety every time I look at it.

image

What about a setting to toggle the indicator on/off? It would look like this:

image

By default the tournament indicator is shown, but you can disable it if you really want to.


drichard:
@anoek What you do think? Useful feature or too many settings?

anoek:
I like it! Sorry I've been a bit awol recently with a home construction project, I'll be back in the saddle soon and will review / merge. Thanks for the PR!

anoek:
Thanks again!

Issue #670, opened 2018-07-08
Option for static window title


GreenAsJade:
### Summary

In [this thread](https://forums.online-go.com/t/name-of-windows/17131) a user requests to be able to have the web page title static, to support a streaming app.

GreenAsJade:
Submitted implementation: #671

PR #671, opened 2018-07-08
Added option to keep web page title static (for streamers) (implements #670)

GreenAsJade:
screenshot 2018-07-08 11 15 55

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -35,6 +35,7 @@ let defaults = {
     "dock-delay": 0, // seconds.
     "double-click-submit-correspondence": false,
     "double-click-submit-live": false,
+    "dynamic-title": true,
     "game-list-threshold": 10,
     "goban-theme-black": null,
     "goban-theme-board": null,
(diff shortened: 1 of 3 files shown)


Spoekenkieker:
IMHO it would _still_ be good to have some server info in the page title, maybe something like …

> Go game on OGS: Hippopotamus vs. Crocodile

?

GreenAsJade:
I pondered that.  I had in mind:

   OGS: Hippopotamus vs. Crocodile

But look at this:

screenshot 2018-07-10 07 39 51

... if there were redundant characters before the playernames, the player names would no longer be visible.

Also, the OGS icon tells you that this tab is an OGS game ... so it seemed to be overall not the right thing to do.

benjaminpjones:
Came here from #1392 I didn't even realize this was an option!

This is actually preferred behavior for me, since I have many observed games going at once during club.

GreenAsJade:
Yay ... at least _someone_ will use this then!   😁

BHydden:
> at least _someone_ will use this then!

I've had it permanently on since it was implemented...

PR #681, opened 2018-08-10
Adds left padding to items in navbar's drawer

konaraddi:
Adds left padding to items in navbar's drawer. Please see the red and green bars in the image below to see the change in padding.

add_padding


anoek:
Indeed :)

Issue #690, opened 2018-09-29
Better label for Score Estimator


GreenAsJade:
### Summary

Better label for Score Estimator, to correctly convey what it does and what it should be used for

 

### Additional Information:

As suggested in the forum:

https://forums.online-go.com/t/score-estimator-fix-it-please/18288/9?u=greenasjade

GreenAsJade:
Implemented by #691

anoek:
Closing because I want the score estimator to be better and for the label to be accurate. This is becoming more of a priority I think

PR #691, opened 2018-09-29
Changed the label of the "Score Estimator" to say "Guess score" (implements #690)

GreenAsJade:
As per issue ticket, we might save some angst by relabeling the Score Estimator.

screenshot 2018-09-29 10 55 11


anoek:
I'd rather fix the score estimator to not suck :) I'm gonna not merge this because it'll cause translation churn ... and because once the score estimator is better, we'd have to revert! :)

PR #715, opened 2018-11-28
Team member update plus incident report layout tweaks

GreenAsJade:
Just call me Eugene :)

Plus:
 - Tweak layout so it doesn't change when you click buttons
 - Add the content of the Note as the default value when editing it


anoek:
Awesome thanks!

Issue #716, opened 2018-11-30
Warn people about time settings used for cheating


GreenAsJade:
### Summary

People cheat all the time with various time settings, most notably Absolute, but also ridiculously fast Blitz.    If we warn people before they accept these, then hopefully less people get sucked in.

PR #717, opened 2018-11-30
Added basic warning for the time settings used by cheats

GreenAsJade:
Implements #716

### DIFF
--- a/src/components/GameAcceptModal/GameAcceptModal.styl
+++ b/src/components/GameAcceptModal/GameAcceptModal.styl
@@ -30,4 +30,7 @@
         padding: 1rem;
     }
 
+    .cheat-warning {
+        color: chocolate;  /* mmm chocolate! */
+    }
 }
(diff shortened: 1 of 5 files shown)


anoek:
I'm certainly open to this idea but I'm not so sure *all* absolute games are cheater games, nor *all* games with minimal main time are either.. for instance Canadian clocks could certainly have a low main time and still have reasonable time settings.

GreenAsJade:
I think it's important to note that this in no way asserts than any given challenge _is_ a cheater game.  It just alerts people that some _might be_ used that way against them.

The other thing to note is that I see this as the framework for building a set of rules about what needs to be warned.   I'm certainly hazy on things like Canadian - I agree that there are bound to be some settings with minimal main time that make sense.

And, I guess that if you were going to cheat with Absolute time at the end, you wouldn't bother making a really long one.

So I'll try to think of some sensible refinements and resubmit :)

anoek:
Cool I think we're almost there, my only note now is that we should make those strings translatable :) That's pretty much just a matter of wrapping them in an underscore call: `_("blah blah")`

GreenAsJade:
Oh doh!  I knew that!

anoek:
Awesome, thanks!

PR #719, opened 2018-12-01
Put Rank column header back into Correspondence Automatch table

### LINKED ISSUE #708: /play created games column order offset
### Summar
The column headers for my created gameoffers are offset one step because "size" is populated by rank-restriction or something

see screenshot
image

GreenAsJade:
Fixes #708


anoek:
Whoops, yeah that was a mistake. Thanks!

PR #720, opened 2018-12-02
Added the PWA service worker.

klueq:
Given that /manifest.json is already
served by OGS, we only need to add
this dummy service worker to enable
the Install PWA button in the Chrome's
settings.

Important note for testing: PWAs work
only over HTTPS with a valid cert or
from localhost.

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -194,6 +194,25 @@ browserHistory.listen(location => {
 
 });
 
+/*** Progressive Web App ***/
+
+try {
+  // This API is available only over HTTPS.
+  navigator.serviceWorker.register('/pwa.js').then(
+    res => console.log('PWA handler registered.'),
+    err => console.warn('PWA registration failed:', err));
+
+  let deferredPrompt;
+
+  window.addEventListener('beforeinstallprompt', e => {
+    console.log('Got the PWA beforeinstallprompt event.');
+    e.preventDefault();
+    // Now we can invoke e.prompt(), but Chrome already has
+    // the Install PWA button in the settings.
+  });
+} catch (err) {
+  console.warn('PWA registration failed:', err);
+}
 
 /*** Some finial initializations ***/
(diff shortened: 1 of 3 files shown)


anoek:
Pretty cool stuff :) Before this is merged I need to understand how and when things are updated, and what else we should do in preparation for whatever pwa-only features are possible (for example, what, if any, permissions should we be asking for?)

klueq:
I wish I knew answers to any of these questions. Most likely I won't have time to work on this PR. The idea is to allow to run OGS as a PWA and see how it goes. Maybe it won't work at all. I don't know.

anoek:
It should work well, my questions aren't so much "can we" but "how do I..." so I can integrate it into my deployment scripts and whatnot. And as for features, that's mainly a "what do we need to ask for so we don't have to ask again in the future as we implement things" kind of a thing, i just haven't researched it enough myself to know the answers to these questions.

comntr:
Are there plans to merge this? It would be nice to have a separate icon on my Android phone. The icon could show if a game needs attention.

anoek:
At present this PR is incomplete, so while it's certainly something to work towards, we need to get things completed before we merge.

kevincoleman:
> I need to understand how and when things are updated, and what else we should do in preparation for whatever pwa-only features are possible (for example, what, if any, permissions should we be asking for?)
I think these are good questions, but the reason they have no answer at this point is that there really aren’t huge changes proposed in this PR. Adding a dummy service worker just helps lay groundwork for other PWA-specific features, right? (Correct me if I’m wrong, @klueq).

I suppose that until we’re adding features that just equates to bloat—but it’s pretty minimal at this point, at least.

So the question is: what should acceptance criteria for this look like? Does it aim to change users’ experience at all, or is it strictly an under the hood addition to encourage other devs to build this into a PWA?

anoek:
Tracking in #976

PR #722, opened 2018-12-09
A better rule for detecting troublesome time settings

GreenAsJade:
Either there has to be enough time for a game, or enough time for each move ongoing.

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -19,8 +19,6 @@ import * as data from "data";
 import {_, pgettext, interpolate} from "translate";
 import {TimeControl, TimeControlTypes} from "./TimeControl";
 
-const QUESTIONABLE_SECONDS_PER_MOVE = 5;  // less than this gets alerted to players about to accept
-
 const times = [ /* {{{ */
     1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12,
     15, 20, 25, 30, 35, 40, 45, 50, 55, 60,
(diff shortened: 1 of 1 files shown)


GreenAsJade:
I fielded the first report about a problematic Fischer game, which prompted me to improve the rule.

src/components/TimeControl/util.ts
```diff
@@ -408,27 +406,42 @@ export function shortShortTimeControl(time_control) { /* {{{ */
...
-            return time_control.main_time < QUESTIONABLE_SECONDS_PER_MOVE && time_control.period_time < QUESTIONABLE_SECONDS_PER_MOVE;
+            return !(
+                time_control.main_time > QUESTIONABLE_ABSOLUTE_TIME ||
+                time_control.period_time > QUESTIONABLE_SECONDS_PER_MOVE
+            );
 
-        // haven't seen silliness, not sure about sensible criteria
         case "fischer":
+            return !(
+                time_control.initial_time > QUESTIONABLE_ABSOLUTE_TIME ||
```
anoek: I've seen a lot of good fisher games that have less than 15 minutes initial time, it's less important so long as the time increment is reasonable - on the flip side being able to build up to 15 minutes and and then wait will definitely not sit well with some people. Maybe something more like 30 seconds as a minimum?

GreenAsJade:
The game is considered cheat if neither

                 time_control.initial_time > QUESTIONABLE_ABSOLUTE_TIME 

NOR
                 time_control.time_increment > QUESTIONABLE_SECONDS_PER_MOVE
        
IE the game is reasonable if the initial time is reasonable OR the time increment is reasonable.  Either one suffices.

The intent of this rule is that if the time increment is reasonable, then you can have as short initial time as you like.

The initial time constraint only kicks in if the time increment is unreasonable, at which point the game is effectively an absolute time game with initial time as the time.

I sidestep "what about the max time" in this proposal by noting that if the time increment is unreasonable, then the max time is academic, because you aren't expecting to accumulate.

anoek:
Hah, yes indeed. :+1:

Issue #730, opened 2018-12-23
Player searchbar on User page accepts input when challenge window is open 


flurmbo:
### Summary

As stated in the title. I think this is caused in src/ogs.styl, which sets the z-index of PlayerAutocomplete components to 3000 and the z-index of modals to 2000.

### Environment

 * Operating system: Ubuntu 18.04.1
 * Browser: Firefox 64

### Steps To Reproduce:

1. Go to a user page
2. Scroll to "game history" and click on another users name
3. Click challenge
4. Challenge window should open. In the background, click on the player search bar. You should be able to type in it

### Additional Information:

*Include any screen shots, links to games/reviews, wonky SGFs, or other information that is useful*

![ogstypingactive](https://user-images.githubusercontent.com/35127144/50381083-f0425d00-0642-11e9-95ea-ae52a2471f28.gif)

schambers:
I can reproduce and have a simple fix. Simple approach is to reorder the PlayerAutocomplete z-index. Will leave it up to dev team to determine if this is an appropriate fix as I'm not sure if this has impact elsewhere.

PR #736, opened 2019-01-12
Issue #730: move PlayerAutocomplete behind model z-index

schambers:
Issue #730


anoek:
Ah indeed, thanks for fixing that :)

PR #738, opened 2019-01-14
Issue 737: add jest support for running tests, add simple first test

### LINKED ISSUE #737: Add example first React test
### Summary

There is currently no tests on frontend online-go.com. This issue serves as a PR that introduces jest and an extremely contrived first example of a test for NotesModal

If the OGS team would not like to go in this direction, or if there is some recommendations on how testing should be performed I'd be open to that.

Ideally, this could be hooked into the shippable builds that run after a PR is submitted and serves as a very basic safety net for anyone adding work to online-go.com.

schambers:
I noticed there are no frontend tests for any of the react components. I thought it may be worthwhile to introduce jest and add a very simple first test.

It was required to mock src/lib/misc.ts as this module inits all of online-go.com and is prohibitive of running tests in isolation. Thankfully mocking out this one component property runs the test.

We may want to leverage enzyme for testing snapshots of the DOM but I've never used it before and wanted to submit this first to see if the OGS team wanted to go in this direction.

Fixes #737


anoek:
I'd love to integrate a test framework :)

I'm not familiar enough with the options myself, so before I merge anything I'd like to see a good set of tests developed to make sure that whatever we pick is going to be effective for our case.

schambers:
I can add a couple more tests. I'm admittedly new to jest myself. 

One challenge I ran into was the module loading is all tied to some files that inherently load the Dom/inject script tags. I managed to mock out the needed pieces but I didn't want to introduce any refactorings as part of using the test framework. This issue showed a very simple test just to show that it can be done. I can create more tests when I have a chance to show that it would provide value. 

I think it would be best to add them as part of this one issue and I can send another pull request.

anoek:
Feel free to re-open if you want to pick this back up

PR #746, opened 2019-02-06
Indicate to moderator if the game is not actually annulable.

GreenAsJade:
The "This game was already annulled" dialog has become bit much :)


GreenAsJade:
Ooops - sorry about the log.

There's still one floating around from when I did the "pause the game list refresh thing" I've noticed as well - I'll get rid of that in the next PR I do, if you haven't already.

PR #752, opened 2019-03-06
Small improvements

kevincoleman:
- Move forgot password link to near form
 - Add static field labels (in case users forget what fields need after they’ve typed)
 - Add email login icon on register form (so users can easily get back to the login form)
 - Adjust spacing & styles slightly


anoek:
I like the new position of the "Forgot password?" text.

The email login icon on the register form is a good touch too.

I'm not convinced the explicit labeling and modified placeholder text is an improvement, and it adds new strings for translators .. not a huge deal, but since I don't think it was a problem to begin with, I'd rather not do.

There's a lot of changes to code that wasn't modified, ie style changes, which I would prefer not to merge since some of the updated styles diverge from what is more common in the rest of the code base.

kevincoleman:
Thanks for looking it over!

[Here’s some of my reasoning for the static labels](https://www.nngroup.com/articles/form-design-placeholders/)—but if you’d still rather not accept that change I can make the updates without them.

Sorry about all those changes that don’t amount to anything but stylistic difference. My env just auto-formats things for me. I thought about battling it, but ¯\\\_(ツ)_/¯.

Do you have a certain set of linters and/or style rules you follow? I’d love to continue contributing to this project, but don’t want to have to manually override things for style’s sake all the time. Perhaps if I know what the styles are I can customize plugins and save those settings locally to override my defaults.

anoek:
Re static labels, the arguments made are more applicable to other types of forms I think, though I dunno, maybe it does make sense. If we do it, I would like to see the same change made to the registration page as well, and maybe some look-n-feel tweaking to make sure it looks good on mobile and on desktop without scrolling (This more notable on the registration page). Also the placeholder text should be simpler or just removed.

Yeah maybe just turn the auto-format stuff off for this project, I don't want to have a bunch of random  automated style updates junking up PRs :)

For linting, yeah we use tslint: https://github.com/online-go/online-go.com/blob/devel/tslint.json is the config. You'll also see the lint errors in the build output, if that's clean then when you commit the build system should be clean as well.

I think I'd be fine with going with MDI or the community augmented set, with the two conditions that it (1) looks as good as the zocial stuff or better when it's integrated (Zocial takes care the button styling as well, including colors and shading, all matching the different branding specifications and whatnot). and (2) doesn't bloat the project by pulling in a bunch of other random material stuff, or thousands of icons we don't need/use.. some of that material stuff I've noticed starts pulling in way too many things and can quickly add hundreds of kilobytes if not megabytes. So so long as those two things are satisfied I'd be fine with it.

kevincoleman:
Re labels: that sounds appropriate to me. I suppose the advantage to making the change on a simple little form like the login would be that it establishes a style precedent for if/when we do have bigger forms. I did actually make the label changes on the registration page too (unless there’s one elsewhere I’m missing). At least from testing in-browser using Chrome’s mobile preview we’re not running into any space issues as far back as the iPhone 5. The iPhone 4 loses a tiny clip off the bottom of the card. Are there particular tests you’d want that to undergo?

screen shot 2019-03-07 at 2 02 55 pm
screen shot 2019-03-07 at 2 03 04 pm

Totally understandable re: autoformatting. I’ll see if I can get it in the config to just ignore that completely while in this workspace.

👌🏻 on TSLint. :)

I’ll look further into the extended MDI library. I know we can implement very lightly if we just include _only_ the icons we want, but then other devs may not get the “magic” of being able to just name an icon and have it work.

Please feel free to disagree, but as I see it simple icons (even just black/gray) would have a nicer aesthetic than the zocial buttons. While it’s cool that they use brand colors, the slight size/alignment issues with them really bug me. I suppose we could try to just fix those individually, but even that’s messy since the library injects the actual icons with `:before`.

I’d suggest making them square:
screen shot 2019-03-07 at 2 07 24 pm
screen shot 2019-03-07 at 2 07 33 pm
And then also adjusting the icon placement a tad (email feels high, facebook feels right).

And, actually, isn’t the Google logo an old one now?

anoek:
Cool, sorry I didn't recall seeing it on the registration page, i overlooked it, sorry about that!

Yeah we actually do label the other forms we have on the site generally speaking, so I guess it brings things more in alignment with that. I'm ok with it. I think we might want to get rid of the placeholders though then, since they don't convey any additional information.

On the one hand, I don't disagree that "cleaner" icons would be more aesthetically pleasing, but I think the UX improves notably by adhering to the branding guidelines for the icons. We *want* them to stand out a bit and for them to register what they are for the user without them having to think about it. That's a much easier thing when you go all in with branding since that's what minds are programmed for. Sticking to a more uniform look might look "cleaner" when they are side by side, but then the user has to parse through them, and in this case I don't think that's a win.

The alignment issue (I presume you're talking about the f for facebook), that's actually part of the branding guidelines for facebook, if you check out their icon on their home page you'll notice it there too. So, while it may bug you if you're staring at it as a coder, I think it's the right thing to do for UX since it registers "Oh yep that's the face book button" in users minds instead of "Hey look an f". 

The g for google is out dated you are right, we should probably update that in any case :)

As for MDI and size and all of that, so basically if we are just using it for social icons, zocial, css + font is 21k.. the MDI woff is 200k, so 10x the size not including css, that seems like maybe not the best trade off. We do already include FontAwesome for all of our other icon needs, and I would only want one of them active for visual consistency, so if we wanted to include all of MDI I would say we should do that if and only if we want to replace FontAwesome with MDI. I'm not opposed to considering such things, but that is a bit of an undertaking, and we should probably get some user input with some before/after shots to see how folks feel about the change.  (FTR FontAwesome 4.7 is only ~75K so is also notably smaller than MDI, though not so much that it would be a deal breaker to switch to MDI based on that)

anoek:
of course https://en.facebookbrand.com/assets/f-logo/?audience=landing makes it look like they really want the f to be on a solid background icon now adays ... it does seem like the zocial stuff is all around out of date for everyone's preferred branding

anoek:
So in light of the fact that both google and facebook icons are out of date, I think we *should* replace zocial with something else (MDI+community is good it looks like, alternatives are also good). If we are not replacing font awesome at the same time, then we should suck in just the icons we need from MDI. If we are replacing all of font awesome, that needs to be in it's own PR and we need to do some before/afters just to make sure that's what we want to do.

FYI for sucking in just what we need from MDI, it looks like they have SVG's of everything, and we have an svg inliner already in our build pipeline (see `svg-load`, there's an example of it in `src/views/Supporter/Supporter.styl`), that might be a good option.

kevincoleman:
Ah! Great thoughts. Yeah, I realized the Facebook icon is supposed to be used off-center, but in that button since it doesn’t bleed to the bottom (and since the rest of the button felt more “generic button” than Facebook brand) it didn’t quite sit right with me. Seems like Facebook wouldn’t love that application of it, either, since it modifies their intended brand mark (the last item in the brand guidelines).

I agree with what you’re saying about color for brand recognition. Maybe with a flagship color for each logo we can make it work similarly with the MD icons. Looks like the SVG inlining is our best bet for now, and we can take a look at MDI inclusion later. I’ll take a stab at inlining the SVGs.

kevincoleman:
Haha looks like I got to merge into some upstream changes. 😬

Hope I did that merge ok. Please feel free to double-check me.

kevincoleman:
An interesting thing I discovered—the SVG inliner only works for adding SVG content post-dom-load, meaning the SVGs can’t then be modified again by CSS. This means that without editing the source SVG files we can’t easily change the colors.

I did find a method that uses CSS filters to modify a black icon into a colored icon, though.

Please check for me: I’m seeing inconsistency in applying any styles to the `.mdi-icon` elements. Sometimes styles load (colored icons), and sometimes they show in the inspector that they’re inheriting a blank set of styles from the selected CSS set (black icons).

kevincoleman:
Found the inconsistency. It was a bracket typo. 🙄

kevincoleman:
Screen Shot 2019-03-08 at 5 17 41 PM
Screen Shot 2019-03-08 at 5 17 54 PM

anoek:
So looking at the branding guidelines for facebook and google it looks like we can't use those icons as is, the g apparently needs to be the colored google g *and* can't be by itself anymore, so we may need to rethink how that sign in / register page present those options:

https://developers.google.com/identity/branding-guidelines
https://developers.facebook.com/docs/facebook-login/web/login-button/
https://developer.twitter.com/en/docs/twitter-for-websites/log-in-with-twitter/guides/sign-in-with-twitter-resources

kevincoleman:
Hmmmm. Ok. Way to throw a curveball, Google!

Maybe a list of material buttons with a similar pattern for each login service? Seems verbose, but at least accessible.

kevincoleman:
Seems like if we follow all of their guidelines we’ll end up with three mismatched styles. 🙄

anoek:
image

For reference, this is how Medium did theirs

kevincoleman:
Yeah! I like that approach. It technically doesn’t meet brand specs, but I bet any designer at Google/Facebook/Twitter would agree it’s better to be consistent.

kevincoleman:
This commit mimics the Medium style more closely, and switches to using `browserHistory` since we have buttons now instead of links. Let me know what you think.

Thanks!

anoek:
Nice looking good :)

I don't think the browserHistory stuff will work though since those login mechanisms aren't handled by the UI, they are supposed to hit the backend and do some redirect magic.

kevincoleman:
Screen Shot 2019-03-25 at 11 31 25 AM
Screen Shot 2019-03-25 at 11 31 15 AM

kevincoleman:
Huh, about `browserHistory`, the routes seemed to resolve just fine. In dev they’re “Not found” pages, but that was the same symptom I got when using regular ol’ links. Can you test them in a staging environment?

anoek:
Yeah they don't work as expected, browserHistory doesn't actually request a new page, it just goes through the client side routing logic, hence why it doesn't work.

kevincoleman:
Ok, no problem. Would be great to get tests (happy to write some) or an env that exposes that shortcoming in dev. Do you have a preferred method for handling routing that’s not triggered by `a` tags?

anoek:
What's wrong with `<a href..>`?

anoek:
I actually didn't realize those links wern't working on the dev environment, i'll fix that up

kevincoleman:
It’s just a semantic issue. Usually it’s not valid to wrap a block element or other inherently clickable element like an HTML button in an inline element like an `a` tag. Functionally, nothing wrong with it, but I always shoot to keep things semantic when possible.

If you prefer I can do it in this case. Just looking for middle ground. :)

anoek:
Ah, right so don't do `<a><button/>...</a>`, just do `<a className='btn'>...</a>`, that is to say, style the link to look like a button or whatever you want it to look. `btn` (in this case probably `xs btn`) are the classses we use for the other buttons on the site, you might try those and see if they look reasonable.

kevincoleman:
Should be good to go now!

anoek:
FTR I opened a pull request against the branch this is coming from when you want to check it out and see if you're happy with the changes

anoek:
Thanks for doing this :)

kevincoleman:
Oh wow. Just got back to see this. Good call on the other improvements here. Looks good deployed. :D Thanks!

PR #753, opened 2019-03-10
Tweaked komi warning

GreenAsJade:
People still fall for trick custom komi.

It appears these get obscured by fast game settings that the acceptor doesn't mind.

So here is a separate warning.


anoek:
:+1:

Issue #758, opened 2019-04-07
Blank screen in OGS's interface


felixthelucas:
Some people are getting a blank screen when they access OGS's interface (as reported in [this thread](https://forums.online-go.com/t/cant-log-into-my-ogs-account-blank-screen/20897/)). So far the problem seems to affect those running iOS 9 (no browser info provided), both in iPhones and iPads, but _not_ iOS 12. 

I'm thinking the issue might be related to the version of their browsers. Though this may be unrelated, someone running Windows XP, trying to use Chrome 49 to access the interface, got this error message:

> No WASM support detected, score estimator falling back to ASM.js mode
es6-shim.js:1827 Uncaught SyntaxError: Invalid flags supplied to RegExp constructor ‘igu’
raven.js:58 SE Initialized

So, he's using an old, pre-ES6 browser, which doesn't support the u flag enabled to block some accented Portuguese words in the updated profanity filter (pull request #757, merged). Maybe the same thing is affecting the others.

In order to ensure backward compatibility, I'll just remove unicode entries from the filter and submit a new pull request. Maybe going forward we can look for a better solution, but this will have to do for now.

felixthelucas:
@anoek, the merge hasn't reached the affected users, they're still getting the previous filter after clearing their browsers' cache. I figure the refresh is due on the server side. Time for a reload or the good ol' fashioned reboot?

anoek:
I think this is good now right?

guest951:
I have iOS 10 and I still have the problem. I tried to clear Safari cache or to use another browser but I still have a blank page.

I tried to run the developper version in my computer and from it I was able to access the site with my iPad. I don't know why it does not work for online-go.com, maybe there is a delay before the developper version is deployed to the web site?

guest951:
Hello, I think I identified the reason it's working with the dev version and not with the production. I get those errors when I try to open online-go.com with Safari:
```
goban.min.js:17 SyntaxError: Cannot declare a let variable twice: 'e'.
ogs.5.1-2310-g270a294-4f27d449f0e13b55befadaa398d32102.js:18 SyntaxError: Cannot declare a let variable twice: 'e'.
```
So the problem comes from the minification. It is related to a bug in iOS 10 which is described here:
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/92#issuecomment-324818437
It should be possible to fix it but I don't know exactly how.

PR #759, opened 2019-04-07
Revised profanity_filter.ts

felixthelucas:
Older, pre-ES6 browsers don't supported the u flag, enabled to handle some accented words in the Portuguese filter. In order to ensure backward compatibility, I've disabled the flag and removed unicode entries from the list altogether.

Not the ideal solution, but it should work for the time being.

More information: Issue #758.


anoek:
If the `u` is useful, you could just do

try { 
profanity_regex["pt"] = new RegExp("...", "igu");
catch (e) {
profanity_regex["pt"] = new RegExp("..", "ig");
}

anoek:
If you decide to do the try / catch stuff I'd be happy to merge that as well :) Thanks for the quick fix here.

felixthelucas:
It is, @anoek. This fix will do for now, but I'll see if I can download some older browsers and run them with WINE to do some testing with a try/catch. Thank you very much and sorry for the trouble.

anoek:
No problem at all :)

PR #768, opened 2019-05-12
Fix mid-word line breaks in PaginatedTables

sinking-point:
Lines frequently broke mid-word on pages like /groups and /puzzles. This was due
to the word-break property being set to break-all. This commit fixes the issue
by removing the setting of this property. word-wrap is still set to break-word,
so extremely long words should still break.


anoek:
I feel like I put this in here for a reason, like maybe to fix tables exploding with really long words.. but I reckon that needs a better solution, because we certainly shouldn't be seeing broken words on those pages. Merging, and presuming we'll figure out what breaks and fix it.

PR #770, opened 2019-05-13
Added option to turn off countdown in main time for byo-yomi and canadian

GreenAsJade:
As recently discussed, and comes up from time to time, it irritates some players to have voice countdown during main time for Byo-yomi and Canadian.

So here is an option to turn that off. 

Because Dwyrin and Collosiuss both complained about this, and they are not great setters-of-settings, I have made it the default to have it turned off, so they get to experience the change for good without effort on their part.

### DIFF
--- a/src/lib/goban.ts
+++ b/src/lib/goban.ts
@@ -54,8 +54,20 @@ export class Goban extends OGSGoban {
         return window.location.pathname;
     }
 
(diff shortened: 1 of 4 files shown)


anoek:
Indeed. The whole audio system needs a good overhaul, this is a good step.

PR #807, opened 2019-06-09
Better challenge cheat warnings

GreenAsJade:
Folk are still missing the warning that they are joining a game with tricky settings.

This change:

1) Moves the warning triangle icon over next to the accept button
2) Causes a popup as they mouse over the cell with the button in it, so they can't press the button without seeing the warning details popup.

Screenshot 2019-06-10 00 30 28


anoek:
Nice

PR #808, opened 2019-06-09
Always save a volume setting change

GreenAsJade:
Background: whenever Dwyrin plays on OGS he opens the settings and turns down the volume.

Sometimes he mutters about how it's always too loud.

Although I have mentioned it, he doesn't seem to catch on that you can save the setting using the archaic "save" icon.

So that got me thinking "Why do we need a save option?  Why would you change it and _not_ want it to persist".

Hence this PR: just persist the volume setting.


anoek:
Pff... archaic. ;p

PR #806, opened 2019-06-09
Kind of lossy fix to the ratings bar chart (issue #466)

tcarette:
Fix #466 

force ratings_x axis range to span full months.
remove today-related logic, i.e. treats current month like past months (code not removed but commented out for eventual easier reuse)

PR #809, opened 2019-06-09
Win loss to weaker stronger bar

tcarette:
Important note: this change is based on branch tcarette:issue_466, subject of the pull request #806 .

Proposal to change the statistics displayed by the bar chart in the profile page from highlighting ratios of strong/weak for either wins and losses to ratios of win/losses for either stronger or weaker players.

Here is a break down of the old bar chart
image

My problem with splitting and normalising the data as above is that it is very sensitive to the ratio of weaker vs stronger player you are playing with. This latter ratio is not intrinsic to the player but rather of OGS's user base.

And here is a break down of the proposed new version
image
edit: fix image according to comment of @GreenAsJade

Note that one of the point is to be able to very quickly see if somebody has a good or bad month as it will be reflected by the fact that you see either a larger unexpectedly losses against weaker players, or an unexpectedly high number of wins against stronger players. Finally, the horizontal size of the right bars indicates how consistent is a player's strength. Those information overlap with the already available statistics, but are distinct and arguably more straightforward.

The vertical dimension still amounts for the total number of games, but the split now reflects the split between games against weaker or stronger players. As argues before, it might not be the most interesting of statistics. My opinion though is that the current version reflects the win/loss ratio which is seen clearly in the pie chart and also depends strongly on opponent's relative strengths anyway.

This version is but a proposal. I'm open to suggestions to show the stats I suggest in a prettier way.


tcarette:
Here are some more examples but I didn't find a test user with heavy game record that might show how it looks more generally

corresponding to the other game records above: a player which is doing good.
image

Multi-month example (DevGaj)
image

GreenAsJade:
I think the blue ellipse is meant to be labelled "Expected" :)

GreenAsJade:
Now that I see this laid out like this, I understand the proposal ( which was originally laid out [here](https://forums.online-go.com/t/win-ratio-visualisation/21629/16 ) 

It makes sense to me.   Interestingly, none of the previously passionate people about these charts chimed into that thread - I might edit the title to make it clearer what the target is.

tcarette:
please do. I'm bad at click baiting.

GreenAsJade:
LOL - that's a nasty term though isn't it?   Really the motivation is to flush out the complainers as soon as possible, so we find out what if any there are.

anoek:
I like it. Nice job jumping into this part of the code @tcarette, it had me scared and I wrote it ;)

tcarette:
Cool, thanks.

GreenAsJade:
So - I think this looks very nice in practice - I can see it on people's profile now, and somehow the bar graph looks much more tractable to "see what it means".

However, there is a problem:

image

Here I have moused over September in the bar chart, which has a great big bright green section, and in the pie chart the bright green section is not correspondingly big.

It appears to me that either the bar or the pie has mixed up the data - the bar's "bright green" appears to be the same data point as the pie's "bright red".

I think the pie chart appears to have the right data, without actually digging in and debugging, because:

Screenshot 2019-06-11 12 56 46

here we see that overall in this month the person's rank went down, so it seems unlikely that they had an "unexpectedly large" number of wins vs stronger.

tcarette:
is it related to #811 ?

GreenAsJade:
Well that was quickly fixed :)   Looks really great now :)

tcarette:
Yeah,  well,  I shouldn't have let this stupid obvious bug through in the first place

GreenAsJade:
Meh - it happens.  It's all worth it: your insight that these bars would look better with expected on the left and unexpected on the right is great.

Issue #810, opened 2019-06-10
Win-loss bar graph gets clipped


flovo:
### Summary

The lower part of the win-loss bar graph gets clipped off. 

### Environment

 * Operating system: Linux / Android
 * Browser: Firefox
 * Browser Version: 67.0.1 (64-bit)

### Steps To Reproduce:

1. Go to Profile page https://online-go.com/user/view/449941
2. The lower part of the Win-loss bar graph is cropped off.

### Additional Information:

This is how it looks:
![image](https://user-images.githubusercontent.com/4864182/59212114-d1646280-8bb1-11e9-8754-b23a6f8517bc.png)

It should look like
![image](https://user-images.githubusercontent.com/4864182/59212181-02449780-8bb2-11e9-973a-1756a90df757.png)
I removed the clipPath to see the whole chart.
![Clippath](https://user-images.githubusercontent.com/4864182/59212325-5ea7b700-8bb2-11e9-92fc-260f594142d0.png)

anoek:
@tcarette might have an idea?

tcarette:
I'll check asap

BHydden:
Am I the only one not seeing the difference? #precoffee

tcarette:
lol. It is there.
I'm searching but I don't think it is related to the latest release. I could reproduce it with a previous version.
But I'm glad to look at it since I have the code open anyway :-)

tcarette:
ok - shot in the dark fix here. Please somebody reviews my interpretation below and tell me if it sounds ok.
https://github.com/tcarette/online-go.com/tree/issue_810

The logic is the following: `outcomes_x` maps the game count domain on a range of y-values and heights. The bar chart height was parametrized with `win_loss_bars_height`, set at 65, but the range of `outcomes_x` was hardcoded to 60. So my guess is it should be set to `win_loss_bars_height` as in the commit.

Seems sound?

Because it still looks like it is a bit clipped. Actually, there is a bunch of things in this graph that seem to not align just right. But hopefully it is not too noticeable; I don't want to do a patch that is illogical "but works".

![image](https://user-images.githubusercontent.com/2481066/59230215-0639df00-8bdd-11e9-82cc-9e4d1b3cbe90.png)

tcarette:
tbh, I don't know where this clipPath is coming from either... 😕

flovo:
I think the lines creating the clipPath are, but I cannot check at the moment
[...]

PR #812, opened 2019-06-10
fix #811: Bar chart is wrong when making a time selection

### LINKED ISSUE #811: Bar chart is wrong when making a time selection
### Summary

Bar chart is wrong when making a time selection

### Environment

 * Operating system: MacOS Mojave 
 * Browser: Chrome 
 * Browser Version: 74.0.3729.169

### Steps To Reproduce:

Normal view seems ok.
image

but a zoom breaks it
image

and it stays broken after deselecting.
image

tcarette:
fix #811

My mistake. I missed the part of the code that recomputed the bar chart in the onTimelineBrush.

### DIFF
--- a/src/components/RatingsChart/RatingsChart.tsx
+++ b/src/components/RatingsChart/RatingsChart.tsx
@@ -945,18 +945,19 @@ export class RatingsChart extends React.Component<RatingsChartProperties, any> {
             return isFinite(x) ? x : 0;
         };
 
+
         this.win_loss_bars[0]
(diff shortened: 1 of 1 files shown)


tcarette:
btw, since I'm working on #810, should I do a single pull request when I'm done?
kinda noobie question here.

tcarette:
I suspect we could just get rid of the edited lines.

anoek:
Awesome thanks for the quick turnaround :) I tend to prefer separate PR's just like you've done, even though they are small, it makes things easier to track in my opinion, so good job!

PR #815, opened 2019-06-13
Fix check for custom komi.

GreenAsJade:
Simple bugfix so that custom komi of 0 gets the warning, like any other custom komi.

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -433,8 +433,8 @@ export function usedForCheating(time_control) {
         case "byoyomi":
             return !(
                 time_control.main_time > QUESTIONABLE_ABSOLUTE_TIME ||
-                time_control.period_time * time_control.periods > QUESTIONABLE_ABSOLUTE_TIME
-            );
+                time_control.period_time > QUESTIONABLE_SECONDS_PER_MOVE
+                );
 
         case "fischer":
             return !(
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -433,7 +433,7 @@ export function usedForCheating(time_control) {
         case "byoyomi":
             return !(
                 time_control.main_time > QUESTIONABLE_ABSOLUTE_TIME ||
-                time_control.period_time > QUESTIONABLE_SECONDS_PER_MOVE
+                time_control.period_time * time_control.periods > QUESTIONABLE_ABSOLUTE_TIME
             );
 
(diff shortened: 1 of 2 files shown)


src/components/TimeControl/util.ts
```diff
@@ -433,7 +433,7 @@ export function usedForCheating(time_control) {
         case "byoyomi":
             return !(
                 time_control.main_time > QUESTIONABLE_ABSOLUTE_TIME ||
-                time_control.period_time > QUESTIONABLE_SECONDS_PER_MOVE
```
anoek: Isn't this going to flag pretty much every live game as questionable? It's quite common to have say 5x30s byoyomi, which if i'm not mistaken, would be flagged as questionable here if the main time wasn't at least 15 minutes

src/views/Play/Play.tsx
```diff
@@ -591,6 +591,9 @@ export class Play extends React.Component<PlayProperties, any> {
 
         let challengeList = isLive ? this.state.live_list : this.state.correspondence_list;
 
+        challengeList.forEach(element => {
+            console.log("komi: " + element.komi);
```
anoek: Probably don't need this log anymore for production

src/components/TimeControl/util.ts
```diff
@@ -433,7 +433,7 @@ export function usedForCheating(time_control) {
...
+                time_control.period_time * time_control.periods > QUESTIONABLE_ABSOLUTE_TIME
```
anoek: Isn't this going to flag pretty much every live game as questionable? It's quite common to have say 5x30s byoyomi, which if i'm not mistaken, would be flagged as questionable here if the main time wasn't at least 15 minutes

src/views/Play/Play.tsx
GreenAsJade: Oh yeah, I will (or you can) get rid of that :)

src/components/TimeControl/util.ts
GreenAsJade: No, I think this is one of these upside down logic things.

This time setting is used for cheating if it is _not_ the case that those OR conditions are true.

So this setting is used for cheating 
```
if  (it is not the case that
      (main time is big enough or 
        there is enough time in the periods))

```

GreenAsJade: Oh wait, you are right.

Hmm - what was I thinking!

I should have just stuck to fixing the actual bug!

PR #817, opened 2019-06-15
Height range adapts to selection for win-loss bars

tcarette:
When doing a time selection, the bar with maximum monthly number of game
is taken as the maximum of the range of the Y axis.

I've noticed that for certain users which have some very busy months in their history, their bar chart are all but invisible for recent activity. e.g.
Screenshot 2019-06-15 at 12 24 41

The idea of this pull request is to just adapt the hight of the bar chart to the selection. It is done on monthly monthly game records to avoid glitches.
Screenshot 2019-06-15 at 12 24 54

It is heavier to compute so the fluidity takes a hit for more active users.
It is to be considered whether it is better to optimise this feature before integrating.

### DIFF
--- a/src/components/RatingsChart/RatingsChart.tsx
+++ b/src/components/RatingsChart/RatingsChart.tsx
@@ -916,12 +916,33 @@ export class RatingsChart extends React.Component<RatingsChartProperties, any> {
         let lower = Math.min.apply(null, this.game_entries.map((d:RatingEntry) => Math.min(d.starting_rating, d.rating) - d.deviation));
         let upper = Math.max.apply(null, this.game_entries.map((d:RatingEntry) => Math.max(d.starting_rating, d.rating) + d.deviation));
 
-        let l = Math.min.apply(null, this.game_entries.map((d:RatingEntry) => (d.ended.getTime() >= this.date_extents[0].getTime() && d.ended.getTime() <= this.date_extents[1].getTime()) ? (Math.min(d.starting_rating, d.rating) - d.deviation) : upper));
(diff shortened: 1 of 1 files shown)


tcarette:
I'm optimizing - just in case, wait a bit.

anoek:
I definitely like the idea, but it is a bit chunky. Since it seems to just be chunky for users with a lot of entries, I'm assuming the chunkiness is coming from iterating over all of the entries every time we we move the mouse.

Would precomputing be possible here? What comes to mind is

  1. When we get our data blob from the server, precompute the range for each month and cache it somewhere.
  2. To compute the range of a selected area, take the min/max month then iterate across all the months in between in our cache to compute the result, that way we are only touching a few dozen things instead of a few thousand.

If there's a notable lurch at data load doing the precomputing, expanding out the elegant `Math.min.apply(null, this.game_entries.map((d:RatingEntry) => is_selected_games(d) ? (Math.min(d.starting_rating, d.rating) - d.deviation) : upper))` into a plain `for` loop would probably help a bit too since you could half the calls to `is_selected_games` (since both maps do that same filtering), as well as eliminate the implicit array allocations that happen with the mapping across the few thousand entries. I don't think that that optimization is going to be too noticeable though, there are other lurches that happen on the page already for folks with large play histories, so it's unlikely to be noticeable... but if you feel it is, it might help a bit :)

tcarette:
I agree that the fluidity of the selection needs to be saved as much as possible.

I made a couple of half assed attempts at going in the direction you propose but I didn't have much time. I will come back more earnestly to this when I have time.

anoek:
Feel free to re-open this if you pick it back up, I think it's a good idea

tcarette:
Yes - I really didn't have the time to take this up again. Also I can blurp some js as I did in other commits, but optimization is another beast altogether. I'm open to pair programming is somebody wants to contribute though.

PR #818, opened 2019-06-19
Fixed up saving the correct volume

GreenAsJade:
Mea culpa.


anoek:
Thanks for the quick fix!

PR #821, opened 2019-06-27
Update to profanity filter for Dutch (nl) language

Adam314159:
Filter cleaned up by Mirreke on forums to better reflect the modern usage.


anoek:
Cool sounds good :)

PR #825, opened 2019-07-06
Enable stone sound on Review and Demo boards

mikheevnickita:


anoek:
We'd need to add some settings to enable this. We used to have this enabled by default for everyone but had a lot of complaints about it, which is why it's off now. I'd be fine with turning it on via a setting though.

anoek:
Feel free to re-open this if you want to take a stab at making it a setting

PR #835, opened 2019-07-16
Expand the capture part of the tutorial

jsonn:
Properly introduce liberties, groups and eyes before asking the user to
figure things out.

### DIFF
--- a/src/views/LearningHub/Capture.tsx
+++ b/src/views/LearningHub/Capture.tsx
@@ -40,7 +40,7 @@ export class Capture extends LearningHubSection {
 class Page1 extends LearningPage {
     constructor(props) { super(props); }
     text() {
-        return _("Capture the white stone");
+        return _("A free space adjacent to a stone is called a liberty. A stone without liberties is captured. Capture the white stone");
     }
     config() {
         return {
@@ -65,7 +65,7 @@ class Page2 extends LearningPage {
     }
 
     text() {
-        return _("Capture the group of white stones");
+        return _("Adjacent stones of the same color form a group and share liberties. They are captured together when they have no liberties anymore. Capture the group of white stones");
     }
     config() {
         return {
@@ -91,7 +91,7 @@ class Page3 extends LearningPage {
     }
 
     text() {
-        return _("You can capture on the side too, capture the white stones.");
+        return _("The border of the field does not count as liberty. Capture the white stones on the side.");
     }
     config() {
         return {
@@ -117,7 +117,7 @@ class Page4 extends LearningPage {
     }
 
     text() {
-        return _("Surrounding isn't enough, you have to fill all the space around the stones. Capture the white stones.");
+        return _("Surrounding is not enough, you have to fill all the space around the stones. Capture the white stones. Notice that you can take even if the newly placed stone has no liberties, because the white stones are taken first.");
     }
     config() {
         return {
@@ -144,7 +144,7 @@ class Page5 extends LearningPage {
     }
 
     text() {
-        return _("Capture the white stones before white can form two \"eyes\"");
+        return _("A group of liberties surrounded by stones of the same color is called an eye. A group with two eyes is alive. Capture the white stones before white can form two eyes");
     }
     config() {
         return {


anoek:
I certainly appreciate the intent here, and for some users it might be better. However, I'm a little hesitant to merge this as I think it explains too much. In short, my observation is the less people have to read the better. With the initial capture "puzzles" as they are, my goal was to make it somewhat of a game to figure out what you need to be doing in this game, thus making it more interesting to jump in and giving you an immediate intuition about the basic flow of the game. I believe the bar is pretty low to figure out, and while I think a sentence about liberties might be justified somewhere in the tutorial, I'm not convinced we need to delve into it in the first section.

jsonn:
Background: There was a long "discussion" about the OGS tutorial during yesterday's US Go Congress stream. The gist is that the current tutorial kind of expects people to have an idea what the rules are. This makes it a lot harder to follow than the old interactive way to go.

An alternative to more verbose instructions would be a new box below the board that covers the rules that are explained interactively. This content would stay the same for each group.

anoek:
I'm certainly very interested in ways to improve the tutorial (and for what it's worth I'm not opposed to content changes in general), however for the initial capture section I'm going to have to respectfully disagree that the tutorial expects people to know what the rules are. 

There is a huge difference in the way games are taught in person vs in an automated interactive way. In person you are taking the position of teacher / student, teacher explains the rules and has the luxury of feeling out their student to know if they are interested in learning new vocabulary (liberties) and whatnot. You have a captured audience, you have the luxury of pushing over any humps of disinterest because they're not going to just get up and walk away. That is not true for computer games. In an interactive way, users do not want to read manuals, they do not want to watch 5 minute videos, they want to jump in and feel smart while figuring things out. If you observe how (successful) video games begin, you'll see this a lot, and if you can find numerous game development blogs and whatnot that talk about it to. The goal is to make learning how to play your game fun, otherwise you just turn people off and you lose them right away. Later when we've captured their attention and they've figured things out, that's the time to teach them vocabulary so they have a basis to understand more Go content. Or if you can slip it in along the way in *very* short blurbs, that can work too, but the key is to keep them moving and discovering the game at a good clip, make them feel like their always accomplishing something (in this case learning a new thing about the game). Explaining too much, particularly before the moment of discovery, takes away that feeling of accomplishment.


I'm somewhat convinced that we are not expecting too much as I just recently with through the process of teaching the game (somewhat) to my 4 year old son, and without really thinking about it, I pretty much followed the initial capture tutorial on the board. I laid down a stone, put 3 around it, and without prompting (while I was thinking about how to teach this) he put the 4th stone down and I got excited and exclaimed he captured it, which he then got excited, I proceeded with 2 stones and stones on the side and it was no issue at all, before you knew it our "game" was placing stones around and capturing them for awhile. 

So while I respect the folks over at the AGA and at the USGC a lot, they are coming largely from a "teaching the game in person" background, and that's very different than what we need to do here.

All that being said, I'm quite sure this tutorial can be improved - but whatever we do needs to move the needle more towards a great interactive-accomplishment style of teaching rather than teaching-the-game-in-person style.

jsonn:
There are fundamentally two different types of learners. Some prefer to be shown the rules step by step and apply them. That's the kinesthetic approach. It is primarily used when dealing with children, but there are also many adults that require this approach. The other approach is learning from the abstract rules and illustrating it with examples. The two choices are completely independent of the media used; they can be applied in person as well as via books or interactive media. Now forcing people to take one approach or the other is a major source of frustration.  Most kids will simply not understand an abstract introduction at all where as many adults will feel ridiculed by a too hands-on approach. It is hard to reconcile them into one set of learning materials too.

Learning by doing doesn't replace explaining the rules though. Case from the tutorial series: why is E5 in Capture/4 a valid move, but A1 in Defend/1 after B1 is not legal? Ignoring everything else about terminology and the approach to learning, just showing rules by examples does not ensure full understanding. That's exactly why all learning systems even for children point out that it is important to put the rules into words at the end or in parallel. This is one area where the current tutorial is seriously lacking. It is also one area where "The Interactive Way To Go" works much and why many prefer to point that one.

anoek:
First I want you to know that I appreciate your taking the time to hash this out with me, it's the most important part of the site and it's certainly flawed right now, so I'm glad to be revisiting it with you to make it better.

I hear you, and I agree we need to explain the rules a better. The points you bring up are good. I do maintain that we are not teaching a class or teaching in person, and while Go eventually straddles being a game and being something you study academically, it is initially a game, and taking pages out of the game development industry regarding introducing your game to players is worthwhile. To this effect, I agree we need to add more to explain liberties and eyes better, but I do feel it's very important to do this in as brief a way as possible, and in a way that minimizes what we take away in terms of their sense of accomplishment of figuring things out during their training (while actually teaching them the rules of the game as you've noted and I'm now seeing with refreshed eyes).

So with that said, here's another cut of text which I try and cover what you wrote but maybe just a little shorter and in some cases shifting the explanations around to attempt to not feed the answer to them right away. Let me know what you think. introducing the concept in small chunks throughout the 5 steps. Maybe something along the lines of:

* Capture the white stone
* A free space next to a group of stones is called a liberty. Capture the group of white stones.
* The edge of the board does not count as a liberty. Capture the white stones.
* You need to remove all of a groups liberties to capture it. Capture the white s tones.
    => Success text: Well done! Usually you cannot place a stone that would be surrounded like that, but because you captured the white stones, it's OK. [next button]
* We call a free space surrounded by one color an "eye". Capture the white stones before white can form two "eyes".

Does that better accomplish teaching liberties and eyes do you think? Or do you think it defers explanation too much?

jsonn:
What about restructuring the capture section a bit?

Step 1: Surround the white stone.
Success: Good job! A free space next to a stone is a liberty. Without liberties, a stone is removed.
Step 2: White wants to capture you, run out!
[Success: Well done. If your stones are next to each other, they count as group and share liberties.]
Step 3: Protect your stone by connecting to your friends! (white on B4, C3, C4, D3, D4; black on B5, B7)
[Success: Excellent. If you connect your groups, they often become stronger!]
Step 4: The edges of the board can be your friend. Capture white! (just a white stone on A1)
[Success: Correct, the edge of the board doesn't count as liberty. Isn't it nice to hunt groups with only one liberty?]
Step 5: Poke the eye to capture! (like current 4)
[Success text: Perfect, but remember: A group without liberties is only allowed if they capture.]
Step 6: Suicide is forbidden. Find a valid move! (checkers pattern for part of the board, group with only one liberty, two connected free spaces in the middle)
[Success: You've mastered the basics for capturing stones. Want to learn survival next?]

anoek:
I'm thinking we might need to setup some A/B testing for this, the primary reason I didn't introduce "Success" messages is because it's a halt and that much more for folks to read and process. I might be off basis, and maybe we need more words on the screen to explain things, but I feel like we might need to put together a quick A/B test and see how new players react because my gut says less is more in this case.

I'm contemplating setting up a test we can control a bit using mechanical turk so we can put theories to test.

anoek:
Going to track this in #975 instead of a pull request

PR #843, opened 2019-07-23
let vs computer challenge modal have disable analysis checkbox

### LINKED ISSUE #842: Show toggle "disable analysis" checkbox on vs computer challenge modal
### Summary

Right now, when you open the challenge modal to play a computer, the checkbox to disable analysis is not shown. 

A workaround for this exists by pressing the "create" button, toggling the disable analysis checkbox to the desired state, creating the challenge and then cancelling it. A subsequent game against a bot will have analysis disabled depending on whether or not the box was checked in the cancelled challenge.

I have submitted a pull request that fixes this issue.

flurmbo:
Fixes  #842. Removes a line that conditionally renders the disable analysis checkbox so that it is always rendered regardless of challenge mode. I've also dedented the jsx that was previously enclosed in the condition.

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -925,64 +925,62 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                     </div>
                 </div>
 
-                {(mode !== "computer" || null) &&
-                    <div>
-                        <div className="form-group" style={{position: "relative"}}>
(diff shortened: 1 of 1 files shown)


anoek:
Perhaps we should instead make it so we always enable analysis when challenging a computer? I feel like that is the expected behavior

flurmbo:
Hi @anoek, personally I like playing the computer with analysis disabled sometimes so I have to practice my reading and not rely too much on the analysis board. I'm not sure what other people do though

anoek:
Ok sure why not, thanks for the PR!

PR #847, opened 2019-07-29
Add chat window to group view

flovo:
At the moment, the group chat is only accessible in the chat view. Adding a chat window to the group page could increase the usage of group chats and increase the activity of users within groups.

This pull request adds an EmbeddedChat window to the group page. 

It is basically a copy and paste of the implementation on the tournament page.

<details>
<summary>Screenshots</summary>

</details>

### DIFF
--- a/src/views/Group/Group.styl
+++ b/src/views/Group/Group.styl
@@ -88,4 +88,14 @@
     button {
         margin: 0 0.5rem 1rem 0;
     }
+
+    .EmbeddedChat {
+        flex-shrink: 0;
+        flex-grow: 1;
+        min-width: 15rem;
+        flex-basis: 15rem;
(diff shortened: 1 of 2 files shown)


flakeman2:
I like this idea a lot.

anoek:
Cool :) We should probably hide it though when the group is private and the user is not a member

flovo:
I added a check for hide_details as well, since group membership is part of the hidden details.

flovo:
Is there still anything I should change?

anoek:
Sorry about that, I think this is good, merging now :)

BHydden:
Gonna have to lift my game! With this merge flovo jumped me on commits ;-)

Issue #851, opened 2019-08-04
Advance through all active games


BHydden:
### Summary

Add all active games to the tail end of the cycle your move button when no active games are awaiting action.

Have the "your move indicator" stay at 0 when you have >0 active games and =0 games to move in, and have the functionality of "on_click" change from "cycle games where your move" to "cycle active games"

### Additional Information:

https://forums.online-go.com/t/advance-through-all-active-games/22208

BHydden:
only task remaining is to fix the notification so that move counter displays 0 if pool is 'active' not 'your move'

BHydden:
Closing via #852

PR #852, opened 2019-08-04
Advance through all active games

BHydden:
This is the start of my attempt to implement #851 

However, my efforts have had the opposite effect... instead of adding cycle active games, it seems the notification doesn't show up at all anymore.

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -34,6 +34,7 @@ declare let Notification: any;
 
 interface Events {
     "turn-count": number;
+    "total-count": number;
     "notification": any;
     "notification-list-updated": never;
     "notification-count": number;
@@ -242,17 +243,17 @@ class NotificationManager {
     advanceToNextBoard(ev?) {{{
         let game_id = getCurrentGameId() || 0;
         let board_ids = [];
+        this.all_boards_waiting = false;
         //notificationPermissionRequest();
         for (let k in this.boards_to_move_on) {
             board_ids.push(parseInt(this.boards_to_move_on[k].id));
-            this.all_boards_waiting = false;
         }
 
         if (board_ids.length === 0) {
             for (let k in this.active_boards) {
                 board_ids.push(parseInt(this.active_boards[k].id));
-                this.all_boards_waiting = true;
             }
+            this.all_boards_waiting = true;
         }
 
         if (board_ids.length === 0) {
@@ -358,6 +359,7 @@ class NotificationManager {
             }
 
             this.event_emitter.emit("turn-count", Object.keys(this.boards_to_move_on).length);
+            this.event_emitter.emit("total-count", Object.keys(this.active_boards).length);
         });
 
         comm_socket.on("notification", (notification) => {
@@ -487,6 +489,10 @@ export class TurnIndicator extends React.Component<{}, any> { /* {{{ */
 
         notification_manager.event_emitter.on("turn-count", (ct) => {
             this.setState({count: ct});
+        });
+
+        notification_manager.event_emitter.on("total-count", (tt) => {
+            this.setState({total: tt});
         });
     }
 
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -245,18 +245,18 @@ class NotificationManager {
             board_ids.push(parseInt(this.boards_to_move_on[k].id));
         }
 
-        for (let k in this.active_boards) {
-            if (this.boards_to_move_on.length !== 0) {
-                return;
+        if (board_ids.length === 0) {
+            for (let k in this.active_boards) {
+                board_ids.push(parseInt(this.active_boards[k].id));
             }
-            board_ids.push(parseInt(this.active_boards[k].id));
         }
-
-        board_ids.sort((a, b) => { return a - b; });
 
         if (board_ids.length === 0) {
             return;
         }
+
+
+        board_ids.sort((a, b) => { return a - b; });
 
         let idx = -1;
         for (let i = 0; i < board_ids.length; ++i) {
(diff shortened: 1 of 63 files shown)


BHydden:
I suspect line 314 is where I went wrong... probably should have paid closer attention when anoek said 
~~~
delete it when game.phase === "finished"
~~~
Will look at tweaking that when I get back to it.

BHydden:
Seems to have returned to how it was before I touched it :P normal 'your move' notifications are working again but the new 'cycle active games' is still totally non-functioning ☹️

BHydden:
discovered that the 'your move' notification shows up, but doesn't work the way it's meant to (clicking does nothing) so I'm not as close to square 0 as I thought :P #justkeephacking

src/components/Notifications/Notifications.tsx
```diff
@@ -242,6 +244,14 @@ class NotificationManager {
         for (let k in this.boards_to_move_on) {
             board_ids.push(parseInt(this.boards_to_move_on[k].id));
         }
+
+        for (let k in this.active_boards) {
+            if (this.boards_to_move_on.length !== 0) {
+                return;
```
flovo: If there are moves to play, do nothing (leave the function without advancing to the next game)

BHydden: Yeah I had a feeling the problem was in this area, just hadn't had the chance to sit down and read through to find what a stuffed up :-P thanks

BHydden:
only task remaining is to fix the notification so that move counter displays 0 if pool is 'active' not 'your move'

src/components/Notifications/Notifications.tsx
```diff
@@ -541,6 +562,7 @@ export class NotificationList extends React.Component<{}, any> { /* {{{ */
     render() {
         return (
             <div className="NotificationList">
+                //I think this line below is messing with display of the notification when set to 0
```
BHydden: needs cleanup

BHydden: fixed

src/components/Notifications/Notifications.tsx
BHydden: fixed with 9177ce2

BHydden:
Clicking on the space where the "your move on x boards" icon should be always works now. 
If it is your turn on >0 boards, clicking it will cycle through them.
If you have games but none of them are your turn, clicking it will cycle through all games.

...but for the life of me, the only way I can get a "0" in the notification is when I accidentally break something and then there's always a "0" there even if you have no games at all, and when I have it working properly I can't get the "0" notification to show :(

BHydden:
very open to any ideas people might have about how I can get it to render correctly :P

src/components/Notifications/Notifications.tsx
```diff
@@ -454,7 +478,9 @@ export class TurnIndicator extends React.Component<{}, any> { /* {{{ */
     constructor(props) {
         super(props);
         this.state = {
-            count: Object.keys(notification_manager.boards_to_move_on).length
+            count: Object.keys(notification_manager.boards_to_move_on).length,
+            total: Object.keys(notification_manager.active_boards).length,
```
flovo: You assign a value to `total` only in the constructor. At this point `total` is always `0`.

For `count` it was done with: 
https://github.com/online-go/online-go.com/blob/aa1fc2028307381aae0aeb427629f0992f20fa17/src/components/Notifications/Notifications.tsx#L36
https://github.com/online-go/online-go.com/blob/aa1fc2028307381aae0aeb427629f0992f20fa17/src/components/Notifications/Notifications.tsx#L360
https://github.com/online-go/online-go.com/blob/aa1fc2028307381aae0aeb427629f0992f20fa17/src/components/Notifications/Notifications.tsx#L488-L490

BHydden: I don't think this is the case, it is still updating somehow (though don't ask me why) because the "it's your turn to move" counter still works properly. Sorry if I wasn't clear but that half still renders, the only thing the doesn't render is the 0 in the circle when you have active games but not any that are your turn...

src/components/Notifications/Notifications.tsx
```diff
@@ -304,6 +321,9 @@ class NotificationManager {
         });
         comm_socket.on("active_game", (game) => {
             delete this.boards_to_move_on[game.id];
+            if (game.phase === "finished") {
+                delete this.active_boards[game.id];
```
flovo: I would remove the `if` and leave the `delete` to always `delete this.active_boards[game.id];` It seems more consistent with what we do for `boards_to_move_on`. If the game is supposed to be in active_boards, we readd it anyway.

BHydden: I had the same feeling, but in the forum thread anoek advised to write it this way, and this part of the code is working as desired. It is only rendering 0 moves while total games > 0 and boards_to_move_on = 0 that is currently not working properly.

src/components/Notifications/Notifications.tsx
flovo: when you change 
`                 <span className={this.state.total > 0 ? "active count" : "count"}><span>{this.state.count}</span></span>`
to
`                 <span className={this.state.total > 0 ? "active count" : "active count"}><span>{this.state.total}</span></span>`

you can see that `total` is always `0`.

BHydden: ok yes I see that you're right about this point and I will fix it now, but I can't see how it would cause the specific rendering issues I'm having... 😕

BHydden: well I still have no idea why, but that magically fixed everything :D haha it's not working perfectly 😆 

Thanks a lot @flovo ❤️

BHydden:
@anoek  thanks to some help from @flovo I think this is finally ready to ship :)

anoek:
Awesome, thanks @BHydden and @flovo :)

PR #853, opened 2019-08-06
Fix for incorrect custom handicap warning

flovo:
As discussed in https://forums.online-go.com/t/please-remove-that-for-games-without-handicap/22447?u=flovo there is a bug for the "Custom handicap" warning when using another language than English.

When using another language, every open challenge had the "Custom handicap" warning attached.

This should fix this.

<details>
<summary>Screenshot</summary>
Before:

After:

</details>


GreenAsJade:
Nice work!  Thanks!

anoek:
Thanks!

Issue #856, opened 2019-08-12
AI marks are visible after "disable AI review"


flovo:
### Summary

When the user disables AI review in Game view, the AI marks are only deleted for the current move. For other moves the marks are still visible.

### Steps To Reproduce:

1. View a game with (full) AI review
2. View multiple moves while AI review enabled.
3. Disable AI review (right panel)
4. The marks for all visited moves (but the last visited move) are still visible.

### Additional Information:

This was reported on the forum at least once https://forums.online-go.com/t/disable-ai-review-does-not-hide-the-winrate-changes/22510?u=flovo

anoek:
Thanks for the fix!

PR #857, opened 2019-08-12
Revome marks from all moves when AI review disabled

flovo:
This is a fix for https://github.com/online-go/online-go.com/issues/856

When the user disables AI review in Game view, the AI marks are only deleted for the current move. For other moves the marks are still visible.

To remove the AI marks for all moves we have to go over the whole tree.


anoek:
I merged because this is better than the current behavior, I think I was hesitant because if the user does happen to have any custom marks, they will get wiped out with this.. but, I think the 99.9% use case is they won't but probably will have stuff left over from AI so probably good to do this.

Issue #858, opened 2019-08-12
"disable AI review" also jumps to the last official move


flovo:
### Summary

When the user disables AI review (right panel), the current move changes to the last official move. 

### Steps To Reproduce:

1. Go to Game View
2. Go to any but the last move
3. Click on "Disable AI review" in the right panel
4. We are now on the last official move.

### Additional Information:

https://forums.online-go.com/t/disable-ai-review-does-not-hide-the-winrate-changes/22510

flovo:
This is probably caused by https://github.com/online-go/online-go.com/blob/93c70b6db82c66c446db7600014043207834e6be/src/views/Game/Game.tsx#L973

anoek:
Thanks for the fix!

PR #859, opened 2019-08-12
Don't jump to last move when disable AI reveiw

flovo:
Don't switch "Play" mode after the user disabled AI review.

When using "Disable AI review" we shouldn't change the game mode, since the user probably want to keep doing what they were on.

Fix for https://github.com/online-go/online-go.com/issues/858


anoek:
Indeed, thanks!

PR #867, opened 2019-08-14
Change top 3 game changing moves to worst 3 moves

flovo:
In my opinion, a game changing move is a move which deviates much from perfect play. Since we don't know what perfect play would look like, we approximate it by using an AI. 

Positive win rate changes are an artefact of the limited AI, not a sing for a game changing move. Therefore, we shouldn't include positive changes in the Top 3. (positive changes are rarely in the top 3 anyway, since their abs is usually small)


anoek:
I think I agree with this, I'm going to update the backend to match for the quick top 3 analysis at the same time as I merge and roll this out I think.

anoek:
Thanks for this! Sorry it took so long to merge

PR #869, opened 2019-08-14
remove trailing whitespaces

flovo:
With this the shippable checks shouldn't fail any more.


anoek:
Oh dang, yeah my mistake - teach me to use the github editor. Sorry about that, usually I get an email yelling at me when I break things... not sure why I didn't this time, unless it was an artifact of again using the github editor..

Anyways, thanks!

Issue #849, opened 2019-08-03
Server Time / Weekend indicator


BHydden:
### Summary

"Can we please add a server date/time indicator somewhere? That way I could know when weekend pauses start and end. Even niftier would be a “weekend starts/ends in _ hours” indicator that only shows on friday/sunday." 
 - smurph

Found this and wanted to throw it up... might be something I'll look into coding up soonish. Thinking it could just be a simple server clock somewhere on the home page and profile page.
I only feel confident coding up the first suggestion so if anyone else feels like implementing the "weekend starts / ends" stuff feel free to chime in.

### Additional Information:

https://forums.online-go.com/t/server-time-weekend-indicator/18731

BHydden:
I got this project up to "ready to submit to anoek for checking over" today... but silly me did it in the same branch as the other project I'm working on... so, if I can split the branches successfully tomorrow... there should be a new PR going up then... :P

anoek:
git cherrypick is your friend :)

BHydden:
Thanks! I was going to attempt some convoluted copy backup - > git revert deal but I'll look into your suggestion first ❤️

BHydden:
#878 merged.

Closing.

PR #878, opened 2019-09-03
Add server time display

BHydden:
Display current server time above active games on profile page as per [Issue #849](https://github.com/online-go/online-go.com/issues/849)

### DIFF
--- a/src/components/ServerTimeDisplay/ServerTimeDisplay.styl
+++ b/src/components/ServerTimeDisplay/ServerTimeDisplay.styl
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2012-2019  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.ServerTimeDisplay {
+    div {
+        padding-top: 50px;
+        padding-bottom: 50px;
+    }
+}
+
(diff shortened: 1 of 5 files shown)


BHydden:
@anoek if you like the look of this I'll begin work on implementing it on the /overview page as well and add in a "weekend starts in x hours / end in x hours" for Friday and Sunday but I wanted to get your feedback on what I've done so far.

BHydden:
I believe this version to be functional and stable and could be merged now if you like what you see, as the additional changes I'm thinking of making are, I believe, QoL feature upgrades and not something to hold back deployment if you like what I've done so far...

But by all means, if you have feedback on optimal wording / location / formatting for the server time display I'm all ears 😄

BHydden:
Currently looks like:
image

BHydden:
@anoek with the latest changes I now think this covers everything I wanted it to be, and everything I interpreted was desired from the initial forum thread. Please let me know if you have any recommendations, requests, or ideas for improvement. Otherwise, I think it's good to go. 😄

anoek:
I like that you implemented that as a component, good job :) I'm not convinced that's the best place for that information to live, but it should be pretty easy to move if we figure out a better place for it because you did that.

The primary issue I see is that the strings need to be translatable, see usages of `_` and `interpolate`.

BHydden:
Regarding location, I just tried to put it as close to where players would be looking at their own games as possible, which to me seems like the main time you would be concerned about weekend presence, but as you say it's easy to move anywhere we want.

Hmm ok... I've not done any research into that yet but I'll see how I go... how deep does that need to go? Do the numbers get translated into Asian numbers? Would the day of the week need to be translatable? ...or is it just the preamble at the start that you want to be translatable?

anoek:
`moment` will with translating times into the various formats and localizing the day names and whatnot. This has two implications: (1) you don't need to worry about translating that part, but (2) `this.state.time.startsWith("Saturday")` doesn't work, you should probably just change that to `new Date().getUTCDay() == 6` (6 for Saturday, 0 for Sunday, 5 for Friday)

You can see some examples of `interpolate` and `_` in https://github.com/online-go/online-go.com/blob/devel/src/views/Supporter/PrettyTransactionInfo.tsx#L57 for the rest of the string translations

BHydden:
I wasn't sure if it would update properly if it wasn't attached to this.state.time I was trying to avoid the case where a user has their browser open all the time and the second half of the message never updates... is that an invalid concern born from my ignorance of the finer details of the language?

BHydden:
Does that look right to you? 😄

anoek:
Looking pretty good I think, might want to change the moment format from `dddd, HH:mm:ss` to `dddd LT` though just so the time is presented nicely dependent on locale

BHydden:
*sigh* 😄 I'm more of an ISO guy myself, but I suppose localisation makes the most sense in this context... roger wilco 👍

BHydden:
@anoek for some reason GitHub didn't kick off the auto-checks with that last push?

anoek:
Thanks for the PR BHydden!

Issue #874, opened 2019-08-22
[AI review] what is the usecase for the heatmap?


flovo:
We are highlighting some moves with bright green squares. The brightness of those squares is determined by the values of LZ policy network (that's what LZ returns as heatmap https://github.com/leela-zero/leela-zero/issues/1571). 

While the values of the policy network are correlated with the number of visits, they are not always similar. 
A rather extreme example is move 156 in https://online-go.com/game/18588826. There the most explored variation is N6 with [1461/2000](https://online-go.com/termination-api/game/18588826/ai_review/335050) visits, while the background appears not to be green at all.

![image](https://user-images.githubusercontent.com/4864182/63546936-3aaa0f80-c52b-11e9-8c77-c4c00040ed4a.png)

This already let to some irritation: https://forums.online-go.com/t/disable-ai-review-does-not-hide-the-winrate-changes/22510/4?u=flovo

anoek:
A fair question. I thought it was interesting to see what leela's hunches were, I took it as "These positions caught Leela's eye". However, now I'm seeing that it can be very confusing because I also was under the impression that this was going to correlate a bit more with how much leela explored a position, and I think normally it does but in cases like this it makes it look like leela really checked out K9 but picked N6 because of whatever reason - where the reality is she explored N6 quite a lot, and also K9 sufficiently, but not nearly as much.

So, now I'm thinking maybe we should ditch the heatmap and instead maybe draw our own heatmap using the visit count on the moves we get, seems like that would align more with the intuition of what people are expecting with a heatmap.

PR #880, opened 2019-09-04
Use visits for heatmap instead of policy network

flovo:
resolves online-go/online-go.com#874

I also changed the normalization of the heatmap to be relative to total visits. With this it's easier to see how many visits the AI spent on a variation. 

1) Policy network:
    image

2) Visits:
    image

3) Visity / total visits
    image


anoek:
Nice, yeah that's much better, thanks!

Issue #865, opened 2019-08-14
AI marks overlap with territory marks


flovo:
### Steps To Reproduce:

1. Go to a game won by score and a full AI review.
3. See that the AI marks and the territory marks are overlapping.

### Additional Information:

![image](https://user-images.githubusercontent.com/4864182/63007585-33c33300-be81-11e9-9eb1-4deb965df5fb.png)

---

A possible fix could be to don't show the AI marks while we are in "Play" state, but only if we are in "Analyze Mode".

flovo:
The marks from Analyze Mode are carried over as well.

We should clear/hide all marks but the territory marks in play mode.

Issue #283, opened 2017-05-18
Last mark style from analysis persists to moves


Capt-Brunch:
### Summary

If you enter analysis mode and pick one of the symbols (square, X, O, triangle) during your move and then return to game, the translucent stone that follows your cursor retains the last symbol you had selected.

### Environment

 * Operating system: Windows 7
 * Browser: Google Chrome
 * Browser Version: 58.0.3029.110 (64-bit)

### Steps To Reproduce:

1. On your move in a game, enter analysis mode
2. Select one of the four symbols X, O, triangle, square
3. Return to the game by clicking "Back to game"
4. Note that the stone following your cursor for your next move now has the last symbol you used superimposed on it.

### Additional Information:

Screenshot attached
![ogs symbol](https://cloud.githubusercontent.com/assets/1940202/26220623/560c6e7e-3be1-11e7-91b5-4358261c4fd3.PNG)

alex-mac:
In development mode, is there a way to create a game against an opponent? I'm having a bit of difficulty recreating this in my dev environment.

anoek:
This issue has been marked stale and will be closed soon without further activity. To keep the issue open, please respond to the comment to keep the discussion going.

PR #884, opened 2019-09-06
Show analyze marks only in analyze mode

flovo:
fixes gh-283
fixes gh-865

please check if I've hidden some marks by accident which should be visible outside of analyze mode.

only analyze mode marks:
- triangle
- circle 
- cross
- square
- numbers/letters
- ai color circles 
- heatmap
- sub_triangle

always shown:
- chat_triangle (tested)
- last move marking (tested)
- altmarking (I don't know what this is)

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -2284,7 +2284,7 @@ export abstract class Goban extends TypedEventEmitter<Events> {
 
         /* Heatmap */
         /* {{{ */
-        if (this.heatmap) {
+        if (this.heatmap && this.mode === "analyze") {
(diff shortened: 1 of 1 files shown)


anoek:
Unfortunately this breaks marks on puzzles and the upcoming joseki board, and it makes it so the ai analysis stuff isn't visible when a game is over until there's user interaction.

Gonna have to ponder more on what we need to do to make things more robust so we can fix these issues and still have a good way of marking up the board outside of the generic analysis mode.

flovo:
Haven't checked puzzles and joseki. 

About hiding the on board AI stuff, they overload the board with information when territory marks are present, so I thought of hiding them as an additional feature.

Issue #813, opened 2019-06-10
feature request: score estimator should be disabled when analysis is disabled


alexmitrani:
Good afternoon

The score estimator is an analysis tool so should be disabled when analysis is disabled in the match options.  

I don't know about other countries but in Chile the online tournament regulations (http://www.igochile.cl/igochile2/home/reglamentos/) forbid the use of analysis tools.  For that reason the Federación Chilena de Go presently prefers KGS for online tournaments.  I would like it to be possible for online tournaments in Chile to use OGS in future therefore I am making this feature request.  

This was also discussed in detail here: https://forums.online-go.com/t/score-estimator-should-be-optional-when-defining-a-match/18052

I look forward to your reply.  

Kind regards

Alex Mitrani
Federación Chilena de Go

anoek:
Yep this came up with AGA/EGF tournaments as well. I think the current plan is to add a new option to disable the score estimator.

PR #885, opened 2019-09-06
Disable score estimator when analysis is disabled

flovo:
resolves gh-813

The score estimator is an analysis tool so should be disabled when analysis is disabled in the match options.


anoek:
Thanks for the PR!

raylu:
if I'm reading this right, this only affects users logged in as one of the players. logging out (in a separate tab) allows you to use the score estimator?

alexmitrani:
many thanks for implementing this @flovo and @anoek

Issue #882, opened 2019-09-04
[Score estimator] individual player scores are useless/wrong in games ended with score


flovo:
### Summary

When using “estimate score” on a finished game (ended by scoring), the individual scores for black and white are just useless/wrong. They are prisoners (+ komi).

This is a cosmetic issue. The estimated score difference is correct.

### Steps To Reproduce:

1. Go to a game won by score (https://online-go.com/game/19350057)
2. Click “estimate score”
3. See that the individual scores don't reflect the state of the board.

### Additional Information:

![Untitled](https://user-images.githubusercontent.com/4864182/64292875-a6e53400-cf6b-11e9-8c39-45c9e168b1bb.png)

flovo:
To prevent any confusion, it would probably be best to hide the captures/points line when SE is on (, or to show the capture count/points of the SE instead).

anoek:
Fair point

PR #886, opened 2019-09-06
Hide captures/points while score estimator active

flovo:
resolves gh-882

The captures/points don't reflect the result of the SE, so best is to hide it.


anoek:
It's true it's always reporting territory which is not good, but when the board is filled in it provides a quick way to count, and the difference in estimated score should be pretty close to what the real difference in score is, so I'd argue while the point count is wrong, the result is still useful at times.

If we can put the estimated difference somewhere then I think we can remove those scores, or just leave them as the official captures+komi or something

flovo:
This doesn't hide the estimated score. The result of the SE is okayish.

It's about the players individual captures/scores.

As you can see in the screenshot, those numbers don't reflect what the SE calculated (neither the total points nor the difference of them).
The numbers are
- if the game hasn't ended yes: the number of captured stones (not including the stones the SE treated as dead)
- if it's a finished game: captured stones + Komi (again not including the SEs estimated deaths)

anoek:
Perfect, sorry I should have tried that first.

Issue #792, opened 2019-05-27
Ability for mods to interact with scoring phase


BHydden:
It would also be nice to have the ability to manually altar the score during scoring phase when we are called for a dispute, possibly included could be a lockout from the players editing while we edit.

BHydden:
auto-scoring with kata now exists, so this is redundant.

PR #887, opened 2019-09-07
allow mods to use Auto-Score button

BHydden:
Partial implementation for #792 

This PR implements the ability for mods to interact with the Auto-Score button to return the board back to a neutral position to better identify living or dead groups in the event of trolls erroneously marking groups alive/dead.

There is room for more to be done in this area, such as enabling mods to mark stones alive/dead and disabling players from being able to mark stones, but I feel what I have already done would be a huge help in such cases.

### DIFF
--- a/src/lib/ogs-goban/Goban.ts
+++ b/src/lib/ogs-goban/Goban.ts
@@ -4245,7 +4245,7 @@ export abstract class Goban extends TypedEventEmitter<Events> {
     } /* }}} */
     public autoScore() { /* {{{ */
         try {
-            if (!window["user"] || !this.on_game_screen  || !this.engine || (window["user"].id !== this.engine.black_player_id && window["user"].id !== this.engine.white_player_id)) {
+            if (!window["user"] || !this.on_game_screen  || !this.engine || (window["user"].id !== this.engine.black_player_id && window["user"].id !== this.engine.white_player_id && window["user"].is_moderator !== true)) {
                 return;
             }
         } catch (e) {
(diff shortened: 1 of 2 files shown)


BHydden:
huh... I just realised my implementation on shows on the mods screen... not sure if that's a bug or a feature :P I'm sure I could track down the trigger to push through the action to the players as well... but it might actually beneficial being able to see the neutral state without the troll knowing?

anoek:
That's because the server as is won't let you change the state of the game since you're not a player :) But it would be nice to update that so moderators could. So keep this around, if you want to work on manually marking dead stones that'd be swell too, i've got a queue of server stuff i'm working on now but if you do the front end i'll get the backend done.

BHydden:
ok if I'm going to be digging into all that, are you for or against mods being able to lock players out of the mark dead stones tool? and would such a feat have a front-end component or just be server-side?

BHydden:
follow-up question; are you totally opposed to this going in in its current purely informational, non-impacting form? I think even the changes as they are could be very useful to the mods while we nut out the bigger changes, and I believe Adam and Conrad shared a similar sentiment in the forum.

anoek:
Score estimator is the same code as the auto-score, so we currently the purely informational form I'd say, no?

BHydden:
Hmm... sounds fiddly... so mods will also need a hard bypass for the new "disable SE when analysis is diabled" 

k I'll start looking into getting the front-end polished off soonish then

flovo:
SE should only be disabled for the players (like analyze game).

anoek:
I'm going to finally close this out PR I think - autoscoring with kata is in the pipe, and as that is rolled out I'll add this capability in, but I'll likely be mucking with the UI along the way so it's likely to conflict.

PR #889, opened 2019-09-08
Turn-indicator: theme update

flovo:
I added a dark theme for the turn indicator
and a grayed version if the turn counter shows `0`

On the dark theme the turn-indicator had a bright white border (#fff). That's much brighter than everything else on the dark theme (text color #ddd)

I changed the color of the turn-indicator circle to match the text-color of the theme. The number within the turn-indicator now has the same color as the top bar.

Since the border would have the same color as the background for both themes, it's now just set to be transparent.

When the turn-indicator  shows `0` it has the same color as the friends icon when no friends are online.

turn-indicator

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -497,7 +497,7 @@ export class TurnIndicator extends React.Component<{}, any> { /* {{{ */
     render() {
         return (
             <span className="turn-indicator" onClick={this.advanceToNextBoard}>
-                <span className={this.state.total > 0 ? "active count" : "count"}><span>{this.state.count}</span></span>
+                <span className={this.state.total > 0 ? (this.state.count > 0 ? "active count" : "inactive count") : "count"}><span>{this.state.count}</span></span>
             </span>
        );
     }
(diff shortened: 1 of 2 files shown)


src/components/Notifications/Notifications.tsx
```diff
@@ -497,7 +497,7 @@ export class TurnIndicator extends React.Component<{}, any> { /* {{{ */
     render() {
         return (
             <span className="turn-indicator" onClick={this.advanceToNextBoard}>
-                <span className={this.state.total > 0 ? "active count" : "count"}><span>{this.state.count}</span></span>
+                <span className={this.state.total > 0 ? (this.state.count > 0 ? "active count" : "inactive count") : "count"}><span>{this.state.count}</span></span>
```
BHydden: I wasn't aware you could add an extra condition to this statement...

is this how it works?
```typescript
this.state.a ? this.state.b ? b.true : a.true : both.false
```
or something else?

flovo: Be aware of the parenthesis. It's simply a `_?_:_` nested into another `_?_:_`.

BHydden: Oh lol thanks I didn't see the parenthesis under the highlighting :rofl:

anoek:
Yeah good stuff, thanks!

PR #890, opened 2019-09-09
Chat improvements

flovo:
Now the chat recognises when an username is followed by a `,` or a `'` as a mention of username.
If the user is mentioned as `player 1`, now this is a mentioning as well.

I also made the space between `player` and the player_id optional. 
Same for `issue` and `pr`.

This still leaves some mentions undetected:
In German `Adam's` would be `Adams` (without Apostrophe). I don't know if there is any language which adds letters at the beginning of a name to indicate anything.
It could be better to remove the checks for word boundaries entirely.

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -56,6 +56,7 @@ data.watch("config.user", (user) => {
     name_match_regex = new RegExp(
           "\\b"  + cleaned_username_regex + "([?:,.!*'\\s])"
         + "|\\b" + cleaned_username_regex + "$"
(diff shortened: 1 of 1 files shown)


flovo:
Before I start adding hundreds of Unicode characters which could follow direct after username, what about simply checking for word boundaries (\b) after a username?

anoek:
Yeah that seems like it should work? I'm not sure why I didn't do that to begin with...

flovo:
It seems to work
image

flovo:
image

anoek:
Thanks for the PR!

Issue #892, opened 2019-09-10
Auto-advance jumps from the your-turn tree into the active-games tree


BHydden:
**Describe the bug**
Post-implementation of PR - #852 it was discovered that the "auto-advance" feature would automatically move a player from their last played move to their first active game.

**To Reproduce**
1. Have Auto Advance turned on and multiple correspondence games, in which at least one is your turn
2. Play a move in all the games that it is your turn
3. See that after your last move is played the server still auto advances you into the new all active games tree.

**Expected behavior**
Auto-advance should only move you to a game if it is your turn to move in that game. Progressing into the all-active-games tree should be a purely manual decision.

**Additional context**
For the life of me, I can't find where the auto-advance class lives in order to fix this 🤦‍♂ 
The issue was first reported in [this forum post](https://forums.online-go.com/t/zero-game-waiting/22781/21) and the 3 following comments.

flovo:
Have you found:

https://github.com/online-go/online-go.com/blob/4c8a23ee74e7a990a4e39890f7198c9f7a4f865d/src/views/Game/Game.tsx#L464

https://github.com/online-go/online-go.com/blob/4c8a23ee74e7a990a4e39890f7198c9f7a4f865d/src/lib/goban.ts#L32-L36

https://github.com/online-go/online-go.com/blob/4c8a23ee74e7a990a4e39890f7198c9f7a4f865d/src/lib/goban.ts#L109-L120

BHydden:
Thanks :) found the first one but didn't find the lib files. Looks like I should be able to fix it with an extra condition in online-go.com/src/lib/goban.ts#L115

BHydden:
pending anoek approval, this should be good to go

PR #895, opened 2019-09-11
Only auto-advance if your turn

BHydden:
PR to resolve issue #892 

WIP - LF feedback @anoek 

I get a Type Error preventing the site from loading when I try to communicate with src/components/Notifications/Notifications.tsx from src/lib/goban.ts

https://i.imgur.com/Vc9mkqJ.png

Looks like some kind of compiling dependency issue of things loading in the wrong order, could you point me in the right direction? I was thinking of moving the whole autoadvance function from lib/goban over to Notifications... is that the right idea or will that not resolve this dependency issue / create new issues?

### DIFF
--- a/src/lib/goban.ts
+++ b/src/lib/goban.ts
@@ -107,10 +107,12 @@ export class Goban extends OGSGoban {
         return getSelectedThemes();
     }
 
+        /*
     getBoardsToMoveOn() {
-        let count = notification_manager.boards_to_move_on.length;
-        return count;
-    }
+        let length = notification_manager.boards_to_move_on.length;
+        return true;
+    } 
+        */
 
     autoadvance = () => {
         let user = data.get('user');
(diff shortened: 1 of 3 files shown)


BHydden:
I can't believe it passes the shippable test when the code breaks the whole site 🤣

anoek:
Yeah that's a dependency problem

You could move that logic to https://github.com/online-go/online-go.com/blob/devel/src/views/Game/Game.tsx#L464 where we listen for that event maybe?

BHydden:
Ok thanks, I'll give that a go

src/lib/goban.ts
```diff
@@ -106,20 +105,6 @@ export class Goban extends OGSGoban {
...
-        if (!user.anonymous && /^\/game\//.test(this.getLocation())) {
-            /* if we just moved */
-            if (this.engine.playerNotToMove() === user.id) {
-                if (!isLiveGame(this.engine.time_control) && preferences.get("auto-advance-after-submit")) {
-                    this.emit("advance-to-next-board");
-                }
-            }
-        }
-    }
-}
```
BHydden: Deleted one too many }

BHydden:
Property 'engine' does not exist on type 'Game'.

BHydden:
Pretty sure this is good to go @anoek 😃 thanks for your help

anoek:
Thanks!

PR #899, opened 2019-09-16
Some defence against unexpected nulls

GreenAsJade:
These changes are intended to prevent the two Joseki related UI Sentry alerts that I have seen.

Each due to apparently-undefined variables that I hadn't realised could be undefined at that point.

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -706,7 +706,10 @@ export class Joseki extends React.Component<JosekiProps, any> {
         .then(res => res.json())
         .then(body => {
             // console.log("Tags Count GET:", body);
-            let tags = body.tags.sort((t1, t2) => (t1.group !== t2.group ? Math.sign(t1.group - t2.group) : Math.sign(t1.seq - t2.seq)));
+            let tags = [];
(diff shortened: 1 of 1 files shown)


BHydden:
Is this related to the sentry report I filled out?

src/views/Joseki/Joseki.tsx
```diff
@@ -1444,6 +1447,10 @@ class EditPane extends React.Component<EditProps, any> {
         // Extract markup for "board marks"
         // maps markup of form "<label:position>"  to an array of {label, position} objects for each mark
 
+        if (description === null) { // I don't see how, but Sentry logs seem to imply there is a way!
```
BHydden: Magic!

GreenAsJade: I'm not aware of a sentry report you filled out: I was responding to an email I got from Sentry, but I guess it could be the same?

BHydden: I used much more detail but essentially it was: 
go to Play tab, select a tag, select a contributor, change tag to the top option (empty) and the page crashes

anoek:
:+1:

PR #900, opened 2019-09-17
Add vacation days accrued to profile 

### LINKED ISSUE #389: Ability to see how many vacation days are to be accrued
### Summary

Displaying how many days are left to be accrued/acquired.
It's just a substraction, it should be a quick-win. And besides the supporter page, I don't know an OGS page that gives us the maximum allowed. 
So, if I've never been interested in visiting the supporter page, I can't know, whatever if I'm a supporter or not.
As a subsidiary point, it would be kind to display the rhythm of acquisition

### Environment

 any

### Steps To Reproduce:

Ask if needed

### Additional Information:

Ask if needed.

Arnaud

BHydden:
Resolves #389

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -601,6 +601,30 @@ export function durationString(seconds): string { /* {{{ */
 
     return ret.trim();
 } /* }}} */
+export function daysOnlyDurationString(seconds): string { /* {{{ */
+    let days = Math.floor(seconds / 86400);
+
+    function plurality(num, single, plural): string {
+        num = Math.round(num);
+        if (num > 0) {
+            if (num === 1) {
+                return " " + num + " " + single;
+            }
+            return " " + num + " " + plural;
+        }
+        return "";
+    }
+
+    let ret: string = "";
+    if (days) {
+        ret += plurality(days, _("Day"), _("Days"));
+    }
+    else {
+        ret += _("0 Days");
+    }
+
+    return ret.trim();
+} /* }}} */
 export function shortDurationString(seconds) { /* {{{ */
     let weeks = Math.floor(seconds / (86400 * 7)); seconds -= weeks * 86400 * 7;
     let days = Math.floor(seconds / 86400); seconds -= days * 86400;
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -606,9 +606,8 @@ export function daysOnlyDurationString(seconds): string { /* {{{ */
 
     let ret: string = "";
     if (days) {
-        ret += days + " " + ngettext(_("Day"), _("Days"), days);
+        ret += interpolate(_("{{number_of_days}} {{day_or_days_plurality}}"), {number_of_days: days, day_or_days_plurality: ngettext("Day", "Days", days)});
     }
-
     return ret.trim();
 } /* }}} */
 export function shortDurationString(seconds) { /* {{{ */
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -16,7 +16,7 @@
  */
 
 import * as data from "data";
-import {_, pgettext, interpolate} from "translate";
+import {_, pgettext, ngettext, interpolate} from "translate";
 import {TimeControl, TimeControlTypes} from "./TimeControl";
 
 const times = [ /* {{{ */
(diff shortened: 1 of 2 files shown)


BHydden:
This should be good to go, let me know if you want me to change anything :)

src/views/User/User.tsx
```diff
@@ -956,6 +968,11 @@ export class User extends React.PureComponent<UserProperties, any> {
 
                     <h2>{_("Activity")}</h2>
                     <Card className="activity-card">
+                        {(!user.supporter && !user.is_moderator && !user.is_superUser) && <h4>{_("Vacation Accrued: (Non-Supporter)")}</h4>}
+                        {(user.supporter || user.is_moderator || user.is_superUser) && <h4>{_("Vacation Accrued: (Supporter)")}</h4>}
+                        {(user.on_vacation) && <div >{_("User On Vacation")}</div>}
+                        {(!user.on_vacation && (user.supporter || user.is_moderator || user.is_superUser)) && <div >{this.state.vacation_accrued_text} {_(" out of 60 Days")}</div>}
```
BHydden: @GreenAsJade suggested it would look less cluttered without displaying the rate of increase. So if you agree, merge this commit otherwise merge the previous commit to keep that info displayed.

src/views/User/User.tsx
```diff
@@ -956,6 +968,11 @@ export class User extends React.PureComponent<UserProperties, any> {
...
+                        {(!user.on_vacation && (!user.supporter && !user.is_moderator && !user.is_superUser)) && <div >{this.state.vacation_accrued_text} {_(" out of 30 Days")}</div>}
```
BHydden: @GreenAsJade suggested it would look less cluttered without displaying the rate of increase. So if you agree, merge this commit otherwise merge the previous commit to keep that info displayed.

anoek: Cool, yeah agreed

src/components/TimeControl/util.ts
```diff
@@ -601,6 +601,30 @@ export function durationString(seconds): string { /* {{{ */
...
+        }
+        return "";
+    }
+
+    let ret: string = "";
+    if (days) {
+        ret += plurality(days, _("Day"), _("Days"));
+    }
+    else {
+        ret += _(" 0 Days");
```
anoek: For translated strings, don't include spaces before or after the string. If you want a space before, do `" " + _("0 Days")`

BUT that's a moot point - this whole case should be dropped, `plurality(days, _("Day"), _("Days"))` is correct when `days === 0` and it saves a translation.

src/components/TimeControl/util.ts
```diff
@@ -601,6 +601,30 @@ export function durationString(seconds): string { /* {{{ */
 
     return ret.trim();
 } /* }}} */
+export function daysOnlyDurationString(seconds): string { /* {{{ */
+    let days = Math.floor(seconds / 86400);
+
+    function plurality(num, single, plural): string {
```
anoek: Remove this declaration, use `ngettext` or `npgettext` instead

src/components/TimeControl/util.ts
anoek: sorry not `plurality`, `ngettext`/`npgettext`

src/components/TimeControl/util.ts
BHydden: thanks a lot! I hadn't looked into how they work, that cleaned up a lot of LOC 😄

src/components/TimeControl/util.ts
BHydden: fixed 😃

BHydden:
Whoops! Silly me! Vacation accrued shouldn't be an interval since it doesn't build at the same rate that it drains... 

I'll try and fix that up tonight before you wake up if I have time after dinner.

BHydden:
@anoek this should be good to go unless there's anything else you wanted me to touch up? 😄

src/components/TimeControl/util.ts
```diff
@@ -601,6 +601,16 @@ export function durationString(seconds): string { /* {{{ */
 
     return ret.trim();
 } /* }}} */
+export function daysOnlyDurationString(seconds): string { /* {{{ */
+    let days = Math.floor(seconds / 86400);
+
+    let ret: string = "";
+    if (days) {
+        ret += days + " " + ngettext(_("Day"), _("Days"), days);
```
anoek: Two more things here: `ngettext` is a translation function, so we don't translate before we call that - it's just `ngettext("Day", "Days", days)`. And second thing - for translations, we can't assume the order X <translations> is the correct order in every language, so we need to use `interpolate` too

such as `interpolate(_("{{number_of_days}} {{day_or_days_plurality}}"), {number_of_days: days, day_or_days_plurality: ngettext("Day", "Days", days)})`

BHydden: That makes sense. I'll fix that up tomorrow morning. Thanks

BHydden:
Thanks for letting me make that change @anoek  😃  I know it probably would have been faster for you to just edit my commit yourself but I feel like I understand translation a lot better after working through that 😄

anoek:
Thanks for the PR!

BHydden:
Yay! ^_^ thanks for the merge :D I was hoping I'd find out before you left whether there was anything else I needed to fix 😛

PR #911, opened 2019-09-21
How to point at the beta joseki explorer server

GreenAsJade:
Instructions about how to point a local OGS client environment at the beta Joseki Explorer server.


BHydden:
Nice one.

GreenAsJade:
I think that given this solution is "set and forget", it's probably good enough, especially since any other solution might be worse for the very important people developing with their own local joseki explorer server ;)

PR #913, opened 2019-09-22
Make WebAssembly feature detection more robust

jreut:
When running Chromium with WebAssembly disabled, the `WebAssembly`
global is still defined, and so a `typeof` check isn't quite enough.
Instead, let's try to allocate a Memory region. This operation will
fail with `RangeError` if WebAssembly is disabled.

See: https://github.com/openbsd/ports/blob/77b70736cf794ac0ff7c420e4321017471799f6d/www/chromium/files/chrome#L57

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -218,10 +218,12 @@
         try {
             var script = document.createElement('script');
             script.id = "ogs_score_estimator_script";
-            if (typeof WebAssembly === "object")  {
+            try {
+                new WebAssembly.Memory({ initial: 0 });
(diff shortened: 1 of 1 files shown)


jreut:
Hello, is there anything you'd like me to do here? I can explain my intentions a little more, I can provide proof this doesn't break the WASM feature for supported browsers, or I can close this pull request if it's not likely to be accepted.

Thank you!

BHydden:
We have just one dev and he's on holiday currently. I'm sure he'll get back to sometime next week :-)

anoek:
Thanks @jreut!

PR #917, opened 2019-09-23
Chat subsitiution: Added /player/$player_id/

flovo:
I just realized that there are 2 different URLs for player profiles. I added the missing one to the chat substitutions.

Btw.: it seems both URLs are leading to the same page. Is there a reason to use 2 different URLs? (3 URLs if one wants to count /player/$id/$user_name as seperate URL)


BHydden:
My favourite, that I wish was used all the time, is /player/$id/$name

I can't really think of a reason why the profile button would need a different link to the external link to your own profile, especially since as you say they load the same page.

flovo:
> My favourite, that I wish was used all the time, is /player/$id/$name

https://github.com/online-go/online-go.com/blob/7b8c2903efb8c41c0bde0bb2e695c4c2dc6e1a90/src/routes.tsx#L120-L121

BHydden:
Currently, I am seeing links like this 

```
Game game 19672452
```

Could you please add in a check for if the user already typed "game" or "Game" then the extra "game" isn't added before $id

flovo:
Have to find out how to do it best. 

The space between “game” and the URL is causing problems.

flovo:
Now, there should be only 1 `Game`, `review`, ...
https://github.com/online-go/online-go.com/blob/7b8c2903efb8c41c0bde0bb2e695c4c2dc6e1a90/src/routes.tsx

BHydden:
champion!

anoek:
Thanks for the improvements!

PR #921, opened 2019-09-24
Joseki "Play mode" upgrade

GreenAsJade:
This brings a new "results" pane to the Play mode, where the user is told information about the number josekis they've played and how successful they are at playing them.

It also makes Play mode select the "Joseki" filter, if none is selected, rather than just opening the filter pane, because doing it this way is a lot more pleasant experience.'


GreenAsJade:
This requires server commit 7c785e6a

PR #924, opened 2019-09-26
fixes #254 UI improvement, Change mute-button...

### LINKED ISSUE #254: UI improvment, Change mute-button location when sidemenu expands
### Summary

Today, if you want to mute sound on a game and you go for the mute/volume-icon in the right side menu, the menu expands and effectively moves the thing you want to click on; the mute-icon/button.
It would be preferable if the mutebutton/icon stayed put while the menu expanded outward

### Steps To Reproduce:
* open a game-board
* try to mute the sound.
* Watch the thing you want to click on troll you

BHydden:
Fixes #254 

## Proposed Changes

  - Add an extra mute toggle on the RHS

(subjective whether this being the only row to duplicate the mute button is aesthetically pleasing, I personally liked it better than moving the icon for every row exclusively to the RHS)

(attempted to make the whole anchor a mute but it had suboptimal interaction with the volume slider: slider still functioned but only if you clicked PRECISELY on the track, one pixel off and it would toggle instead)

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -2590,14 +2590,20 @@ export class Game extends React.PureComponent<GameProperties, any> {
                 }
 
                 <a>
-                    <i className={"fa volume-icon " +
-                        (this.state.volume === 0 ? "fa-volume-off"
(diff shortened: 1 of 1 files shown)


anoek:
Cool sounds good, thanks!

PR #927, opened 2019-09-30
Fix title on Leaderboards page to be less confusing

BHydden:
## Proposed Changes

  - Fix title on Leaderboards page to be less confusing

Old:
image

New:
image

note:
The only actual change I made was: 
 - remove line 127 
 - add lines 128-132

This is my first PR since turning on [Prettier](https://github.com/prettier/vim-prettier) which auto-formats the code to be more readable. 
If you don't like it, let me know and I'll turn it off, rollback, and resubmit this PR with just the change to the text.

### DIFF
--- a/src/views/LeaderBoard/LeaderBoard.tsx
+++ b/src/views/LeaderBoard/LeaderBoard.tsx
@@ -16,16 +16,18 @@
  */
 
 import * as React from "react";
-import {_, pgettext, interpolate} from "translate";
-import { get} from "requests";
-import {errorAlerter} from "misc";
-import {Player} from "Player";
-import {Card} from "material";
+import { _, pgettext, interpolate } from "translate";
+import { get } from "requests";
+import { errorAlerter } from "misc";
+import { Player } from "Player";
+import { Card } from "material";
 
-interface LeaderBoardProperties {
-}
+interface LeaderBoardProperties {}
 
-export class LeaderBoard extends React.PureComponent<LeaderBoardProperties, any> {
+export class LeaderBoard extends React.PureComponent<
+    LeaderBoardProperties,
+    any
+> {
     constructor(props) {
         super(props);
         this.state = {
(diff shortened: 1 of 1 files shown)


BHydden:
If you approve the formatting changes Prettier makes, in the future formatting and content changes will be in isolated commits.

BHydden:
Thanks for the feedback! If anoek likes it I'll certainly spilt it up but I wanna hear from him first as he may simply say "I like it how it is, just add content not coding style" which would be totally fair so until then I'll wait and see :-) 

it's a really simple change to fix so once I've got his word one way or the other it'll certainly be <24hr turnaround :-) plus I also need his feedback on whether he likes the new title so two birds with one stone :-)

BHydden:
FYI @anoek I'm pretty sure there are customisable options for it too, so if your feelings are "in general I like it but this one thing is a deal breaker" I can look into that too :-)

anoek:
I do like the formatting, I think I'll install that on my end too, and it's awesome it supports multiple editors. We should synchronize our prettier settings I reckon so we don't keep twiddling them back and forth. 

However I don't want a bunch of commits that include beautifications mixed in with actual changes, so just make sure that when you make changes, beautifications come in a separate PR. 

The very last change in this PR reformats a `{{{` to three separate lines, if you see that, we should just remove the `{{{`, those I had there for doing code folding, but I'm migrating away from that and just using syntax folding.

I'm going to close this PR so you can re-open with just the title change separately, and meanwhile I'm gonna go nuke a thousand `{{{` braces and play with some prettier settings. If you have your prettier config that you like shoot it over to me?

BHydden:
That all sounds awesome! And separate PRs makes even more sense than separate commits, sorry I didn't think of it!

FWIW I currently use the default configuration that came with tpope's bundle I linked above, so I'm happy to use whichever settings you prefer :-)

PR #934, opened 2019-10-01
Show username in chat instead of player 123

flovo:
Instead of showing `player 1` when someone links a user profile, show the username instead.

This pull request needs a close review. I'm not sure if I broke anything in a subtle way (I encountered no obvious problems)

## Proposed Changes

  - Replace `<Link>` with `<Player>`: I had to enable CacheUpdate to get the username. (There are repeated requests if no user with id exists, else it looks like every user is requested only once)
  - Add another Chat shortcut `@" "` to allow link mentions by name. I had to modify [Player.tsx](src/components/Player/Player.tsx) and [player_cache.ts](src/lib/player_cache.ts) to allow for fetching of `user` by username.
  - Modify [tab-completion](src/lib/tabcomplete.ts) to automatically enclose completed usernames in `@" "`. 
(I tried to enclose every typed username after it's completely entered, but was not able to make it work for *all* usernames (special chars and spaces are problematic))

image

----

I tried to create a test account containing a 💖 to test for special characters, but it's not possible on the beta server. But on the main server users exist using this character https://online-go.com/player/85719/. Are special characters deprecated. And is somewhere a list which chars are allowed?

### DIFF
--- a/src/lib/tabcomplete.ts
+++ b/src/lib/tabcomplete.ts
@@ -170,7 +170,7 @@ function onKeyPress(e, options) {
                      /* Space should not be added when there is only 1 match
                             or if there is already a space following the caret position */
                      let space = (match.matches.length > 1 || last.length && last.substr(0, 1) === " ") ? "" : (first.trim().length === 0 ? ": " : " ");
-                     $this.val(first + match.value + space + last);
+                     $this.val(first + '@"' + match.value + '"' + space + last);
                      setCaretToPos(this, sel.start - text.length + match.value.length + space.length);
                  }
 
--- a/src/lib/tabcomplete.ts
+++ b/src/lib/tabcomplete.ts
@@ -170,7 +170,7 @@ function onKeyPress(e, options) {
                      /* Space should not be added when there is only 1 match
                             or if there is already a space following the caret position */
                      let space = (match.matches.length > 1 || last.length && last.substr(0, 1) === " ") ? "" : (first.trim().length === 0 ? ": " : " ");
-                     $this.val(first + '@"' + match.value + '"' + space + last);
+                     $this.val(first + '@"' + match.value + '/' + player_cache.lookup_by_username(match.value).id + '"' + space + last);
                      setCaretToPos(this, sel.start - text.length + match.value.length + space.length);
                  }
 
--- a/src/lib/tabcomplete.ts
+++ b/src/lib/tabcomplete.ts
@@ -129,6 +129,40 @@ function matchName(input, nicknames) {
     return { value: "", matches: matches };
 }
 
+function matchFullName(input, nicknames) {
+    let matches = [];
+    let i = 0;
+    let letter;
+    let letters = "";
+    $.each(nicknames, (index, value) => {
+        if (input.lastIndexOf(value) === input.length - value.length) {
+            matches.push(value);
+        }
+    });
+
+    if (matches.length === 1) {
+        return { value: matches[0], matches: matches };
+    } else if (matches.length > 1) {
+        for (; i < matches[0].length - length; i = i + 1) {
+            letter = matches[0].toLowerCase().substr(length + i, 1);
+
+            $.each(matches, (index, value) => {
+                if (value.toLowerCase().substr(length + i, 1) !== letter) {
+                     letter = "";
+                     return false;
+                }
+            });
+            if (letter) {
+                letters += letter;
+            } else {
+                break;
+            }
+        }
+        return { value: input + letters, matches: matches };
+    }
+    return { value: "", matches: matches };
+}
+
 /* tslint:disable */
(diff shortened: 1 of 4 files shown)


BHydden:
Special characters ARE deprecated, though only for new usernames, existing ones were allowed to continue. I'm not sure of the exact list.

BHydden:
Worth testing whether or not the link breaks if a players name is changed after the text is submitted (it shouldn't but it's worth confirming)

flovo:
It's impossible to resolve references by name (`/u/`, `/user/`, `@" "`) after the name changed (I guess). The old name would be visible as standard user, but not linked. 

If the user is resolved and changes their name, the link should work while you have the chat window open. After you close the window, we have to map name => id again, which will obviously fail.

BHydden:
In that case, maybe this should either 2 serparate functions ('Player #####' links to $id and '@xxxxx' links to $username)

or... if you want it to be one function, maybe both inputs ('Player #####' and '@xxxxx') could link to  "xxxxx #####" (eg. "Player 445315" and "@BHydden" both resolve to "BHydden 445315")

in this way  there would still be a way to permalink to a user, since $id cannot be changed

flovo:
I haven't removed any functionality. It's still possible to use `player id` or a link to the profile page (`/user/view/id`, `/player/id`).

I thought about doing the name => id before it's posted, but I didn't came up with a good way to do it yet. Maybe we could modify tab-completion to add the player-id to the `@" "` tag (e.g. @"flovo/449941").

BHydden:
I think I just thought of one! Currently /player/id formats itself to look like `"Player" + $id` right? 
What if instead, it formatted itself to look like `$name $id` instead?

flovo:
This PR changes the formating from `/player/id` => `player id` to `/player/id` => `name`, but the name is still linked to the user_id. If the submitted chat line contains the player_id, we are using it. Only if no id is available, we have to use the name.

BHydden:
oh... that should be fine then... unless there's some way to find and match name with id so that even when they enter a name it automatically links the id... but I don't think that's strictly necessary given what you've already done...

flovo:
When using tab-complete, the id is added to the `@"name/id"` tag. 
When the `@"name/id"` tag contains an id, use the id to resolve the username, else try using the username to resolve the id.

I encountered a problem when setting both username and id: When the username doesn't match the id, the wrong username isn't replaced by the correct one. So I had to add a check: if an id is present, don't pass the username to the Player React. Without this workaround, the cache was poisoned with a wrong username <=> id relationship.

PS: After changing a username back and forth multiple times, it doesn't work reliable for that specific user anymore. For the other tests it looks fine.

btw. it seems the Player React doesn't update the username if the first try failed.

flovo:
I think I found the bug. If a player with id doesn't exists, the termination-api returns a error message instead of the player data. That's even the case if we requested several ids and only 1 doesn't exist:
> https://beta.online-go.com/termination-api/players?ids=1.999999
> {"error":"Player 999999 not found in cassandra"}

src/lib/tabcomplete.ts
```diff
@@ -170,7 +170,7 @@ function onKeyPress(e, options) {
                      /* Space should not be added when there is only 1 match
                             or if there is already a space following the caret position */
                      let space = (match.matches.length > 1 || last.length && last.substr(0, 1) === " ") ? "" : (first.trim().length === 0 ? ": " : " ");
-                     $this.val(first + match.value + space + last);
+                     $this.val(first + '@"' + match.value + '/' + player_cache.lookup_by_username(match.value).id + '"' + space + last);
```
anoek: I don't mind making @ completion a thing, but would it be possible to support the existing name completion *and* @ prefixed tab completion? Also, when I type '@a<tab>` I get @@"anoek/1" is that expected, or should it just be with one @? (I don't really care in either case)

flovo: > but would it be possible to support the existing name completion _and_ @ prefixed tab completion?

An option would be to do the @"" thing only when the username is already completed (e.g. `a =tab> anoek =tab> @"anoek/1"`), but I wasn't able to check if a completion is needed. I don't understand the tab-completion code well enough to find out how to do it.

> when I type `@a<tab>` I get `@@"anoek/1"` is that expected

The `@@"anoek/1"` is the linked equivalent to `@anoek`, so I think leaving the `@ is the expected behavior. I thought  about what to do when `@"a =tab>`, but decided to go the easy way by not checking for it.

flovo:
Now not existing player_ids and usernames are handled by storing default values if fetching fails.

The tab completion now works as before. By pressing tab a second time the username is enclosed in `@"username/id"`. It would be nice be able to do both on mobile as well :man_shrugging:

anoek:
Thanks for the PR, pretty neat!

PR #937, opened 2019-10-02
es5 reversion bugfix (can't edit joseki descriptions)

GreenAsJade:
Apparently you can't do that :)

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -1597,7 +1597,7 @@ class EditPane extends React.Component<EditProps, any> {
             this.currentMarksInDescription(this.state.new_description));
     }
 
-    currentMarksInDescription = (description: string) => {
+    currentMarksInDescription = (description: string): Array<{label: string, position: string}> => {
         // Extract markup for "board marks"
         // maps markup of form "<label:position>"  to an array of {label, position} objects for each mark
 
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -1597,7 +1597,7 @@ class EditPane extends React.Component<EditProps, any> {
             this.currentMarksInDescription(this.state.new_description));
     }
 
-    currentMarksInDescription = (description) => {
+    currentMarksInDescription = (description: string) => {
         // Extract markup for "board marks"
         // maps markup of form "<label:position>"  to an array of {label, position} objects for each mark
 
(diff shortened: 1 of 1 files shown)


anoek:
Oye, it goes a bit farther, the reason that doesn't work is because what's returned by `matchAll` is a `RegExpStringIterator` instead of an array. `matchAll` itself is also actually a problem too because it's not part of es5. The reason typescript isn't complaining is because there's no type being used, but if you start adding some types in, such as `description:string`, you'll start to see that things are not too happy. I guess this is a pretty good argument for going through and finally adding types to everything that doesn't have them so we can finally turn on the no-implicit-any flag so these kinds of things are caught at compile time.

BHydden:
I'm always an advocate for clean code; sounds good!

GreenAsJade:
Yeah - from time to time I've thought "geez this is slack how I'm not typing" but not actually done anything about it.  I'm all for it.

I don't quite understand why typing makes a difference to whether matchAll is "caught" and what caught means in this case.   I mean... it's working???

How is it that it is "not part of es5", and es5 is specified, but it still works?

BHydden:
My understanding is since typescript is basically a superscript of JS, when it's not typed it isn't picked up by the stricter tests and just passes as regular JS

GreenAsJade:
Yes, but if the function isn't available at build, because it's not part of the Javascript spec that is supported then ... how does it build?

BHydden:
Magic... Duh! 😉

BHydden:
Haha

anoek:
typescript tries to infer types, but there's only so much it can reasonably do. In this case we have a (to the compiler) generic callback, so the function parameters can't (?) really be inferred and it just has to trust the programmer. I think that's pretty much always true with named functions, and I also imagine that if we go through and slap types on function parameters and any class variables, I imagine we'll be in pretty good shape in terms of empowering typescript to infer the rest of the types for us so it can catch stuff like this.

As for the issue at hand, apparently `string.matchAll` is pretty bleeding edge stuff with no support in Edge or Safari currently. I *think* the spread operator is fine in this case, as that compiles down to something that typescript creates for us for es5 compatibility. I think what we need to do is not use `matchAll` unfortunately. I don't think we can use a shim in this case for it either since the native implementation is returning something that the non-native spread operator has trouble with, and any shim should use the native implementation if it's available.

anoek:
And apparently I should have looked at your commits before typing that up because you already did that :+1:

PR #938, opened 2019-10-02
Add page link to chat channels

BHydden:
## Proposed Changes

  - add page link to group and tournament chat channel icons

Clicking on the icon for a tournament or a group in the chat channel list will now take you to it's respective page.


BHydden:
> making the icon an actionable thing is too hidden I think

I can see that... From memory, the main problem was UI real estate so I wanted to maximise the effect of what we already had. I did add an on-hover text to the icon though, so for the people whose mice do tend to wander over to the far edge of the screen, it is likely they'd see the hover text and think "what's that for" which would hopefully minimise both not knowing it exists and accidentally pressing it.





> (1) we could make a link button that appears when the mouse is over, much like the leave button.

Possible, but that might exacerbate the potential problem of clicking it by accident since a dedicated button would almost certainly be bigger than those tiny existing icons we have on the side.



> (2) get rid of both the link and the leave button because they aren't very mobile friendly and are kind of a weird pattern to begin with and instead add a header space above the chat window itself that has actions like leave, and visit group/tournament. That would give us some room for adding channel title topic kind of stuff too.

I can only speak from my own personal experience on this, I use chat on my Samsung S7 pretty much every day. The leave buttons are easy to press when I need to, and I don't believe I've ever accidentally pressed it (though even if I did a second tap easily resolves that). On the phone the icons sit flush against the edge of the screen, personally I'd have to go out of my way to touch such a place so I don't think I'd be accidentally pressing it. Still, I think it should be possible to use on the phone. 

Regarding your header idea though:
Firstly, moving all of the "leave" buttons into a channel-specific header would mean that you would have to enter each channel before leaving it. Currently, you can quickly and easily drop any channels you want without leaving the one you're interested in. A bonus if you have a slow connection.
Secondly, with the limited screen real estate on a phone, the space required for a header which contains two buttons and a bulletin text would take away precious space needed by the already restricted chat window. I have a very small font size as default on my phone that would be unusable for most, and even I only have 27 lines that are only about 7-8 words each, so I would be against putting anything in place that further reduced that.

Since my own experience alone is hardly grounds to impose a feature on the whole community, I'm thinking now (if not earlier) is the right time to make a forum thread to discuss the feature in general and also possible solutions.

BHydden:
[Discussion thread on the forum](https://forums.online-go.com/t/page-links-from-chat-list-groups-tournaments/23153)

BHydden:
sorry about that :P I'll stop being lazy now...

src/components/Chat/ChatDetails.tsx
```diff
@@ -0,0 +1,179 @@
...
+    }
+    goToTournament = (_ev) => {
+        browserHistory.push('/tournament/' + this.state.channelId);
+        this.close_all_modals_and_popovers();
+    }
+
+    render() {
+        let user = data.get("user");
+
+        console.log("I'm rendering");
```
BHydden: this line gets printed but none of the divs show up on the screen

BHydden: @anoek  I'm trying to use the player name menu as a template to build the chat options menu but it's not rendering even though the render() code is being run... am I missing something obvious?

flovo: They are rendered at the bottom left of the page. If you click the ▼, the site suddenly has a scroll bar.

BHydden: > am I missing something obvious?


> bottom left of the page. ...  the site suddenly has a scroll bar.

Well, that certainly qualifies!!! 🤣 





> You specified the below property in `popover` as below chat window.

Hmm, odd. I copied that from display_details in Player.tsx and they load directly under the username. (Hence, I figured, "popover" rather than "popinline")

src/components/Chat/Chat.tsx
```diff
@@ -815,6 +836,63 @@ export class Chat extends React.Component<ChatProperties, any> {
...
+        //         cur = cur.parent();
+        //     }
+        // } catch (e) {
+        //     console.error(e);
+        // }
+
+        console.log("just 2 string " + channel_idd);
+        popover({
+            elt: (<ChatDetails chatChannelId={channel_idd} />),
+            below: this.refs.elt,
```
flovo: replace with
```below: event.currentTarget.parentElement```

BHydden: Huh... Alright I'll give that a try, thanks :-) still confused as to why it works for player and not chat haha but thanks a lot for your help

flovo: For Player the whole element is just the line with the player name and rank, while for Chat it's the whole chat view.

flovo: Every element on the page is a distinct element (they are at least distinguishable by their occurrence in the html tree). And everything™ created with a tag is an [element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element).

BHydden:
Just found [this YouTube video on arrow functions in ES6](https://www.youtube.com/watch?v=h33Srr5J9nY&t=331) which talks about their affect on `this` and can't help but feel like it would solve all my problems :P but, of course, we just downgraded to ES5 🤣

BHydden:
*excited squee* it wasn't ES6 only, I was just typing it wrong 😄 🎆

anoek:
We compile to es5 but we get to use the entire TypeScript language, which has pretty much all the latest and greatest in terms of syntax, including arrow functions and the expected behavior of `this` :)

BHydden:
I just got full functionality for this working... I'll try a little bit to improve the display of the dropdown menu but I should be ready for final review shortly :)

BHydden:
Dropdown menu is now fully functional.

I'd like it better if the dropdown menu didn't overflow past the scrollbar onto the chat area, but I can't figure out how to lock it inside the div instead of the window.

BHydden:
If you have tips on how to keep it in the div, or else make all popovers draw down and left instead of down and right, I'd be happy to follow that up... But otherwise I think this is good to go.

anoek:
Cool thanks!

Issue #939, opened 2019-10-02
PaginatedTable component shows wrong results when typing too quickly


flurmbo:
**Describe the bug**
This is a small bug I noticed on a user's game history. When typing a page number into the input of the PaginatedTable component, it will only show the page corresponding to the first digit of the page number. So for example, if you type in "25", it will show page "2" of the results. See the attached gif for an example.

This is because when the PaginatedTable first notices the input has been changed to "2", it sets `this.updating` to true, which blocks the update() function from making an ajax request after the "5" has been typed in (unless you type very slowly!) 

I've made a PR for this but someone more familiar with the code should probably take a look at it.

![paginatedtable](https://user-images.githubusercontent.com/35127144/66011725-d82d2300-e489-11e9-89e2-a7c9c698b468.gif)

anoek:
Implemented in #940

PR #940, opened 2019-10-02
debounce updates in paginated table

flurmbo:
My go at fixing #939. Not sure if this is the best implementation, but it does fix the issue with the wrong page being shown sometimes.

## Proposed Changes

  - remove the blocking `this.updating` class member on the PaginatedTable component
  - remove `this.needs_another_update`
  - debounce calls to update() by 100 ms

### DIFF
--- a/src/components/PaginatedTable/PaginatedTable.tsx
+++ b/src/components/PaginatedTable/PaginatedTable.tsx
@@ -137,24 +137,18 @@ export class PaginatedTable extends React.Component<PaginatedTableProperties, an
         return post(this.source_url, query); // TODO: Check the URLs and typify the result again
     }
 
-
-    needs_another_update: boolean = false;
     filter_updated() {
         this.setPage(1);
-        if (this.updating) {
-            this.needs_another_update = true;
-        } else {
-            this.update();
-        }
+        this.update();
     }
 
-    updating: boolean = false;
+    update_debounce = null;
     update() {
-        if (this.updating) {
-            return;
-        }
-        this.updating = true;
-        this.needs_another_update = false;
+        clearTimeout(this.update_debounce);
+        this.update_debounce = setTimeout(this._update.bind(this), 100);
+    }
+
+    _update() {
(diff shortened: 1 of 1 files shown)


GreenAsJade:
This has needed to be fixed for ages!  Yay!

Is it conceivable to count the "bounces" and actually advance by that much?   I haven't thought about it too much, but it is irritating to know you want to go 5 pages back and have to click to 5 slowly.

There is actually the same problem with the back button in the Joseki Explorer: it goes too slowly because it waits for each update before it lets you click again.

BHydden:
Surely you could implement both with a local count variable that increments and decrements based on requests and updates?

GreenAsJade:
Don't call me Shirley!

BHydden:
Ok, but I am serious ;-)

anoek:
The `updating` / `needs_another_update` pattern code you see in there is solving a different problem - if you're typing numbers / paging through results, sometimes each page takes a long time to load. The updating pattern makes it so if a request is in flight, we don't update until it comes back. If we need to update again when it comes back, we do so then.

The problem with the debounce solution you have is that if someone is paging through results and things are taking awhile (such as looking through games on a bot record, sometimes those take a few seconds to complete even), it's quite possible that it's going to start new requests before the old ones have finished, and then as those requests complete it's going to refresh the page results with the last one completed.

Thanks for the attempt though, and I certainly welcome future attempts at fixing this rather annoying bug!

flurmbo:
Thanks for the explanation @anoek, is the problem you referring to this sort of behavior in the gif below (tested with my changes)? The interval between some requests is greater than the debounce interval, so we see multiple updates in quick succession.

One way I thought of dealing with that is keeping track of the last ajax request made and then after receiving data from server only updating if that data is from the most recent request made. Could be a little complicated and might make UI a bit unresponsive though

ogs

BHydden:
What if you added a half second wait after each number edit was pressed before the request is sent? That gives you time to enter the full desired number and it can shoot off a single request

PR #942, opened 2019-10-02
Adding clocks to zen mode

### LINKED ISSUE #484: Zen mode should display clocks, undo, pass & resign buttons
### Summary

To effectively play live & blitz games players need to see the clocks. They may also need to access `pass` and `resign` buttons. Further, an `undo` button should be visible so that players can request an undo or see that their opponent has requested one.

Without this information, especially the clocks, Zen mode is unusable for live & blitz games.

### Environment

 * N/A

### Steps To Reproduce:

1. Start a live or blitz game.
2. Enable Zen mode

### Additional Information:

* This has been discussed (briefly) in [this forum post & poll](https://forums.online-go.com/t/poll-clocks-prisoners-in-zen-mode/14257).

flovo:
Fixes #484 

## Proposed Changes

  - Adding clocks to zen mode
  - Introduce a separate portrait layout for zen mode
  - move controls to the right (below clocks), to make goban as big as possible
  - some related layout changes

I changed the background color of the clocks to be monochrome in zen mode. The linear-gradient attracts too much attention in zen.

zen landscape
zen portrait

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -392,12 +392,65 @@ goban-view-bar-width=400px
 
     &.zen {
         themed background-color bg
-        .left-col, .right-col {
+        .left-col {
             display: none;
+        }
+        .right-col {
+            padding-right: 0.5rem;
+            min-width: 150px;
+            max-width: 300px;
+            padding-top: 2.8rem;
+            justify-content: start;
+
+            .game-state {
+                display: none;
+            }
+            .play-controls {
+                display: block;
+                flex-grow: 0;
+            }
+            .players {
+                flex-direction: column-reverse;
+            }
+            .player-container {
+                width: 100%;
+            }
+            .black.player-container {
+                background: #333 !important;
+            }
+            .white.player-container {
+                background: #ddd !important;
+                .score-container {
+                    color: #000;   
+                }
+            }
         }
 
         .leave-zen-mode-button {
             display: inline-block;
+        }
+        .Dock {
+            display: none;
+        }
+    }
+
+    &.portrait.zen {
+        .center-col {
+            padding-top: 2.8rem
+            display: block;
+
+            .game-state {
+                display: none;
+            }
+            .black.player-container {
+                background: #333 !important;
+            }
+            .white.player-container {
+                background: #ddd !important;
+                .score-container {
+                    color: #000;   
+                }
+            }
         }
     }
 
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -72,7 +72,7 @@ interface GameProperties {
 /* TODO: Implement giving voice and control over to players in Reviews */
 /* TODO: Implement mobile interface for reviews */
 
-export type ViewMode = "portrait"|"wide"|"square"|"zen";
+export type ViewMode = "portrait"|"wide"|"square";
 type AdClass = 'no-ads' | 'block' | 'goban-banner' | 'outer-banner' | 'mobile-banner';
 
 export class Game extends React.PureComponent<GameProperties, any> {
@@ -131,7 +131,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
         this.game_id = this.props.match.params.game_id ? parseInt(this.props.match.params.game_id) : 0;
         this.review_id = this.props.match.params.review_id ? parseInt(this.props.match.params.review_id) : 0;
         this.state = {
-            view_mode: "wide" as ViewMode,
+            view_mode: this.computeViewMode(),
             squashed: goban_view_squashed(),
             undo_requested: false,
             estimating_score: false,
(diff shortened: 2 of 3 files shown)


BHydden:
I'm almost certain this needs to be an optional preference you can turn on from settings. I think if we just dumped this on all the zen players we'd get a lot of kickback. Especially from the correspondence zen players who have absolutely no need to ever look at their "clocks"

anoek:
I'm hesistent to merge this, I seem to recall clocks being specificially requested *not* to be in there. And moving the controls to the right side messes up the balance of the board in that mode, which was another feature request for that mode.

That being said I really don't have a good handle on the current state of what people think of that mode, maybe some discussion would beneficial on the forums before we change things too much?

Spoekenkieker:
Couldn’t it be a tiny checkbox, maybe in form of an analog stopwatch?

Also, I think the clocks above have too much contrast for being “Zen”, here’s a quick and (very) dirty alternative suggestion:
image

But I don’t use Zen mode anyway, I just think the clocks above are too bright and distracting.

Another alternative for Zen mode I could imagine could be to only show the player’s clock and not the opponent’s.

dragon96:
I don't use Zen mode myself, but from this forum post (https://forums.online-go.com/t/poll-clocks-prisoners-in-zen-mode/14257), it seems like having a clock in Zen mode is a pretty heavy in demand. I'd be happy to help out with this feature if it makes sense.

@anoek Is there a process for making potentially controversial feature changes to OGS? Maybe we could use a few iterations on design and discussion in the forums.

anoek:
35 votes with 0 saying they don't want clocks and only a few saying they don't care, so I'd say that's not controversial at all, lets do it.

I do like @Spoekenkieker 's coloring suggestion though, do other's concur?

BHydden:
> 35 votes with 0 saying they don't want clocks

Quite uncontroversial, indeed! :O 



> I do like @Spoekenkieker 's coloring suggestion though, do other's concur?

My only concern was whether they were too subtle to read clearly, but if trohde can read it it must be fine ;) I too am in favour of the more subtle colour scheme.

anoek:
> My only concern was whether they were too subtle to read clearly

Yeah might be, somewhere in between perhaps? And I imagine we'd need a toned down black for the light theme.

BHydden:
Well I can read it fine but I have good eyesight. I suspect it's not far off the money. And yes, light theme should also be adjusted (or just disabled) ;-) #unbiased

flovo:
The reason it's not themed is, I had trouble to overwrite the themed css. Therefor there is a hard overwrite (`!important`) in the css at the moment. If anyone (@dragon96) wants to continue on this, feel free to do so.

Spoekenkieker:
> if trohde can read it it must be fine ;)

😂

As I wrote, it was just a quick’n’dirty sketch …
BUT keep in mind that we’re just looking at a small cutout here, with all the github whitespace around, and it’s probably harder to read here than it would be on-site where a dark background is all around.
But yeah, I realize now that I was coming from dark theme only here … I’ll try to do something similar for light theme later today.

Anyway, it was just a suggestion for more a subtle colour scheme.

Spoekenkieker:
Even dirtier sketch for “day” theme:

image

I think the black might be even a bit lighter but you folks will know better about the details.

Another thing that came to my mind: some (long) time ago there were some threads on L19 about other servers and/or SGF viewers/editors, etc., and one thing I remember was that there were a few people who said that a **full-screen view should make the board as big as possible**, i.e. top to bottom of the screen, and since the space left/right of the screen is free anyway, I’d suggest to put all info and controls there, thus allowing the board to grow as large as the screen allows.

dragon96:
> since the space left/right of the screen is free anyway, I’d suggest to put all info and controls there, thus allowing the board to grow as large as screen height.
I concur with this. I'm a big fan of how [lichess](https://ogs-forums.s3.amazonaws.com/original/2X/f/fe4f9d43951ac3682dd039e58f8a6d8abe59ae67.png) arranges the panel. 

I don't have any strong opinions about the colors, but I think the text should be spaced out a little bit differently. I did some imperfect tweaking in the Chrome console, with the original clock on the left and the zen clock on the right. What does everyone think?
Image

BHydden:
I like the idea of simplifying it down... it makes sense that zen mode wouldn't need things like player name and icon which can be distracting but simply clocks and captures on a plain background. Great job.

anoek:
I too like the simplified look

dragon96:
Here's a preliminary sketch; the dark mode monochrome looks a little off to me, especially for white. Any suggestions on that?

Screenshot 2019-10-12 16 58 57

BHydden:
I really like them all, my only minor feedback is I think light mode black could be a little lighter and dark mode black might look nice with a slightly lighter shadow around the edges

dragon96:
Thanks for the feedback @BHydden. How does this one look?

I'd like to save this for separate pull request. I think the play buttons might look better in the right column instead of under the board (as shown in the top left panel), and wanted to gauge some other opinions on this first. :)

Screenshot 2019-10-14 08 44 26

flovo:
Yes, it's better to move the buttons to the right (in landscape view). Some users requested the goban as big as possible (no elements above or below the goban)

anoek:
noted to be abandoned in #942

PR #943, opened 2019-10-03
External goban

anoek:
This PR is going to split off the goban code and use it from a published NPM, `goban`. The source code has been moved to https://github.com/online-go/goban where we'll be doing several updates to fix some outstanding issues with rendering, refactor the code base so it's not such a mess, and make it usable for things like embedded sgf viewers and the like.


BHydden:
Wow merged already good work :-)

PR #951, opened 2019-10-14
Staging PR to keep track of the Goban overhaul changes

anoek:
This PR will be updated as we update the Goban code. When the Goban code is stable enough to be ready for production we'll merge this code.


BHydden:
oh snap 😮  nicely done

Issue #123, opened 2017-03-09
Puzzle URLs should encode transformation options so they can be linked to


anoek:
### Summary

Currently if you link to a puzzle, the new viewer will get a random set of transformations. We should encode the transformation options on the url in some way so when folks link to them, viewers get a consistent view of the puzzle.

ajcsimons:
+1

PR #952, opened 2019-10-14
chat substitution: allow # in front of numbers

flovo:
now matches `player #123`, ...
don't `match #123` => `game 123` anymore.


BHydden:
will this break its implementation in PM? I use #{gameId} very often in my mod PMs

flovo:
Not tested, but if my observations are correct, none of my changes applies to pm and game chat. I'm planning to make it work through.

BHydden:
In theory, if the things you add don't bleed through, the things you remove shouldn't either... but definitely worth testing to make sure before it goes live :)

flovo:
I don't get it to match `#<number>` as `game<number>` only when it's not matched by anything else without rewriting the whole `chat_mtach`.

anoek:
Thanks!

PR #953, opened 2019-10-15
Adds a clock for the active player in Zen mode.

dragon96:
For spectators, there will be no clock. The clock only displays the time remaining for that period, and the number of captures by both sides. Perhaps in future changes, the number of remaining periods should be included.

See discussion in https://github.com/online-go/online-go.com/pull/942, and fixes https://github.com/online-go/online-go.com/issues/484.


anoek:
This has good potential I think. I would like to hear some other people's inputs regarding only showing your clock. The look and feel certainly feels zen so I certainly like it from the aesthetics department, but for those playing live games, do we need to include the other clock too - or do we just say if you don't like it then use not-zen?

Feedback on the current implementation:

image

In the past the general concensous from the users is that the board should fill as much of the screen as possible, I think it does this pretty well on wide screens as is, but when we're not in wide screen mode some more filling would be requested I think. Also we need to think about what we do for mobile devices, since having a right column isn't practical on mobile.

My personal opinion too is that if we're going to add the clock over on the right, lets put the pass/resign/undo/"your move" text over there too (or do we even need "your move?" in zen)

flovo:
In #942, I got the impression that neither of us (github users) is using the zen mode, so I created a forum topic requesting for feedback.
https://forums.online-go.com/t/zen-mode-design-change-looking-for-feedback/23317?u=flovo

---

About mobile phones/portrait mode:

We should show the clock above or below the goban. I'ld prefere the clock on top, but it might be only because I'm already used to it being there.

(See #942 Goban.tsx for zen portait mode (use `this.state.zen_mode` to determine if we are in zen_mode and never set `this.state.view_mode` to zen))

---

> or do we even need "your move?" in zen

> [On the general topic of hiding things in Zen mode, I noticed that it still displays whose turn it is. This is completely unnecessary as the information is already given by the last-move marker on the board (or absence after a pass).](https://forums.online-go.com/t/hide-messages-while-being-in-zen-mode/23026/15?u=flovo)

flovo:
Days is only needed for correspondence games. I don't think the clock is relevant there anyways, so maybe we could it just show in live games. 

> although personally I'd love to avoid having any English on the display if possible.
It's translated to the user's language ;-)

Btw.: #942 is abandoned (It tries to solve the same things as this one), but I can open a seperate pull request for seperating `zen_mode` and `view_mode` it needed.

dragon96:
I've made a few more changes, and here are what the dark/light themes look like:
Screenshot 2019-12-17 19 45 14
Screenshot 2019-12-17 19 45 16

The main changes are:
 - Two clocks. For byo-yomi and Canadian, the periods and stones remaining are not displayed. I want to leave that for a separate PR.
 - No more game state below the board. (I feel like someone else already removed it before me.) 
 - Play controls are on the right. Leaving changes to make the buttons look nicer for another PR.
 - Factored in some of the recent changes while I was AWOL.

I haven't had the time to massage out all of the edge cases. Correspondence games with longer times will still stretch the clock size, and small windows are buggy when a move is played. All of the other common use cases in live games should be covered though.

Let me know how it looks.

BHydden:
> Let me know how it looks.

Look good in horizontal mode, can you throw up some screenshots of what it'll look like in vertical / mobile display?

dragon96:
What's the easiest way to enter portrait mode? I've tried resizing my browser window, but Window.Game.state.view_mode is still 'wide'. Am I looking at the right property?

Screenshot 2019-12-18 08 56 09

dragon96:
The backend must be down or something, but here's what the clocks would look on a phone. (Just learned about the mobile device toolbar!)

Screenshot 2019-12-18 10 05 16

dragon96:
@anoek Thought I'd ping you directly, since I know this PR has been open for quite a while :)

anoek:
Yeah this looks really nice. Thanks!

I'm not sure how the margins will be received on mobile, it looks nice but I makes it harder to place stones accurately, however, I'm also not sure how many people toggle on zen mode on mobile anyways so maybe it's a moot point. I'm going to merge and we'll see.

Thanks for the PR, it's a great improvement!

anoek:
This is up on the beta site for anyone who want's to play around with it.

flovo:
No margins please. The goban get's to small. 
It'd be best if the goban is as big as possible. We have users concerned about pixel loss each time we discuss any design change (even for desktop computers).

BHydden:
100% agree with flovo, no margins on mobile.

PR #955, opened 2019-10-16
add middle click support to chat links

BHydden:
popups still being blocked for some reason

### DIFF
--- a/src/components/Chat/ChatDetails.tsx
+++ b/src/components/Chat/ChatDetails.tsx
@@ -49,29 +49,25 @@ export class ChatDetails extends React.PureComponent<ChatDetailsProperties, any>
         this.props.partFunc(c, false, false);
         this.close_all_modals_and_popovers();
     }
-    goToGroup = (): string => {
+    goToGroup = (ev) => {
         this.close_all_modals_and_popovers();
 
         let url: string = '/group/' + this.state.channelId.slice(6);
-        return url;
-        // if (shouldOpenNewTab(ev)) {
-        //     console.log("goToGroup()");
-        //     window.open(url, "_blank");
-        // } else {
-        //     browserHistory.push(url);
-        // }
+        if (shouldOpenNewTab(ev)) {
+            window.open(url, "_blank");
+        } else {
+            browserHistory.push(url);
+        }
     }
-    goToTournament = (): string => {
+    goToTournament = (ev) => {
         this.close_all_modals_and_popovers();
 
         let url: string = '/tournament/' + this.state.channelId.slice(11);
-        return url;
-        // if (shouldOpenNewTab(ev)) {
-        //     console.log("goToTournament()");
-        //     window.open(url, "_blank");
-        // } else {
-        //     browserHistory.push(url);
-        // }
+        if (shouldOpenNewTab(ev)) {
+            window.open(url, "_blank");
+        } else {
+            browserHistory.push(url);
+        }
     }
 
     render() {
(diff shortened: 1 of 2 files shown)


src/components/Chat/ChatDetails.tsx
```diff
@@ -66,18 +83,22 @@ export class ChatDetails extends React.PureComponent<ChatDetailsProperties, any>
             <div className="ChatDetails">
                 <div className="actions">
                     {this.state.channelId.startsWith("group") &&
-                        <button
-                            className="xs noshadow"
-                            onClick={this.goToGroup}>
-                                <i className="fa fa-users"/>{" "}{group_text}
-                        </button>
+                        <a href={this.goToGroup}>
```
BHydden: What have I done wrong here? 
Expected functionality: a href="/group/$group_id"
Observed functionality: a href="    goToGroup = (): string => {
        this.close_all_modals_and_popovers();

        let url: string = '/group/' + this.state.channelId.slice(6);
        return url;
        // if (shouldOpenNewTab(ev)) {
        //     console.log("goToGroup()");
        //     window.open(url, "_blank");
        // } else {
        //     browserHistory.push(url);
        // }
    }"

BHydden: why is the whole function getting squished in there instead of just the return value?

BHydden: ERROR 
`TS2322: Type '() => string' is not assignable to type 'string'.`

anoek: `href` takes the url only, you're passing a function to be called, not the string url.

BHydden: Yeah but if i only put the href there I can't figure out how else to call the close popover command and it just sits there :-( 
should I try adding an event listener instead or something?  Or go back to trying to get window.open to work like it does it PlayerDetails?

BHydden: I thought this fix was going to be so simple but it's really thrown me against the wall :-(

flovo: To react to middle button, you can use `onAuxClick` (https://stackoverflow.com/questions/1795734/triggering-onclick-event-using-middle-click, https://stackoverflow.com/questions/43057982/html-js-define-onclick-and-onauxclick-at-the-same-time). 
Safari and IE (Edge) don't support `onAuxClick` (https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onauxclick). `onClick` should still do the trick for them, so additional testing for IE and Safari might be needed (or we just ignore them).

This works on Firefox: https://github.com/flovo/online-go.com/commit/6f6fbb992e6d616ae7516d2f1d50bb883d365052

BHydden: thanks flovo that looks interesting... ideally I'd have full support for all browsers, especially since it's been done on the site already with profile links... but at this point this is sounding like a perfectly fine workaround :P lol

@anoek , how do you feel about flovo's suggestion to throw IE and Safari under the bus? 😉

BHydden: if you're ok with the workaround, I can have the PR up later today

flovo: I think it'll work for IE and Safari as well (the onClick should handle both cases for them). It would need testing (or knowlege) though.

BHydden: the shouldOpenNewTab() is meant to handle it, but currently it triggers a pop-up warning.
anoek implemented a workaround to make the browser handle it so that the pop-up warning would trigger for clicking on a profile name... but, despite thinking I implemented it the same way as he did, mine still triggers the pop-up warning (though, if that is enabled, then the functionality works fine) but, again, ideally I'd like to not trigger the warning since I know it's possible. But, I guess we'll see what anoek thinks.

flovo: Yes, if they don't have onAuxClick, onClick should already do the trick. onAuxClick was introduced to mitigate the problem of developpers not checking which button was clicked, wihch let to unexpected behaviour: https://developers.google.com/web/updates/2016/10/auxclick

flovo: I don't get warnings in neither Firefox nor Chromium

BHydden: @flovo I just tried implementing your onAuxClick workaround just in case anoek likes it and it's no good, it does the same as shouldOpenNewTab() it still triggers the pop-up warning.

BHydden: > I don't get warnings in neither Firefox nor Chromium

hmm... odd... well, I'll push what I have I guess, and anoek can test whether he gets popup warnings or not? I guess at the end of the day, a warned popup is better than no functionality at all...

BHydden:
Since flovo says the code works properly on his end, maybe it's simply some error with my environment causing the popups. Could you please check and confirm on your end anoek?

anoek:
This seems to work on chrome, but not firefox, I haven't tested ie or edge at this time.

BHydden:
Never envisioned such a simple idea becomming such a headache 🤣

flovo:
> but not firefox

In my Firefox it seems to work fine. Middle click on “Group Page” in the dropdown opens a new tab with the groups page. No popup warnings (nor any warning at all).

BHydden:
> No popup warnings

Do you, by any chance, have popups set to allow within FF?

flovo:
“Block pop-up windows” is checked (Blocked) for OGS :man_shrugging:

BHydden:
Oh nice, thanks

PR #957, opened 2019-10-21
Always floor/ceil rank

flovo:
When the rank of a player is near threshold, there is a discrepancy between the PlayerDetails (rank rounded to 1 decimal point) and the rank displayed in [] behind the players name (rank ceiled with no decimal places)

This came up in the [forum](https://forums.online-go.com/t/rank-wrong-presentation-calculation-solved-rounding/23343) for a player whose precise rank would be (7.0301…k). His rank is shown as [8k] and 7.0k in PlayerDetails, which led to enough irritation to create a forum post.

My proposal is to floor/ceil the decimal rank as well (to 7.1k in this case). It's more consistent and should mitigate such inconsistencies in the future.

Before:
Before
After:
After

### DIFF
--- a/src/lib/rank_utils.ts
+++ b/src/lib/rank_utils.ts
@@ -185,7 +185,7 @@ export function rankString(r, with_tenths?:boolean) {
 
     if (r < 30) {
         if (with_tenths) {
-            r = Math.ceil((30 - r) * 10) / 10;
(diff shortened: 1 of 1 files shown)


anoek:
Sounds good, though we still need to do `.toFixed(1)` after the `ceil`/`floor` has been applied so that the result always includes the tenths place

anoek:
Thanks for the PR!

PR #958, opened 2019-10-21
seperating view_mode and zen_mode

flovo:
This is a needed step to enable different layouts for zen mode in
portrait and landscape mode. (as mentioned in #953)

With this change `view_mode` stores the screen orientation, and `zen_mode` if zen mode is enabled.


anoek:
Sounds good, sorry for the delay in merging.

PR #961, opened 2019-10-24
Remove lingering old position information 

GreenAsJade:
when an exploratory position is clicked on.

(and start adding more typing)


GreenAsJade:
Note: this is a user-visible irritating bug at the moment: when you click a non-joseki move it should not tell you that this position has tags and a source.

anoek:
Thanks for the fix!

PR #963, opened 2019-10-28
Correct Player id handling in Player React

flovo:
The joseki code is using -1 as fallback userid.

The changes I made earlier to [Player.tsx](https://github.com/flovo/online-go.com/blob/e7ca023e62d71641381f5355e9aeaa556798df49/src/components/Player/Player.tsx) caused notpositive ids to be interpreted as usernames and tried to fetch users with that name (what luckily failed).

This pr restores the original 'do nothing' behavior for invalid (not positive) userids again.


anoek:
Cool, thanks for the fix!

PR #964, opened 2019-10-28
Allow user to select filters that don't have tags included. 

GreenAsJade:
**Requires server 5e5aecbb**

Folk have been saying they want to search by source alone, etc.

Fair enough too.

This change lets them do that.   

If we get bazillions more sequences, there may be performance issues to be fine tuned.  Worst case scenario will be  "to get better performance, select a tag to filter by".


anoek:
Nice, thanks!

PR #977, opened 2019-11-04
Fix game history speed icon for no timelimit

flovo:
Games with no time limit (`time_per_move===0`) were accidentally shown as blitz games in the game history on the profile page. This pr should fix it, showing it as correspondence.

Using `time_control_parameters.speed` as default source should improve the detection of the right game speed further.

Additionally, I added checks to ensure that negative `time_per_move` isn't interpreted as “Live”, but I doubt we get negative values. If neither corr, live nor blitz is detected, no icon will be displayed for game speed.

### DIFF
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -567,15 +567,25 @@ export class User extends React.PureComponent<UserProperties, any> {
                   }
                 }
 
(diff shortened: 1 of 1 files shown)


anoek:
Thanks!

PR #978, opened 2019-11-05
Fixes bug introduced with game history table update

mattetress:
Fixes issue with reviews and demos no longer showing


anoek:
Thanks for the quick fix!

PR #980, opened 2019-11-06
remove win/loss color for annulled games

mattetress:
Annulled games shouldn't get win/loss color in the history table


anoek:
Good change, thanks!

PR #985, opened 2019-11-09
Quick fix for Play mode, broken with caching

GreenAsJade:
Caching broke Play mode.

This disables caching except for Explore more, where it's most useful anyhow.

Some code resequencing needed to make Play mode work with caching (due to cache causing move processing in-line instead of after call to server).


anoek:
Thanks for the fix!

PR #992, opened 2019-11-19
clarify analysis setting

### LINKED ISSUE #991: Clearer wording needed for 'always disable analysis' in site settings
I have come across multiple users who do not understand that this setting only affects them and that their opponents will still be able to use analysis if it is enabled in the game settings. 

Clearer wording could prevent future confusion.

mattetress:
Fixes #991 

## Proposed Changes

Added clarifying text to analysis setting in settings menu

Screen Shot 2019-11-18 at 10 24 34 PM


anoek:
Thanks for the improvement!

PR #995, opened 2019-11-23
add email to moderator page

BHydden:
## Proposed Changes

  - Attempt to add an email column for new users in moderator page


BHydden:
I confirmed that user.email is the correct name for the field, but the column isn't populating in my tests. Any thoughts on what I'm missing?

mattetress:
It's hard for me to help without being able to test it out directly for myself, but does the email heading show and not the emails or do you get neither?

BHydden:
The header shows its just the data that doesn't populate. Eugene helped me realise that email isn't an element of X as I had wrongly assumed, looking into an alternate path to access it.

BHydden:
Still working on this but nothing wish pushing up yet, mostly walking in circles sadly.

anoek:
I am pretty sure this is going to need some server side support to add that field for you, I don't typically return that field anywhere since it's sensitive information.

BHydden:
Yeah I was concerned that was the case. It can be accessed through the moderator modal on the profile page so I was hoping I could replicate that somehow.

anoek:
Thanks for the PR, sorry it took so long to get to the backend part

BHydden:
I didn't realise you were planning on doing it so soon! Thank you!!! 😍😍😍

PR #996, opened 2019-11-25
Fix blitz games displaying live icon

RubyMineshaft:
Some blitz tournament games were getting assigned the live game icon in the game history table.  I've increased the time per move threshold to 15 to catch these games


flovo:
I fear it's not as easy as just increasing the threshold. Games with `time_per_move >= 10` are already live games. See for example https://beta.online-go.com/api/v1/games/6321, where `speed: "live"`
```json
    "time_per_move": 10,
    "time_control_parameters": "{\"system\": \"byoyomi\", \"time_control\": \"byoyomi\", \"speed\": \"live\", \"pause_on_weekends\": false, \"main_time\": 0, \"period_time\": 10, \"periods\": 5}",
```

mattetress:
Hmm yeah seems that some tweaking may be needed in the api. The increased threshold did catch the majority of games I was able to find, since the speed attribute wasnt present on most

mattetress:
What if we had games with `time_per_move <= 15` override the speed attribute for the icon?

mattetress:
I added some code to quickly and dirtily catch games that are categorized as live by the api and display the blitz icon in game history.   I don't know if it's an acceptable solution, but it does correctly override the live categorization.  Perhaps a temporary fix?

GreenAsJade:
I think this should be fixed in the back end.

You can create quite slow Blitz games and very fast Live ones.

From a moderation perspective, I personally won't chase timeouts in Blitz as "escaping", so it does make a difference.

anoek:
Ok I finally tracked this down on the backend I think, they should be in sync now. It's not retro-active though unfortunately.

PR #999, opened 2019-11-30
Update Joseki.tsx

aesalon12:
## Proposed Changes
  - A new node will be blank when added

## Why
  - Less clutter when various contributors add nodes one at a time. Either they erase ## Joseki or leave it in and someone has to erase it later.


anoek:
Thanks for the improvement!

PR #1004, opened 2019-12-04
There's really no need to double check the "clear marks" action

GreenAsJade:
... in a recent Dwyrin video I noticed what a pain it is for users to have a centre-screen swal dialog every time they want to clear marks.

This PR asserts that it's really not needed at all - clearing marks is a lightweight deletion that needs to be easy for a reviewer, and with low consequences of accidental use.

If it turns out some sort of confirmation is needed, let's find that out by putting this in, then adding the next thing if/when needed and when we know what it should be.

### DIFF
--- a/src/components/Clock/Clock.styl
+++ b/src/components/Clock/Clock.styl
@@ -37,10 +37,6 @@
             animation-iteration-count: infinite
         }
 
-        &.paused-correspondence {
-            color: #888;
-        }
-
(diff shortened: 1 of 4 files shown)


GreenAsJade:
^^ nuts the white container colour change accidentally got on this branch as well.

Pull this one to get both :D

anoek:
:+1:

PR #1007, opened 2019-12-06
Use number value handicap instead of sting d.handicap

flovo:
Fixes https://forums.online-go.com/t/wins-losses-against-stronger-weaker-opponents-again/23952?u=flovo

btw: Is it possible to turn this kind of type conversion off in javascript?


flovo:
Closed in favor of https://github.com/online-go/online-go.com/pull/1008

PR #1008, opened 2019-12-06
Fix bug in handicap adjustment for rating graph

flovo:
Fixes https://forums.online-go.com/t/wins-losses-against-stronger-weaker-opponents-again/23952?u=flovo

In 
https://github.com/online-go/online-go.com/blob/ac381c26f952e344574b5a69e7e8d3d340b12c7f/src/components/RatingsChart/RatingEntry.ts#L106
handicap is passed as string, resulting the math going wrong. (effective rating ~ real rating)

Besides the coloring on the game history used the real ratings to determine stronger/weaker.

## Proposed Changes

  - move the code to determine the effective outcome to rank_utils.ts
  - use the new code for rating graph
  - use the new code for game history

### DIFF
--- a/src/components/RatingsChart/RatingEntry.ts
+++ b/src/components/RatingsChart/RatingEntry.ts
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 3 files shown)


anoek:
Awesome thanks for this!

PR #1009, opened 2019-12-08
Read JWT from config every request

GreenAsJade:
Fixes (hopefully) the problem of the jwt not being ready occasionally, and the potential problem of users changing their identity after starting to use OJE.

(Note that the front end will still have the wrong idea about their permissions if they change to an account with different permissions, but this would be very rare, and caught server-side anyhow).


anoek:
Awesome, thanks!

PR #1010, opened 2019-12-08
Do == null not === null because undefined can happen.

GreenAsJade:
Fixes rendering problems when something is undefined - this can happen if the server didn't supply it for whatever reason.

Generally, use the == null not === null form, because we're never trying to discriminate against undefined.

### DIFF
--- a/src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
+++ b/src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
@@ -58,9 +58,9 @@ export class JosekiVariationFilter extends React.PureComponent<JosekiVariationFi
             body.forEach((id, idx) => {
                 // console.log("Looking up player", id, idx);
                 const player = player_cache.lookup(id);
-                contributor_list[idx] = {resolved: player !== null, player: player === null ? id : player};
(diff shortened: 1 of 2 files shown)


BHydden:
Good to hear we're avoiding discrimination.

anoek:
Coming down the pipe I'd like the code base to be moving to more and more strict typescript settings as it helps catch more bugs. As such, we definitely don't want to be removing our `===`'s in favor of `==` as we'll just end up having to undo that as we turn flags on.
 
What I have found in my efforts to increase type safety (see the goban repo, which now has all flags turned on) is that I ended up avoiding nulls in almost all cases. Instead, any field that could be unset was properly marked with the conditional field flag: `field?: type;`, and when strictNullChecks is on it ensures that field is only ever either undefined or a valid type value, so no null checking necessary. null checks went from `obj.field !== null` to `!obj.field` or `obj.field !== undefined` or `field in obj`, take your pick with whatever makes sense for your object.

If you do still have a field that could be null *or* undefined, we should make that an explicit check so we know what we're dealing with, but really those should be pretty rare, I never had a single instance of needing or wanting both states.

GreenAsJade:
I wonder why I started using null.  It appears that was a mistake - will fix.

GreenAsJade:
Yeah, nah.

PR #1011, opened 2019-12-09
Add a setting to hide ranks and ratings

stefankoegel:
Some players (including me) find it more relaxing to not see their opponents' or their own rank and rating.

This commit adds a setting called "Hide ranks and ratings".
When the setting is activated:
* Ranks next to player names are hidden
* Ranks in the mini goban are hidden
* Ranks and ratings in the player details and profile card are hidden
* Ratings in the rating table on the profile page are hidden
* The rating chart on the profile page is hidden
* There is a button on the profile page that toggles the visibility of the rating chart and rating table, as well as the rank next to the player name.


anoek:
Looks good, nice work, thanks!

GreenAsJade:
Thats awesome!

PR #1012, opened 2019-12-09
Correct stronger/weaker labels in game history and rating graph

flovo:
I confused the stronger/weaker labels. The coloring of all games in history and graph where reversed. 

This changes the colors back to the right ones.


BHydden:
Fixes #1013

anoek:
Thanks for the quick fix!

PR #1020, opened 2019-12-11
Get rid of null, use undefined for undefined things.

GreenAsJade:
Fixes use of null

## Proposed Changes

Get rid of null :)

Could we have this branch on beta, to get a few more eyes on it?   I think I've checked everything, but a hard change to be sure about.

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -43,18 +43,18 @@ const server_url = data.get("joseki-url", "/godojo/");
 
 const position_url = (node_id: string, variation_filter?: any, mode?: string) => {
     let position_url = server_url + "position?id=" + node_id;
-    if (variation_filter !== null) {
-        if (variation_filter.contributor !== null) {
+    if (variation_filter) {
+        if (variation_filter.contributor) {
             position_url += "&cfilterid=" + variation_filter.contributor;
         }
-        if (variation_filter.tags !== null && variation_filter.tags.length !== 0) {
+        if (variation_filter.tags && variation_filter.tags.length !== 0) {
             position_url += "&tfilterid=" + variation_filter.tags;
         }
-        if (variation_filter.source !== null) {
+        if (variation_filter.source) {
             position_url += "&sfilterid=" + variation_filter.source;
         }
     }
-    if (mode !== null) {
+    if (mode) {
         position_url += "&mode=" + mode;
     }
     return position_url;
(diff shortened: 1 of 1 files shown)


flovo:
> use undefined for undefined things

You should check for undefined instead. The `if` statement implicitly casts everything to Boolean. https://www.w3schools.com/JS/js_booleans.asp

GreenAsJade:
@flovo  Can you elaborate on what I need to learn/adjust/improve here?

What anoek recommended is "don't use null to represent 'hasn't got a value yet', use undefined instead".

This appears particularly useful advice, because the APIs can return undefined at times - if you have nulls around as well, then suddenly you find yourself needing to cater for both.

My initial way of dealing with this was to say "if foo == null" instead of "if foo === null".    Anoek pointed out that strict compare is highly desireable, and asked not to have loose compare at all if possible.

By getting rid of the idea that "null represents I don't have a valid value for this yet" and using undefined for that instead, the code simplified in other places as well as those just dealing with possible undefined from APIs.

That's the change that is in this PR.

Unfortunately, I can't quite see what you mean "check for undefined instead" ... instead of what?

And I'm not sure why you have stated that "if casts to Boolean": I know this basic fact, but I'm missing what you are trying to convey.

Thanks!

flovo:
If you want to know if variable `x` is undefined https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined:
```javascript
if (x === undefined) 
```

```javascript
if (x) 
```
is false if either `x` is
- undefined
- null
- ""
- 0
- -0
- an object
- and probaly more 
so the new check is more loose than the old one. See 'Everything Without a "Value" is False' and 'Everything With a "Value" is True' on https://www.w3schools.com/JS/js_booleans.asp

> What anoek recommended is "don't use null to represent 'hasn't got a value yet', use undefined instead".
:+1: 

>  ... instead of what?
instead for what you did.

GreenAsJade:
I see that semantically `if (x)` is also a loose comparison, which is what you are saying.

However, it is a useful one in the contexts it is being used, because it specifically is intended to mean "if x is not undefined and not null and not equal to zero"

I don't see the improvement in changing

`{ this.state.thing && <div> thing: {this.state.thing}</div> }`

into

`{ (this.state.thing !== undefined && this.thing !== 0) &&  <div> thing: {this.state.thing}</div> }`

I'm not sure this code would trigger the lint warnings that we hope to avoid by not having loose `==` around the place, would it?

If it would, then that's an argument why the extra code is worth it: it stay lint clean for other reasons why we have this check.

But if this code is allowed by lint (etc) while `==` is not, then I prefer the concise thing that means what I mean it to mean, along with the useful catchall that it brings (for example, I don't want that div displaying if thing accidentally gets null or an object that is false in it).

GreenAsJade:
Ooo - I missed this bit (maybe added later?)

"Would it be helpful to define an Interface for variation_filter, so we don't need to use any as type."

For sure!   I've been gradually adding typing, tidied up a few smaller things in this PR, but that's a good one.

anoek:
This is live on beta for any testing you wanted to do

PR #1024, opened 2019-12-18
Updates for new AI version

anoek:


BHydden:
Is this up on the beta server? I'd love to have a play around with it :-)

anoek:
Yep sure is

BHydden:
The new interface is absolutely GORGEOUS 😍😍😍

BHydden:
Occasional bug where pass at end of game is considered a huge blunder when it isn't. Exists in implementation on main site also but more obvious with the new interface.

https://beta.online-go.com/game/6682

flovo:
Nice. Looks fantastic. :heart_eyes_cat: 

If I see it right, we don't use the score and territory estimations of KataGo. Any reasons to not use score estimation? In handicap games wr is not helpful:
win rate graph for handicap game. wr is always close to top.

anoek:
well, that's output from lz and kata.. not really sure what they are saying with that, but it's not really our "bug" persay I don't think.

anoek:
> If I see it right, we don't use the score and territory estimations of KataGo. Any reasons to not use score estimation? In handicap games wr is not helpful:

It's not quite that simple in practice as the end of a game and doing score estimation is a fairly time sensitive thing - the AI servers though get rather busy when they are grinding away at a game or two, it could be a couple of minutes, in the worst case more, before getting a score estimation or end of game result, which isn't acceptable.

BHydden:
I believe in this case flovo isn't asking about using katago to replace the current score estimator feature (although that would be ideal further down the track) 

By default KataGo, when running analysis, has been designed to return both a win percentage and an estimate of the score difference. I believe flovo is saying this estimate of score difference would be more useful to graph than the win percentage for a handicap game.

anoek:
Ahh I see, I will be happy to explore that, might be rather cool - but i'm gonna merge and release this first, that'll be part of a future update if that works out

PR #1028, opened 2019-12-20
Right-arrow to move back forwards through previously visited moves

GreenAsJade:
Provide a right arrow button with right arrow key binding to let users go forwards again through moves that they just backed up from using the left arrow.


GreenAsJade:
Goban 0.5.3 update looks fine.

PR #1039, opened 2020-01-19
Make the disconnect timer not cover up other important things

GreenAsJade:
As per this thread https://forums.online-go.com/t/player-leaves-match-no-way-to-contact-them/24574

the disconnect timer covers up more than one important function.

## Proposed Changes

Just cover up the player's avatar.

This change also has a separate commit that reindents this part of the code, some of it was on 2-space, now it's all 4.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -2509,16 +2509,16 @@ export class Game extends React.PureComponent<GameProperties, any> {
                     }
                     return (
                     <div key={idx} className={`${color} player-container`}>
-                        {this.state[`${color}_auto_resign_expiration`] &&
-                            <div className={`auto-resign-overlay`}>
-                                <i className='fa fa-bolt' />
-                                <CountDown to={this.state[`${color}_auto_resign_expiration`]} />
-                            </div>
-                        }
 
                         <div className="player-icon-clock-row">
                             {((engine.players[color] && engine.players[color].id) || null) &&
                                 <div className="player-icon-container" style={player_bg}>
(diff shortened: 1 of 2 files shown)


anoek:
:+1: 

I like it on desktop, mobile needs a tweak :)

image

GreenAsJade:
Screen Shot 2020-01-27 at 1 43 09 pm (2)

anoek:
Hmm, this makes the whole screen jump since the text and bolt enlarge the icon space. Any ideas on how to make that better? Should we just show a bolt there and either replace or add the timer part near the other timer?

Maybe we should be doing that in general anyways, just show a bolt over the icon, then replace the countdown timer with whatever is less - the normal play timer that is still running, or the Zeus timer

GreenAsJade:
My feeling was "yeah, that jump is a little untidy, but who really cares, it's an untidy situation anyhow".

One of the purported advantages of this change is that it leaves the game timer visible, which is the truest representation of what is going on: both timers are counting.   So showing only the timer that is less would be a backwards step in that regard.

The other thing I considered is simply having the current behaviour for mobile - don't change it at all.     This would at least allow the incremental improvement for web.

I guess an easy thing to try would be to make the disconnect timer lay out as a row in mobile and reduce the text size further - it might fit alongside the normal timer.  I'll see if that works.

GreenAsJade:
Screen Shot 2020-01-28 at 7 41 02 am (2)


(the game timer ran out just as I took the screen shot, but you get the idea :) )

anoek:
Thanks for the patch and the revision, much better!

PR #1050, opened 2020-03-02
New OGS audio

anoek:
This PR contains a revamp of the OGS audio system, including lots of new samples and better way of playing audio that should eliminate most playback errors, and adds support for some audio effects like adjusting the balance on stone placement sounds depending on where they are placed on the board.


BHydden:
This having a run on beta first or going straight to the main site?

PR #1051, opened 2020-03-06
replace tsumegodojo with tsumegododo in all chats

flovo:
Replace tsumegodojo with tsumegododo as proposed in chat.

This is a rather aggressive implementation, not only replacing URLs but all occurrences of tsumegodojo.

It works in chat and PMs


anoek:
Thanks :)

PR #1059, opened 2020-03-10
Prefetch joseki for display speed

GreenAsJade:
Requires Joseki Server 069eb3ba

## Proposed Changes

Cache prefetched results for all the next-nodes from each position that the player views.


anoek:
Cool, thanks for your continued work on this!

PR #1068, opened 2020-03-22
Bug/group links

### LINKED ISSUE #1066: In group chat, clicking a link to another group results in an invalid state
**Describe the bug**
I was having an online group meetup and a member was talking about another group and posted a link to that group. When I clicked the link, some of the cards on the page were for the group he linked to and some of the cards were for the group I came from.

**To Reproduce**
1. Go to the page for any group (`/group/123`)
2. Copy the url for that page
3. Go to the page for another group (`/group/234`)
4. Paste the url you copied into group chat
5. Click the link
6. The page will now be in a weird, invalid state.
7. Now refresh the page.
8. The page should be in a valid state again.
9. Hit the browser back button.
10. The page is in an invalid state again.

**Expected behavior**
The page should cleanly navigate to the page for the group being linked to.

**Screenshots**
Mar-22-2020 10-22-52

 - OS: Mac OSX Catalina
 - Browser: Chrome
 - Version 80

**Additional context**
Without poking around the code, it looks like the page is doing some work in `componentDidMount` that isn't being correctly handled in `componentDidUpdate`. Or a `useEffect` is missing some dependencies.

Edit: Screenshot

airjp73:
Fixes #1066 

## Gif with fix
_Note_: The link in the gif is something I stubbed in locally
Mar-22-2020 16-05-22

## Proposed Changes
I was right about my hunch. There are side-effects in a couple components that are being handled on mount, but not when important props change.

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -902,7 +902,7 @@ export class Chat extends React.Component<ChatProperties, ChatState> {
 
 export class EmbeddedChat extends React.PureComponent<ChatProperties, {}> {
     render() {
-        return <Card className="Card EmbeddedChat"><Chat {...this.props} /></Card>;
+        return <Card className="Card EmbeddedChat"><Chat key={this.props.channel} {...this.props} /></Card>;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Chat/Chat.tsx
```diff
@@ -902,7 +902,7 @@ export class Chat extends React.Component<ChatProperties, ChatState> {
 
 export class EmbeddedChat extends React.PureComponent<ChatProperties, {}> {
     render() {
-        return <Card className="Card EmbeddedChat"><Chat {...this.props} /></Card>;
+        return <Card className="Card EmbeddedChat"><Chat key={this.props.channel} {...this.props} /></Card>;
```
airjp73: The `Chat` component has more complicated side-effects than the table. I figured if we just reload the entire component when the channel changes, that was the approach with the least risk of breaking things. I'm happy to take a stab at fixing it properly in the `Chat` component itself if you want, though.

anoek:
Awesome, thanks for the fix!

airjp73:
Happy to help! 😊

PR #1072, opened 2020-03-24
Made left navbar spacing b/w icons and text uniform.

evoluteur:
Fixes #
It is not very noticeable but I see it so I wanted to fix it.
Icons are closer to text for Tournament, Ladders, Groups... for other menus on hover there is an underlined space before the text.

## Proposed Changes
  - removed space before menu text (in HTML)
  - added margin after icon (in CSS)

navbar-current

navbar-aligned


anoek:
Thanks!

PR #1083, opened 2020-04-01
fix settings thumbnail list threshold

flovo:
The selection of a new value wasn't accepted by the selector.

It turned out that the `ev` value is a number, not an Object.


anoek:
Doh, thanks for the fix!

PR #1084, opened 2020-04-01
Alernative: Use numeric input for thumbnail list threshold

flovo:
The ideal number of MiniGobans depends on the usecase, display size, …. In my case 8 would be a good number (2 rows of 4 games on the home page), but it isn't in the list. A numeric input is more flexible.

image

Conflicts with https://github.com/online-go/online-go.com/pull/1083. `updateGameListThreshold` works in this case as expected.


anoek:
Sounds good :)

PR #1090, opened 2020-04-06
Chat indicator

flovo:
Adds a chat notification Icon to the NavBar. The Notification Icon shows the total count of unread messages and if the user is mentioned in any chat room selected by the user.

image

image

image

---

To achieve this, the following changes are made:
-   The chat receiving logic moved to a central location (`chat_manager`), to keep the different components in sync. 
    The `chat_manager` seems to be a good place for this. It's called chat_manager after all.
-   The read timestamp for each chat room is stored in local storage, to be able to notify the user if there was chat activity while they were offline. Due to backend limitations the backlog is limited to 40 messages.
-    A copy of the Chat Room List (left hand site on Chats page) is used as dropdown for the Chat indicator. A click there on the chat room opens the selected room on the chat page.
    With some minor abstractions, the new ChatList can replace the list on the ChatList, without changing the feel of the ChatPage.
-    Adding a settings page for chat notification defaults (join all group/tournament chats). Those settings are overwritten by room specific settings.
    To keep the settings in sync, there are some hacky snippets in `ChatIndicator.tsx` (line 28-92). There is probably a better way to do it (probably an eventEmiter like in data).

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -18,28 +18,23 @@
 import * as React from "react";
 import {Link} from "react-router-dom";
 import {_, pgettext, interpolate} from "translate";
-import {post, get} from "requests";
 import {Player} from "Player";
 import {profanity_filter} from "profanity_filter";
 import {comm_socket} from "sockets";
 import * as data from "data";
 import * as preferences from "preferences";
-import {emitNotification} from "Notifications";
 import {Flag} from "Flag";
 import {Card} from "material";
 import {TabCompleteInput} from "TabCompleteInput";
-import * as player_cache from "player_cache";
 import { PlayerCacheEntry } from "player_cache";
 import {string_splitter, n2s, dup} from "misc";
 import {SeekGraph} from "SeekGraph";
 import {PersistentElement} from "PersistentElement";
-import {users_by_rank} from 'chat_manager';
+import {chat_manager, users_by_rank, ChatChannelProxy, global_channels, group_channels, tournament_channels, resolveChannelDisplayName} from 'chat_manager';
 import * as moment from "moment";
-import cached from 'cached';
 import {popover} from "popover";
 import {ChatDetails} from './ChatDetails';
 import {shouldOpenNewTab} from 'misc';
-import { ActiveTournamentList, GroupList } from 'types';
 
 
 declare let swal;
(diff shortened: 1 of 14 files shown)


anoek:
Neat stuff!

There seems to be a couple of thoughts and issues I'll itemize here:

 1. The increment doesn't seem to be working on my end. I've joined a channel in the chat screen so it shows up in my indicator drop down, then wandered off to another part of the site. When I chat in that room with another account, it doesn't light up in blue nor increment, however when I click on it, I do see the new message count in the list. 

  2. If we aren't going to allow people to join global channels there (which is probably a good idea, otherwise those channels will get swamped with people who don't care about chat), should we just hide that list on the indicator dropdown?

  3. Should we make the lists non-collapsible? The collapsing in the chat page has the function of hiding the channels we don't want to chat in, but it seems here like we don't ever show those to begin with, so might as well just have that header and and show the channels they're joined perhaps?

  4. The font used in the action list that pops up when you click the down arrow next to the channel is not the site font, should probably make that match.

That's all I can think of for now - great work!

flovo:
1) ups. Should work now.
2) I'd like to be able to follow Help chat, and get notified when mentioned in shadowban.
    I'm not sure, if people join global chats just for fun. There could be some users adding `mentions` to all chats. 
3) At the moment, the collapes list shows only chats with notifications, while the expanded list shows all chats we subscribed to, but I can change it.
4) done

anoek:
This is definitely cool.

When I opened it, since it shares the same look as the chat channel list, I expected it to work like it, and it took me several tries to get one of the channels to disappear. It wasn't immediately obvious to me that I needed to click both unfollow mentioned and unfollow unread for it to be removed from the list.

What about instead of visibility being dependent on those, we simply have the same leave/join system that we have in the chat page, so a big red button to "Leave channel", and unjoined channels grayed out in the list - hidden by default by the fact that the list is only showing joined channels? So basically, match the UX that is in the chat page.. not that that's the best UX, but I think we should be consistent if nothing else.

flovo:
I think there might be some interest in being notified when mentioned, but not about unread messages.

anoek:
Oh I agree, I wasn't saying to get rid of those options, just don't use them as a basis to hide or show the channel in the list, I think it's a bit too obscure is all I was thinking.

BHydden:
I think it's fine if the channel auto-closes if you unfollow both message types... but I agree that it would be nice to also have a "leave channel" button that you could easily click once per channel, rather than _needing_ to unfollow both message types to leave the channel... I think that's the gist I got from anoek's train of thought (and to which I agree)

anoek:
I don't think it should disappear until the user explicitly clicks leave channel, toggling options shouldn't make something simply vanish, we want that to be something the user is trying to do, not does by accident.

flovo:
Now featuring a big red button.

image

Also show all channels if the list is expanded. When collapsed, there are only the subscribed channels with unread messages.

anoek:
Nice :)

OK I think this is ready to merge, but I am a little scared of the extra load on the chat server code, it's been bumping into some limits so I don't want to risk overloading it. I am currently working on it though, so it should be running smoothly again here soon, at which point I'll click merge.

PR #1091, opened 2020-04-07
User: Make links <Link> instead of <a>

flovo:
This prevents the page from reloading all resources.  At the moment, those links cause the browser to reload everything.


anoek:
Whoops, thanks!

PR #1095, opened 2020-04-11
tutorial: put board presentation first, add simpler examples of captu…

### LINKED ISSUE #975: Do some A/B testing for our tutorial section
As per discussion in pull request #835 , several folks thought our tutorial needs to be expanded. While I'm in favor of improvements, I want to make sure whatever we do doesn't do the opposite for the current generation of new players, basically I want to try and straddle the line between being brief enough that people get the idea's and continue through the tutorial, but detailed enough so that they actually get the idea's and continue through the tutorial.

jfbaraton:
…re and seki. Add variations for the last ladder page

no extra translations needed

Fixes #975

## Proposed Changes

  - put the existing board presentation as first into the first section
  - add more straightforward examples of capture and seki
  - add variations and refutations of moves in the last ladder page. to make it more understandable why a move is wrong

### DIFF
--- a/src/views/LearningHub/Capture.tsx
+++ b/src/views/LearningHub/Capture.tsx
@@ -30,6 +30,9 @@ export class Capture extends LearningHubSection {
             Page3,
             Page4,
             Page5,
+            Page6,
+            Page7,
+            Page8,
         ];
     }
 
@@ -60,6 +63,7 @@ class Page1 extends LearningPage {
         };
     }
 }
+
 class Page2 extends LearningPage {
     constructor(props) {
         super(props);
@@ -86,6 +90,7 @@ class Page2 extends LearningPage {
     }
 
 }
+
 class Page3 extends LearningPage {
     constructor(props) {
         super(props);
@@ -112,6 +117,7 @@ class Page3 extends LearningPage {
     }
 
 }
+
 class Page4 extends LearningPage {
     constructor(props) {
         super(props);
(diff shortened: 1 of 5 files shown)


anoek:
Nice work! I like the additions a lot.

For discussion, please feel free to refute:

I think "The Board" being first not what we want. I know it seems fundamental but I'd argue it's not, it's the foundation of some theoretical stuff, but if they don't even know the game is about surrounding each other with stones, talking about different board sizes and areas of the board is a bit abstract and ahead of the concepts we need to introduce, and honestly probably a bit boring since it's three pages of text with nothing on the boards.

My thought was that as users progress they'll probably notice some things about the board, star points, wondering if it's always 9x9, that sort of thing - and that quick section is there to answer those questions they've raised themselves, which is always more interesting than blurbing in the abstract about questions they don't have yet. 

However, my thoughts on this are largely theoretical, so if you have some other thoughts on the matter I'm definitely open to them.

jfbaraton:
ok, how about, I put back the "Boards" where it was, and create a new section "Board" that would come first.
it would contain a single page that says that:
- stones are placed on crosslines and don't move
- the goal of the game is to surround the largest areas
- stones can be captured

after that, the next section is the capture section.


If you don't want it, I can merely put pack the "Boards" section where it was and PR the puzzle improvements

anoek:
Sure that sounds good, lets try it :)

jfbaraton:
done

some new texts in the Intro page => need translation (i can contribute, don't know how though)

anoek:
Looks good, thanks!

Translation happens on https://translate.online-go.com/projects/ if you happen to speak any other langauges! The strings should show up there soon.

PR #1096, opened 2020-04-12
Show game timings option for moderators

GreenAsJade:
Provides a toggle for moderators to present the game timings.


BHydden:
Awesome <3 I don't know how hard it would be to implement, but when I'm looking at game timings of ongoing games one thing I really wish I had was 'time since the last move was played'... is that possible to include in this PR?

I suspect it could be possible by doing game start time + sum of all time taken per move... buy maybe there's an easier way?

GreenAsJade:
@BHydden  I could try that and see if it gives a reasonable answer :D

yarn.lock
```diff
@@ -456,6 +456,13 @@
   resolved "https://registry.yarnpkg.com/@types/markdown-it/-/markdown-it-0.0.4.tgz#c5f67365916044b342dae8d702724788ba0b5b74"
   integrity sha512-FWR7QB7EqBRq1s9BMk0ccOSOuRLfVEWYpHQYpFPaXtCoqN6dJx2ttdsdQbUxLLnAlKpYeVjveGGhQ3583TTa7g==
 
+"@types/moment-duration-format@^2.2.2":
```
BHydden, line 459: double-check with @anoek  but I don't think we're supposed to submit edits to yarn.lock and rather it should auto-build on compile? sorry if that's mistaken

anoek: Nope commits to `yarn.lock` are expected

anoek:
Neat :)

yarn.lock
BHydden: Thanks

PR #1098, opened 2020-04-13
It got to me in the end...

GreenAsJade:
Fixes #

VSCode linter complaining about type of debounce variable

## Proposed Changes

Ask the compiler what frickin type it is.


anoek:
I hadn't looked into it before, but check this out as per  https://stackoverflow.com/questions/51040703/what-return-type-should-be-used-for-settimeout-in-typescript

Use `ReturnType<typeof setTimeout>` for the type, as in

`const timer: ReturnType<typeof setTimeout> = setTimeout(() => '', 1000);`

PR #1099, opened 2020-04-13
Chart of Joseki visit data

GreenAsJade:
Initial implementation - a few rough edges due to react-charts being in beta, but good enough to be better than the table of numbers, I think.

(At least, that's how it looks on beta, I guess we'll see with the real numbers in it!)


anoek:
Is there a joseki server update that goes along with this? I'm not seeing any charts :(

GreenAsJade:
No server update for this one.

The chart is supposed to be hidden away under the "Details" button on the "Updates/Admin" page:

Screen Shot 2020-04-15 at 7 10 58 am (2)

(this chart looks wierd because people hardly ever visit OJE on beta :D )

anoek:
Ah cool :)

Merging because it's behind admin so not a big deal, but FYI the chart labels are black in dark theme, that and the chart does this really slow expandy thing when it's opened up, but I reckon you can patch those as they bug you :)

PR #1101, opened 2020-04-15
More protection against backsteppers in joseki

GreenAsJade:
... and fix the superfluous 's' on the game timings page.

## Proposed Changes

Don't accept a back button press while we're waiting for a server response.

(Was supposed to be like that already, but not enough protection in place to achieve it)


anoek:
Hmm, maybe we can fix this a different way? People get frustrated when they can't hammer buttons :)

How about this: When you make a request, remember the current move, then when the result comes back, check to see if the engine is on the move it expects - if not, discard the results.

I stuffed a unique `id` in each `MoveTree` node for this exact purpose, so I think what you're looking for is `goban.engine.cur_move.id` - store and check that, and I think you'll be good to go to let people hammer away.

GreenAsJade:
Would you be amenable to putting this in while I figure out the enable-hammer?

PR #1106, opened 2020-04-18
Hide game history colors when ratings are hidden

### LINKED ISSUE #1092: Suggestion: When ratings are set to hidden, hide the colors in the profile game history
**Problem:** At the moment, when ratings are set to hidden, you can still see whether an opponent was stronger or weaker than you by looking at the background color of the game in your game history (lighter colors for stronger, darker for weaker). I think it would better fit the spirit of the 'hide ratings' option to have these colors hidden.

Annotation 2020-04-07 163620

**Solution:** Remove the game history colors when 'hide ratings' is selected.

**Additional context:** I'm trying to do a challenge on OGS where I play 10 games with ratings hidden, and then look back on the results to see if there were any surprises, but the colors make this difficult. It's hard to avoid because I still need access to my game history.

stefankoegel:
Fixes #1092

## Proposed Changes

  - Add a new css class .library-hidden-result that is used for all results in the game history when ratings are hidden.

### DIFF
--- a/src/views/User/User.styl
+++ b/src/views/User/User.styl
@@ -402,11 +402,11 @@
     }
 
     .game-history-table, .review-history-table {
-        .name, .date, .player, .library-lost-result, .library-won-result, .library-tie-result, .library-lost-result-vs-weaker, .library-lost-result-vs-stronger, .library-won-result-vs-stronger, .library-won-result-vs-weaker, .library-won-result-unranked, .library-lost-result-unranked {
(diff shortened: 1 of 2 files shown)


anoek:
Nice improvement, thanks for the PR, and sorry for the long merge time!

GreenAsJade:
@stefankoegel @anoek  It appears this removes game history colours unconditionally, not only if ratings are hidden,

GreenAsJade:
Hmm - not sure what happened there.  This was reported in #English, and seemed to be true, but now I go back to it, it all seems to be working.


🤷

PR #1117, opened 2020-04-25
Fix: Make threshold settings accept 0

flovo:
Fixes: Thumbnail threshold setting didn't save 0. When `parseInt` evaluated to 0 (false), we skipped the save settings part.

## Proposed Changes

- Check if `parseInt` evaluates to NaN. `parseInt(undefined)` and `parseInt("")` both evalutate to NaN, therefor we can skip the check for undefined.
- Also, when the preference is changed, set the text field to the stored value. This makes it more transparent, which value is used.


anoek:
Whoops, indeed, thanks for the fix!

PR #1121, opened 2020-04-26
fix tabcompletion match every username thats's 1 letter longer than input

flovo:
Fixes: In some cases `matchFullName` matched multiple different usernames. This resulted in various occurrences of `@" /0"`. The result was either additional `@"/0"` cluttering the chat line, or worse, the chat substitution to interpret is as user 0. The later results in hiding the username.

This fix prevents the mismatch from happening. 

----

The error happens if the input is 1 char shorter than any username. In this case `lastIndexOf` retruns -1, since the username isn't a substring of input. Since input.length was 1 letter shorter than the username, the result of `input.length - value.length` was -1 (equal to `lastIndexOf`). 

An additional check if the username is a substring of input solves this issue.

### DIFF
--- a/src/lib/tabcomplete.ts
+++ b/src/lib/tabcomplete.ts
@@ -135,7 +135,8 @@ function matchFullName(input, nicknames) {
     let letter;
     let letters = "";
     $.each(nicknames, (index, value) => {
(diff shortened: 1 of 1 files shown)


BHydden:
Wow fast work, well done ❤️

PR #1124, opened 2020-04-27
Navigate to last visited puzzle of collection

### LINKED ISSUE #1112: Save puzzle number in localstorage
**Is your feature request related to a problem? Please describe.**

When working on a large puzzle set (e.g. [Cho Chikun’s Encyclopedia of Life and Death](https://online-go.com/puzzle/10600)) it is annoying to constantly find where I left off the last time.

When I click the link that takes me to a puzzle set I have previously opened in my browser, instead of it taking me to the first puzzle, it should take me to the last puzzle I visited.

A user can simply bookmark the puzzle before they leave, but that also annoying and doesn’t help if the browser suddenly closes or the user accidentally closes the tab.

Off course we could have a more involved system which marks which puzzles have been solved, which puzzle sets we have initiated etc. (proposed in #1031)

**Additional context**

None.

runarberg:
Fixes #1112

## Proposed Changes

  - Save a visited puzzle of collection in local storage
  - If a collection is found in local storage, navigate to it instead of the starting puzzle.

### DIFF
--- a/src/views/Puzzle/Puzzle.tsx
+++ b/src/views/Puzzle/Puzzle.tsx
@@ -206,6 +206,7 @@ export class Puzzle extends React.Component<PuzzleProperties, any> {
                 this.setState(state);
                 this.onResize(true);
                 window.document.title = state.collection.name + ": " + state.name;
+                data.set(`puzzle.collection.${state.collection.id}.last-visited`, state.id);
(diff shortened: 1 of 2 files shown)


anoek:
Nice job, thanks for the PR!

PR #1133, opened 2020-05-07
Give mods a button to copy a PM session to the player's moderator log

GreenAsJade:
It's mildly a nuisance to copy and paste the dialog with players into their moderator log.

Here is a button that does it in one click.


BHydden:
Honestly, I could see myself misclicking this far too often, especially since we have  no way to delete mod logs... I can't say I'm realy in favour of this... plus there are all the times you don't necessarily want the whole chat history, if you've spoken with them about unrelated matters prior.

GreenAsJade:
It could have a confirm button, though mod logs are cheap.

If you don't want the whole lot, maybe copy and paste as normal...

BHydden:
Mod logs aren't cheap, they're permanent. They're one of the few things on the site we can't correct. My issue is not at all "I wouldn't use it" my issue is gravely and severely "I might unintentionally use it with no means of correction"

GreenAsJade:
Mmm - maybe we do need a delete-mod-log button.

BHydden:
> Mmm - maybe we do need a delete-mod-log button.

This would have my undying support ❤️ if we had that, I would have no issues at all with your current PR :-)

GreenAsJade:
But ... since that's probably API dependent - what if we had a confirmation dialog where you could edit what's going into the log if you want?

BHydden:
Yeah that sounds pretty good actually, I'd probably use that

GreenAsJade:
OK, that was a good improvement, I like it.

GreenAsJade:
Screen Shot 2020-05-08 at 6 37 49 pm (2)

BHydden:
> Cease thy trolling!

Never!!! 🤣

anoek:
Looks great, thanks!

PR #1134, opened 2020-05-08
Default custom komi value

andreportaro:
Based on: https://forums.online-go.com/t/cant-start-live-bot-games/27265/4, creates simple validation when trying to create challenges using an invalid custom komi.

## Proposed Changes

  - defaults the custom_komi to 0
  - Prompt user to correct custom komi settings in case they sent a null value (probably from bad session having being saved in the state)


anoek:
Thanks for the fix!

PR #1136, opened 2020-05-12
Fix #1126, turn puzzle names into links

### LINKED ISSUE #1126: Make puzzle link an actual `<Link>`
**Describe the bug**
Currently the puzzles in the puzzle-list table are programmatically linked to inside a `click` event listener—as opposed to with an `<a href="/puzzle/id">` tag. This produces a number of annoyances which includes a lack of keyboard navigability, a lack of context-menu for the link (right mouse click), a pop-up blocked center mouse click, etc.

**To Reproduce**
1. Go to '/puzzles'
2. Scroll down to 'puzzle table'
   - Notice you cannot <kbd>Tab</kbd> to a desired puzzle.
3. Hover over any row
   - Notice it is *not* an actual link.

**Expected behavior**
* I should be able to right click the link (say to bookmark, or copy URL).
* I should be able to middle click the link to open in new tab without my pop-up blocker stopping it.
* I should be able to navigate to the puzzle using the <kbd>Tab</kbd> key on my keyboard.

**Screenshots**
mynd

 - OS: Ubuntu 19.10
 - Browser Firefox 75.0
 - Version 75.0 (64 bit)

**Additional context**

Note that the tournament table doesn’t have the same row click handler, and favors traditional anchor links. Marking inconsistency between views.

Note that the rating column is not interactive but is visually identical to an interactive rating in an open puzzle. Trying to rate the puzzle collection will simply open the collection—which is a little confusing—because of the row click handler.

Some Users might find it preferable to be able to click anywhere on the row to open the puzzle. So there might be a valid reason not to implement this (or add the link without removing the row click handler).

Also note that it is a bit pedantic to complain about the lack of keyboard navigability—especially since the puzzle it self cannot be solved using the keyboard (as far as I know at least).

cassidoo:
Fixes #1126

## Proposed Changes

  - Makes Puzzle table names into links so that people can right click the links
  - Fix minor typos in contribution guide

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,18 @@
 # Development Environment
 
-Getting setup is easy, you'll need to have [node.js](https://nodejs.org/) installed,
+Getting setup is easy, you'll need to have [Node.js](https://nodejs.org/) installed,
 then simply clone the repository and within the working directory run the following:
 
 ```
-# You only need to run this the first time
+# You only need to run this the first time if you don't already have yarn installed
 sudo npm install -g yarn
 yarn install
 
 # Run this to start the development server and build system
-npm run dev
+yarn run dev
 ```
 
-If you're on linux, you can simply type `make` and it will do all this for you as well.
+If you're on Linux, you can simply type `make` and it will do all this for you as well.
 
 Once running, you can then navigate to [http://dev.beta.online-go.com:8080/](http://dev.beta.online-go.com:8080/)
 which loads the interface from your local server that you just started with gulp, and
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -1,57 +1,62 @@
...
 
 ```
-# You only need to run this the first time
+# You only need to run this the first time if you don't already have yarn installed
 sudo npm install -g yarn
 yarn install
 
 # Run this to start the development server and build system
-npm run dev
+yarn run dev
```
BHydden, line 12: May I ask what the actual difference is? I used `npm run dev` to set up my environment and it seemed to work fine, what changes when you use `yarn run dev` instead?

cassidoo: It works pretty much exactly the same in this case. But, if you're going to install `yarn` right before this, you might as well use it, otherwise there's no point in having it installed in the lines above.

cassidoo:
@anoek? @GreenAsJade? Any thoughts?

BHydden:
> Any thoughts?

anoek is a little busy right now with [this](https://forums.online-go.com/t/re-correspondence-games-lost-by-disconnection/26703/261?u=bhydden) but once that's resolved he should have more time to get back to github issues and PRs :)

src/views/PuzzleList/PuzzleList.styl
```diff
@@ -38,6 +38,14 @@
         font-weight: bold;
     }
 
+    .puzzle-name a {
+        color: black;
```
anoek, line 42: This should probably be something like

`themed color link-color`

anoek:
Because of the `onRowClick` handler in that `PaginatedTable` instance you're modifying, just switching them to links doesn't work since the row handler triggers regardless of what you're trying to do when clicking the link. I think you could update the handler to prevent that behavior though, probably by checking the event with `shouldOpenNewTab` found in `lib/misc.ts`.

anoek:
(Overall I like the intent of this PR though, thanks!)

anoek:
Feel free to re-open this if you want to tackle getting the row handler / link stuff is working

PR #1152, opened 2020-05-30
Fix bug where stone placed right after sequence reset in play mode

### LINKED ISSUE #1151: Playing a stone in Joseki 'Play Mode' immediately after resetting Joseki
**Describe the bug**
If you go to the Joseki explorer, and play a move right after resetting the Joseki sequence, it causes weird errors.

**To Reproduce**
As above, just see this video:
[https://www.loom.com/share/af35f7fa811b49bdbcb596c9a11a0d8a](https://www.loom.com/share/af35f7fa811b49bdbcb596c9a11a0d8a)

Also happens whenever a move is played too quickly, even if it isn't the first move. When it's not the first move, people probably encounter it less often (since they have to think about their next moves)

**Expected behavior**
This doesn't happen

ofekih:
Fixes #1151

It just forces user to wait until fetch before placing first stone. Fixes bug

See below video: [https://www.loom.com/share/ff2b8c42db1f4dee988c7d42c2383449](https://www.loom.com/share/ff2b8c42db1f4dee988c7d42c2383449)

I'm clicking actively, but it doesn't place the stone until sequence loaded

It may be possible to have this work where it doesn't say "That move isn't listed!" if it sees the fetch didn't happen yet, and perhaps just wait until the fetch occurs. I could try implementing that if desired, but this quick solution works

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -324,7 +324,8 @@ export class Joseki extends React.Component<JosekiProps, any> {
     }
 
     resetJosekiSequence = (pos: string) => {
-        // ask the server for the moves from postion pos
+        // ask the server for the moves from position pos
+        this.goban.disableStonePlacement(); // prevents a bug where move is played before moves are fetched
         this.fetchNextMovesFor(pos);
     }
 


src/views/Joseki/Joseki.tsx
```diff
@@ -324,7 +324,8 @@ export class Joseki extends React.Component<JosekiProps, any> {
     }
 
     resetJosekiSequence = (pos: string) => {
-        // ask the server for the moves from postion pos
+        // ask the server for the moves from position pos
```
ofekih, line 327: fixed a minor typo in this line (postion -> position)

ofekih:
Turns out this problem happens whenever a move is played too quickly as well, I should address everything.

GreenAsJade:
Thanks - great that you're tackling this!

I think it's perfectly fine to disable placement while we're resetting the board.

It would be sad to insert more waits, particularly, if they end up user visible, though.

Another approach I have had in mind is to make sure that the response we get from the server is in fact the one that we're currently waiting for.

I _think_ that these bugs are caused by another request going out before the previous one has returned, and the code getting the reply thinking it will be for the latest request, but actually it is for the one before.

At least - is this something worth thinking through if you are staring at this part of it!

ofekih:
Sorry for the slow reply, I was busy. So these bugs are all caused by the fact that `processPlacement` is called before the fetch finishes, so `next_moves` is not yet updated, and then:

```ts
const chosen_move = this.next_moves.find(move => move.placement === placement);
```

inside `processMoves` messes up. I think also other things mess up afterwards, but it's more of the same type of mess up. So you're completely correct in your thoughts.

`processPlacement` is called from `onBoardUpdate`, which is called whenever a piece is placed. So the board sends a request to the Joseki server to get the next moves. Then, the user plays a move before the request is returned and `next_moves` is updated, causing the erroneous `That move isn't listed!` messages, and other stuff to go wack.

So your suggested approach wouldn't exactly work, since everything is happening as per expected. The response from the fetch is correct. The `onBoardUpdate` is correctly called. The only problem is that `onBoardUpdate` is called before the correct fetch results are returned. It's just a timing thing.

So an easy solution would just be: don't place moves while we're waiting from a fetch request. We're not ready to hear your moves yet, we wouldn't know if it's correct or not, so just wait a moment.

Another solution would be in the `onBoardUpdate`, right before calling `processMoves`, check to see if we parsed our fetch results yet, and if not, wait to call `processMoves` until we're ready to handle. For the user, this would look like they could immediately play their moves, but they would have to wait a moment before they see any response, be it `That move isn't listed!` or another response, but when they do see a response, it will be accurate.

The second approach may be a bit nicer for the user, but not by much. As long as they don't play joseki instantly, then the fetch would naturally complete by the time they want to play a move. I needed to constantly spam a joseki move very quickly in order to reach this bug, so making them wait a split second isn't that bad.

Curious to hear your thoughts :)

ofekih:
hey, just saw your pm, yeah, it's probably better to just make this issue impossible rather than fix it on a case-by-case basis by preventing users to make moves. lemme know if you have suggestions. I'm busy tonight, but probably sometime tomorrow or later this week I'll be able to try implementing stuff

GreenAsJade:
Busy lives eh!

"Another solution would be in the onBoardUpdate, right before calling processMoves, check to see if we parsed our fetch results yet, and if not, wait to call processMoves until we're ready to handle. For the user, this would look like they could immediately play their moves, but they would have to wait a moment before they see any response, be it That move isn't listed! or another response, but when they do see a response, it will be accurate."

I really like this.  I think the scenario where a person is playing faster than the responses are coming is where by definition they don't care what the response is, they want to get to the next position.   

I can't think of any downside, if it's going to work! 

I _think_ you mean that in the case where the user got ahead of the responses, then processMoves is going to get called after a response, rather than immediately: IE we will see a conditional call to processMoves in the response handling code, which is called if there is processMoves outstanding...

ofekih:
yup, busy lives indeed haha

so I mentioned many more points in a reply to your pm on the forums.

the problem is, if we're in onBoardUpdate, we see that we didn't fetch results yet, and we want to wait, the problem is how would we wait? a busy wait wouldn't work because js is synchronous. the other solution would be a recursive setTimeout method, which may work, but is pretty messy, and as we mentioned on the forums is probably not the best solution. so I'm not sure what an elegant way to wait is ;(

anoek:
Feel free to re-open if this is still applicable / when you feel it's ready

PR #1163, opened 2020-06-06
Add support for iOS 10 in minimized scripts

### LINKED ISSUE #758: Blank screen in OGS's interface

guest951:
Fixes #758 (for iOS 10)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,6 +4,7 @@ var path = require('path');
 let fs = require('fs');
 var webpack = require('webpack');
 const pkg = require('./package.json');
+const TerserPlugin = require('terser-webpack-plugin');
 
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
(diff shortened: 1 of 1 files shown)


anoek:
Awesome, thanks! I copied this over to the goban repository too: https://github.com/online-go/goban/commit/aa3011729c2fdd458a109ec12bbdbd453759b895

anoek:
Apparently you had a PR for it there already, thanks and sorry about that I should have checked!

guest951:
Sorry, I should have precised. Thanks for the merge!

Issue #1113, opened 2020-04-23
Katago Point Estimation instead of Winrate


Zacharias030:
Could you please enable a setting to alternatively display the points-estimation from katago instead of the winrate delta on moves? Seems  like a cool feature for site supporters :)

I'd like it best, if  the  points total could be displayed on moves with negative numbers meaning that white is leading.

koshogs:
Particularly useful suggestion in handicap games as the %win changes can be almost meaningless.

Mentioned here: https://forums.online-go.com/t/ai-review-game-score-graph-feature-request/24422

Zacharias030:
Any news on this one?

anoek:
Thanks @djma for the implementation!

PR #1164, opened 2020-06-07
AI reviews graphing scores: First PoC

djma:
Gives a shot at #1113

https://forums.online-go.com/t/proposal-score-based-ai-review

## With Win % estimates
image

## With Scores estimates
image

### DIFF
--- a/src/views/Game/AIReviewChart.tsx
+++ b/src/views/Game/AIReviewChart.tsx
@@ -32,6 +32,7 @@ interface AIReviewChartProperties {
     updatecount : number;
     move_number : number;
     setmove     : (move_number:number) => void;
+    use_score   : boolean;
 }
 
 const bisector = d3.bisector((d:AIReviewEntry) => { return d.move_number; }).left;
--- a/src/views/Game/AIReviewChart.tsx
+++ b/src/views/Game/AIReviewChart.tsx
@@ -55,6 +55,7 @@ export class AIReviewChart extends React.Component<AIReviewChartProperties, any>
     width?:number;
     height?:number;
     max_score?:number;
+    min_score?:number;
     win_rate_line_container?:d3.Selection<SVGPathElement, unknown, null, undefined>;
     win_rate_area_container?:d3.Selection<SVGPathElement, unknown, null, undefined>;
     win_rate_line?: d3.Line<AIReviewEntry>;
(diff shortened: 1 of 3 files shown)


anoek:
Awesome, great progress! And thanks for tackling this, it's been on my todo list forever!

Some things I noticed:

* When viewing a leela zero reviewed game, that option is still there - we should hide it when it won't display anything useful
* Ideally it would remember the toggle state
* The red "analyzing" progress balls disappeared somewhere along the line
* Those balls should be present on this graph too
* The "waiting for processing" animated graph should be present for the score estimate too while we are awaiting the initial data from the servers

djma:
Thanks for the quick review!

> When viewing a leela zero reviewed game, that option is still there - we should hide it when it won't display anything useful

Fixed c30a73e 

> Ideally it would remember the toggle state

I thought about that too, but I concluded that it might be preferable to keep it defaulted to win% because scores aren't always available. And if somebody had "score-mode" set, they would see a different chart depending on the AI review.

What do you think?

> The red "analyzing" progress balls disappeared somewhere along the line
> Those balls should be present on this graph too

How do you get into an "analyzing" state where you see those red balls?

image

> The "waiting for processing" animated graph should be present for the score estimate too while we are awaiting the initial data from the servers

Fixed d051d57

anoek:
> I thought about that too, but I concluded that it might be preferable to keep it defaulted to win% because scores aren't always available. And if somebody had "score-mode" set, they would see a different chart depending on the AI review.
> 
> What do you think?

I'd say if they have it set to score, and they load one where it's not available, it should display the win% regardless of the setting (and the setting will be hidden), then when they next view a review that has the point estimation, it should display the point estimation because that's what was last set still.

Seem reasonable, or do you think that gets confusing to the user?


> How do you get into an "analyzing" state in dev where you see those red balls?

Easiest way is to click the + KataGo button or + LeelaZero button from the drop down on a game, that'll kick off a new review and select it by default, so you'll see the red balls.

To do that though you'll need to become a beta site supporter, so head to site supporter, click Stripe as your checkout method, plug in any email, 

4111 1111 1111 1111

as the credit card number, any future expiration date, any ccv, fill in the rest of the fields with whatever, buy, then presto you're a site supporter and those buttons should work for you.

djma:
LeelaZero, no scores:

image

Full Katago with scores:
image

Full Katago with win%:
image

Free Katago with scores (no branch analysis):
image

Free Katago with win% (no branch analysis):
image

Free Katago with scores (on the red dot):
image

Free Katago with win% (on the red dot):
image

Analyzing squiggly lines (sorry for fast screen cap. It only lasts for milliseconds):
image

src/views/Game/AIReviewChart.tsx
```diff
@@ -284,11 +251,23 @@ export class AIReviewChart extends React.Component<AIReviewChartProperties, any>
...
+          && this.props.ai_review.scores != null
+          && !this.props.ai_review.scores.includes(0);
+
         if (this.props.entries.length > 0) {
-            entries = this.props.entries.map(x => x);
-            entries.unshift({win_rate: 0.5, move_number: 0, num_variations: 0});
+            entries = this.props.entries.map((x, i) => {
+                return {
+                    win_rate: x.win_rate,
+                    score: x.score === 0 && use_score_safe ? this.props.ai_review.scores[i] : x.score,
```
djma, line 262: The fact that `entries*.score` and `ai_review.scores` could be different, was honestly really confusing.

djma:
@anoek do you have time for a re-review next time you're working on this repo? Thanks!

anoek:
This is great @djma , thanks for the PR!

PR #1169, opened 2020-06-10
SingularPointMiniGoban

ajcsimons:
Switch the vertical current score display on mini goban, as on my/observe games page, to use ngettext so as not to say "1 points". Also changed state object to contain the raw number and string building in the view as I think that's better practice, don't know what you OGS folks think. Assumes numbers come before unit, which may not be valid in all languages.

Fixes: silly looking "1 points"


ajcsimons:
Here's a new version that runs it through interpolate so you can put the number after the word for points, but because these new strings aren't in the translation catalog, whereas 'point\u5points' already was, other languages won't get the translation until those new translation strings are added.

anoek:
Great, thanks!

PR #1173, opened 2020-06-13
chat_manager: fix ignoring of blocked users

flovo:
Last weeks https://github.com/online-go/online-go.com/pull/1160 had a bug, resulting in (almost) all messages are considered from blocked users. Therefore they didn't increased the unread chat count and mentioned status.

This time I checked it both ways. 
1) Blocked messages don't count
2) Not blocked messages count


anoek:
Thanks for the fixes!

PR #1178, opened 2020-06-18
Fix #1177 Group name search set to 'contains' instead of 'startswith'

### LINKED ISSUE #1177: Searching for group by name should be "contains" instead of "start with"
I have tried to search for the group "One Arbitrary Group" a lot of time but faileds, because I remember the name as "An Arbitrary Group".

Similar situation with famous global initiative "Teaching Ladder". Every go server have a "[Something] Teaching Ladder", but you can't search for "Teaching" here. You must search for "OGS" because the group name "OGS Teaching Ladder" start with OGS.

My point is, many times the signature word of a group's name is not the first letters. Changing from "start with" to "contain" will make the search more bloated, but at least the list will include groups that user was hoping to find.

AVAVT:
This is my proposed PR for #1177

### DIFF
--- a/src/views/GroupList/GroupList.tsx
+++ b/src/views/GroupList/GroupList.tsx
@@ -55,7 +55,7 @@ export class GroupList extends React.PureComponent<GroupListProperties, any> {
                         <SearchInput
                             placeholder={_("Search")}
                             onChange={(event) => {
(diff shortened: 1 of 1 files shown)


anoek:
I feel like we had it like this once upon a time and there was a reason why we switched, but I'm not sure why, it seems like it's better to have it as `icontains`.

Thanks for the PR!

PR #1179, opened 2020-06-21
External goban image

Adam314159:
**WARNING: This PR is co-dependant with [this one](https://github.com/online-go/goban/pull/6) on goban**

---

The basic idea is to allow users to style their goban with a custom image in the form of supplying a link to an external picture. 

- I have changed the look of the plain board preview in the selector, to more intuitively represent the cutomizable nature
Screenshot_2020-06-21 OGS

- Upon selecting the custom option an input form appears together with the color selectors, once the text is replaced with an image link, it will automatically update the board background. 
Screenshot_2020-06-21 OGS(1)

- in the future I was thinking of creating a similar option for the stones, together with two buttons allowing for playing one-color-go and blind-go easily. Let me know what you think about that.

---

Please note, this is my first barbecue with React. I mostly wanted to have a look around with a simple/fun project. If the code is not up to par, or if you think it was a stupid idea to begin with, don't be afraid to tell me to shove it :)

### DIFF
--- a/src/components/GobanThemePicker/GobanThemePicker.styl
+++ b/src/components/GobanThemePicker/GobanThemePicker.styl
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2012-2017  Online-Go.com
+ * Copyright (C) 2012-2020  Online-Go.com
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as
--- a/src/components/GobanThemePicker/GobanThemePicker.styl
+++ b/src/components/GobanThemePicker/GobanThemePicker.styl
@@ -28,15 +28,19 @@
         flex-direction: row;
         flex-wrap: wrap;
         justify-content: center;
-		text-align: center;
-		font-size: 10px;
+        text-align: center;
+        font-size: 10px;
+        }
 
         .selector {
             margin: 1px;
             cursor: pointer;
             border: 1px solid transparent;
             opacity: 0.5;
-			color: white;
+            color: white;
+            word-wrap: break-word;
+            overflow: hidden;
+            text-transform: uppercase;
         }
 
         .selector:hover, .selector.active {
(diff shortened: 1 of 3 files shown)


src/components/GobanThemePicker/GobanThemePicker.styl
```diff
@@ -28,12 +28,15 @@
         flex-direction: row;
         flex-wrap: wrap;
         justify-content: center;
+		text-align: center;
```
anoek, line 31: I see you converted tabs to spaces in the .ts files but looks like the .styl file was missed, spaces here too please :)

src/components/GobanThemePicker/GobanThemePicker.tsx
```diff
@@ -118,11 +123,15 @@ export class GobanThemePicker extends React.PureComponent<GobanThemePickerProper
...
                                 ...theme.styles,
-                                ...(theme.theme_name === "Plain" ? {backgroundColor: boardCustom} : {})
+                                ...(theme.theme_name === "Plain" ? {backgroundImage: "linear-gradient(-45deg, orange,yellow,green,blue,indigo,violet)"} : {})
                             }}
                             onClick={this.selectTheme["board"][theme.theme_name]}
                             >
-                            <PersistentElement elt={this.canvases.board[idx]} />
+
+                            {theme.theme_name === "Plain"
+                                ? <span>CUSTOM</span>
```
anoek, line 132: Everything's gotta be translatable, for this I would use something like

`<span>{pgettext("Custom board theme", "CUSTOM")}</span>`

and also make sure the span has some css applied that makes that text either hides the overflow (`overflow: hidden`) or wraps (`word-wrap: break-word`), whichever you think looks better

src/components/GobanThemePicker/GobanThemePicker.styl
Adam314159: Curses and appologies, should be fixed.

src/components/GobanThemePicker/GobanThemePicker.tsx
Adam314159: Yup, I actually did both, wrapped the text and hidden any overflow on the off chance the translation was that long. 

I think both the custom selector and the default text in the form should now be translatable. Thanks for checking it out!

anoek:
Thanks Adam, nice work!

PR #1183, opened 2020-06-28
Set OGS theme to match user's OS theme on first visit only

### LINKED ISSUE #1047: Use CSS prefers-color-scheme to automatically select dark mode
OGS already supports a dark theme. It would be nice if it was enabled automatically if the user enable dark mode in his operating system.

This can be achieved with the new `prefers-color-scheme` CSS media query. Example:

```
@media (prefers-color-scheme: dark) {
   /* here are css rules that are only interpreted in dark mode */
}
```

Documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

tsonge:
Fixes #1047 

This simply sets the OGS light/dark theme preference to match the user's own operating system's theme preference, only for the first visit. 

Note that afterwards if the user changes the light/dark theme from within OGS, that preference will remain on subsequent visits.

EDIT: To be honest I don't know if we want to sync up the OGS theme with the user's OS theme on _every_ visit, or just the first visit, or even perhaps implementing a new option entirely called "Set OGS theme to sync with OS". In any case for this PR I just went with first visit only sync.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -105,7 +105,18 @@
   <body>
       <script>
         try {
-            document.body.className = JSON.parse(localStorage.getItem('ogs.theme') || '"light"');
+            let theme = localStorage.getItem('ogs.theme');
+            if (theme) { // user has been on the site before
+                document.body.className = JSON.parse(theme);
+            } else { // first time user is on the site, so set initial theme according to his operating system theme.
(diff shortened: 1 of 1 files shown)


anoek:
Very slick, thanks this has been on my todo list to tackle, I didn't think of solving it this way - much cleaner than what I was thinking we'd have to do.

Issue #946, opened 2019-10-05
[Chat] right->left layout seems to be broken


flovo:
**Describe the bug**
The ordering of the chat line elements seems to be wrong for languages with right->left writing direction. 

**To Reproduce**
1. Go to Arabic chat
2. Write multiple words in chat
4. See: The order of the elements in the chat line is strange. (The username is on the left, the flare on the right). Spaces in the username split the username around the message.
Additionally, the cursor jumps back and forth while writing, but it could be caused by inputting lattin letters.

**Expected behavior**
The username should be on the right, left to the time and the flare.

**Screenshots**
![image](https://user-images.githubusercontent.com/4864182/66254590-aa412a80-e778-11e9-8706-42028d5ffc96.png)
![image](https://user-images.githubusercontent.com/4864182/66254706-6e0ec980-e77a-11e9-8db8-05ae091848e0.png)

user with space in name
![image](https://user-images.githubusercontent.com/4864182/66254716-8e3e8880-e77a-11e9-963d-0aa3b504d72a.png)
![image](https://user-images.githubusercontent.com/4864182/66254713-82eb5d00-e77a-11e9-93fa-91e895891bf0.png)

**Desktop (please complete the following information):**
 - OS: ArchLinux
 - Browser: Firefox 69.0.2 and Chromium 77.0.3865.90

**Smartphone (please complete the following information):**
 - Device: OnePlus 6
 - OS: Android 9
 - Browser Firefox 68.1.1
 - Version 68.1.1

anoek:
This issue has been marked stale and will be closed soon without further activity. To keep the issue open, please respond to the comment to keep the discussion going.

PR #1185, opened 2020-06-29
Prevent rtl chats from losing rtl class when scrolled

tsonge:
I was looking at #946 and while I couldn't reproduce the error he was having, I did nevertheless find the rtl (right to left reading languages) chat channels a bit buggy. 

Initially when we click on an rtl chat channel (like, say Arabic), the text is displayed correctly, sticking to the right hand side. As soon as we scroll, the div with the `chat-log` class loses the `rtl` class, which ends up displaying the text messages sticking to the left hand side (incorrect for rtl languages).

This fixes it.


anoek:
Nice fix, thanks!

PR #1190, opened 2020-07-04
Anull hammer fall

GreenAsJade:
When games became un-anullable, we lost the ability to see if they are annulled by hammer status.

## Proposed Changes

Make the hammer fall when the game is annulled, so we can see that it has been done.

Also fix up the status of the Dock after annullment has happened (the info and hammer status need to be refresher)


anoek:
Nice, thanks!

PR #1193, opened 2020-07-07
AIReview: draw out ghost branch even when on trunk

### LINKED ISSUE #785: Show AI playout ghost stones when selecting the triangle move
For top 3 moves, if you select the triangle move and it has been analyzed, we currently cannot see the moves the AI has played out. We need some way of seeing that, instead of simply progressing to the next move.

tsonge:
Fixes #785 

Most of the highlighted file changes are simply due to reindentation. But essentially I'm wrapping the code where we generate the AI marks in a for loop to use the AI generated branches from previous trunk moves as well, to fix the bug (see fillAIMarksBacktracking method).

Screen Shot 2020-07-07 at 4 59 46 AM

You can test this by visiting http://dev.beta.online-go.com:8080/game/8598 @ move 2/3 (where the triangle move is part of an AI branch). Please use the account username: huh, password: woo to view the game since it is private.


anoek:
That's great, thanks for figuring out how to do that!

PR #1198, opened 2020-07-12
Don't leave joseki throbber running after a server response we couldn't parse

GreenAsJade:
Fixes #

If you enter a bad URL looking like a joseki position, the server will respond with an error.

This is caught by the joseki client code, but it leaves the throbber running, so the user has no idea that actually there's nothing else going to happen.

## Proposed Changes

Turn off the throbber when a server error is received.


anoek:
Thanks for the continued maintenance on this awesomeness :)

PR #1199, opened 2020-07-12
Allow more time settings

flovo:
On the forum, we got multiple requests to allow a wider range of time settings. One request is to increase the time cap for fisher time, the other to allow live games with high main time and blitz like time increment. 

https://forums.online-go.com/t/fixed-max-time-in-fischer-blitz-settings/29003?u=flovo
https://forums.online-go.com/t/suggestion-allow-disabling-time-cap-for-fischer-time/21665?u=flovo

Besides adding additional time options, I merged live and blitz. A distinction in the selection dialogue is difficult. Already right now time settings like 5m+9s are only available in the blitz category, but count as live game for `computeAverageMoveTime` it is a live game. So I would just drop the distinction between the two and just distinguish between live and correspondence games when creating a game.

### DIFF
--- a/src/components/TimeControl/TimeControl.ts
+++ b/src/components/TimeControl/TimeControl.ts
@@ -16,7 +16,7 @@
  */
 
 export namespace TimeControlTypes {
-    export type TimeControlSpeed = "blitz" | "live" | "correspondence";
+    export type TimeControlSpeed = "live" | "correspondence";
     export type TimeControlSystem = "fischer" | "byoyomi" | "canadian" | "simple" | "absolute" | "none";
 
     export interface Fischer {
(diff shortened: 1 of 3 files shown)


BHydden:
While you're in there, can't you just delete simple time altogether? 😅

anoek:
I'm on board with adding more time options. Removing blitz all together might have some farther reaching ramifications, I don't want to make that change without taking some time to think that one through a bit more.

GreenAsJade:
Just chiming in to say I find the blitz distinction helpful, and would prefer to see it refined and unified (same meaning everywhere) rathe than removed.

It is particularly helpful because "there is no such thing as an escape (timeout) in blitz".   That is in some ways the whole difference.  In blitz you _expect_ your opponent to not be able to make it in time during the game, responding in time is the challenge.

anoek:
Yeah, I think the blitz distinction is good. Feel free to re-open with alternative timing options though.

PR #1200, opened 2020-07-12
Active games list: do not overlap with activity view

flovo:
On the profile page, sometimes the active games list overlaps with the activity list.


anoek:
Awesome thanks, that was bugging me :)

PR #1208, opened 2020-07-21
Adding "Israeli Go Association" into Go resources

Adam314159:
Adding Israeli Go Association as requested in: https://forums.online-go.com/t/adding-a-link-to-other-go-resources/29408


anoek:
Thanks!

BHydden:
For PRs like this, I like to add the new item to the second last position, rather than the bottom.
This way, you don't need to edit the previous bottom to add a comma to it.
Just my 2c :) thanks for the quick response to the forum request!

PR #1210, opened 2020-07-26
Fixing games list wider than view

flovo:
Games list on home and profile page was cropped on the left and right.

With this change, horizontal scroll is available when the list is wider than the available space.


anoek:
Oh nice thank you, that was bugging me too

PR #1211, opened 2020-07-26
Maximise Goban in Zen Mode

flovo:
There is a strange margin in portrait zen mode:

Screenshot_2020-07-26 Tournament Game Kuksu Main Title Tournament 13th Cycle - League D3 (62330) R 1 (flovo vs IcedEarth)

This PR changes the css to maximize the goban in zen mode when possible.


anoek:
This messes up the vertical centering in wide mode, but improves all the normal use cases so I'm gonna just merge it, though vertical centering in wide mode would be nice at some point if that's an easy fix :)

PR #1212, opened 2020-07-26
Show play controls in portrait zen mode

flovo:
In portrait zen mode, ply controls are missing right now. 

Since they are needed to Submit move and resign, we should show them there as well.


anoek:
Yikes, thanks. I think the last person passionate about zen mode removed them, I should have tested that more thoroughly.

PR #1213, opened 2020-07-26
Align clocks above each other in horizontal zen mode

flovo:
In horizontal zen mode we have enough vertical space to place the clocks above each other. 

This allows the user to reduce the windows width even further while keeping the goban at maximum size.


anoek:
Looks like this conflicted after I merged the other zen mode update(s), i couldn't quite see what to do immediately though and I am thinking this might not work the same now that the controls are active again - maybe? Punting it back to you to play with and resolve.

anoek:
Feel free to re-open when ready

PR #1214, opened 2020-07-26
Hide accept undo button in zen mode

flovo:
A suddenly appearing button has the potential to break concentration and
therefore, is not zen.


anoek:
Cool thanks

PR #1215, opened 2020-07-30
game/id/move_number skips to move

ajcsimons:
Add ability to add move number on end of game url to automatically skip to that move upon loading game, e.g. online-go.com/game/24010741/123 will skip to move 123. setTimeout was initially me being lazy to find where to put code to run after the normal logic to skip to last move of game, but actually gives nice visual clue that you are not viewing last move but skipped earlier IMO.

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -117,7 +117,7 @@ export const routes = (
                     }
                 }
 
-                // Make sure it shows up in the left pannel
+                // Make sure it shows up in the left panel
                 joined[channel] = 1;
                 data.set("chat.joined", joined);
 
@@ -125,6 +125,7 @@ export const routes = (
             }}/>
             <Route path="/observe-games" component={ObserveGames}/>
             <Route path="/game/view/:game_id" component={Game}/>
+            <Route path="/game/:game_id/:move_number" component={Game}/>
             <Route path="/game/:game_id" component={Game}/>
             <Route path="/review/view/:review_id" exact component={Game}/>
             <Route path="/review/:review_id" component={Game}/>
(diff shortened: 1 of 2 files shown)


Turbokretzsche:
Is it clear from the api definition that the second number after game ID is always move number? Else might want to update the definition or make a query parameter out of this.

ajcsimons:
> Is it clear from the api definition that the second number after game ID is always move number? Else might want to update the definition or make a query parameter out of this.

Testing on the beta server, the server appears to just ignore the extra /move_number on the end of the game/game_id url, though it is plausible that if the routing did stricter checking it might get upset and barf. I didn't go for the query parameter approach /game/1234?move=123 or hashbang /game/1234#!/move/123 just because /123 is more convenient for users to type, and also more intuitive and guessable I think for non-technical users. Maybe the fact we are using a different root url for what is really a client-side navigation within the same game resource has some bad effect on caches? @anoek

anoek:
I like the concept. Can you update the implementation to use the routes system? (See `routes.tsx`), and you'll want to add the `move_number` parameter to the `GameProperties` interface.

After that, instead of doing a `setTimeout`, I believe if you add an event handler to `this.goban` in `Game` and listen for the `gamedata` event, and do your `nav_goto_move` in there that'll get you what you want without a delay or a possible timer misfire. I think you'd only want to do it the first `gamedata` event though, so maybe `this.goban.once("gamedata", () => { ... })`? Something along those lines.

Nice work~

flovo:
`game/1234?move=123` would have the advantage we could do the same for variations as well. But `game/1234/123? variation=acbf` could work as well.

ajcsimons:
Will look at making those updates in next few days.

ajcsimons:
@anoek updated with suggestions

anoek:
Nice work, thanks!

PR #1223, opened 2020-08-22
add win/loss colors for games when ranks hidden

mattetress:
Fixes https://forums.online-go.com/t/rank-display-disabled-and-profile-page/28624

## Proposed Changes

  - Adds win/loss colors to games in user game history table when ranks are hidden without revealing the relative strength of the opponent


anoek:
Nice work, thanks for the patch!

PR #1238, opened 2020-10-12
Use an icon in the game history table to show clearly who won in each game

GreenAsJade:
## Proposed Changes

Put an icon next to the winner's name in the game history table, so you can see at a glance who won.

Dark:
Screen Shot 2020-10-12 at 6 47 21 pm (3)

Light:
Screen Shot 2020-10-12 at 6 47 05 pm (3)

### DIFF
--- a/src/ogs.styl
+++ b/src/ogs.styl
@@ -145,6 +145,8 @@ light.strong-l-text                     = light.bg
 light.weak-w-text                       = light.fg
 light.weak-l-text                       = light.fg
 
+light.winner-trophy                     = #e6c525
+
 light.shade0                            = lighten(light.fg, 20%)
 light.shade1                            = lighten(light.fg, 40%)
 light.shade2                            = lighten(light.fg, 60%)
@@ -170,7 +172,6 @@ light.react-select-selected-bg          = #2684FF
 light.react-select-selected-fg          = #FFFFFF
 light.react-select-focused-bg           = #DEEBFF
 light.react-select-focused-fg           = light.fg
-
 
 /** Navbar **/
 
@@ -287,6 +288,7 @@ dark.strong-l-text                      = dark.bg
 dark.weak-w-text                        = dark.fg
 dark.weak-l-text                        = dark.fg
 
+dark.winner-trophy                     = #e6b500
 
 /** React Select colors **/
 
--- a/src/views/User/User.styl
+++ b/src/views/User/User.styl
@@ -459,8 +459,13 @@
             max-width: 9rem;
             overflow: hidden;
         }
-    }
 
+        .game-history-winner {
+            themed color winner-trophy;
+            padding-left: 4px;
+            vertical-align: middle;
+        }
+    }
 
     .moderator-log {
         td {
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -904,8 +904,8 @@ export class User extends React.PureComponent<UserProperties, any> {
                                         {header: _("Date"),   className: () => "date",                            render: (X) => moment(X.date).format("YYYY-MM-DD")},
                                         {header: _("Size"),   className: () => "board_size",                      render: (X) => `${X.width}x${X.height}`},
                                         {header: _("Name"),   className: () => "name",                            render: (X) => <Link to={X.href}>{X.name || interpolate('{{black_username}} vs. {{white_username}}', {'black_username': X.black.username, 'white_username': X.white.username}) }</Link>},
-                                        {header: _("Black"),  className: (X) => ("player " + (X ? X.black_class : "")), render: (X) => <Player user={X.historical.black} disableCacheUpdate />},
-                                        {header: _("White"),  className: (X) => ("player " + (X ? X.white_class : "")), render: (X) => <Player user={X.historical.white} disableCacheUpdate />},
+                                        {header: _("Black"),  className: (X) => ("player " + (X ? X.black_class : "")), render: (X) => <React.Fragment><Player user={X.historical.black} disableCacheUpdate />{X.black_won ?  <i className="fa fa-trophy game-history-winner"/> : ""}</React.Fragment> },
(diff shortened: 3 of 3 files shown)


anoek:
Good one, thanks!

BHydden:
There are not words to describe how redundant and tacky this feels to me 🤣 oh well 🤷‍♂️

GreenAsJade:
@BHydden I guess you use OGS on a desktop where the whole game history can be seen without scrolling the pane.

Even in that situation, you are forced to flick with your eyes from the Result string, after you have interpretted it, to the correct column of who won to see who won.

If you are wondering who won, or even wanting an impression of win-rate, that is UX labour.

Note that the overview of a person's performance we get from the green/red of the right hand column is not available to colour blind people.

This change addresses these issues.

BHydden:
Indeed I forgot about mobiles and red-green colour blind people :P without those, it feels like it only helps for annulled games, but yes, accessibility is certainly worth it 👍

GreenAsJade:
The original person who requested this suggested it as an "Option".

I didn't make it a "Settings" option yet because I didn't want to open the can of worms of how far do we go with display customisability via options.

It's a possibility though.

Station_11 also suggested that the mark be on the left of the winner's name, with the goal of making them align vertially.  This would call for an extra space in the loser's column (for the missing mark) which I wasn't keen on.

If we would take this opportunity to move to fontawesome 5, we could use the medal for the mark, which is slightly less visually intrusive than the fat trophy.

All food for thought.

BHydden:
Aaah, without a "resolves issue xxx" or a link back to a forum discussion, I wasn't aware this was a requested feature. All becoming clearer now :) thanks

GreenAsJade:
So if someone else requested it, it's OK, but if I thought if it then it's redundant and tacky?   I see...

BHydden:
If someone else requested it, they likely gave a reason why it was useful. If you had even done that much, it would have been fine too :P all I saw was "I made a thing" and my privileged, colour-seeing ass couldn't think of a reason why you'd done it 😛

BHydden:
the tacky part was only in regards to the trophy, which you yourself admitted wasn't ideal haha

why not just make the winner's name bold?

GreenAsJade:
You're right, I should have put the reason in an Issue.  🤷‍♀️ 

Bolding is possibly an option, though till a bit obtuse and risky for vision impaired people.

I'm experimenting Station_11's suggestion, which was a star on the left.

GreenAsJade:
I tried it and didn't like it.   Unfortunately nothing says "winner" like a trophy.

anoek:
What about the trophy in a column right before 'Result', only present when the player is the winner?

GreenAsJade:
The thing is then you need to engage "processing power" to work out what that new column is trying to say, and then also compute which of the names in the list is actually the player.    That's not always easy, especially with people called 友谊赛 and 友配赛 playing each other.

I'm contemplating making it a setting so when people complain, at least they can turn it off.  

I guess that if Bhydden found it objectionable, then certainly a fair percentage of others will, although I don't...

anoek:
I think you have a "processing power" overhead right now too, there's a bunch of unaligned tournament trophy icons floating around, what do they mean? They won a tournament? Surely not, that's too many tournaments. Maybe they won the game? Yeah that looks right. Point being, it's not necessarily as obvious as us developers think it should be, unfortunately few things are with UI/UX. That being said, I don't know that a trophy column would be any better, just throwing out alternative ideas.

Question though, as it is, what happens when the names are long enough to be truncated by that table, do the trophy icons get pushed off?

FTR I'm open to re-imagining that information in very different ways if it helps us get to an ultimately better look and feel.

BHydden:
as an idea, what if we had winner and loser columns instead of black and white columns, and then we could colour code the names to represent black or white (or append a ⚫ and ⚪ next to the respective name)

GreenAsJade:
My first reaction was negative, but this could have legs.  I'll experiment with that

GreenAsJade:
@BHydden  what if it is a draw?

BHydden:
If it's a draw just have whoever's profile you're looking at listed in the winner's column I guess... tie by moderator is the only way to have  a draw currently anyway because even when integer komi is selected the system still says B+0.0 or W+0.0 for whatever reason

GreenAsJade:
Fun fact - there's code in there that looks like once upon a time the winner's name in the name columns was given special treatment - a css class is set for it .... but it is no longer used - that css class isn't defined.

If I have to guess, I'd say @RubyMineshaft felt it was redundant after he put in the red/green for indicating winner ... understandably!

GreenAsJade:
Here's the latest Bhydden suggestion:

Screen Shot 2020-10-14 at 9 35 21 pm (2)

TBH I like the original submitted as much as any variant.   Any specific choice will have its detractors.

I'm getting inclined to have a set of options and the whingers can chose which they hate least... would be easy enough to do.

GreenAsJade:
Screen Shot 2020-10-14 at 10 03 23 pm (2)

BHydden:
I like the last one best, with colours to the left of names... is it possible to make the white one smaller? I'm sure they're probably the same size, but the white one looks bigger to me because of that optical illusion thingo (if it's too much hassle don't bother, just asking in case it's easy)

anoek:
Maybe we could put it to a vote on the forums? If there's a wide spread on preference, maybe then make it a preference, but if there's a general convergence then we can just go with whatever option that is. 

I'm thinking just throw up those screen shots you have there, option of trophy icon next to winner, winner/looser columns, and original view (ie no change).

BHydden:
Gonna make that Forum thread GaJ? or working on some extra PoC Designs?

BHydden:
Would still love a winner / loser column option <3

GreenAsJade:
I ran out of steam to start a forum, thread - I like it how it is anyhow - but might gain steam again if there were an officially sanctioned thing to do.

GreenAsJade:
Note: if Anoek was not averse to having it as a setting selector, we could just wack it in and let people chose what they want.

BHydden:
Fair enough, BTW KoBa asked in site supporter chat whether it might be a good idea to not display a trophy for an annulled game?

GreenAsJade:
I totally agree that's a must fix bug.   Someone should raise a bug report ;)

BHydden:
> Someone should raise a bug report ;)

See #1249

PR #1242, opened 2020-10-25
Added a show dismissable rating explanation link to PlayerDetails

### LINKED ISSUE #1003: Imposible to copy user name
**Describe the bug**
After selecting the username, right click on it doesn't bring up the context menu, but follows the link to the profile page. This prevents the username from being copied by using the context menu.

**Expected behavior**
Right click should bring up the context menu

The problem happens in Chrome.
Firefox works fine

**Additional context**
https://forums.online-go.com/t/adjust-right-clicking-behavior-for-usernames/23915?u=flovo

### LINKED ISSUE #1241: Have a link to an explanation about ratings
This thread made the suggestion, in the face of ongoing confusion about how ranks move:

https://forums.online-go.com/t/ratings-moving-opposite/30683/21?u=eugene

GreenAsJade:
Fixes #1241

## Proposed Changes

Dismissable link on PlayerDetails - with player ranking present:

Screen Shot 2020-10-25 at 10 38 45 am (2)

Dismissable link on PlayerDetails - with player ranking provisional:

Screen Shot 2020-10-25 at 10 38 56 am (2)

Dark mode:

Screen Shot 2020-10-25 at 10 39 07 am (2)

### DIFF
--- a/src/components/Player/PlayerDetails.styl
+++ b/src/components/Player/PlayerDetails.styl
@@ -76,4 +76,16 @@
             }
         }
     }
+
+    .rating-explanation {
+        display: flex;
+        justify-content: flex-end;
+        align-items: center;
+        .fa {
+            padding-left: 2px;
+        }
+    }
+    .hide-rating-explanation {
+        display: none;
+    }
 }
--- a/src/components/Player/PlayerDetails.tsx
+++ b/src/components/Player/PlayerDetails.tsx
@@ -85,6 +85,7 @@ export class PlayerDetails extends React.PureComponent<PlayerDetailsProperties, 
             ui_class: "...",
             country: "un",
             error: null,
+            show_rating_explanation: data.get("show-rating-explanation", true)
         };
     }
     resolve(player_id) {
@@ -226,6 +227,12 @@ export class PlayerDetails extends React.PureComponent<PlayerDetailsProperties, 
         .then(() => termination_socket.send('chat/remove_all', {player_id: this.props.playerId}))
         .catch(() => 0);
     }
+
+    dismissRankingExplanation = () => {
+        data.set("show-rating-explanation", false);
+        this.setState({show_rating_explanation: false});
+    }
+
     render() {
         let user = data.get("user");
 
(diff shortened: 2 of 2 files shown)


BHydden:
> Fixes #1003

I think you meant to say #1241 ?

GreenAsJade:
Yeah dunno what happened there LOL

benjaminpjones:
I'm pretty new to this, so I don't want to make any out of place suggestions, but have you considered using a small icon for this?  It would be persistent, but also take up less real estate:

PlayerDetailsQuestionMark

BHydden:
Nice idea @benjaminpjones it does indeed look tidier. However, one concern I have with your proposal is that we already use a ? to indicate provisional rank, and adding a second one could lead to confusion and a lot of questions on "how do I remove the ? next to my rank"

image

benjaminpjones:
That's a good point. Another option is the "i" for "info" icon that's pretty widespread (even though "info" doesn't start with "i" in all languages!)

image

GreenAsJade:
These are nice ideas, but I think ultimately it seems preferrable to have more attention to the availability of an explanation, and then be able to clean that up completely.

anoek:
I don't know how I feel about having this link here, this window is supposed to be a quick view of someone's information, and some quick action buttons, a link to "how does this work?" isn't what I'd expect here.

I feel like this is (1) a documentation issue, ie if people aren't finding this information easily enough in the documentation wiki, perhaps we need to move that page to be more visible, and also (2) probably self resolving when the next iteration of the ranking system drops, which will definitely eliminate the confusing "i won but my rank went down", we didn't really seem to have a lot of questions before that so I think the priority of this information will drop as that question goes away.

GreenAsJade:
Maybe we can deploy this as a temporary helper for us all until the next ranking system does come along eh?

benjaminpjones:
Just want to point out that the wiki link in this PR is not created yet.  Did you mean to link to [Getting Started: Your Rating](https://github.com/online-go/online-go.com/wiki/Getting-Started-%7C-Creating-and-Adjusting-Your-Account#your-rating) instead?

benjaminpjones:
Also, one more thought- aren't there other places besides the player card where ranking is shown?  For instance, would a link on the User page have similar effect, without cluttering the PlayerDetails card?

Screen Shot 2020-10-27 at 12 44 15 PM

GreenAsJade:
Good catch, I have no idea where I got that wrong URL from.   I'll fix that ASAP.

I had considered putting it right where you suggested - and I like it - except it's not the place that people are looking when they are wondering "why did my rank go down".

anoek:
I'm going to close this, new rating system shouldn't be too far away now

PR #1245, opened 2020-10-25
Mark annulled games more clearly.

### LINKED ISSUE #1244: Annulled games not clearly marked in the UI
**Describe the bug**
If you go to an annulled game, you have to dig a little to see that it is annulled.  Often, annulled games have some funny business, so it can be disheartening to see an erroneous winner.

**To Reproduce**
1. Go to https://online-go.com/game/27029807
2. Observe that you cannot see that the game is annulled without clicking the info button.
3. Also note that on profiles, occasionally the Game Result column is too far to the right to see 'annulled'

**Expected behavior**
A notice that the game was annulled.  Should be at least as visible as the game result, given that the result is not valid.

**Screenshots**
https://online-go.com/game/27029807
Screen Shot 2020-10-24 at 11 10 22 PM
Screen Shot 2020-10-24 at 11 05 46 PM

benjaminpjones:
Fixes #1244 

## Proposed Changes

  - In the Game view, strikethrough the game result and add "GAME ANNULLED"
  - Strikethrough the game result in the Game History pane if the game has been annulled.

Screen Shot 2020-10-24 at 11 42 43 PM
Screen Shot 2020-10-24 at 11 42 23 PM

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -2518,7 +2518,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
                 </div>
                 <div className='annulled-indicator'>
                     { state.annulled &&
-                        'GAME ANNULLED'
+                        pgettext("Displayed to the user when the game is annulled", "Game Annulled")
                     }
                 </div>
                 {/* } */}
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -2503,7 +2503,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
                     }
 
                     {(state.mode === "play" && state.phase === "finished" || null) &&
-                        <span>
+                        <span style={{textDecoration: state.annulled ? 'line-through' : 'none' }}>
                             {state.winner
                                 ?
                                 (interpolate(pgettext("Game winner", "{{color}} wins by {{outcome}}"), {
@@ -2514,6 +2514,11 @@ export class Game extends React.PureComponent<GameProperties, any> {
                                 (interpolate(pgettext("Game winner", "Tie by {{outcome}}"), {"outcome": pgettext("Game outcome", this.goban.engine.outcome)}))
                             }
                         </span>
+                    }
+                </div>
+                <div className='annulled-indicator'>
+                    { state.annulled &&
+                        'GAME ANNULLED'
                     }
                 </div>
                 {/* } */}
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -71,7 +71,7 @@ function getGameResultRichText(game) {
     if (game.annulled) {
         result = <span>
                      <span style={{textDecoration: 'line-through'}}>{result}</span>
-                     <span>, {pgettext("Game has been annulled", "annulled")}</span>
+                     <span>, {_("annulled")}</span>
                  </span>;
     }
 
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -29,13 +29,11 @@ import {PlayerIcon} from 'PlayerIcon';
 import {GameList} from "GameList";
 import {Player} from "Player";
 import * as preferences from "preferences";
-import {updateDup, alertModerator, getGameResultText, ignore} from "misc";
+import {updateDup, getGameResultText, ignore} from "misc";
 import {longRankString, rankString, getUserRating, humble_rating, effective_outcome} from "rank_utils";
 import {durationString, daysOnlyDurationString} from "TimeControl";
 import {openModerateUserModal} from "ModerateUser";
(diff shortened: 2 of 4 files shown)


src/views/User/User.tsx
```diff
@@ -29,13 +29,11 @@ import {PlayerIcon} from 'PlayerIcon';
 import {GameList} from "GameList";
 import {Player} from "Player";
 import * as preferences from "preferences";
-import {updateDup, alertModerator, getGameResultText, ignore} from "misc";
```
BHydden, line 32: don't think you meant to remove the alertModerator import here? it wasn't related to displaying the endstate of the game

benjaminpjones: You are correct that it is unrelated, but this it is an `alertModerator` is an unused import (courtesy of my linter).  I'm guessing it fell off whenever that functionality was moved into PlayerDetails.

If you think I should remove that kind of stuff for better chance at being approved, I will, but I thought it was a good idea to tidy up while I was looking at that file.

BHydden: just caught my eye, I can't recommend one way or another... anoek may advise whether cleanup should have its own PR or not, frankly it's not my call I just wanted to see why it was included

benjaminpjones: If you're curious, [commit b88bc3369f17e5036d309ff96cf1e559b83bc2e1](https://github.com/online-go/online-go.com/commit/b88bc3369f17e5036d309ff96cf1e559b83bc2e1#diff-3634b4c362041964a17503f2ba9d107297f02376af3788b4648644d880de5887) is when `alertModerator` was removed from the profile page :)

src/views/Game/Game.tsx
```diff
@@ -2516,6 +2516,11 @@ export class Game extends React.PureComponent<GameProperties, any> {
                         </span>
                     }
                 </div>
+                <div className='annulled-indicator'>
+                    { state.annulled &&
+                        'GAME ANNULLED'
```
anoek, line 2521: Any text we display to the user needs to be translatable, to do that simply wrap it with an `_` or `pgettext` call, such as `_("Game annulled")` or `pgettext("DIsplayed to the user when the game is annulled", "Game annulled")` (the first argument is a comment string for the translators)

I like the strike through on the profile page, on the game history, but I think I don't like it on the game page.

And last but not least, lets `text-align: center` and remove the all caps, make it just Game Annulled` on the game page

src/views/User/User.tsx
```diff
@@ -60,23 +58,22 @@ interface UserProperties {
...
+function getGameResultRichText(game) {
+    let result = getGameResultText(game);
 
-let Rank = (props: {ranking: number, pro?: boolean}) => (<span>{rankString(props)}</span>);
+    if (game.ranked) {
+        result += ", ";
+        result += _("ranked");
+    }
+    if (game.annulled) {
+        result = <span><span style={{textDecoration: 'line-through'}}>{result}</span><span>, annulled</span></span>;
```
anoek, line 72: Translate here too, something like: `pgettext("Game has been annulled", "annulled")`

anoek (changes requested):
Nice improvement!

src/views/Game/Game.tsx
anoek: Scratch the style change comments, my css didn't correctly, I see it was centered (and red), and with that style the strike through looks good too.

So, just the translation thing, ignore the rest! :)

benjaminpjones: Ah, yes I saw the i18n support, but wasn't sure how to use it.  Updates incoming!

benjaminpjones:
I *think* I fixed it, however I am not sure how to test translations:

Screen Shot 2020-10-26 at 11 56 24 AM
Screen Shot 2020-10-26 at 11 56 47 AM


I assume 'annulled' is already in the dictionary... does the context parameter for `pgettext` need to be more generic?

anoek:
Yep "annulled" is already in there, as is "Game has been annulled", "Annulled", and "game annulled". So "Game Annulled" isn't in there yet, but close variations are. I would say if "Game has been annulled" looks good there (like it's not too big or doesn't look funny) lets go with that, otherwise we'll just add "Game Annulled" to the list and give the translators some more work to do :)

anoek:
I guess for "annulled" just make that `_("annulled")` and you should see it translated, that's the variant that's already translated.

benjaminpjones:
Okay, I went ahead and submitted the change to 'annulled':
Screen Shot 2020-10-26 at 1 44 56 PM

I did not submit the change to "Game Annulled" because changing it to "Game has been annulled" did not translate when I tested in Spanish, and therefore (hopefully) not too much work to translate both in one go.

anoek:
Awesome thanks!

GreenAsJade:
Nice one!

Issue #1187, opened 2020-07-01
English profanity filter includes some strange entries


PiDelport:
While implementing #1186, I noticed that the English profanity filter includes a couple of strange entries:

* `ethical\W+slut` – A reference to the [_The Ethical Slut_](https://en.wikipedia.org/wiki/The_Ethical_Slut)? The book is a non-pornographic book about polyamory and relationship ethics, so I'm puzzled why this specific entry exists: `slut` is separately included, so there should be no reason to additionally censor "ethical" here. Remove?
* `queaf` – A misspelling of "queef"? There's no corresponding entry for "queef", so I'm guessing this should either be corrected to "queef", or be kept in addition to "queef"?
* `snowballing` – This is uncommon slang for a sexual act, but it's also a benign and common word that can legitimately be used in Go discussion, such as talking about a move, joseki, or position snowballing in some way. I think it could be removed from the list?
* `women\W+rapping` – I have no idea why this is included: I can't find any profane definition of it, and I don't think discussion of actual women rapping is profane. Remove?

benjaminpjones:
You don't find women rapping offensive??

anoek:
I have no objections to any of these changes FWIW. I imagine rapping was added because if you removing one p you have raping, but yeah I don't think we have a problem in our community with that, and I could also see folks talking about women rappers while having no intention of bring up rape.

PiDelport:
(@benjaminpjones: Is that humour, or a sincere question?)

PiDelport:
@anoek: Ah, I hadn't considered that.

It still seems hard to explain, though, since there are no rape-related entries otherwise, except for the phrase "date rape" specifically?

Anyhow, thanks for the feedback! I'll make a PR with these changes.

benjaminpjones:
@PiDelport Sorry I really need to work on my internet writing style. I just thought that entry was particularly funny 😆

PiDelport:
(@benjaminpjones: No worries! 😄️ I read it as humour, first, but then it occured to me later that some people might actually sincerely have a problem with it, somehow.)

PiDelport:
I spotted one more entry that probably needs tweaking:

* `girl\W+on(\W+top)?` → `girl\W+on\W+top`

This avoids generically matching "girl on" which has many benign uses that would probably puzzle people to be filtered.

BHydden:
[Poe's Law](https://en.wikipedia.org/wiki/Poe%27s_law)

[...]

PR #1252, opened 2020-10-29
Update English profanity_regex

PiDelport:
This implements the changes discussed in issue #1187.


PiDelport:
@anoek: Ah, good catch, sorry!

Should I squash the two commits, for tidiness?

anoek:
Nah we're good, thanks for the patch!

PR #1253, opened 2020-10-31
fix syntax error in profanity filter

RubyMineshaft:
There was an unmatched ")" in English profanity filter causing the entire regex to be spit into the console. 

Screen Shot 2020-10-31 at 4 05 17 PM


BHydden:
Great catch RM ❤

anoek:
Thanks for the fix RM, I obviously clicked merge without actually testing that one ;)

PR #1255, opened 2020-10-31
Take into account that white goes first in japanese handicap games

GreenAsJade:
Fixes the problem that the moderator game timings table had black and white reversed in Japanese handicap games


anoek:
Awesome thanks :)

Issue #1257, opened 2020-11-02
Plurals not handled correctly for some languages


benjaminpjones:
**Describe the bug**
Not all languages follow the **1 => singular, !1 => plural** paradigm.  One example is Russian where you use different plural forms depending on if the number ends in a 2/3/4 or not. However, looking at translate.ts, we do not take this into account, even though there is evidence that there were plans to do this (`pluralidx()`, storing the plural forms in an array).

**To Reproduce**
1. Switch to Russian.
2. Find a game with more than 4 captures:

<img width="353" alt="Screen Shot 2020-11-01 at 11 45 01 PM" src="https://user-images.githubusercontent.com/25233703/97830826-4b602480-1c9c-11eb-84c1-3948cff2f8f4.png">

3. Note that it says "камня" instead of "камней"

I don't actually speak Russian, but I think it should say the latter.  In addition, note that the translators are putting in 2 plural forms, so nothing needs to change on the translation side:

<img width="1035" alt="Screen Shot 2020-11-01 at 11 40 50 PM" src="https://user-images.githubusercontent.com/25233703/97830919-8f532980-1c9c-11eb-8d7a-719c77d149bf.png">

BHydden:
Welcome to the rabbit hole ❤

https://youtu.be/0j74jcxSunY

benjaminpjones:
Okay, I took a look at the [CLDR's Language Plural Rules](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html) and it looks like we support 7 languages that have more than 1 plural form.  Here is a brief summary:

el : Greek : [one, other]
eo : Esperanto : [one, other]
en : English : [one, other]
vi : Vietnamese : [all]
ca : Catalan : [one, other]
it : Italian : [one, other]
cs : Czech : [one, few, many, other]
eu : Basque : [one, other]
et : Estonian : [one, other]
es : Spanish : [one, other]
ru : Russian : [one, few, many, other]
nl : Dutch : [one, other]
pt : Portuguese : [one, other]
zh-tw : Chinese (Traditional) : [all]
tr : Turkish : [one, other]
zh-cn : Chinese (Simplified) : [all]
ro : Romanian : [one, few, other]
pl : Polish : [one, few, many, other]
fr : French : [one or zero, other]
hr : Croatian : [one, few, other]
de : German : [one, other]
da : Danish : [one*, other]
fi : Finnish : [one, other]
hu : Hungarian : [one, other]
[...]

Issue #1248, opened 2020-10-26
Document the i18n functions


benjaminpjones:
Internationalization and localization is important, but there are a lot of moving parts, even for the smallest of fixes.  I think a good start for documentation would be:

- an i18n page on the wiki<sup>1</sup>
- jsdoc comments for _() and pgettext().

<sup>1</sup>if there is a better place for contributor-focused documentation, let me know!

benjaminpjones:
Hmm not sure how to assign myself. I fully intend to do this!

BHydden:
FWIW, the rest of us coders tend to just add the _() or pgettext() and leave the translations to the translators :P you're the first I've seen to actually test it



> not sure how to assign myself. I fully intend to take charge of this!

we don't normally get assignments, but in any case only anoek can do them... mostly he doesn't bother since we have such a small team

benjaminpjones:
>the rest of us coders tend to just add the _() or pgettext() and leave the translations to the translators

I will probably just do that in the future! But for the first time, I wanted to get it right. That's part of the reason documentation on pgettext would be useful 😀

>we don't normally get assignments, but in any case only anoek can do them... mostly he doesn't bother since we have such a small team

Got it. I guess I didn't want to make it sound like I expected someone else to do this (unless they really want to!)

BHydden:
apart from anoek we're all a volunteer team, so nothing to fear there haha, none of us do anything unless we really want to :P it's a passion for all of us who simply want OGS to be the best it can be

benjaminpjones:
https://github.com/online-go/online-go.com/wiki/Internationalization-and-Localization

PR #1258, opened 2020-11-09
More translate updates

benjaminpjones:
## Proposed Changes

  - Add comments to translate.ts functions (Fixes issue #1248)
  - Support plurals for more languages (Fixes issue #1257)
  - Update *yarn.lock*. Not sure what info this file contains that *package.json* doesn't, but I think it needed to be updated after I added the jest dependency.
  - Add `npx jest` to *shippable.yaml*. **I don't know how to confirm that I did this right - feedback welcome!**

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -12,6 +12,7 @@ build:
     - node_modules/.bin/gulp min_styl
     - node_modules/.bin/tslint --project tsconfig.json 
     - node_modules/.bin/webpack --optimize-minimize --devtool=source-map
+    - npx jest
 
   post_ci:
     - gzip -9 dist/ogs.min.js | wc -c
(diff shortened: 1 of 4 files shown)


anoek:
Confirmed, `npx jest` is working: https://app.shippable.com/github/online-go/online-go.com/runs/2183/1/console (seen by clicking that green check mark), if you expand the `build_ci` section you can see the execution there :+1: 

And yeah if we update `package.json` we need to update `yarn.lock`, `package.json` specifies roughly what versions of things we need, `yarn.lock` stores **exactly** what versions of our packages and all dependent packages are currently being used so that builds across systems are consistent and we minimize situations where "it works for me!" but some version of something is slightly different on another system that causes it not to work the same.

anoek:
Great work with this, thanks for tackling it!

benjaminpjones:
Awesome, thanks for clarifying!

PR #1260, opened 2020-11-10
Put annul-opacity (in game history table) into the theme

GreenAsJade:
In the thread on this topic, it was complained that the opacity of annulled games in the game history table was not enough - I believe this is referring to light mode, and I agree with that.

## Proposed Changes

Theme the opacity, so it can be greater in light mode.


anoek:
Yep, nice thanks!

PR #1262, opened 2020-11-10
Update Puzzle.tsx

### LINKED ISSUE #1261: [feature]Puzzle: different board sizes
It would be nice to have different board sizes for different level of puzzles. E.g. for puzzles with counting territory (e.g. https://online-go.com/puzzle/22759) 

Fix is to have different options in the dropdown (in Puzzle.tsx:807)

mih4ajlo:
different board sizes for puzzles

Fixes #1261

## Proposed Changes
Board sizes: 7, 11, 15 and 17.

### DIFF
--- a/src/views/Puzzle/Puzzle.tsx
+++ b/src/views/Puzzle/Puzzle.tsx
@@ -805,8 +805,12 @@ export class Puzzle extends React.Component<PuzzleProperties, any> {
 
                                 <select value={this.state.puzzle.width} onChange={this.setPuzzleSize}>
                                     <option value={19}>{_("19x19")}</option>
+                                    <option value={17}>{_("17x17")}</option>
+                                    <option value={15}>{_("15x15")}</option>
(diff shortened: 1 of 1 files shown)


BHydden:
I kinda feel like it would be best to leave puzzles as just 9x9 and 19x19 (and maybe 13x13)

Adding in non-standard board size puzzles could be confusing for beginners

benjaminpjones:
I see no reason why there shouldn't be alternate sizes. Beginners shouldn't have to mess with sizes, since that option is set by the puzzle author.

That said, if this was added, I think id prefer to see an interface more similar to the "Create Game" dialog:

19x19
13x13
9x9
Custom

Where custom opens fine tuned settings

anoek:
Thanks @mih4ajlo ! I'll merge this, I guess I don't see a problem with it either. I'd be good with a Custom Size system too fwiw @benjaminpjones

PR #1263, opened 2020-11-11
Added display of the time of last thing that happens 

GreenAsJade:
As per Vsotvep request here: https://forums.online-go.com/t/move-timing-go-public/31770/9?u=eugene

Calculate and show how much time was spent in the game after the last move was played (this is either score phase or timeout time)

Screen Shot 2020-11-11 at 10 26 19 pm


GreenAsJade:
Note: this seems to work well enough for moderators to check it out.  Not extensively tested though.

anoek:
Awesome thanks :)

PR #1266, opened 2020-11-12
Show the user's rank for each game in the history table 

GreenAsJade:
The recent changes had two problems raised:

1) We can't see the rank of the player at each game anymore
2) It's not clear who's colour the colour indicator is

## Proposed Changes

https://forums.online-go.com/t/update-of-game-history-table/31742/91?u=eugene


anoek:
Awesome thanks!

PR #1271, opened 2020-11-18
Language Settings Updates

### LINKED ISSUE #656: Language picker in settings has a bad UX design
1) The language setting is the only "one of many" setting fields that does not use a dropdown menu. Ex: "Game thumbnail list threshold", "Board labeling", etc... all use a dropdown. This creates an inconsistency in user experience.

2) There is no indication that the currently chosen language is clickable. It just looks like a display of the currently chosen language.

3) Because the Profanity Filter setting that follows shows many languages, it gives the false impression that there aren't many languages for the language setting. 

4) Because you click on languages to choose them in the Profanity Filter, it's not instinctive to think that you have to click on the language in the language setting to change it (opposite behavior). One is click to pick, the other is click to change.

The best fix would be to show a drop down menu like all other "one of many" settings.
- - -

I just realized today after using the site for many years that the interface could be changed to other languages. How did I realize? I told a friend on the site that I thought the site would get more asian players if it was translated into chinese, japanese, etc... I honestly thought English was the only language.

benjaminpjones:
Fixes #656 

## Proposed Changes

  - In Settings, change the Language Picker to a dropdown
  - Change the profanity switches to a multi-dropdown

Screen Shot 2020-11-18 at 12 41 36 AM

### DIFF
--- a/src/views/Settings/Settings.styl
+++ b/src/views/Settings/Settings.styl
@@ -50,44 +50,22 @@
         }
     }
 
-    .PreferenceDropdown {
+    .PreferenceDropdown 
+    {
         min-width: 10rem;
         .PreferenceDropdown-value-container {
             padding-left: 0.5rem;
             padding-right: 0.5rem;
         }
     }
-    .ProfanityFilterCard {
-        display: inline-flex;
-        width: 20rem;
-        max-width: calc(100vw - 4rem);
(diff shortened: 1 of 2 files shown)


anoek:
Yep, that's better. Thanks!

PR #1275, opened 2020-11-19
Control the spacing between the color marker and the rank explicitly,…

GreenAsJade:
… and tweak it better

Fixes 

The gap ended up too small between the colour marker and the rank.   Not sure why I didn't pick that up in Beta ... we sure could do with some more realistically populated accounts in there to test this stuff better.

Anyhow, sorry to have to tweak it again

## Proposed Changes

Put the colour marker into a span, so we can control the spacing, and set the spacing better.

### DIFF
--- a/src/views/User/User.styl
+++ b/src/views/User/User.styl
@@ -460,6 +460,10 @@
             overflow: hidden;
         }
 
+        .span {
+            padding-left: 0.5em;
+        }
+
         .user_info  {
             white-space: nowrap;
         }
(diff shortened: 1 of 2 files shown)


GreenAsJade:
Hmm - FTR, I just discovered why I didn't spot this unfortunate spacing before.  It only happens on my large monitor!

On my Macbook monitor, the spacing for some reason is fine!

I totally don't get it yet, and will keep poking around, but we might as well have this fix anyhow, because that spacing should be controlled in this way.

PR #1282, opened 2020-11-22
Fixed oversized element causing mobile horizontal scroll

JDNdeveloper:
When viewing my own games I noticed that there was horizontal scroll on mobile. See this screen capture: https://streamable.com/pbzr4e

It looks like the implicit left/right margin in `analyze-mode-buttons` was causing the issue:
image

This fixes the problem by explicitly only setting top/bottom margin.


anoek:
Awesome thanks!

PR #1284, opened 2020-11-23
Fixed wrong display of final board after using score estimation or AI review

### LINKED ISSUE #881: Wrong display of final board stated after using some features of the game view
### Summary

In a finished game, when we use the “estimate score” or “disable AI review”, we temporarily lose the information of which stones were marked dead in the score removal phase. Reloading the page restores the correct state.

### Steps To Reproduce:

1. Go to a game ended with score (https://online-go.com/game/19350057)
2. Click estimate score
3. Click back to game
3. See that all stones are now alive and territory marks are correct for the new state, but not the original state.

### Additional Information:

This affects the individual scores of black and white as well, but the final score difference is correct. 

Reported at https://forums.online-go.com/t/bug-report-completely-wrong-score-display/22760?u=flovo

---
<details><summary>Screenshots</summary>
Before

After

</details>

JDNdeveloper:
Fixes #881 

Made the following changes to fix wrong display of final board after using some features:
- Sync state after disabling AI review.
- Remove `this.goban.engine.clearRemoved();` when exiting scoring estimate mode, which seemed to corrupt the final board state requiring a refresh to fix.

Also changed the "Back to Game" button for score estimation to "Back to Board" to reflect that it has different behavior than the "analysis toolbar" "Back to Game" button after https://github.com/online-go/goban/pull/14 merges (new behavior is to return to previous board position rather than the last official move). To get back to final board state after score estimation will require two presses: (1) "Back to Board", (2) "Back to Game", or alternatively just hit `esc` twice. I'm happy to name it something different if "Back to Board" doesn't feel appropriate, or drop this altogether (although I think having two buttons named the same that do different things is rather confusing). NOTE: If you're already in "play" mode (e.g. viewing the final board) pressing "Back to Board" or hitting `esc` once will take you back to the current game state.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -1438,6 +1438,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
                 move_tree.clearMarks();
             }
             this.goban.redraw();
+            this.sync_state();
         }
         this.setState({ ai_review_enabled: !this.state.ai_review_enabled });
     }
@@ -1886,7 +1887,6 @@ export class Game extends React.PureComponent<GameProperties, any> {
             estimating_score: false
         });
         this.goban.setScoringMode(false);
-        this.goban.engine.clearRemoved();
         this.goban.hideScores();
         this.goban.score_estimate = null;
         this.sync_state();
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for all your work in these areas, these are not easy bugs to tackle!

PR #1287, opened 2020-11-24
Fixed player details popup not loading on first press

### LINKED ISSUE #1286: User profile popup not loading on first press
When browsing my game history and clicking on a player I noticed the first time I press it's not loading properly:
image

Then if I click outside of the popup and press again it loads:
image

In the console I'm seeing these errors when this occurs:
```
Attempted to fetch invalid player name:  ...
invalid player name
```

JDNdeveloper:
Fixes #1286 

Two main changes were required to get this fixed:

### `PlayerDetails` fix to get grayed out buttons enabled

I fixed the `resolved` state property logic in `PlayerDetails` to get the buttons to not be disabled. The previous semantics of `resolved` were inverted, where the buttons were disabled if the state was resolved as opposed to the other way around. 

There was a crazy bug here that was causing it to "work" before despite the semantics being wrong: for some reason the `player` object (of type `PlayerCacheEntry`) was getting the `PlayerDetails` state properties added to the object! So on first `resolve()` for a player we set `resolved: true` which disabled the buttons, then on second `resolve()` the player had `resolved: false` property set (false due to initial blank state), so `Object.assign({resolved: true}, player as any)` had `resolved: false` set! 

I still haven't figured out how this is possible, but it seems related to react's `setState()` internals (likely a bug). To fix this I changed it to `Object.assign({}, player as any, {resolved: true})` to make sure the final state update had `resolved: true`.

### `Player` race condition causing player name to be `...`

This was the sequence of the race:
1. `PlayerDetails` component is created and set to initial blank state.
1. In `PlayerDetails` constructor, player is looked up in the cache but isn't present.
1. `Player` component inherits the default props from its parent `PlayerDetails`, including `{username: '...'}`.
1. `Player.componentDidMount()` attempts to fetch the player, and since `player_id` is undefined (due to inheriting props) it attempts to fetch by username `...`.
1. `PlayerDetails.resolve()` is called from `PlayerDetails.UNSAFE_componentWillMount()` which fetches the player (using `playerId`) which leads to updating the user props of `Player`.
1. `Player.UNSAFE_componentWillReceiveProps()` is called which updates its user state from the new props (all good at this point!).
1. The username player fetch from `Player.componentDidMount()` returns, and overwrites the state to the username captured by the promise callback closure which was the original `...`.

This race leads to the username in state being set to `...` despite the props having the proper username.

To fix this I just prevented fetching by username if username is `...`. @BHydden mentioned that at one point there was a user with the username `...` but that was taken care of so I think this should be safe (would be good to get confirmation on that).

### Other misc fixes

It looks like there was a bug in `UNSAFE_componentWillReceiveProps` where it was using the username from the old props rather than the new ones, so I fixed that even though it likely won't affect anything (didn't notice any changes with or without the change).

### DIFF
--- a/src/components/Player/Player.tsx
+++ b/src/components/Player/Player.tsx
@@ -165,7 +165,7 @@ export class Player extends React.PureComponent<PlayerProperties, any> {
             else if (player_id && player_id <= 0) {
                 // do nothing
             }
-            else if (username && username !== '...') {
+            else if (username && username !== "...") {
                 player_cache.fetch_by_username(username, ["username", "ui_class", "ranking", "pro"]).then((user) => {
                     if (this.unmounted) {
                         return;
--- a/src/components/Player/Player.tsx
+++ b/src/components/Player/Player.tsx
@@ -92,7 +92,7 @@ export class Player extends React.PureComponent<PlayerProperties, any> {
             else if (player_id && player_id <= 0) {
                 // do nothing
             }
-            else if (username) {
+            else if (username && username !== "...") {
                 player_cache.fetch_by_username(username, ["username", "ui_class", "ranking", "pro"]).then((user) => {
                     if (this.unmounted) {
                         return;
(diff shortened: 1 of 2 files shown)


BHydden:
> To fix this I just prevented fetching by username if username is `...`. @BHydden mentioned that at one point there was a user with the username `...` but that was taken care of so I think this should be safe (would be good to get confirmation on that).

@anoek  as part of this PR, if you could simply add that username to the blocked list, that would probably be simplest

anoek:
Thanks once again @JDNdeveloper !

PR #1288, opened 2020-11-24
Fixed stale rank on homescreen

### LINKED ISSUE #1277: Rank displayed on homescreen can be out of date
**Describe the bug**
Sometimes the rank displayed on the OGS homescreen for the logged in user is stale.

Possibly related issues:
- https://github.com/online-go/online-go.com/issues/280
- https://github.com/online-go/online-go.com/issues/919

I think this may be some sort of client-side caching issue.

**Screenshots**

Old rank of 10.0k is displayed on homescreen:
image

Correct rank of 9.6k is shown on profile page and "user card":
image

Sometimes after refresh it will display the new value, but not always!:
image

JDNdeveloper:
Fixes #1277 

I noticed that the `ProfileCard` user prop had stale data in it and didn't match the child `Player` component's user ratings. Looks like this is just because we assign the user once during initial render and `data.get("config.user")` when the page first loads could be out of date. This fix watches for `config.user` changes and forces re-render if it occurs, preventing stale homescreen data.

### DIFF
--- a/src/views/Overview/Overview.tsx
+++ b/src/views/Overview/Overview.tsx
@@ -70,6 +70,7 @@ export class Overview extends React.Component<{}, any> {
         }
 
         this.state = {
+            user: data.get("config.user"),
             overview: {
                 active_games: [],
             },
@@ -91,6 +92,7 @@ export class Overview extends React.Component<{}, any> {
     componentDidMount() {
         this.setTitle();
         notification_manager.event_emitter.on("turn-count", this.setBoardsToMoveOn);
+        data.watch("config.user", (user) => this.setState({"user": user}));
         this.refresh().then(ignore).catch(ignore);
     }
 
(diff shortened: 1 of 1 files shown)


BHydden:
At this rate we're going to run out of issues! haha, thanks for all your hard work @JDNdeveloper  you're powering out some impressive improvements that have gone long overlooked ❤️

src/views/Overview/Overview.tsx
```diff
@@ -91,6 +92,7 @@ export class Overview extends React.Component<{}, any> {
     componentDidMount() {
         this.setTitle();
         notification_manager.event_emitter.on("turn-count", this.setBoardsToMoveOn);
+        data.watch("config.user", (user) => this.setState({"user": user}));
```
anoek, line 95: When we watch something in a component we need to `unwatch` it in `componentWillUnmount` too

anoek:
Yeah seriously, thanks for all these fixes :)

src/views/Overview/Overview.tsx
JDNdeveloper: Done!

PR #1296, opened 2020-11-28
Hide user's own rank when "hide ranks" is active (and alignment tweak)

GreenAsJade:
The user's rank should not be shown if they have selected "Hide ranks" option.

## Proposed Changes

Don't show it if they have selected hide ranks option 👍 

I also took the opportunity to vertically align the user's rank with the rest of the row.


anoek:
Cool thanks :)

PR #1297, opened 2020-11-29
Fixed mobile rotation view_mode bug

### LINKED ISSUE #1292: On mobile during a game, the game board size is not optimized for screen when you switch from horizontal view to vertical view.
**Describe the bug**
When you play on the phone, if you starting from a vertical view, the game board is properly optimized for the screen, but if you switch to horizontal by accident and change back /or starting from horizontal view and change to vertical view, the game board doesn't go back to the original optimized size for the screen which will become too small, need to refresh or tab to regain normal size.

**To Reproduce**
Steps to reproduce the behaviour:
1. View any game,
2. Starting from a horizontal view or staring from a vertical view and change to horizontal view
3. Switch to Vertical view
4. The game board size is too small for the screen

**Expected behavior**
It should go back to the original optimized screen size. 

 - Device: [e.g. iPhone11 plus]
 - OS: [e.g. iOS14]
 - Browser [e.g, safari]

**Screenshot** 
img_4686

**Additional context**
This is my first issue. Apologies if there are any mistakes. ps. recently discover this go platform. Absolutely love it.

JDNdeveloper:
Fixes #1292

It looks like the `onResize=...` callback passes in some arguments, so `skip_state_update` arg in `this.onResize()` had integer values (maybe the window height or width?) which are "true" values (as opposed to the default false for the argument), causing the view mode to not be re-evaluated when rotating screen. Wrapping in a lambda to discard the args and use default ones fixed the issue. Made the change in all `onResize=...` places to avoid the bug in other places.

### DIFF
--- a/src/components/LadderComponent/LadderComponent.tsx
+++ b/src/components/LadderComponent/LadderComponent.tsx
@@ -183,7 +183,7 @@ export class LadderComponent extends React.PureComponent<LadderComponentProperti
 
         return (
             <div className="LadderComponent">
-                <ReactResizeDetector handleWidth handleHeight onResize={this.onResize} />
+                <ReactResizeDetector handleWidth handleHeight onResize={() => this.onResize()} />
 
                 <UIPush event="players-updated" channel={`ladder-${this.props.ladderId}`} action={this.updatePlayers} />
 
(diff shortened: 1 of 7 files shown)


anoek:
Hah, subtle, nice find!

PR #1298, opened 2020-11-29
Added mobile portrait puzzle layout

### LINKED ISSUE #1034: Puzzles mobile template
Is it bug with mobile view of puzzles page or there is just no responsive template on this page? It is kind of hard to figure out what is going on on the board.

**Screenshots**
image

**Smartphone**
 - OS: Android 9
 - Browser FireFox 68.3.0

JDNdeveloper:
Fixes #1034

Modeled after the Joseki portrait layout.

image

### DIFF
--- a/src/views/Puzzle/Puzzle.styl
+++ b/src/views/Puzzle/Puzzle.styl
@@ -31,6 +31,10 @@ puzzle-controls-width=400px
     overflow: hidden;
 
 
+    @media screen and (max-width: 800px) {
+        flex-direction: column;
+        overflow-y: auto;
+    }
 
     .left-col {
     }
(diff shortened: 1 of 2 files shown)


anoek:
Hey very cool! 

I'm think though we need to let that bottom portion scroll so it doesn't shrink the board. Right now the board jerks to different sizes as the buttons pop into existence. 

Ideally I think those buttons and the "Correct" message would be moved to be right under the board too, and the puzzle select and puzzle information be underneath the buttons and maybe even below the puzzle description when it exists. It might be nice to reserve the space for the buttons and success message too so it doesn't jerk the bottom content around, though I guess we'd have to see how that looks.

JDNdeveloper:
@anoek I can't seem to repro the board shrinking/resizing. Currently the bottom portion (`.right-col`) is set to `overflow: visible` so that the whole screen can scroll, rather than just the bottom part (matching UI of Game and Joseki views). If you know how to repro this in the dev setup, or what change should be made to fix it, let me know and I'll be happy to test it out or make the change.

I'll look into reordering the bottom content.

anoek:
It looks like it works as expected on chrome and chrome mobile, but not in firefox or firefox mobile.

JDNdeveloper:
@anoek It looks like Joseki is also broken in Firefox, I'll try to fix that too.

image

benjaminpjones:
If you fix that in this PR, go ahead and add #1300 in the description!

JDNdeveloper:
@anoek I've moved around the element order in the portrait layout and reserved space for the undo/reset buttons. Reserving space for the correct/incorrect elements was more awkward (they're pretty large) so I decided against it, and it should pose less of a usability issue than the undo/reset buttons.

To ensure rotation worked I had to change how view_mode is set in Puzzle (similar to what we do in Game), and I also changed the CSS to reference portrait rather than screen width so that the CSS will match the React HTML view mode.

@anoek @benjaminpjones I've decided against fixing the Firefox issue in this PR since the new portrait UI should still be better than the current one, and it's an existing bug for Joseki as well. In the next couple of days I'll fix the Firefox issues for Puzzle and Joseki and post a new PR. In the meantime it would be great if this could be merged, unless there's other feedback.

anoek:
Sounds reasonable, thanks for all of your work @JDNdeveloper!

PR #1301, opened 2020-11-30
Added a space between play icon and play in the Navbar

zeemai:
In the navbar, there's no space between the play icon and the play word, which looks odd compared to the rest of the items in the menu. I think this might be a typo?

Just added a fa in the className in Navbar.tsx so it added the extra space and look consistent.


anoek:
Thanks for the patch @zeemai !

PR #1303, opened 2020-12-01
Escaper warning initial implementation

GreenAsJade:
As discussed, we need automated help warning escapers.

This first client-side implementation warns people who disconnect from games.

## Proposed Changes

When someone goes to play a game, check if they have a recent disconnect, and warn them with a modal.

When they acknowledge, note the time that they acknowledged, and don't hassle them again unless they disconnect again after this.

I really would have liked to add a mod note that they received the warning, but unfortunately that can't be done client side from the user's account.

Screen Shot 2020-12-01 at 4 44 01 pm

### DIFF
--- a/src/components/EscaperWarningModal/EscaperWarningModal.styl
+++ b/src/components/EscaperWarningModal/EscaperWarningModal.styl
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2012-2017  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.EscaperWarningModal {
+    text-align: center;
+
+    padding: 1em 0 1em 0 !important;
+    border-color: red !important;
+
+    div {
+        padding: 0.1em 2em 0.1em 2em;
+    }
+
+    button {
+        width: fit-content;
+        align-self: center;
+        margin-top: 1em;
+        margin-bottom: 1em
+    }
+}
--- a/src/components/EscaperWarningModal/EscaperWarningModal.tsx
+++ b/src/components/EscaperWarningModal/EscaperWarningModal.tsx
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2012-2017  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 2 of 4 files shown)


BHydden:
I would include at the end "If your connection is unstable, please consider playing correspondence instead of live"

Since this has no effect other than to inform those accidentally DCing (since we can't include mod notes), then it might as well offer a helpful workaround for legitimate disconnects.

GreenAsJade:
I think there's more text there than I'd like already.

If someone gets this warning and feels it is unfair because they have an unreliable connection, they can raise that with us or in the forum and get help.

BHydden:
Aren't we trying to reduce work for ourselves, not add in more features that people are going to feel like they need to defend themselves against?

GreenAsJade:
Hmm - I feel clear in my mind that the key thing I'm trying to do is get in automatic  communication to people that abandonning is not OK, in direct response to them abandonning, instead of having to do that myself every time it is reported.

I'm willing to take the followup load, such as it may be, in exchange for keeping it really clear and on point what the message is.

BHydden:
I'm of the opinion that legitimate disconnections exist, everyone loses internet from time to time, and to be greeted with this message each time may make innocent users feel vilified. Just my 2c

GreenAsJade:
They face this possibility either way - either this message, or an approach from us after a report.

All they have to do is click OK, and hopefully think to themselves "darn, I'll try not to let that happen next time".

anoek:
As I was thinking through this, I think I've come to the conclusion that we can't do this sort of thing on the user interface alone, we need a good amount of API support. So I'm closing this for now, we can re-open as we figure out what we need from the backend and implement it there, then implement the front end for it.


Here's what went through my head: First, disconnects happen a lot, there are about 250 of them a day. I don't have a good feeling about how many are malicious vs. the user just didn't know how or that they should resign vs. connection / hardware / software problems,  however even if the majority of those daily cases are malicious, pretty much every nice player is going to have a problem that is out of their control at some point or another and run into this. So, by volume of players that this system will affect, the good nice players are going to vastly outweigh the troublemakers. As we care very much about the experience of those good nice players, whatever we do here needs to be optimized thinking of innocent players first and foremost. A popup mildly shaking our finger at them telling them not to abandon games is a bit user hostile for those cases, so I would want to rethink when that automated dialog with them is triggered and how it is worded. My thinking is maybe we need to first have a free pass system, as in we don't bother people if they've disconnected just once recently, especially if they play a lot of normally terminating games. The second, is when we word anything, it needs to presume innocence. Maybe we include some instructions or a link to instructions that tell them how to resign (believe it or not that has come up many times, people just don't know they need to resign or don't see the buttons). I think we don't include accusatory text ever in an automated system.


There are also a couple of non obvious problems we'd run into with how this currently works:

1) The data system relies on `localStorage` system, which a small but still significant portion of the population has disabled. I believe Safari has (or maybe just had?) it disabled under a common privacy setting, and anyone who uses private browsing or various privacy addons / extensions. For those users, while they are used to logging in every time and their settings never sticking, I feel like seeing a popup like this every time they went to play might be more than annoying, especially if they were an innocent case.

2) That `players/${id}/games/` is actually kind of expensive. It's fine on the profile page because it's a rare page, but the play page is called a lot, I have good reason to believe it would crush the servers if we called it every time someone visits that page since we've felt DoS effects when popular profiles are linked. That is largely my problem to fix, and I do have a plan, but it's a notable amount of work as a very large migration needs to happen and a good bit of code needs to be touched to account for the schema changes. However, even if that were fixed though I feel like maybe this should have a proper backend implementation anyways, have the game server set flags and track/log when users have been informed / warned that sort of thing.


So, ultimately I think we need to figure out what we want the backend to do and where it stores what flags and how those are updated, and then we can come back to how and when that is presented to the user.

PR #1304, opened 2020-12-02
Fixed various Puzzle bugs

JDNdeveloper:
**_Note_**: this doesn't fix the Firefox issues, I'll work on that soon and post a new PR for it. Was planning on doing that tonight but started shaving the yak instead...

### Fixed goban container blocking buttons on mobile

Moved center col (i.e. goban) behind right-col elements so the overly long goban container didn't block transformation and undo/reset buttons (using `z-index`). This only was an issue on mobile for certain goban dimensions, so wasn't really a problem until the portrait mode was added in #1298.

### Fixed page jumping down when transform button pressed on mobile

Prevented jumping when transformation button pressed (issue was when we tried to focus on a different element (`$("#selected_puzzle").focus().blur()`) so that button didn't look improperly selected, we ended up causing page to jump). Changed to blur the just-pressed button instead, using react refs to get the button element. In desktop mode the puzzle element was usually always in frame so it wouldn't cause the jump, but in the mobile layout this was more prominent.

### Fixed puzzle solution text boxes not being cleared on transformation

Cleared `Correct`/`Incorrect` box when transformation button pressed (`this.reset()` -> `this.doReset()` which properly resets the state rather than just the Goban).

### Handled all caps color text on color invert transformation

I noticed in a puzzle dialog box that on color invert "BLACK" was being changed to "white" (case did not match). This was caused by code in `transformMoveText()` replacing any color that didn't match camel-case with all lowercase. I enhanced this to handle all caps, so now it properly handles: (1) camel-case (e.g. first letter capitalized) (2) all lower case, and (3) all upper case. Anything that doesn't match those will default to lower case.

Original text:
image

After color transformation WITHOUT this fix:
image

After color transformation WITH this fix:
image

### DIFF
--- a/src/views/Puzzle/Puzzle.styl
+++ b/src/views/Puzzle/Puzzle.styl
@@ -48,6 +48,7 @@ puzzle-controls-width=400px
         flex-direction: column;
         align-items: stretch;
         overflow: hidden;
+        z-index: 1;
     }
     &.portrait .center-col {
         height: 100vw;
@@ -63,6 +64,7 @@ puzzle-controls-width=400px
         overflow-x: hidden;
         min-height: auto;
         padding: 0.5rem;
+        z-index: 2;
     }
     &.portrait .right-col {
         overflow-y: visible;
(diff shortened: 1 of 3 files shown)


anoek:
Excellent fixes, thank you!

PR #1310, opened 2020-12-07
Fix puzzle scroll jump and puzzle collection mobile

### LINKED ISSUE #1306: "My Puzzles" collection list doesn't render well on mobile
See below:

image

JDNdeveloper:
### Fix scroll jump on Puzzle after solving

After solving a puzzle, if the text field is too large the `Next` button is off screen. When we focus on `Next` it would cause the screen to jump. Fixed by ensuring we don't scroll after focusing on `Next` by restoring previous scroll position.

### Made Puzzle Collection page layout better on mobile

Fixes #1306 

Before (desktop, mobile):

image
image

After (desktop, mobile):

image
image

The mobile screen is still horizontally scrollable so we can rearrange puzzle list entries and edit them:

image

### DIFF
--- a/src/views/Puzzle/Puzzle.tsx
+++ b/src/views/Puzzle/Puzzle.tsx
@@ -320,7 +320,9 @@ export class Puzzle extends React.Component<PuzzleProperties, any> {
             show_wrong: false,
         });
         setTimeout(() => {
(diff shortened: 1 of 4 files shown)


anoek:
Awesome, thanks!

PR #1312, opened 2020-12-08
Fixed supporters card text overlapping on support page in safari

zeemai:
Fixed on the Support page, the supporter card text overlapping in safari browser. Change flex: 0 to flex: 0 0 auto to make it the same as Chrome

Screenshot for overlapping text (Safari)
Screenshot 2020-12-08 at 16 48 42

### DIFF
--- a/src/views/Supporter/Supporter.styl
+++ b/src/views/Supporter/Supporter.styl
@@ -293,7 +293,7 @@ supporter-page-width=60rem;
             }
 
             .title {
-                flex: 0;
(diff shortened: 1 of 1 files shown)


benjaminpjones:
> Fixes # On Support page

Did you mean to link an issue?

zeemai:
No. Don't think there's an issue for it. Sorry, my bad. Will correct it now.

anoek:
Thanks for the fix!

PR #1317, opened 2020-12-11
Added variation stone transparency to game settings

JDNdeveloper:
I've found it difficult when looking at an AI review to tell the current board state apart from the projected moves in a variation.

This change adds a setting to control the transparency level of stones in a variation (default is 0.6, my personal preference is 0.2). For the sake of not changing existing behavior I've kept the default the same, but if others agree that a lower value is better I am happy to change the default.

This diff requires this goban PR to merge first (shippable test will fail until goban PR is merged and version bumped): https://github.com/online-go/goban/pull/19

Below are some screenshots with different variation stone transparency settings values:

0.0:
image

0.2 (my personal preference):
image

0.3 (maybe would be a good happy medium for a new default?):
image

0.6 (default):
image

1.0:
image

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -35,6 +35,7 @@ let defaults = {
     "dock-delay": 0, // seconds.
     "double-click-submit-correspondence": false,
     "double-click-submit-live": false,
+    "variation-stone-transparency": 0.6,
     "dynamic-title": true,
     "function-keys-enabled": false,
     "game-list-threshold": 10,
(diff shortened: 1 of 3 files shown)


anoek:
Thanks!

JDNdeveloper:
@anoek I don't see this change on `online-go.com` yet. Can you redeploy? Thanks!

anoek:
It's up on beta, validating some things before we push the recent changes live

BHydden:
It's very common for there to be a few days, or even a week, before GitHub merges get pushed live.

PR #1318, opened 2020-12-13
Use a new tab for advance to next board if this game is live.

GreenAsJade:
Fixes the problem that people click on the TurnIndicator in live games and thereby disconnect from their current game, which we don't want them to do.

## Proposed Changes

Create a new tab for the next game if it is different to the current one.

(Note that once people have a tab for each live game they are playing at the same time, they should not click on the turn indicator, they should just switch tabs.   If they click on the turn indicator they'll get another tab, but there is not much we can do about that other than let them learn not to do that)

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -239,6 +239,7 @@ class NotificationManager {
         this.auth = data.get("config.notification_auth");
         this.connect();
     }
+
     advanceToNextBoard(ev?) {
         let game_id = getCurrentGameId() || 0;
         let board_ids = [];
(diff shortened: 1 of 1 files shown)


BHydden:
Don't forget to test that functionality remains the same for corr games, otherwise looks good 👍

GreenAsJade:
Yes it works nicely for Correspondence games - if you have one correspondence and one live, each waiting for a turn, then if you are on the correspondence one and click it, the live one appears in the same tab.  Then if you click it again it opens the correspondence one in a new tab leaving the live one open.

BHydden:
sounds perfect! thanks for the improvement on this feature :D

anoek:
Nice, thanks for the enhancement!

PR #1328, opened 2020-12-19
Display final action time in appropriate resolution

GreenAsJade:
Fixes the problem that "final action" timing can only be accurate to a second

## Proposed Changes

Don't display sub second amounts.


anoek:
Cool, thanks :)

PR #1334, opened 2020-12-23
Fix translation of a string - don't interpolate inside _()

GreenAsJade:
Fixes missing translations

## Proposed Changes

 * Use " not ` for simple string inside _()
 * Add _() where it is missing


anoek:
Thanks for the fixes! Yeah I wonder if I can figure out how to get a warning/error produced for any ` usage in any of the translations functions.. I'll have to ponder that.

PR #1337, opened 2020-12-27
Don't create a new tab on notification click if the game is finished

GreenAsJade:
Fixes [the problem of creating a new tab on notification click if the live game has finished](https://forums.online-go.com/t/clicking-on-black-dot-with-of-moves-waiting-opens-in-new-tab/32626/4?u=eugene) .

## Proposed Changes

Don't create a new tab on notification click if the live game on the current tab is finished.


anoek:
Awesome thanks :)

PR #1338, opened 2020-12-28
Add option to show '?' on latest move if there is an undo request for it

GreenAsJade:
Fixes [the problem some people have knowing that they got an undo request](https://forums.online-go.com/t/wish-better-noticeable-undo-requests/964/70?u=eugene)

## Proposed Changes

Add a player preference for showing a visual indicator of undo pending.

Show a '?' instead of the circle for the most recent move if there is an undo request for it and the preference is turned on.

This implementation requires the Goban support provided in https://github.com/online-go/goban/pull/23 .

This implementation does _not_ attempt to colour the '?' red, as requested by some, because that is a can of worms to get right visually (think themes, custom boards, highlights on stones etc).

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -36,6 +36,7 @@ let defaults = {
     "double-click-submit-correspondence": false,
     "double-click-submit-live": false,
     "variation-stone-transparency": 0.6,
+    "visual-undo-request-indicator": false,
     "dynamic-title": true,
     "function-keys-enabled": false,
     "game-list-threshold": 10,
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -581,6 +581,7 @@ export class Game extends React.PureComponent<GameProperties, any> {
                     this.goban.double_click_submit = preferences.get("double-click-submit-correspondence");
                 }
                 this.goban.variation_stone_transparency = preferences.get("variation-stone-transparency");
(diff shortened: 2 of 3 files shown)


GreenAsJade:
Screen Shot 2020-12-28 at 1 41 05 pm (2)

BHydden:
Default on or default off? I vote for on.

Great addition!

GreenAsJade:
LOL default is off, for the reasons argued in that thread.

If someone has trouble noticing that they have undo requests, then they can go looking for this.

I personally dislike it, and I suspect most people who aren't in favour of undo requests would also.

Especially, I don't want to make anything default that implies that you have to accept an undo.

Screen Shot 2020-12-28 at 1 49 33 pm (2)

GreenAsJade:
(shippable is failing because it needs the updated Goban)

anoek:
Very cool, thanks for that undertaking @GreenAsJade !

Spoekenkieker:
“If someone has trouble noticing that they have undo requests, then they can go looking for this.”

Problem is, IF someone has trouble noticing this, the won’t notice they have this trouble ;-)

BHydden:
> “If someone has trouble noticing that they have undo requests, then they can go looking for this.”
> 
> Problem is, IF someone has trouble noticing this, the won’t notice they have this trouble ;-)

There's already a new PR reversing this to be default on.

PR #1342, opened 2021-01-01
Updates to TimeControl.durationString()

benjaminpjones:
After translation updates, this will fix one of the German language localization issues presented in the OGS Forum thread [message not localized (unremovable branch, time control)](https://forums.online-go.com/t/message-not-localized-unremovable-branch-time-control/32557/15)

## Proposed Changes

  - updates to internals of `durationString`
  - `durationString` now returns lowercase (if not translated)
  - Remove some unused imports, functions

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -15,7 +15,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import * as data from "data";
 import { computeAverageMoveTime } from 'goban';
 import {_, pgettext, ngettext, interpolate} from "translate";
 import {TimeControl, TimeControlTypes} from "./TimeControl";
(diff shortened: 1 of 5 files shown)


benjaminpjones:
Initially, I thought this PR might create undue load on translators.  However, with the last two commits, their work should be simplified.

src/components/TimeControl/util.ts
```diff
@@ -501,70 +452,46 @@ export function isLiveGame(time_control) {
...
 
-    function plurality(num, single, plural) {
-        num = Math.round(num);
-        if (num > 0) {
-            if (num === 1) {
-                return num + " " + single;
-            }
-            return num + " " + plural;
+    let coarse_fine_time_template = pgettext("Duration strings using two units (e.g. \"1 week 3 days\", \"2 hours 45 minutes\"). This is localizable because some languages may need to change the order of the coarse and fine times.",
+                                             "{{coarse} {{fine}}");
```
anoek, line 462: coarse is missing a `}`

src/lib/translate.ts
```diff
@@ -155,10 +155,10 @@ export function ngettext(singular: string, plural: string, count: number) {
 
     /* If we don't have a ngettext translation entry at all, but
      * we do have some stand alone translations, use those */
-    if ((count !== 1 || !(singular in catalog)) && plural in catalog) {
+    if (count !== 1 && plural in catalog) {
```
anoek, line 158: I'm not entirely sure if the way we were doing it was particularly useful, so I am fine with this change as it's "more correct" I reckon - but just curious, was this a noted problem somewhere or why the change?

anoek:
This looks great, thanks for the updates around this :) One quick fix so the string replacement works and a quick sanity check on the translation fallback code and I think we're good to go.

src/lib/translate.ts
benjaminpjones: Granted, there was no issue before the updates in this PR.  However, it manifested itself like so:

image

The issue was that `ngettext` was detecting a _() translation of "%s days" but not "%s day".  I think it was borrowing the translation from `util.mktime()`, which has literals _("1 day") and _("%s days"). To be honest, I wasn't sure about the best way to fix this.

### Things I considered

- Check the language in ngettext `if (currentLanguage() != 'en' && ...)
- Add a translation in English
  - I see catalog is populated in English/debug, but I couldn't figure out how this works since there is no English project on translate.online-go.com
- Call interpolate("1 day", "%s days"
  - This would work, but would produce a warning
- Remove some of the "smarts" of `ngettext`

I went with the last option because it seemed the simplest and would prevent spooky-action-at-a-distance errors like the one I ran into.

src/components/TimeControl/util.ts
benjaminpjones: Arg. Fixed, and added tests so I don't do that again!

benjaminpjones:
Thanks for the review!  That commit should address your first note. Hopefully my comment on the second note explains what I was trying to do, but I'm open to discussion on a different fix.

anoek:
Awesome, yeah that all makes sense, and thanks for the details!

PR #1344, opened 2021-01-04
Preserve type for dup() and add some logging for Notifications error

benjaminpjones:
## Proposed Changes

  - Signature of dup() changed from `dup(obj: any): any` to `dup<T>(obj: T): T`
  - Fix type errors that arose from above change
  - Log an error when `game_id` is missing from a notification

The third item may seem separate from the first two, but I came across `dup()` while debugging the stone removal notification error (hence the branch name).  Happy to break this PR up if necessary.


anoek:
Nice, thanks!

PR #1353, opened 2021-01-11
Align clocks above each other in horizontal zen mode

flovo:
Rebase of #1213.

In horizontal zen mode we have enough vertical space to place the clocks above each other.

This allows the user to reduce the windows width even further while keeping the goban at maximum size.


anoek:
Nice, thanks!

PR #1369, opened 2021-01-28
Show rank instead of rating in rating-grid

flovo:
With the breakdown ratings now being comparable with the overall rating — they are the would be rating if the player had only played the games with the speed+boardsize — we can show the ranks in the breakdown chart instead of the ratings. Most users do not know how to read the ratings, and we are using rank everywhere else on the site already.

## Proposed Changes

image

### DIFF
--- a/src/lib/rank_utils.ts
+++ b/src/lib/rank_utils.ts
@@ -60,6 +60,10 @@ export function rating_to_rank(rating:number) {
     return Math.log(Math.min(MAX_RATING, Math.max(MIN_RATING, rating)) / A) * C;
 }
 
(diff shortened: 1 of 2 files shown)


rajehya:
Speaking of which, isn't it better and more intuitive if integer kyu ranks look the same as the actual rank? For example, 4.1k to be 4k instead of 5k. That way it's consistent with dan ranks and everything looks more intuitive.

benjaminpjones:
Is it still possible to see the Glicko rating somewhere for these?

anoek:
Thanks @flovo !

And yeah we do need a toggle, I'll add it in real quick

PR #1370, opened 2021-01-28
Make game name a link again

flovo:
In the game history make the game title a link again, but keep the row clickable. This way one can use the link context menu when right clicking on the game name.


Adam314159:
Thank you <3

anoek:
Thanks for the fix :)

PR #1371, opened 2021-01-29
Mod note improvements

RubyMineshaft:
## Proposed Changes

  - Removes 1024 character limit for displayed messages which was causing some previously readable mod notes to show as `<Message too long>`
  - When annulling games, words "white" and "black" will now be replaced with player links.

### DIFF
--- a/src/components/Chat/chat_markup.tsx
+++ b/src/components/Chat/chat_markup.tsx
@@ -184,7 +184,7 @@ const global_replacements: TextReplacement[] = [
     { split: /(https?:\/\/[^<> ]+)/i, pattern: /(https?:\/\/[^<> ]+)/i, replacement: (m, idx) => (<a key={idx} target="_blank" href={m[1]}>{m[1]}</a>) },
 ];
 
-export function chat_markup(body: string, extra_pattern_replacements?: Array<TextReplacement>, lengthLimit?: integer): Array<JSX.Element> {
+export function chat_markup(body: string, extra_pattern_replacements?: Array<TextReplacement>, lengthLimit?: number): Array<JSX.Element> {
     let replacements = global_replacements;
 
     if (lengthLimit && body.length > lengthLimit) {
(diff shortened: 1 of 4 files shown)


BHydden:
I would be very cautious to blanket replace all usage of the words "black" and "white" with player links...

[further 'reading' :)](https://www.youtube.com/watch?v=CcZdwX4noCE)

RubyMineshaft:
I did think about whether or not it would cause problems but couldn't come up with another use of "white" or "black" in the context of an annulment note.

I suppose we could use something like `#black` and `#white` if there's real concern that we'd run into issues.

BHydden:
I missed that it was isolated to the game file and not all mod notes... you may be right, at this time I can't think of other reasons to use "black" or "white" specifically in an annulment message without intending to refer to the player...

src/components/Chat/chat_markup.tsx
```diff
@@ -187,10 +187,6 @@ const global_replacements: TextReplacement[] = [
 export function chat_markup(body: string, extra_pattern_replacements?: Array<TextReplacement>): Array<JSX.Element> {
     let replacements = global_replacements;
 
-    if (body.length > 1024) {
```
anoek, line 190: I believe this had a purpose, if I recall correctly some game's had exceedingly long messages (put in by bad people), basically they were so long it'd crash some browsers. I'm cool with adding another parameter to this function, but we should have the default check still here as a safe guard.

RubyMineshaft: What if we add something like a `lengthLimit` parameter which could just be set arbitrarily high for mod notes and kept at 1024 for chat?

anoek: That sounds great :)

RubyMineshaft:
Should look a bit better now. I added an optional lengthLimit parameter to the chat_markup function, and set the limits for chat back to 1024. 😃

anoek:
Cool, thanks!

PR #1380, opened 2021-02-01
Increase length limit for notes in mod center

RubyMineshaft:
Some notes in the mod center are showing as `<Message too long>`.  I increased the limit of those to match what we're using for notes on user profiles.


anoek:
Thanks!

PR #1382, opened 2021-02-06
Bugfix in choosing whether to open new game tab or not

GreenAsJade:
Fixes [the problem of opening a new tab if on a game that is finished](https://forums.online-go.com/t/clicking-on-black-dot-with-of-moves-waiting-opens-in-new-tab/32626/15?u=eugene) which should not happen

## Proposed Changes
 
Look in the right place for window's game status.

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -30,6 +30,7 @@ import {FabX, FabCheck} from "material";
 import {TypedEventEmitter} from "TypedEventEmitter";
 import {toast} from 'toast';
 import {sfx} from 'sfx';
+import { Goban } from "goban";
 
 declare let Notification: any;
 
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Not sure how that originally got through my testing 🤦‍♀️

src/components/Notifications/Notifications.tsx
```diff
@@ -280,7 +280,7 @@ class NotificationManager {
         const goban = window["global_goban"];
```
benjaminpjones, line 280: I strongly recommend specifying type when accessing variables from `window`:

```
const goban = window["global_goban"] as Goban;
```

GreenAsJade: Done

anoek:
Thanks for the fix!

PR #1383, opened 2021-02-06
Show number of byo-yomi periods in zen mode

### LINKED ISSUE #1373: Byoyomi and Zen Mode
**Describe the bug**
Byoyomi Periods not shown in Zen Mode

OS Windows / Chrome Browser 

Not showing remaining periods in japanese byoyomi

benjaminpjones:
Fixes #1373

## Proposed Changes

  - Show number of byo-yomi periods left once byo-yomi has started
  - Some code style fixes: mixed tabs/spaces
  - Some css simplification: elements inside byo-yomi-container were handled separately, even though they share the same style in all cases.

Note: Byo yomi periods do not show up until main time is used up.  If desired, it would be easy to add byo-yomi during main time, but I tried to maintain the minimalist philosophy of zen-mode.

image

image

### DIFF
--- a/src/components/Clock/Clock.styl
+++ b/src/components/Clock/Clock.styl
@@ -126,17 +126,7 @@
 
 
 .MainGobanView {
-    .period-moves {
-        font-size: 10pt;
-    }
-    .period-time {
-        font-size: 10pt;
-    }
-    .byo-yomi-periods {
-        font-size: 10pt;
-    }
-
-    .periods, .period-time, .byo-yomi-periods {
+    .byo-yomi-container, .canadian-container {
         font-size: font-size-mid;
     }
 }
--- a/src/views/Game/Players.styl
+++ b/src/views/Game/Players.styl
@@ -340,110 +340,110 @@
 }
 
 body.dark .MainGobanView.zen .players {
-	.white {
-		&.player-container {
-			background: linear-gradient(#2C2C2C 0%, #383838 100%);
-			border: 1px solid #181818;
-			color: #DDD;
-		}
-	}
-	#game-white-clock,
-	.white .score-container {
-		color: #DDD;
-	}
+    .white {
+        &.player-container {
+            background: linear-gradient(#2C2C2C 0%, #383838 100%);
+            border: 1px solid #181818;
+            color: #DDD;
+        }
+    }
+    #game-white-clock,
+    .white .score-container {
+        color: #DDD;
+    }
 
-	.black {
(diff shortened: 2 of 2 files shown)


benjaminpjones:
Tagging @dragon96 and @flovo in case you have comments as well (this is related to PR #953)

benjaminpjones:
I just realized I made the byo-yomi clock smaller than it was originally.  Here's what the last commit makes it look like:

image

BHydden:
Sudden Death in big red letters... aah yes, very zen :)

benjaminpjones:
Hmm that's a fair point. Do you think the zen users would want (1) instead? I imagine being on your last byo-yomi isn't a very calm experience either way.

BHydden:
I was opposed to the clocks being added in the first place, so I'm afraid I can't help you with any ideas on how to display the B-Y periods 😛 I think if the game is fast enough that the time settings matter in a meaningful way, it's probably not all that zen in the first place 🤷‍♂️

anoek:
I think red SD is fine, it's consistent, zen mode is more about getting all of the clutter out of the way I think, not necessarily avoiding clocks. (I wouldn't be opposed to changing it for the record, I just don't think it's a problem)

I also wouldn't be opposed to having a zen setting that doesn't show clocks, but I don't think that should be the default, and is obviously out of the scope of this PR.

BHydden:
Obviously, it's a design and layout nightmare, but maybe we could consider having these features on a toggle. That way each person could craft their ideal zen environment based on what they find helpful and not.

dragon96:
No strong opinions here. I wanted to add byo-yomi count in the original change, but couldn't figure out a way to make it work with the zen mode simplicity. During the discussions on whether to add a clock, a few folks were worried that adding a clock would make the board even smaller than it already is on a mobile device, so I would say that's the main thing to be mindful of.

One idea I had for a less intrusive zen-mode clock is to have a bar that shows the fraction of your main time you have left, like what lichess has with the green bars. The tricky part is making this work for all the time control styles. Some initial thoughts I have so far:
* byoyomi: maybe a main time bar + a small bar for each byo-yomi period you have left
* canadian: a main time bar with some ticks to show the number of stones left in the period
* fischer without a max-time: maybe only show the fraction of original starting time? not sure.

But that's a topic for a separate FR :) 

Screen Shot 2021-02-08 at 10 50 40 AM

anoek:
Thanks for the improvement!

PR #1389, opened 2021-02-07
Adding sorting case for board size

### LINKED ISSUE #1388: Sorting active games by size does not work
**Describe the bug**
In the list of active games on the home page, I can sort the games by title, opponent etc.. However, when I try to sort by size, the order is seemingly random (but reproducible, so probably not really random).

**To Reproduce**
1. Go to the home page, while having more than ten acitve games.
2. Click on the column label "Size" above the list of games.
3. The order of games changes and an arrow appears indicating the games are sorted by size.

**Expected behavior**
I would expect all 9x9 games to be on top, then 13x13, then 19x19 on the bottom (or the reverse of this)

**Screenshots**
Screenshot from 2021-02-07 10-21-11

 - OS: Ubuntu 20.10
 - Browser Firefox
 - Version 85

 - Device: Moto G7 Power
 - OS: Android 10
 - Browser Firefox
 - Version 85

**Additional context**
I never worked with TypeScript, but I have a little experience with JS. I'd be willing to try to fix this myself, if somebody can point me in the right direction.

TiDreyer:
Fixes #1388 

## Proposed Changes

  - Adding a sorting case for board size to fix bug #1388

It seems to be working as expected, when I check with some dummy games on http://dev.beta.online-go.com:8080/.

I would be happy to receive comments, since I'm not familiar with the code base.

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -144,6 +144,21 @@ export class GameList extends React.PureComponent<GameListProps, any> {
                         }
                     });
                     break;
+
+                case '-size' :
+                case 'size' :
+                    lst.sort((a, b) => {
+                        try {
+                            let a_size = a.goban ? a.goban.width * a.goban.height : 0;
+                            let b_size = b.goban ? b.goban.width * b.goban.height : 0;
+
+                            return a_size - b_size || a.id - b.id;
+                        } catch (e) {
+                            console.error(a, b, e);
+                            return 0;
+                        }
+                    });
+                    break;
             }
 
             if (this.state.sort_order[0] === '-') {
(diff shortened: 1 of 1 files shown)


mmartin:
I'm nitpicking, but maybe `width * height * 100 + width` would be a better formula. There are weirdos who play on non-square boards. :smiley: 

Also, I don't think You need to use `goban` object, this should be fine:
```js
let a_size = a.width * a.height;
let b_size = b.width * b.height;
```

TiDreyer:
Thanks, good points @mmartin. I adjusted the code accordingly.

anoek:
Awesome, thanks @TiDreyer !

PR #1393, opened 2021-02-10
Add game metadata on Goban Div

### LINKED ISSUE #1392: Provide game metadata in the Goban div
People would like to [customize their experience based on the context of the game they are playing.](https://forums.online-go.com/t/features-everyone-secretly-want-on-ogs-but-will-never-be-implemented/32125/63?u=eugene)

They want metadata in css classes:

```
<div class="Goban" data-game-id="123" data-tournament-id="1234">
<div class="Goban" data-game-id="123" data-ladder-id="1234">
```

GreenAsJade:
Fixes #1392

## Proposed Changes

Make PersistentElement take a hash of extra properties to put on the element, and put the game metadata on the Goban in this way.

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -397,6 +397,7 @@ goban-view-bar-width=400px
         align-items: stretch;
         width: 100%;
 
+        // Why? ...
         .goban-container {
         }
 
--- a/src/components/PersistentElement/PersistentElement.tsx
+++ b/src/components/PersistentElement/PersistentElement.tsx
@@ -17,7 +17,13 @@
 
 import * as React from "react";
 
-export class PersistentElement extends React.Component<{elt; className?}, any> {
+interface PersistentElementProps {
+    elt: HTMLElement;
+    className?: string;
+    extra_props?: any;  // hash of new props to put on the element
+}
+
+export class PersistentElement extends React.Component<PersistentElementProps, any> {
     container:HTMLDivElement;
 
     componentDidMount() {
@@ -28,6 +34,7 @@ export class PersistentElement extends React.Component<{elt; className?}, any> {
     }
 
     render() {
-        return <div className={this.props.className || ""} ref={e => this.container = e} />;
+
+        return <div className={this.props.className || ""} {...this.props.extra_props} ref={e => this.container = e} />;
     }
 }
--- a/src/components/PersistentElement/PersistentElement.tsx
+++ b/src/components/PersistentElement/PersistentElement.tsx
@@ -20,7 +20,7 @@ import * as React from "react";
 interface PersistentElementProps {
     elt: HTMLElement;
     className?: string;
-    extra_props?: any;  // hash of new props to put on the element
+    extra_props?: object;  // hash of new props to put on the element
 }
 
 export class PersistentElement extends React.Component<PersistentElementProps, any> {
@@ -34,7 +34,6 @@ export class PersistentElement extends React.Component<PersistentElementProps, a
     }
 
     render() {
-
         return <div className={this.props.className || ""} {...this.props.extra_props} ref={e => this.container = e} />;
     }
 }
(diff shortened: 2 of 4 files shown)


src/views/Game/Game.styl
```diff
@@ -397,6 +397,7 @@ goban-view-bar-width=400px
         align-items: stretch;
         width: 100%;
 
+        // Why? ...
```
benjaminpjones, line 400: Haha [I just delete these when I see them](https://github.com/online-go/online-go.com/pull/1383/commits/e395201c3d24bd7277895343ff3dc55a78525577)... if there's a reason for them, I'd like to know as well!

pdg137:
Here's the HTML I get from this, on a *game* page:

```html
<div class="Goban" game-id="10021" ladder-id="10">
  <div class="Goban" style="width: 609px; height: 609px; background-color: rgb(220, 179, 92); background-image: url(&quot;https://cdn.online-go.com/5.1/img/kaya.jpg&quot;); top: 197.5px; left: 8.5px;">
  ...
```

The info is on the outer Goban div instead of the one inside of it that seems to be the more important one. Is the outer one even necessary at all?

Also, can you get this into Gobans within a "GameList"? I'm hoping this info will be useful for distinguishing games within lists.

GreenAsJade:
@pdg137 It's pretty ugly that we have a Goban inside a Goban isn't it?   But it would be a braver person than me to attempt to flatten that - goodness knows what css would break.   

Given that you know that the structure is like this, do you care?  You get the info you need from the outside one, you set what you need to on the inside one...

It's a heck of a lot cleaner to add it on the outside one, but I can poke around on adding it onto the inside one if "I have to" 😁

GreenAsJade:
... and it will be a lot easier to add it to the game list as well, if it is on the outer thing.  

Though regrettably the class of the outer thing is "small board" with a space.  I'm not sure what the "small" does - maybe font stuff, but the board is used.

We could add "Mini-goban", so it becomes "Mini-goban small board".

Either that or we have to do lots of yucky stuff to put those properties on the Goban inside, in both Game and GameList...

GreenAsJade:
@anoek this should get "work in progress" on it.  Not sure how to do that.

I'm starting to think that really the attributes should be on the real Goban div...

pdg137:
I can probably handle having them on the outer div, but then I'm worried about this change legitimizing that element.  Right now the outer "Goban" one doesn't seem to be doing anything, and it's going to be harder to remove it once it has some useful properties.

The "small board" one seems to be where the green highlighting is applied when it's your move, not sure whether it does anything else.

GreenAsJade:
@pdg137  For the record, I tried collapsing out the outer Goban, as feared it breaks the CSS, so that job is non-trivial.

GreenAsJade:
@pdg137  @anoek  I think this is what was requested, now, for the Goban.

MiniGoban suport for the same thing, in GameList will, require a back end change to provide the ladder and tournament id with each game in the game list, so that's better in a separate request.   Maybe let's see if this takes hold first.

src/views/Game/Game.styl
anoek: Yeah those are always safe to delete, that's just slop - looks like that came in during the initial import into the open source repo 4 years ago - slop.

anoek:
Sounds good, looks good, I'll do some testing and merge it soon.

I *think* it should be possible to remove that outer Goban, all of that was implemented before we were comfortable using `flexbox` so that can likely be gotten rid of, although I think we still need the `Goban.inner-container` to deal with layout of the text around the board, and providing a `relative` environment for the canvas div container. But yeah, unless there's a particular reason to go down that rabbit hole, it's almost certainly not worth it.

src/views/Game/Game.styl
BHydden: We still love you anoek-senpai ❤ thanks for your hard work!

GreenAsJade:
@pdg137  Over to you ;)

pdg137:
@GreenAsJade okay [here you go](https://forums.online-go.com/t/how-to-customize-boards-for-each-tournament-ladder-game/34459)!

PR #1421, opened 2021-03-07
Added hadler for ctrl+c shortcut

### LINKED ISSUE #1407: Ctrl+C blocked by undocumented keyboard shortcut in game view
**Describe the bug**
You can't copy any text out of an OGS game, likely due to an undocumented and obscure "copy branch" keyboard shortcut.

**To Reproduce**
Steps to reproduce the behavior (in an OS/browser combination that uses Ctrl+C to copy):
1. Go to a game.
2. Select some text not in a text input box.
3. Press Ctrl+C.
4. Try pasting the text.
5. Observe that it was not copied correctly.

**Expected behavior**
Ctrl+C should not be blocked while text is selected.  If the Ctrl+C shortcut is to be activated, that should only happen when analyze game is focused, and at that point any selection should be deactivated.

Ctrl+C should be documented in the [wiki](https://github.com/online-go/online-go.com/wiki/Advanced-Tips-&-Tricks#keyboard-shortcuts).

**Screenshots**
Example text that I can't copy:

image

 - OS: Windows
 - Browser: Brave
 - Version 1.20.103 Chromium: 88.0.4324.152 (Official Build) (64-bit)

**Additional context**

Likely caused by [this line](https://github.com/online-go/online-go.com/blob/9e923e1e7d64f2a964c310352c484b6ed7676334/src/views/Game/Game.tsx#L3217).

The Ctrl+V shortcut is also suspect, but all the textboxes available in games seem to disable it already.

KRostyslav:
Fixes #1407

## Proposed Changes

  - In **KBShortcut** class  was added handler for ctrl+c shortcut

### DIFF
--- a/src/components/KBShortcut/KBShortcut.tsx
+++ b/src/components/KBShortcut/KBShortcut.tsx
@@ -197,7 +197,7 @@ $(() => {
                 /* Allow escape through to other handlers, such as SWAL to close modals */
                 return true;
             }
-            
+
             if (shortcut === "ctrl-c") {
                 /* Allow copy text on Ctrl+C in modal */
                 document.execCommand("copy");
--- a/src/components/KBShortcut/KBShortcut.tsx
+++ b/src/components/KBShortcut/KBShortcut.tsx
@@ -81,6 +81,7 @@ let keymap = {
     34: "page-down",
     35: "end",
     36: "home",
+    67: "c",
     112: "f1",
     113: "f2",
     114: "f3",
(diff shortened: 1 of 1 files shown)


pdg137:
Is there a way to let it just run the browser default action instead?  (What if your browser uses Ctrl+C to do something different?)

KRostyslav:
> Is there a way to let it just run the browser default action instead? (What if your browser uses Ctrl+C to do something different?)

hm... I'll check

anoek:
I *think* if you return true there it'll just fall through to the browser handler, which should copy

anoek:
(I would like to see this changed to fall back to the default behavior, I believe there's a number of issues and cases where this won't work. Safari doesn't support it at all, and there are cases when chrome and Firefox refuses it as well, all n all I'd rather just let the browser deal with it so it always works and shouldn't break as browsers change things)

anoek:
Thanks for getting to the bottom of that!

Issue #1097, opened 2020-04-12
Provide an option to disable notification (dropdown)


jsonn:
**Is your feature request related to a problem? Please describe.**
For streamers, it is often difficult to close the notifications in the top right and even then, it is distracting.

**Describe the solution you'd like**
A profile setting to either disable notifications completely or at least don't show the dropdown with explicitly clicking on the button.

Issue #591, opened 2018-03-21
Ability to filter out unwanted notifications


arnaudgo:
### Summary

Sometimes you get five or more notifications for the same event when you are not interested in even a single one.

I would like to be able to mute all notifications minus some of certain types (server maintenance,..) forever, for hours or I don't know.
Of course, I would love to be able to unmute them before the deadline and I would like to be able to access to the notifications history including the missed/ignored ones.

Obviously, it would be super if muting were not session-bound.

### Environment

 Any

### Steps To Reproduce:

Any

### Additional Information:

Ask if needed

anoek:
To be clear, you're talking about the announcements that pop up in the upper right that have links to games or twitch channels or whatever? As opposed to the "game started!" notifications in the drop down?

arnaudgo:
Thank you for asking, indeed ”event" can refer to many things and your assumption is correct: I was excluding "Game started!" or "Game has ended"...

I was referring to the kind of events you mentioned.

Arnaud

PR #1423, opened 2021-03-10
Ability to block announcements from specific users

benjaminpjones:
Related to #1097 and #591, but I can't say it is the exact fix that either has asked for.

## Proposed Changes

  - Discreet dropdown menu next to each announcement in the ActiveAnnouncements card.
  - One option to block the poster (mods excluded).  This can be reversed by unchecking the "Ignore chats" option on the Block dialog.
  - An option to just hide a single post

In both cases a sweet alert dialog is launched, so no one does this by mistake.

Note:

I hijacked the "Ignore chats" block functionality for this.  My reasoning is this: you probably don't want/need to chat with a streamer whose notifications you have blocked anyway.  In the future, I imagine there would be an additional block_announcements field in `players/%%/block`, but for now, I think this is the best solution.

chrome-capture (2)


benjaminpjones:
More notes:

I hand rolled the popup menu, as I know size was a concern for bringing in a library.  I tried my best to make it as responsive, but getting tooltips/popup menus completely right is not so simple.  However, rest assured that I tested the design for mobile screen sizes, and it should at least be usable.

Screen Shot 2021-03-10 at 5 47 47 PM

Another thing that could use polish is the fact that I haven't added any listeners.  This means that if you manually block a user, you need to refresh before the announcements from that user disappear.  However, I felt that it would be appropriate to submit what I have as the core functionality is there!

BHydden:
make sure it's not possible to block anoek

src/components/Announcements/ActiveAnnouncements.tsx
```diff
@@ -64,7 +80,48 @@ export class ActiveAnnouncements extends React.PureComponent<ActiveAnnouncements
 
         return (
             <Card className="ActiveAnnouncements">
-                {lst.map((announcement, idx) => (
+                {lst.map((announcement, idx) => {
+                    let user = data.get("user");
+                    let can_block_user = !user.anonymous &&
+                        (user.id !== announcement.creator.id) &&
+                        announcement.creator.ui_class.indexOf('moderator') < 0;
```
benjaminpjones, line 87: @BHydden

benjaminpjones:
I added a line to prevent blocking mods.  I didn't have a great way to test, but I think @anoek should have that ui_class?

Worst case scenario, a user will try to block @anoek, and the backend will say "lol no".  This is one of the benefits of piggybacking off the "Ignore Chat" functionality

BHydden:
awesome yep, moderator class includes anoek... I wasn't sure if I should suggest *all* of us couldn't be blocked, but certainly, anoek's posts are usually site critical. If including all of the mods with him is the easiest way, that should be fine.

benjaminpjones:
Are a lot of you streamers? As long as you all aren't spamming announcements, I'm fine with giving you a pass haha.

From a logistics standpoint, yes, it's easier to exclude moderators for now because that's how "ignore chat" works.

BHydden:
Almost none of us stream :P but we do sometimes receive and act on requests by other streamers to announce them. This also is fairly rare though as most of the main streamers can make their own announcements.

benjaminpjones:
Makes sense.  I can live with that.  Being able to reduce notifications to just the very important site announcements would be a nice-to-have, but requires more involved API support.

anoek:
I've added backend support for `block_announcements` so we no longer have to piggyback on `block_chat`, FYI

I haven't gone through and updated your code to account for that, except I think in `getBlocks` when I merged. I shoved the updated code in to this PR, so you'll want to pull FYI. 

I've also added a toggle to the block window where you can toggle that flag for a given user.

anoek:
As for eliminating non-site announcement notifications, I think that could be done with the existing preferences system, no? Just add a toggle in the settings page for it so they can add it back in if they want. Might also be cool to have "Block all twitch announcements" or something, and just match that against the domain, that way people can still opt to receive announcements to non streamer things, such as big go events or some notable game on ogs or whatever. (Just a thought, not a requirement for merging or anything)

benjaminpjones:
Oh that's really great, thanks @anoek! Just looked at your changes and I think I only need to switch out `setIgnoreAnnounce` with the new `setAnnouncementBlock`, which I can do later today.

> As for eliminating non-site announcement notifications, I think that could be done with the existing preferences system, no? Just add a toggle in the settings page for it so they can add it back in if they want. 

The thing I'm having trouble seeing is what criteria says whether it's a site announcement?  `ui_class == admin`?  I think the most natural field to look at is `announcement.type`, but currently all the announcements are of type "system".

>Might also be cool to have "Block all twitch announcements" or something, and just match that against the domain, that way people can still opt to receive announcements to non streamer things, such as big go events or some notable game on ogs or whatever.

I didn't even think about looking at the URL!  I think that's a great idea.

benjaminpjones:
That ought to do it.  I've added a new setting page to manage blocked users and muted players.  It's a little rough around the edges, but it does the job.  Also, the ability to mute based on type is now in!

Settings:
Screen Shot 2021-03-12 at 12 18 48 AM

The active announcements pane:
Screen Shot 2021-03-12 at 12 41 49 AM

anoek:
Thanks for the feature, I'm quite sure it'll be loved by many!

benjaminpjones:
Absolutely! And I see you got Announcement History in there too- great stuff.  Thanks for working with me on this!

PR #1425, opened 2021-03-14
Fix data id causing announcements not to go away.

benjaminpjones:
I had copied `"announcements.cleared"` and did not change it to `"announcements.hard_cleared"`.  This meant that we were saving hidden announcements from the card, but actually overwriting the hidden pop-ups.

Apologies for the issues.  Thanks @anoek for looking into it and thanks @pdg137 for reporting!

### DIFF
--- a/src/components/Announcements/ActiveAnnouncements.tsx
+++ b/src/components/Announcements/ActiveAnnouncements.tsx
@@ -39,7 +39,7 @@ for (let k in hard_cleared_announcements) {
         delete hard_cleared_announcements[k];
     }
 }
-data.set("announcements.cleared", hard_cleared_announcements);
+data.set("announcements.hard_cleared", hard_cleared_announcements);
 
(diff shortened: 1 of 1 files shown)


benjaminpjones:
One question about user experience: when we hide the ActiveAnnouncements, do we want to hide the relevant popup as well?  As the code stands (after this fix), it is possible to hide the bar, but leave the pop-ups up.

Screen Shot 2021-03-13 at 10 26 58 PM

Any ideas on what we want here?

BHydden:
I would expect the bar to remain even if you dismiss the pop-up... but I think it would be strange to delete the bar and have the pop-up remain.

anoek:
Yeah I agree, if the user clicks 'hide' in the bar, the optimal experience would be for it to clear the popup one too I think

anoek:
Thanks for the fix @benjaminpjones !

PR #1434, opened 2021-03-24
Remove animation which causes constant 60fps repaints

artursapek:
Fixes issue I reported in the forums (sorry - will use Github Issues in the future)

https://forums.online-go.com/t/web-client-performance-issues/35191/8

## Proposed Changes

There is apparently an infinite webkit animation in the unread notifications UI which causes the browser to continuously repaint at 60 frames per second, causing high CPU/GPU usage.

This PR simply removes the animation, because as far as I can tell it's not even really noticeable to begin with. This fixes the CPU load issue.

See below the difference (pay attention to the "Frames" row)

## Before

image

## After

image

### DIFF
--- a/src/components/Notifications/Notifications.styl
+++ b/src/components/Notifications/Notifications.styl
@@ -18,13 +18,6 @@
 
 indiciator-size=1.5rem
 
-@keyframes pulsate-notification {
-    0% { opacity: 0.5;  }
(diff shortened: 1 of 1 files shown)


anoek:
That's awesome, thanks for digging into that!

artursapek:
@anoek np, thanks for the fast merge. And it looks like it's already live on production? Wow.

PR #1440, opened 2021-03-29
Remove "user has recently timed out of a game"

Trevoke:
Credit to benjito in link beneath

https://forums.online-go.com/t/has-recently-timed-out-of-a-game-uh-does-years-ago-count/31064/12

This display is erroneous, and the backend fix, it seems, is non-trivial. While we wait for a backend fix, this removes the message entirely and frees the users from any potential negative impact on their reputation.

Fixes #

## Proposed Changes

  -
  -
  -


anoek:
Thanks! I might just remove that all together, I'm not sure it's actually beneficial in any way..

PR #1448, opened 2021-04-10
Player notes - initial implementation

GreenAsJade:
This adds the requested ["friend note"](https://forums.online-go.com/t/feature-suggestion-friend-note/35514)

## Proposed Changes

Put a button into PlayerDetails that opens a textarea modal where the user can save freeform text about that player.
  
Note: I didn't see any reason to have this available only for "friends".

Note: this implementation is "initial" in the sense that it was primarily to test out the remote-storage API, but actually I think it is perhaps quite useable anyhow...

Screen Shot 2021-04-10 at 11 32 38 pm

Screen Shot 2021-04-10 at 11 33 32 pm

### DIFF
--- a/src/components/PlayerNotesModal/PlayerNotesModal.tsx
+++ b/src/components/PlayerNotesModal/PlayerNotesModal.tsx
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2012-2021  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as React from "react";
+import {_} from 'translate';
+import {Modal, openModal} from "Modal";
+import { get, set } from "remote_storage";
+declare let swal;
+
+interface Events {
+}
+
+interface PlayerNotesModalProperties {
+    playerId: number;
+}
+
+export class PlayerNotesModal extends Modal<Events, PlayerNotesModalProperties, any> {
+
+    playerNotes = {};
+
+    constructor(props) {
+        super(props);
+        this.state = {
+            notes: undefined
+        };
+    }
+
+    componentDidMount = () => {
+        super.componentDidMount(); /* this.close() doesn't work if you don't do this */
+        get("player-notes")
+        .then(
+            (player_notes) => {
+                this.playerNotes = player_notes || {};
+                if (player_notes && player_notes[this.props.playerId]) {
+                    this.setState({ notes: player_notes[this.props.playerId]});
+                }
+            },
+            (err) => {
+                console.log("Player notes fetch error: ", err);
+            }
+        );
+    }
+
+    updateNotes = (ev) => {
+        this.setState({notes: ev.target.value});
+    }
+
+    saveNotes = () => {
+        this.playerNotes[this.props.playerId] = this.state.notes;
+        set("player-notes", this.playerNotes)
+        .then(
+            () => {console.log("Saved player notes..."); },
+            (err) => {console.log("Whoa, error writing player notes!", err); });
+        this.close();
+    }
+
+    render() {
+        return (
+          <div className="Modal PlayerNotesModal" ref="modal">
+              <div className="body">
+                <textarea placeholder={_("(no notes yet)")} value={this.state.notes} onChange={this.updateNotes} />
+              </div>
+              <div className="buttons">
+                <button onClick={this.close}>{_("Cancel")}</button>
+                <button className="primary bold" onClick={this.saveNotes}>{_("Save")}</button>
+              </div>
+          </div>
+        );
+    }
+}
+
+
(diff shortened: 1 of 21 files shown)


anoek:
I like this, though I'm wondering if / what else the expectation around notes is going to be? With this, you'd have to do a few clicks to see if you have any notes on someone. I feel like for folks using this, the next thing they're going to ask for is some sort of indicator next to someones name showing that you have some notes stored for that player, which might have some implications on how we want to store this data. Not sure though, it seems like the interface could get even more cluttered with note indicators - also is a note indicator enough - should there be flags along with that? Good notes / bad notes kind of thing?

I'm just trying to think through what the next asks are going to be for this line of features to try and head off any notable migration work before we merge this feature.

src/components/PlayerNotesModal/PlayerNotesModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            notes: undefined
+        };
+    }
+
+    componentDidMount = () => {
+        super.componentDidMount(); /* this.close() doesn't work if you don't do this */
+        get("player-notes")
```
benjaminpjones, line 44: @anoek I just realized the interface for `remote_storage` and `local_storage` are the same.  Is this desirable, given that the status quo for pulling in modules is `import {x} from 'y'`?  Hard to tell what is being done without scrolling up to the imports...

benjaminpjones:
>I'm wondering if / what else the expectation around notes is going to be? With this, you'd have to do a few clicks to see if you have any notes on someone.

I think I will use this feature, and personally would appreciate some indication that something is stored without opening the modal.  Something as simple as changing the color and/or style and/or icon of the button in the player card would suffice for me.  Maybe bolding "player notes" when !empty?

src/components/PlayerNotesModal/PlayerNotesModal.tsx
anoek: Could probably use some pondering and improvement, currently, for no good reason other than "that's how it's always been", `data`, `preferences`, and now `remote_storage` all use a `import * as data from "data"` style import, whereas everything else uses `import { x } from 'x'` style, except `ITC` which uses `import ITC from "ITC"` style.

We should probably normalize those a bit, possibly grouping and refactoring the functionality along the way. Currently we have `data`, `preferences`, `remote_data`, and `ITC`, all of which are used for managing persistent state and synchronization in different ways. It seems like maybe we could refactor that code into a tidier state management library or something. But that's it's own can of worms..

anoek:
Yeah I think a special color or / dot / icon on player names would make the feature notably more useful. This has a notable implication that I think we need to revise the `remote_storage` api to maybe deal with namespaces and have some index capabilities to the client can retrieve an index of all all players they have notes for, and then pull up individual notes on demand.

src/components/PlayerNotesModal/PlayerNotesModal.tsx
benjaminpjones: Oh of course, I forgot that `data` gets pulled in like that!  In that case, it probably makes sense to do the same with `remote_storage`.

src/components/PlayerNotesModal/PlayerNotesModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as React from "react";
+import {_} from 'translate';
+import {Modal, openModal} from "Modal";
+import { get, set } from "remote_storage";
```
benjaminpjones, line 21: To avoid naming collisions:

import * as remote_storage from "remote_storage";

See discussion at https://github.com/online-go/online-go.com/pull/1448/files/04f8b723374c9c4f7d8edc572b15ffab6735a77f#r612600953

GreenAsJade: I ended up doing it that way in #1451 in data.ts, for obvious reasons :)

I agree it belongs here (and uniformly).

GreenAsJade:
(for the record, part of this review, the size of the text being entered should also be validated along the way - I'll do that too)

GreenAsJade:
@anoek  I might dense this morning ... why do we need to extend the API to be able to do "notes presence indicator"?

Is it because we don't want to fetch a whole darn hash of all the player notes every time?

Is this a problem even with the current implementation? ... the PlayerNotesModal gets all the notes that the person has entered on all people, every time, and choses the correct one from that hash.

I had also wondered whether you'd want this indication against lists of people for example the friends list - I imagine so.

anoek:
It's probably not a problem for the vast majority of the time, but seems suboptimal as we end up punishing those that use the feature a lot by having an ever increasing non cacheable blob to fetch every time they visit the site from any device. I could see that adding several seconds of loading time for someone who enjoys the feature and is loading up their games on a mobile device when they don't have great signal, or if they just have slow internet, as a surprising amount of the population does.

I also limited each key size to 65k. We could increase that for this type of implementation I reckon, but I think I might spend some extra cycles here and rethink the remote_storage api so it's a bit more efficient for the edge cases and try to mitigate / avoid some of the race conditions of setting blobs of things, which would affect both this and perhaps more applicably any sort of synchronized preferences.

GreenAsJade:
I guess that's what these prototype PRs are here to flush out :)

anoek:
OK at your convenience check out https://github.com/online-go/online-go.com/pull/1453 and see if that fits our need a bit better.

With that version I think you'd want to set player notes keyed by player, and to have an indicator of some sort for a player name should just be a quick get to the remote storage - which will now be immediate instead of going back to the server as remote changes will be pushed out to all devices and a copy stored locally.

GreenAsJade:
( I guess it's obvious, but this pulls in #1451 and it's updates to #1453 )

PR #1451, opened 2021-04-12
Prototype implementation of persisting theme choice into remote-storage

GreenAsJade:
Not intended for production yet - just for @anoek  review of direction

### DIFF
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -28,7 +28,6 @@ let store = {};
 let event_emitter = new TypedEventEmitter<Events>();
 let remote_persisting = false;
 let to_be_persisted = {}; // not-yet confirmed-persisted key value pairs
-let hydrating_from_remote_storage = false;
 
 // keys in this list will be persisted remotely when written by set()
 const remote_persist_list = new Set([
@@ -55,10 +54,14 @@ export function setWithoutEmit(key: string, value: any | undefined): any {
 }
 
 export function set(key: string, value: any | undefined): any {
+    _set(key, value, false); // external callers don't get to decide what is persisted
+}
+
+function _set(key: string, value: any | undefined, disable_remote_persist: boolean) {
     setWithoutEmit(key, value);
     event_emitter.emit(key, value);
 
-    if (remote_persist_list.has(key) && !hydrating_from_remote_storage) {
+    if (remote_persist_list.has(key) && !disable_remote_persist) {
         if (!remote_persisting) {
             // we're not already underway persisting something, so we have to first read the remote values, then update and write back the new...
             // ... taking into account that we might get asked to persist more values while this is happening...
@@ -219,16 +222,13 @@ try {
     remote_storage.get('persisted-local-storage')
     .then(
         (persisted) => {
-            hydrating_from_remote_storage = true;
             for (const key in persisted as {}) {
-                set(key, persisted[key]);
+                _set(key, persisted[key], true /* don't try to remote persist this value we just read from remote! */);
             }
-            hydrating_from_remote_storage = false;
         },
         (err) => { console.error("Error retrieving persisted local storage settings:", err); }
     );
 } catch (e) {
     console.error(e);
-    hydrating_from_remote_storage = false;
 }
 
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2012-2020  Online-Go.com
+ * Copyright (C) 2012-2021  Online-Go.com
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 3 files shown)


src/lib/data.ts
```diff
@@ -43,9 +46,39 @@ export function setWithoutEmit(key: string, value: any | undefined): any {
     return value;
 }
 
-export function set(key: string, value: any | undefined): any {
+export function set(key: string, value: any | undefined, persist: boolean = false): any {
```
anoek, line 49: How about white-listing what key's should persist, that way we don't have to remember in code which keys we should be passing `true` here to? So instead of checking if `persist === true`, we check if the `key` is in the set of keys we persist?

src/lib/data.ts
```diff
@@ -170,3 +204,29 @@ try {
 } catch (e) {
     console.error(e);
 }
+
+// update local data store from persisted values (which may have been modified by other devices)
+
+try {
+    remote_storage.get('persisted-local-storage')
```
anoek, line 211: If I'm not mistaken, having this `get` here and the one up above in `set` could result in a bit of a race condition. I'm thinking we should probably have a single place where we manage that synchronization, what do you think?

src/lib/data.ts
```diff
@@ -43,9 +46,39 @@ export function setWithoutEmit(key: string, value: any | undefined): any {
...
+export function set(key: string, value: any | undefined, persist: boolean = false): any {
     setWithoutEmit(key, value);
     event_emitter.emit(key, value);
+    if (persist) {
+        if (!persisting) {
+            // we're not already underway persisting something, so we have to first read the remote values, then update and write back the new...
+            // ... taking into account that we might get asked to persist more values while this is happening...
+            persisting = true;
+            to_be_persisted[key] = value;
+            remote_storage.get("persisted-local-storage").then(
```
anoek, line 58: Instead of getting every time we set, maybe we can just do a `get` when we first load or need to sync? This is presuming we add a mechanism to be notified when we need to update, which I think we can do just by using our existing inter-tab communication system, https://github.com/online-go/online-go.com/blob/devel/src/lib/ITC.ts

src/lib/data.ts
benjaminpjones, line 49: nit: I wonder if there is a clearer parameter name here than `persist`.  This concern based on my understanding that localStorage is also "persistent" (outlasting the lifetime of the session). Perhaps `persist_remotely`?

benjaminpjones (commented):
Cool stuff!

benjaminpjones:
One question (not necessarily specific to this PR) on intuitiveness of the user settings: is there going to be a "rule of thumb" on which settings will be persisted to remote storage?  I'm just wondering from a user side, will it be possible to guess what will be persisted remotely without testing each setting?

src/lib/data.ts
GreenAsJade: Yeah naming eh!   I changed that a few times already, happy to change it again.

GreenAsJade: Yeah - I was initially thinking that this would go in `preferences.defaults`  

But of course preferences aren't the only kind of data that gets stored.

So I copped out of making another list.

But I think you are right.

src/lib/data.ts
GreenAsJade: Oh, are you pointing out that code can start setting stuff before this initialisation even finishes?

Right - I hadn't considered that case.   I'll tidy that up.

src/lib/data.ts
GreenAsJade: I had steered away from needing to be notified, hoping that the elegantly-simple remote-storage API would be good enough.

Do you have an API addition in mind, to support notification of remote-storage data changes?

If there was such a thing, I'd be inclined to flatten the persisted-remotely localstorage data, so we get updated when individual elements change, and never have to pass the whole set around.

GreenAsJade:
@benjaminpjones  I am sure that the question of what settings do get persisted will be a controversial one, and we certainly will have the conversation of "Can I have an option to indicate which of my preferences should persist, on a per preference basis?  On a "type" basis?  Not at all?"

At this point I've been more inclined to see what the underlying storage implementation will need to be, and thus how far away we are from having anything at all, before asking about those tricky questions :)

GreenAsJade:
@anoek  is the shippable problem mine or yours?   So far I've been hoping anoek magic will make it go away :)

benjaminpjones:
@GreenAsJade looks like it's something to do with the `chat_markup` test. It's now pulling in sockets.ts (through data->remote_storage)

```
ReferenceError: io is not defined

      41 | };
      42 | 
    > 43 | export const termination_socket = window['websocket_host'] ? io(window['websocket_host'], io_config) : io(io_config);
         |                                                              ^
      44 | export const comm_socket = termination_socket;
      45 | 
      46 | let ai_host = '';

      at Object.<anonymous> (src/lib/sockets.ts:43:62)
      at Object.<anonymous> (src/lib/remote_storage.ts:20:1)
      at Object.<anonymous> (src/lib/data.ts:18:1)
      at Object.<anonymous> (src/lib/preferences.ts:18:1)
      at Object.<anonymous> (src/lib/profanity_filter.ts:18:1)
      at Object.<anonymous> (src/components/Chat/chat_markup.tsx:22:1)
      at Object.<anonymous> (src/components/Chat/chat_markup.test.tsx:1:1)
```

I can look into it.  I'm hoping there's an easy way to mock `io` in chat_markup.test.tsx.

PR #1453, opened 2021-04-14
Updated remote storage system to synchronize with other devices

anoek:
This is take 2 on our new remote storage system.

This version synchronizes all remote data to the local device for immediate and fast access. When keys are updated, the updates will be pushed out to other devices immediately. When reconnecting, updates for any changes made while the device was not connected will be downloaded.

* Set will immediately update our local copy of the key and send the update to the server which will then sychronize to other devices pretty quickly.

* Get will immediately return the local copy of our key, and relies on our synchronzation system to take care of updating our local copy so it's ready when we need it.

* Remove will immediately remove our local copy of the key and clear the value from other devices.

* Remote storage only returns meaningful values when the user is authenticated, otherwise all gets will return undefined.

KNOWN ISSUES:

  Right now we set our local state so we can access it immediately, however
  if the send doesn't happen correctly we can end up with clients being out
  of sync with one client having the local state set but the remote storage
  having never received that data. I think we *do* want the local set to happen
  immediately for UI convenience, which means I think the item on the TODO to
  fix this is to have persistent queue for sets and removes that gets retried
  when we reconnect to the server.


GreenAsJade:
Why does the client of remote_storage.set need a promise?   Won't error handling now be managed by remote_storage, so remote_storage.set should return immediately ?

GreenAsJade:
Am I right in understanding that remote_storage has now become "a mirror of the specified subset of whatever you specified to store in localStorage"?

It seems to me that remote_storage assumes that the right way to tell the client about an updated of the remote value is by updating localStorage directly - IE it assumes that these values that are being remotely persisted are always set by data.set()

This in turn assumes that the client of remote_storage API is the OGS client, which cooperates in this way.

Is that what we want?

Note that PlayerNotesModal currently doesn't use the data API at all, it talks direct to remote_storage...

GreenAsJade:
... having thought some more about this, I think the answer is "Yes, all OGS client modules should use data.ts for their data storage needs, and remote_storage underpins the persistence of that using the server remote_storage API.    Other clients can borrow the remote_storage implementation, or do what they like with the server remote_storage API"

GreenAsJade:
... and third update ... I had the wrong end of the stick.    I see that remote_storage.ts is using data.ts to keep it's local cache of the value it was asked to persist remotely.

This feels odd because data.ts is being used by the rest of the client to store that value as well.   So the client says "please set the value of 'foo', which data.ts stores in local as ogs.foo ... then if foo is to be persisted, remote_storage asks data.ts to store the local copy of the remote value of foo as 'remote-storage.{user_id}.foo', which is the entity that is then updated if someone remotely updates 'foo'.

This leaves me wondering how does the client 'watch for' these changes?

If the client has a watch on 'foo', how will the update from remote_storage cause that to fire?

I can see that the implication is that data.ts needs to call remote_storage.get to get the latest value of persisted items, but I can't see how the watch works...

GreenAsJade:
I pushed a suggested change into #1451 - and it works apparently well.

This change is to make it so that the synchronised value _is_ the value that data.ts uses.   remote_storage.ts's job is to do the push to the server for remote persist, and to "tell data.ts" when there is an update received from the server.

anoek:
> Why does the client of remote_storage.set need a promise?

Yeah if/when we transition to having a persistent retry mechanism that should go away probably, maybe have a callback in case the caller wants to know when something was successfully stored, but likely most of the time we won't particularly care.

> This leaves me wondering how does the client 'watch for' these changes?

We should likely implement a light wrapper around `data.watch`, I don't think the right answer is for code to call `data.watch` directly for remote data since we have the implementation detail about prefixing with `remote-storage.{user_id}` - we shouldn't put that throughout the code, that should be something `remote_storage` only needs to know about and deal with.

> bunch of stuff about why we use data.ts instead of localStorage directly

Mainly because it's there and will allow a little bit of code re-use, like we did with `preferences` - `preferences.watch` is just a wrapper around `data.watch`, now `remote_storage.watch` can be the same.

GreenAsJade:
The implementation I pushed views things the other way around for remote_storage: that it is simply the implementation that data.ts uses to persist remotely those thigns that are whitelisted for remote persistence.

The client of data.ts doesn't have to know whether this is happening or not.

By calling data.set to update the local storage with sync updates, data.watch "just works".

GreenAsJade:
Come to think of it - why were you storing with {user_id} ... I ended up without that, because I couldn't see what it was doing.   remote-storage is user-specific anyhow, isn't it?

benjaminpjones:
>remote-storage is user-specific anyhow, isn't it?

But local storage isn't!

anoek:
If two users or accounts are using the same device, we don't want remote storage data shared between them. We should be doing this for preferences too, and probably some but maybe not all data? I'm not sure, I'd have to do an audit. We should have been doing this all along.

Otherwise, if you don't key by user id, notes and preferences from one player will show up for another if they use the same computer.

GreenAsJade:
Oh, that's bogus eh.   I had somehow let myself think that because preferences etc weren't worrying about it, localStorage must be user specific, without actually asking myself "wait, is that right?"

Obviously it isn't.     Curiously, then, we've never had a complaint about that.

I think it works "properly" from my perspective, which is that the multiple accounts that I have all behave the same from a preference point of view - OGS magically knows that I want the same things happening in each acocunt.  I never actually noticed that before 😁    But it must be true, eh?


I like the way it ended up in my latest suggestion, with data.ts presenting the API for use in the client, and remote-storage acting behind it (and updating data's storage, so remote-storage.get is not even needed).

That's much nicer than "oh darn, I want to store foo remotely as well, now I have to update all the places foo is used", which I think is the implication of making remote-storage to be like preferences.

GreenAsJade:
If I have some time soon, is it worth me updating "the way I did it" to handle the user-specific bit, or do you really not like remote-storage being  "under" data?

anoek:
Making whether a value is stored remotely or locally opaque to the caller, (ie hiding `remote_storage` from the user and just doing everything through `data` or not), I think is a discussion to be had.

On the one hand, I don't like having three systems for storing key value pairs. Even having data and preferences seemed redundant. There's certainly a part of me that wants to unify these in some way.

On the other hand, as a user of the data and preferences systems, it's proven to be kind of nice from a maintenance perspective when going through the code - you know when calling a `data` operation you're dealing with adhoc data, and if you're calling a `preferences` operation you're dealing with well defined keys that always have default values set, so different assumptions when you're reading through the code. Similarly, even though it's yet another system, I think it might be better to have `remote_storage` usage be it's own thing, so you explicitly know when you're making calls that are "expensive" and subject to change from other clients, and you explicitly know that it's something that's going to change on all devices not just the current one.

There's the further issue that if we change a key from being local to remote, what's the implication there for replication? Should we proactively push whatever value may be stored locally that when the client first updates to the version of code where that value is now remote? We'll need some migration code to handle that. What about when the user id changes? Conversely, if we roll back a decision to have a key be remote, and make it local, we need to think about that too. In that sense, I think it's much simpler to just have the explicit remote_storage system - every client starts from a blank slate, all keyed by user id, and we don't have to deal with migrations.

Ultimately, while I think it's kind of slick to just add a key name somewhere in data and magically things start replicating, I think there might be some migration headaches, and long term I imagine it'll be nicer to maintain if we are explicit in code when we are dealing with remote data.

anoek:
> If I have some time soon, is it worth me updating "the way I did it" to handle the user-specific bit, or do you really not like remote-storage being "under" data?

Certainly up for discussion, like I said there's certainly a part of me that wants to unify things, a clean path to that objective isn't clear to me yet though.

If we're going to go with the `data` route though we should address the migration issue, I haven't thought that one through much as I've been coming from it the other direction.

GreenAsJade:
What if you want to remote-persist the preferences?

It seems that the preferences code has to be come aware of remote-storage, in this case, and chose to use it instead of data?

That's actually similar to the initial implementation I did, which had 'remote-persist' as a parameter to data.set, I guess.

I'm just checking if this is what we'd want, rather than the 'whitelist' of "remote persist these" that was the suggestion back then?

anoek:
>What if you want to remote-persist the preferences?

> It seems that the preferences code has to be come aware of remote-storage, in this case, and chose to use it instead of data?

I'm thinking we might want to get a bit fancier with preferences, and maybe here is where we want to blend in some what what you were doing with your set style and not.

Following this rabbit trail a bit, I think depending on the preference we'll want to a) store it locally only and never remotely b) store it remotely always, or c) have an option where you can have a setting on one device that overrides the default remote setting, but still be able to set the default setting for other devices. For example, you might want dark theme enabled by default on all of your devices, except for your tablet where maybe you want to have the light theme instead.

The interface and method that comes to mind would be to add a "save this to all of my devices" kind of button/icon next to a preference that you're setting, so changing a preference would update the local copy as normal, saving would update the remote copy. When reading a preference we'd read from local, if it doesn't exist, we'd read from the remote, if that doesn't exist, we'd read from the defaults.

Anyways, that's a bit of a can of worms. Back to whitelisting vs explicit remote_storage calls, I think a good argument could indeed be made for whitelisting some of the preferences - so maybe that's the place to put the automagic remote storage stuff you were doing? Do you have any examples of where having that functionality reside in data would be preferable to having it in preferences?

GreenAsJade:
I dunno about preferable, but the current implementation of theme-choice goes straight to data :)

Maybe it should actually be routed through preferences as well!

GreenAsJade:
Oh, and of course player-notes is not amenable to being a preference.

anoek:
> I dunno about preferable, but the current implementation of theme-choice goes straight to data :)

Uhg, actually looks even better, it goes through preferences *and* data, preferences for what standard theme to use, and data for the customization stuff.

anoek:
Ok, after further conversations on the topic, I've become convinced that @GreenAsJade's approach is better.

The remote storage functionality has been combined into the `data.ts` file and hidden from normal use. Now, `data` and `preferences` `set/remove` functions have an additional `replication` parameter to control if, when, and how the data being set is replicated to other clients. Applicable excerpt from the comments:


```
/* Usage:
 *
 * That data module is an interface that aids with storing local and remote
 * data. It is the underlying storage mechanism for the `preferences` system
 * which adds key name checking and ensures default values are set.
 *
 * The main methods are get, set, remove, setDefault, watch and unwatch
 *
 *   set(key, value, replication?)
 *       Sets the key value pair locally and optionally replicates the value to
 *       other devices used by the player. The replication parameter is
 *       described below.
 *
 *   remove(key, replication?)
 *       Removes the key value pair locally and optionally replicates the
 *       removal of the key to other devices used by the player. The
 *       replication parameter is described below.
 *
 *   setDefault(key, value)
 *       Sets a prioritized default value to be returned by get if no other
 *       value for the key is available, see get for more details on
 *       prioritization.
 *
 *   get(key, default_value?)
 *       This computes and returns the appropriate value for the given key.
 *       The value of a key can reside in several places. We check in order
 *       and return the first defined value we find.
 *
 *          If the value has been locally set (via set), that value is used.
 *          Else if the value has been set remotely, that value is used
 *          Else if the value has a default value set by setDefault, that value is used
 *          Else the default_value is returned, which may be undefined.
 *
 *   watch(key, cb: (value) => void, call_on_undefined?, dont_call_immediately?)
 *      When the value of a key (as determined by get(key) changes for any
 *      reason to some defined value, the callback is called with the new
 *      value. If you want the callback to be called even if the value is
 *      undefined, pass true for call_on_undefined. Normally, the callback
 *      is immediately called when you call watch to avoid having to duplicate
 *      a get and a watch statement, however if this behavior is not desired,
 *      pass true for the value of dont_call_immediately.
 *
 *   unwatch(key, cb: (value) => void)
 *      Removes a callback from being called when key is updated
 *
 *
 *   REPLICATION
 *
 *   The optional set and remove `replication` parameter can be set to one of
 *   the following values if replication to other devices used by the currently
 *   authenticated player is desired. (Note that if the user is not logged in,
 *   no replication will take place.)
 *
 *   values for replication:
 *
 *      Replication.NONE
 *          No replication is performed. This is the same as not passing this paramter.
 *
 *          When to use: When you don't want the value replicated to other devices
 *
 *
 *      Replication.LOCAL_OVERWRITES_REMOTE
 *          Replicate the value set/remove, however if the other device has already
 *          set the given value locally, use that value instead of the replicated value.
 *
 *          When to use: When you want to update an account wide default value, but
 *             not override any per-device defaults
 *
 *      Replication.REMOTE_OVERWRITES_LOCAL
 *          Replicate the value set/remove and overwrite the locally set values on
 *          other devices.
 *
 *          When to use: When you all devices and tabs to have the same value
 *
 *      Replication.REMOTE_ONLY
 *          Replicate the value set/remove but do not set/remove the local value
 *
 *          When to use: When you want to update the account wide default value, but
 *             not override any per-device defaults, including the current device
 *
 */
```

In this way all existing code that reads and watches data will benefit without modification. When we set a value, if we want it replicated we update the `set` code to specify the replication behavior.

I'll probably revisit this code again before merge to clean it up a bit more and do another round of tests with fresh eyes.

benjaminpjones:
This interface looks really nice (look at y'all writing documentation 😋)  One thing I'm trying to wrap my head around is the `setDefault` function.  Not that I don't understand what it does, but that I don't really understand how it should be used in practice.

I feel like optimal would be for all the `setDefault` calls to happen sometime around page load.  But if that's the case, wouldn't it make sense to hold it all in one config file (perhaps `defaults.json`) and load the map all at once?

anoek:
`setDefault` is indeed used typically (always?) on load, but not always with static data. Preferences uses it to set all of the defaults, which is indeed in a static map object, however our chat system uses it to set what the default chat channels should be based on navigator language, and I think main sets some defaults for configuration parameters based on globals that exist from index.

PR #1455, opened 2021-04-15
Player notes on friends list

GreenAsJade:
Adds an indicator of player notes presence on the friends list.

The little icon is clickable to open the PlayerNotesModal.

(building on the work of previous PRs)

Screen Shot 2021-04-15 at 9 28 48 pm

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,65 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the devel branch
+  push:
+    branches: [ devel ]
+  pull_request:
+    branches: [ devel ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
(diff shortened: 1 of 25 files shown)


anoek:
I'm wondering if we should put that indicator code in the Player component, that way we can turn it on/off anywhere where we use that? Might be good to have it default on so it's just kinda everywhere.

If we go that route, I'm not sure we should put a full note pad icon near the name though, maybe either an indicator dot or a color change or something?

BHydden:
I disagree... I think the clipboard icon does a good job of representing notes... certainly a dot should not be used as it would be confused with the online presence indicator

GreenAsJade:
Doh - definitely it should be in the Player component.  As BHydden, I do like the notepad: in the absence of that, it takes a fair bit of "intuition" to work out why the name is bold (or whatever).

anoek:
Sounds good, notes icon it is.

GreenAsJade:
This now has the latest remote-storage, being called explicitly as needed.

It's a lot more work to make something pre-existing use remote-storage now : you have to find all the places it was referenced.  But that may not be a bad thing, because it makes you think about what if this thing changes unexpectedly in each of those places.

This doesn't have player note icon extracted into Player ... separate branch for that, I'm not totally wild about the result.

PR #1456, opened 2021-04-16
Notes indicator in player

GreenAsJade:
As suggested, this puts the render of the "we have player notes" icon into the Player component, as an option.

The option is turned on only in Friends list in this commit.


GreenAsJade:
(I said I wasn't wild about it, but I guess it turned out OK, though as usual color/theme questions abound!)

PR #1457, opened 2021-04-17
"First remote storage features" update

GreenAsJade:
Put players notes and theme replication on top of the updated remote-storage in #1453.

** Includes suggested fix for emitting replication update logic

*** Includes workaround for `row.replication` not being populated in remote-storage/synch


src/lib/data.ts
```diff
@@ -522,6 +522,7 @@ termination_socket.on('remote_storage/update', (row:RemoteKV) => {
     safeLocalStorageSet(`ogs-remote-storage-store.${user.id}.${row.key}`, JSON.stringify(row));
 
     if (last_modified < row.modified) {
+        safeLocalStorageSet(`ogs.${row.key}`, row.value);
```
anoek, line 525: This needs to be moved up to the above if statement so we only store the value if `if (row.replication === Replication.REMOTE_OVERWRITES_LOCAL) `

anoek: Also that needs to be json stringified I think

GreenAsJade: 🤦‍♀️  Of course, done.

GreenAsJade: Oh, I missed the stringify comment, I'll take a look

GreenAsJade:
(I couldn't resist tidying up the commit tree 😊  )

src/components/NavBar/NavBar.tsx
```diff
@@ -55,10 +58,11 @@ function previewTheme(theme) {
 function exitThemePreview() {
     _update_theme(data.get("theme"));
 }
+
 function setTheme(theme) {
-    data.set("theme", theme);
-    _update_theme(theme);
+    data.set("theme", theme, data.Replication.REMOTE_OVERWRITES_LOCAL); // causes _update_theme to be called via the data.watch() in constructor
```
anoek, line 63: I'm thinking this should be LOCAL_OVERWRITES_REMOTE so that if you have multiple devices, if you set it manually on any particular device it'll stay stuck for that device?

Note that this will make more sense as time goes on and you begin to use new devices or browser local storage gets cleared for whatever reason.. right now, all devices will have the local value for theme set so it'll be a bit harder to notice this, but I think in time `LOCAL_OVERWRITES_REMOTE` is the right way to go.

Come to think of it, that name is a misnomer, it should be `LOCAL_OVERRIDES_REMOTE` or something, it doesn't do any writing..

src/components/PlayerNotesModal/PlayerNotesModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    updateNotes = (ev) => {
+        const new_notes = ev.target.value;
+        if (new_notes.length < 5000) {
+            this.setState({notes: ev.target.value});
+        }
+    }
+
+    saveNotes = () => {
+        data.set(`player-notes.${this.props.playerId}`, this.state.notes, data.Replication.REMOTE_OVERWRITES_LOCAL);
```
anoek, line 52: Here, maybe we should trim `this.state.notes`, and if it's an empty string, do a `data.remove` instead?

src/components/NavBar/NavBar.tsx
GreenAsJade: Now we are coming to "what people would want" instead of "make it work", at last 😄 

I had it set with REMOTE_OVERWRITES_LOCAL so that I definitely saw the changes every time, as much as for any other reason (also this was the equivalent of how it worked before those new funky options were added).

Overall, it seems like your logic is correct - if it's the case that when I log into a new device, I get the remote setting automatically (should be true, needs to be confirmed) then you would argue that if I change that locally, then I want it different here.

BUT what if I just changed my mind about what I want everywhere?

src/components/PlayerNotesModal/PlayerNotesModal.tsx
GreenAsJade: We could.   Interestingly, it  seems it would make no difference to the functionality at this point -  "" evaluates to false anyhow, so it appears to all intents and purposes like the person "has no notes".   So one could argue it would be "more code to achieve the same thing"...

src/components/NavBar/NavBar.tsx
GreenAsJade: ... I feel more inclined to release it with it always changing everywhere.   When someone comes up with "hey I want liht mode on my phone" we could look at this problem at that point.  In the mean time, everyone would experience cool synching  😁

GreenAsJade:
So ... these setting are still notably not-user-dependent when stored in localStorage @anoek

GreenAsJade:
... and since the only way in and out of localStorage is via safe read and safe write in data.ts, maybe it's now as simple as putting the user in there?

src/components/NavBar/NavBar.tsx
anoek: But if it's `REMOTE_OVERWRITES_LOCAL`, you *can't* have separate themes for different devices, pretty sure that's going to annoy some folks.

I'm pondering if there's a better way to approach this rather than having remote actually stomp on the local values too.. might noodle on it a little bit more here today.

I'll get this merged and we can feel it out a bit on beta before we release.

src/components/PlayerNotesModal/PlayerNotesModal.tsx
anoek: It just makes it so the key will be removed eventually, tidier db / replication, not that big of a deal

anoek:
That's a little trickier because there are some things we want to store when the user is not logged in (such as theme, language, volume). That's definitely a different PR, we'll probably want a migration to copy over everything from the non-user-id specific keys to user specific keys, and figure out that whatever we do makes sense when you visit the site without being logged in, and how that follows you as you create an account and log in, log out, log in as a different user, etc.

anoek:
Thanks for all the work @GreenAsJade !

GreenAsJade:
Man, that "lets have a simple key-value store for clients to use" sure turned out to be more than I had imagined!
:)

anoek:
Yeah I figured it'd be a can of worms, that's why I hadn't gotten around to it. But it's good to have it moving along now, it'll be great to have in place.

PR #1459, opened 2021-04-18
Save sync update in localStorage

GreenAsJade:
Just for completeness on this branch, if we want this feature update on a branch as finished.

This same fix is in #1457.


anoek:
Awesome thanks, yes indeed :)

PR #1466, opened 2021-04-25
Update Chat.tsx

### LINKED ISSUE #1465: Change placeholder text on game page
As per the suggestion in help chat:

> ronin3: Kosh: apropos the "chat etiquette" - I noticed other sites have a placeholder that goes "Please be nice in chat!" that I like quite a bit - here we have "say hi!" which is okay, but that other message kind of sets the tone a bit more

https://github.com/online-go/online-go.com/blob/8a8d763db1ecc71465d31ac69eac2836932c4a00/src/views/Game/Chat.tsx#L239

Suggest we implement suggestion to have placeholder text say `Please be nice in chat.`

BHydden:
Fixes #1465

## Proposed Changes

  - As per the suggestion in help chat, have placeholder text say `Please be nice in chat.` rather than `Say hi!`.


anoek:
Something about that exact verbiage doesn't sit too well with me. I think it's that it somewhat implies that we think the player needs to be told that, where as I would like it phrased more like we're encouraging their natural friendly behavior. Said another way, I'd like it to come across that we are making the assumption that they are a good nice person, not that they would presumably be not nice in chat and thus need to be asked to be nice.

BHydden:
How about just something like `Friendly game chat.`?

More of a title than a request, which the first two were...

anoek:
I feel like it could still be better...

What do we feel was lacking or otherwise inadequate about "Say hi!"? was it not clear enough that that was the chat area, or something else? I'm not opposed to changing it, but we should know what we're trying to fix so we can decide if whatever replacement we come up with does the trick.

BHydden:
It was noted in the following discussion in Help chat that we at OGS do not require or except users to chat. 

Some may see the text "Say hi!" and get the impression that a greeting is expected and that if one is not offered that player is being rude.

Shy people should not feel pressured by out placeholder text to speak if they prefer not, and equally chatty people should not feel entitled or offended by a silent opponent.

These are the main concerns this PR is attempting to correct; that our placeholder text in the chat box better reflects the actual ideals that our server stands for. 

If the above two examples are not acceptable, I will continue thinking of how best we can identify this field...

I feel like the ideal impression/understanding I would want to succinctly send is "this is the area which you can use to chat kindly with your opponent, if you wish to do so"

BHydden:
Below is the full chat from Help chat in case my summary missed any key details...

```
ronin3: Kosh: apropos the "chat etiquette" - I noticed other sites have a placeholder that goes "Please be nice in chat!" that I like quite a bit - here we have "say hi!" which is okay, but that other message kind of sets the tone a bit more
GreenAsJade: Where do we have something that says "Say hi"? Actually, we don't require or expect people to say hi in game
BHydden: It's the grey default text in the chat text box on a game screen
BHydden: placeholder text*
BHydden: ronin3: I liked your idea and created an issue and pull request for it on GitHub; we'll see what anoek thinks :) GH-1465
BHydden: As GreenAsJade said, we don't require or expect people to say hi in game... so I imagine the change may go through :)
GreenAsJade: Wow - an oversight eh! I've seen it of course, and just mentally translated it as "if you want to say hi, you can do it in this window", rather than "please say hi here" !
BHydden: yeah same I think... as a naturally talkative person I never really gave the "say hi!" text a second thought... but now that it's pointed out, it does seem to be a bit incongruous with our stated ideals
GreenAsJade: Fo sho
```

darkronin:
Even nice and excellent people can hurt with words or misinterpret or see a non-reply as an act to ignore. If the exact verbiage "Please be nice in chat" does not sit well, here are a few alternatives, taking inspiration from other popular sites with a chat box.

```
Please be excellent in chat!
Say something...
Say something... or not.
```

I quite like "Say something..." because it sets the expectation: instead of "saying hi" (meaning, implying a conversation is about to happen) it states that you are free to speak _on your side_.


There are also a few more "impersonal" ones:

```
Type your message here
Send a message 
Chat at this table 
Discuss at this table
```

To the above it stands to reason to add (or not): `(warning: opponent may respond or not)`

anoek:
Ok I'm warming up to the idea of having the something like `Please be nice in chat!` or `Plase be excellent in chat!`

My go to places to look are lichess and chess.com, I'd be curious what other sites do.

Lichess uses the original suggestions: `Please be nice in chat.`
Chess.com uses: `Message...`
Discord: `Message #ChannelName`
Slack: `Message #ChannelName`

What if we did something like `Message ...` in the input field, and a italicized centered text at the top of the chat window that disappears when the user talks that says either `Please be excellent in chat!` or `Please be nice in chat!`or something like that?. For the main chat page and for private chats we then switch to `Message #English` and `Message Username` placeholders? 

That might serve a few purposes as it'd more clearly designate that part of the screen as where chat output goes with the message "Please be nice in chat!" that attempts to set the tone of the chat, and then the "Message ..." style input placeholder that they've probably seen other places where they chat

> seeing this forum thread makes me think the only way to do this PR properly is perhaps to have a different version for each culture/flag of the user.

I'd think our translation system should take care of this implicitly, each translator can phrase things in a way that makes sense and attempts to deliver a similar sentiment. And when there are differences in cultures within languages, we do technically support translations on language/region bases too, like `en-au`, `en-uk`, `en-us` - but we haven't needed to use that yet I don't think.

PR #1470, opened 2021-05-04
Unify "master" and "override" volume settings

rileylark:
Hi there,

I'm Riley, and I've been a customer of online-go.com for about a year. I really enjoy the site, but I frequently find myself thinking the master volume is set to zero, but then surprised that the sounds are playing after all.

One potential area of confusion is the separation of the concepts of the "override" volume and "master" volume, I think. I was interested in checking it out and so I humbly submit this sacrificial PR.

## Proposed changes
* Merge the volume override setting visible on game boards with the master volume setting visible on the settings page
* Drop the `sound-volume` and `sound-enabled` preference keys. If deployed, this will cause existing user preferences to revert to their "master volume" setting.
* Overall, this means when a user adjusts the volume on a game screen, it ALSO adjusts the master volume. Conversely, adjusting the master volume means adjusting the volume for individual games. Before, those were two separate controls, which I personally have found confusing over the past year.

### DIFF
--- a/src/lib/configure-goban.ts
+++ b/src/lib/configure-goban.ts
@@ -85,8 +85,8 @@ export function configure_goban() {
         getShowVariationMoveNumbers: ():boolean => preferences.get("show-variation-move-numbers"),
         getMoveTreeNumbering: ():"none" | "move-number" | "move-coordinates" => preferences.get("move-tree-numbering"),
         getCDNReleaseBase: ():string => data.get('config.cdn_release'),
-        getSoundEnabled: ():boolean => !!preferences.get('sound-enabled'),
-        getSoundVolume: ():number => preferences.get('sound-volume') as number,
+        getSoundEnabled: (): boolean => data.get('sound.volume.master') > 0,
+        getSoundVolume: (): number => data.get('sound.volume.master'),
 
         watchSelectedThemes: (cb) => preferences.watchSelectedThemes(cb),
         getSelectedThemes: () => preferences.getSelectedThemes(),
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -74,10 +74,8 @@ let defaults = {
     "show-tournament-indicator": true,
     "show-variation-move-numbers": false,
 
-    "sound-enabled": true,
     "sound-voice-countdown-main" : false,
     "sound-voice-countdown": true,
-    "sound-volume": 0.5,
 
     "sound.volume.master": 1.0,
 
--- a/src/lib/sfx.ts
+++ b/src/lib/sfx.ts
@@ -517,12 +517,6 @@ export class SFXManager {
         return data.get(`sound.voice-enabled.${sprite_name}`, true);
     }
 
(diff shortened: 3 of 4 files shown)


anoek:
The expected behavior is that if you set your master volume to zero, when you visit any game the default volume for that game should also start off at zero. If that's not the case, that's certainly a bug - can you confirm what behavior you're seeing?

rileylark:
That's the behavior I was expecting too. I can confirm that my master volume is currently set at zero, but when I start a game the volume is NOT at zero. Furthermore, going into preferences and RE-setting master volume to zero does not affect that in-game volume. 

In `sfx`, it looks like that in-game volume setting overrides master volume throughout the entire app, so that in-game volume basically behaves as the real master volume. My PR here just kinda unifies them so that master volume and in-game volume are the same. After my PR, changing in-game volume also changes master volume, and changing master volume changes the in-game volume.

In both cases it still takes a refresh for the setting to appear after my PR.

Are you hoping that games could have a _per-game_ override?

anoek:
Yeah per-game override, which was certainly working at somepoint, not sure when that would have changed. I'll have to go look through the history.

We used to have it all one setting like you're PR here is doing, however people frequently spectate games and the request from them is they want to be able to change the volume on some games.

What we *really* need to do is implement a settings modal that allows you to adjust your sound settings from within the game page, and from there allow you to either control the master volume or the "this game" volume setting. But that's a bit more of a project..

rileylark:
Ah ok, my PR certainly does not restore that behavior. Here's a demo of the current behavior, where you can see a new Demo inheriting the previous game override instead of the master volume. The same behavior exists between games.

The `sfx` module uses a single location in `data` to store the override, and always gives that preference over the master volume.

Feel free to close this PR if you'd rather keep current behavior. Would you be interested in more work from me thinking about the volume settings to achieve your goal of per-game overrides?

https://user-images.githubusercontent.com/1630424/117166603-256f2600-ad8c-11eb-9153-2774d2b8a239.mov

rileylark:
Random thought: one quick fix would be to keep the current "override" code _but just not store it to `data`_, so that it will effectively override _the current tab_ but not save in preferences.

1. Master volume would still be used in other tabs, but
2. Of course if you refreshed or changed games you'd have to re-override

anoek:
OK I'm coming around here. My understanding of the situation is that we have a fancy sound settings page which of which the volume levels are all completely ignored because we override the volume with whatever the last value of `sound-volume` was on the game page. That implies to me that this patch makes a good deal of sense since apparently no one seems to be complaining too much about not having per-game sounds these days since they haven't had it. This patch doesn't change the current user experience, it just ditches the `sound-volume` setting in favor of using the fancy sound settings system we have.

So, I'm all for it.

My only request is that since we removed the volume override system, lets remove the `setVolumeOverride` and `getVolumeOverride` methods too since they no longer make any sense, just do `getVolume("master")` style calls instead.

anoek:
Awesome, thanks for the fix and sticking with me through the conversation!

PR #1473, opened 2021-05-08
Systematic fix to presence/flare alignment in Player

GreenAsJade:
Fixes ongoing hassles of alignment of elements of `Player`

## Proposed Changes

- Don't use absolute positioning for presence indicator, use flex to line it up on the centreline no matter what size it is
- Use flex to put all the elements on the baseline in the case of "chat", where we have flare instead of presence indicator
- Put the primary parameters in `Player` and specialise only where it needs to be specialised (eg PrivateChat)
- This also fixes the squished private chat header.

Screen Shot 2021-05-08 at 12 36 44 pm

### DIFF
--- a/src/components/FriendList/FriendList.styl
+++ b/src/components/FriendList/FriendList.styl
@@ -17,7 +17,6 @@
 
 .FriendList {
 
-
     .show-offline {
         input, label {
             cursor: pointer;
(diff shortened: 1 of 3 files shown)


anoek:
Awesome thank you :)

Issue #1476, opened 2021-05-10
CPU usage high just sorting chat lists for windows I'm not looking at


rileylark:
**Describe the bug**
When someone else joins or parts a chat channel, the entire member list is sorted twice. For me, mid-day in the USA, this takes on average about 35ms of CPU.

![image](https://user-images.githubusercontent.com/1630424/117707416-2fca5f00-b194-11eb-8c69-3e827dc75d56.png)

It's happening at a rate of something on the order of 20Hz

![image](https://user-images.githubusercontent.com/1630424/117707553-4ffa1e00-b194-11eb-966e-26de4cf178f9.png)

On my MBP this means about **25% CPU usage** for this tab - even though I'm not even looking at the chat screen. On my phone, this means my battery is draining really fast when I'm just contemplating a static Go board for minutes.

**To Reproduce**
1. Join these chat channels
![image](https://user-images.githubusercontent.com/1630424/117707710-7d46cc00-b194-11eb-9651-28c98dcd9926.png)

2. Wait for mid-day USA
3. Go to a static game, not a chat room

**Expected behavior**
Low CPU usage 

**Actual behavior**
High CPU usage!

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Brave (Chrome)
 - Version 1.24.82 Chromium: 90.0.4430.93 (Official Build) (x86_64)

**Smartphone (please complete the following information):**
 - Device: Pixel 3
 - OS: Android
 - Browser: Brave

anoek:
Yep any/all of those would be awesome. Thanks for hunting that down, you'll make a lot of people happy if you can make that go away :)

If you're seeing chat sorting happening when you're not on the chat screen, I suspect there might be something going on stemming from the use of the `ChatIndicator` system. That system doesn't need any sorting at all I don't think, nor an active list of the users even, but I think those things probably just happen because it's re-using a lot of state and code from the chat system that drives the chat page, which does need those things.

rileylark:
Great, I might check it out this week!

rileylark:
So I can see the big lists to sort, it might be nice to develop this against production APIs - will it work to just swap in production values to the dev environment?

rileylark:
I made progress on this - switching to an insert/remove instead of re-sorting each time saves a buncha CPU. Will post PR later today

rileylark:
I think this is resolved - at least down to "it's not the biggest CPU issue anymore." Ok for me to close it? Let me know if you'd rather keep it open!

PR #1477, opened 2021-05-11
keep users list sorted instead of re-sorting from scratch each time

rileylark:
Fixes (improves, anyway) #1476 

## Proposed Changes

The chat_manager was re-sorting channel user lists twice every time someone joined or parted the channel. This led to the CPU graph below even when sitting idle on the home screen, where you can see a single call to `_update_sorted_lists` was taking about 20ms

image

I changed the code to either insert joining users or remove parting users, with the assumption the sorted lists were already sorted. This made the CPU graph look more like this:

image

I'm not totally sure what the gray hashes mean, but I think it's like "idle" or some such. The overall tab CPU usage dropped for me on my 2013 MBP from 25% CPU to 2% CPU.

## Caveats
I only tested on chromium on Mac - maybe Firefox or Safari have different optimizations that would see different CPU impact. It's hard to imagine this making things WORSE for any browser, but I've said things like that and been wrong before ;) I'd understand if you'd want to test this more widely before merging. I personally do not want to do that testing ;)

## Potential next steps
There's still quite a bit of optimization _possible_ - inserting items into the list still takes my CPU 2ms for each user, and there's no need I could see to maintain sorted lists of every channel. I thought a 20% drop in absolute CPU usage for this would be worth merging by itself.

Also, the chat window itself still completely pegs my CPU. This doesn't really improve that situation much. It should mainly help folks who are looking at games on their phones and not chatting.

----
I believe this could be ready to merge, let me know what you think and if you'd like any adjustments!

### DIFF
--- a/src/lib/chat_manager.ts
+++ b/src/lib/chat_manager.ts
@@ -27,7 +27,7 @@ import { ActiveTournamentList, GroupList } from 'types';
 import {_, interpolate } from "translate";
 import { shadowban } from "src/views/Moderator";
 import { getBlocks } from "BlockPlayer";
-import { string_splitter, n2s, dup, Timeout } from "misc";
+import { insert_into_sorted_list, string_splitter, n2s, dup, Timeout } from "misc";
 
 
 export interface ChatMessage {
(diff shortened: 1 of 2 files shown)


anoek:
Very cool, thanks a ton!

PR #1478, opened 2021-05-12
Guard against timing resulting in moderator banner not being created yet

GreenAsJade:
Fixes Moderator chat PM window not opening in some cases.

## Proposed Changes

Guard against moderator banner not being defined yet.

### DIFF
--- a/src/components/PrivateChat/PrivateChat.tsx
+++ b/src/components/PrivateChat/PrivateChat.tsx
@@ -280,6 +280,8 @@ class PrivateChat {
         }
     }
     updateModeratorBanner() {
+        if (!this.banner) { return; }
+
         if (this.player.ui_class.match(/moderator/)) {  // surely would be better to use player.is_moderator, but not available!
             this.banner.removeClass("banner-inactive");
             this.banner.empty();


src/components/PrivateChat/PrivateChat.tsx
```diff
@@ -280,6 +280,8 @@ class PrivateChat {
         }
     }
     updateModeratorBanner() {
+        if (!this.banner) { return; }
+
         if (this.player.ui_class.match(/moderator/)) {  // surely would be better to use player.is_moderator, but not available!
             this.banner.removeClass("banner-inactive");
```
anoek, line 286: This would throw an error if you removed that condition and it was true, no?

GreenAsJade: Yes that's why under some circumstances at the moment the chat doesn't actually pop up when it should - somehow a chat is being received before this.banner is set.

anoek: Hah yeah, sorry about that, I obviously should refrain from reviewing first thing in the morning. Thanks for the fix!

PR #1479, opened 2021-05-12
Notify of late chats

GreenAsJade:
Fixes the problem where we never have a reason to go and look at the chats people leave in our games after we have left.

## Proposed Changes

Send a PM to the players in the game who are not connected at the time that a chat is left in a finished game.

Note that a "nicer" implementation might be a notification, or at least more "system like" message in the PM box, but these would require backend support.

(I had thought to use the key-value store, except it would need to be extended to allow for shared data)


src/components/PrivateChat/PrivateChat.tsx
```diff
@@ -280,6 +280,8 @@ class PrivateChat {
         }
     }
     updateModeratorBanner() {
+        if (!this.banner) { return; }
```
anoek, line 283: As mentioned in the other PR, not having this will make it explode, best to fix it somewhere else I suspect?

src/views/Game/Chat.tsx
```diff
@@ -235,7 +268,7 @@ export class GameChat extends React.PureComponent<GameChatProperties, any> {
                             ? _("Login to chat")
                             : !data.get('user').email_validated ? _("Chat will be enabled once your email address has been validated")
                                 : (this.state.chat_log === "malkovich"
-                                    ? pgettext("Malkovich logs are only visible after the game has ended", "Visible after the game")
+                                    ? pgettext("Malkovich logs are only visible to your opponent after the game has ended", "Visible after the game")
```
anoek, line 271: I don't think we should churn this text for this as it's not displayed to the user ever - that's just a translator context string

anoek:
Have you tested this with a demo board (where player ids are undefined etc)?

Ultimately, what would be the best solution do you think? I feel like a notification, while probably better than a PM, might be pretty easy to miss too. I'm wondering if we need a "last read message" system so we can have an indication for ongoing games, an indication on the game page that a new unread message is available (a feature that has been requested in the past since it's easy to miss messages as you're tabbing through your correspondence games), and have that same system push up old finished games to a game history section or something, games that are over but have new chats or reviews on them, something like that.

I'm not sure, I feel like we should conceptualize how we would want this all to work in an ideal world, and then figure if that's something we can do in the nearish future or if we should put this bandaid on. (Problem with bandaids is they tend to stick around forever..)

src/components/PrivateChat/PrivateChat.tsx
GreenAsJade: Ah I see what you are asking now.  You are reading this the wrong way around: this is being added :)

src/views/Game/Chat.tsx
GreenAsJade: Ah that's fair.

GreenAsJade:
I think that a notification feels "right".  It's obviously not desparately important that these are seen right away... otherwise we'd have more complaints about it.   It's more like "unfortunate that this chat is not seen" - so if it's a notification that persists until the relevant game is visited, then at some point the person will see it and go there, or dismiss it.

I don't think a last message read system is warranted, because when a game is in progress you get to read the messages when you go to the game to play it - that's the time at which your mind is on that game.  Being "alerted" that someone said something in play would be tiring, I think.

I wasn't aware that a "has chat" indicator for a game has been requested  ... I guess that's more passive, and can make a lot of sense.  It'd be independent of "chat delivered to finished game" though, because you aren't looking at finished games in the same way that you are looking at ongoing ones.

src/components/PrivateChat/PrivateChat.tsx
anoek: lol.. teach me to review PR's first thing in the morning. Yes, this is good then!

GreenAsJade:
BTW, the reasons I think that putting it in PM is not best are:

 - It's not a message from the other person at all, it's a messsage from the system - feature mis-use
 - It doesn't go away when you visit the game
 - It appears in the PM box of the person who left the message as well

anoek:
Yeah I agree a notification would be better. Let me look at wiring that up, it's a little different than other notifications because we don't want 1 per message but just 1 per game until they visit it kind of thing.

GreenAsJade:
I think the ideal thing would be a list of games with outstanding chats in them - a non-zero list causing the dismissable notification to pop up the list of links.

If you were going to put effort in, you might well chose to detect the unread-post-game-finished chat on the serverside as well ... not sure if that would be better (OTOH in theory there should be nothing wrong with the way it is, after I test it on a demo board)

anoek:
Yeah indeed, I reckon on the server side would be better so it works with the mobile clients too

GreenAsJade:
Hah - so it's sounding like this has been a "flush out requirements" exercise, but code not required?

(For the record, Demo Boards would work because they can't get into the state "finished".  If somehow they did, then it would sent a chat/pm to an undefined player_id, which I imagine would at worst cost an error message in a log somewhere)

anoek:
> Hah - so it's sounding like this has been a "flush out requirements" exercise, but code not required? 

Yeah I think so :) I'm going to leave this PR open for a little bit all the same until I can get into the server code and make sure there's nothing blocking doing that without too much pain, pretty sure it'll be reasonably straight forward though.

PR #1481, opened 2021-05-15
Improve report-raising modal

GreenAsJade:
Fixes the incorrect instruction, and the amount of space for typing.

Tries, but fails, to fix the problem of closing prematurely when accidentally clicking outside.

## Proposed Changes

 - Use a textarea instead of an input
 - Say "clear" instead of "brief"
 - Specify closeOnClickOutside to be false ~~(done, but doesn't work)~~

Screen Shot 2021-05-15 at 12 32 27 pm


anoek:
Thanks for the improvement :)

PR #1484, opened 2021-05-16
Change the social login buttons' text from "Login..." to "Sign in..." to comply with Apple's Guideline

honganhkhoa:
This is to change the texts on the social login buttons from "Login with ..." to "Sign in with ...", since Apple just rejected my iOS app again due to the text on the AppleID login button. 😔

Screen Shot 2021-05-16 at 11 01 41

According to [Apple's guideline](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/), there are only three options: "Sign in with Apple", "Sign up with Apple", or "Continue with Apple".

## Proposed Changes

  - Change the texts on the social login buttons on Sign in and Register page from "Login with ..." to "Sign in with ...".

### DIFF
--- a/src/views/Register/Register.tsx
+++ b/src/views/Register/Register.tsx
@@ -144,11 +144,11 @@ export class Register extends React.PureComponent<{}, any> {
                     </form>
 
                     <div className="social-buttons">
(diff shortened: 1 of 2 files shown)


anoek:
Thanks for the patch!

PR #1485, opened 2021-05-16
Set win/loss counts to zero on special initial rating RankEntry

### LINKED ISSUE #1483: Total game count in Profile page is off by one
**Describe the bug**

The total game count reported over the pie chart in the Profile page is one less than it should be.

**To Reproduce**
Steps to reproduce the behaviour:

* Go to a player's Profile (ideally one with few games)
* Look at the total games counted over the pie chart.
 * See that it is one less than the ranked games they have played.

**Expected behavior**

Correct total displayed

**Additional context**

The problem is that the initial rating RatingEntry is being created on the back-end with the weak-wins value set to -1:

Screen Shot 2021-05-16 at 11 29 33 am

GreenAsJade:
Fixes #1483

## Proposed Changes

Set win/loss counts to zero on detection of special "initial rating" RatingEntry

### DIFF
--- a/src/components/RatingsChart/RatingEntry.ts
+++ b/src/components/RatingsChart/RatingEntry.ts
@@ -107,7 +107,14 @@ export function makeRatingEntry(d:any):RatingEntry {
         outcome = effective_outcome(opponent_rating, rating, handicap);
     }
 
-    return new RatingEntry({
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for the fix!

PR #1497, opened 2021-06-20
Add support for the move number suffix of game urls in the chat windo…

ajcsimons:
Add support for the move number suffix of game urls in the chat window linkification function, so that "https://online-go.com/game/10509/45" becomes that link with text "game 10509 move 45" rather than a link to just the game "game 1050945" with the move number not part of the link mashed on the end.

This is a postscript to https://github.com/online-go/online-go.com/pull/1215 which added support for these urls with move suffix, but had no update to the chat regex so the chat window mangles them.

### Fixes
Chat window mangling the game urls with move suffix

## Proposed Changes
Adds optional move number to the chat link replacing regex for game urls.


anoek:
Awesome, thanks for the patch!

PR #1503, opened 2021-07-06
Update GoResources.tsx

BHydden:
Remove IWTG - non-functional due to outdated Flash components


anoek:
Alas.

Thanks for the patch!

BHydden:
I know 😔 it was my favourite... Such elegant simplicity...

PR #1391, opened 2021-02-09
Accessibility theme

GreenAsJade:
Fixes the problem that we use very unfriendly colours for colourblind people.

## Proposed Changes

Add a theme and tweak the colours.

PR #1504, opened 2021-07-10
Restore the old indicator's color after changes in #1391

mmartin:
Restore the old color from https://github.com/online-go/online-go.com/pull/1391#issuecomment-870628520


anoek:
Indeed, thanks for the fix!

PR #1509, opened 2021-08-04
PM Reporting

### LINKED ISSUE #1507: Ability to report private messages
It would be great to have a report button in the PM window that provides moderators access to the conversation being reported.  

We currently have to rely on players to copy and paste the conversation or provide a screenshot, which opens the possibility of the reporter spoofing parts of the conversation or removing things.

RubyMineshaft:
Fixes #1507

## Proposed Changes

- I've added a report button to the top of the pm window
Screen Shot 2021-08-03 at 11 16 03 PM

- Clicking the report button will open this report modal (There's a typo in "moderation" in this screenshot, but it's been corrected)
Screen Shot 2021-08-03 at 10 54 43 PM

- PM reports will include a link to View Reported Conversation
Screen Shot 2021-08-03 at 10 55 03 PM

- Which in turn opens a modal containing the chat
Screen Shot 2021-08-03 at 10 55 13 PM
Screen Shot 2021-08-03 at 10 55 29 PM

### DIFF
--- a/src/components/IncidentReportTracker/IncidentReportTracker.styl
+++ b/src/components/IncidentReportTracker/IncidentReportTracker.styl
@@ -63,6 +63,15 @@
         padding-top: 0.7rem;
     }
 
+    #conversation {
+        font-weight: bold;
+
+        &:hover{
+            cursor: pointer;
+            text-decoration: underline;
+        }
+    }
+
     .incident {
         padding-bottom: 1.0rem;
         margin-bottom: 1rem;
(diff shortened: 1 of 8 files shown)


anoek:
That was fast, nice job!

GreenAsJade:
I wonder why you would not be able to do this for a moderator?

Up till now we have allowed reporting of moderators, and have investigated those reports diligently...

BHydden:
I assumed that meant moderators couldn't use the button (since we already have a make note from message button) rather than moderators couldn't be reported... But I'll admit i didn't actually look at how the code was implemented.

RubyMineshaft:
> I assumed that meant moderators couldn't use the button (since we already have a make note from message button) rather than moderators couldn't be reported... But I'll admit i didn't actually look at how the code was implemented.

This.  Sorry, I should have written that a bit more clearly.  I just took the button off of the moderators' side because the buttons were too crowded and we don't really need it.

PR #1510, opened 2021-08-04
increase reported conversation modal width for smaller screens

RubyMineshaft:
## Proposed Changes

I adjusted the modal width and amount of padding for smaller screens to make reading more comfortable.


anoek:
Thanks for the fix!

PR #1511, opened 2021-08-06
Plot rating by game

GreenAsJade:
## Proposed Changes

Add a toggle that allows the rating graph to be plotted by game.

When selected, 
 - the win-loss bars are removed (because they don't make sense without timeline granularity)
 - the x axis becomes "games played"
 - the pie chart shows the date for the games selected by the lower sub-selection area 
 
Screen Shot 2021-08-06 at 10 49 44 pm (3)

### DIFF
--- a/src/components/MiniGoban/MiniGoban.styl
+++ b/src/components/MiniGoban/MiniGoban.styl
@@ -17,6 +17,7 @@
 
 .MiniGoban {
     display: inline-flex;
+    flex-direction: column;
     width: 22rem;
     height: 25rem;
     max-width: 100vw;
@@ -31,6 +32,17 @@
 
     &.link {
         cursor: pointer;
+    }
+
+    .minigoban-title {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        margin-bottom: 0.5rem;
+
+        .game-date {
+            font-size: smaller;
+        }
     }
 
     .inner-container {
--- a/src/components/MiniGoban/MiniGoban.tsx
+++ b/src/components/MiniGoban/MiniGoban.tsx
@@ -108,7 +108,6 @@ export class MiniGoban extends React.Component<MiniGobanProps, any> {
     }
 
     sync_state() {
-        console.log(this.goban.engine);
         const score = this.goban.engine.computeScore(true);
         const black = this.props.black || "";
         const white = this.props.white || "";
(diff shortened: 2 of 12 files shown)


anoek:
Nice feature :)

GreenAsJade:
Note: I didn't have any data in Beta to try an account with seriously large numbers of games.

anoek:
This is working great, my only note is the placement of the toggle leaves a lot of white space to the right, so I'm wondering if there would be a better place for it? I'm not sure, but what about in the upper right of the graph or something?

image

BHydden:
> leaves a lot of white space to the right

Sure would be a shame to have any horizontal whitespace 😏 

image

anoek:
Hey now ;)

GreenAsJade:
I like your idea where to put it anoek.  So far I have been daunted about how to place other elements over the top of the graph.

However, I'm about to tackle that for the  [SICK feature](https://forums.online-go.com/t/add-game-level-curve-besides-time-level-graph/38221/16?u=greenasjade), so maybe I will then know how to put the toggle there.

anoek:
Yeah that's going to be a little messy since it's a d3 thing...

I think the way I'd try is to add an element in `RatingsChart` that is positioned relatively to `ratings-graph`, and in `onResize` and `initialize` set the x/y for that element manually, those methods both compute the width of that graph, and the y should be static, so it should be relatively easy to do there. Do that, add some padding to the text, and I think it shouldn't be that bad in the end, albeit a little messy having to manually position things.. but at least it's in the midst of code where we do that already for the d3 chart.

GreenAsJade:
OK - a moderate amount of change in various places in that commit ... let me know...

GreenAsJade:
Overview of what's there:

https://user-images.githubusercontent.com/61894/129358925-cf35f2ed-765c-48c7-a38a-b7211433bb9b.mp4

anoek:
Very cool, that inline game when you click on the chart is hot :) Great job, I'll look over and merge in a little bit here.

BHydden:
Now that you've added on click game previews, people will want them for game history and game links in chat too 😉

GreenAsJade:
You don't even have to click, you just hover ;)

(Phone people might say "hey we can't hover!".   But I think the phone form factor doesn't support the pie either.)

I did think about game history, and game links in chat is interesting.

The biggest problem is placement of the preview.

True story: I found the current placement next to the rating chart by fluke!   I had been dreading trying to make it work placing it right by the cursor, because this is full of edge cases, and for some reason I just typed "top: 0 right: 0" as the first stab before making it float.    I had meant left: 0,  it was a typo !      Then the preview popped over the pie chart and I thought "Oh, cool!!"   :D 🎉

GreenAsJade:
(That said, it may be that floating it by the cursor is the best overall idea, especially if it is going to appear in more places ... we can cross that bridge at that time)

BHydden:
I like where it is for the profile page, but yeah cursor float is probably best for in chat ❤️

PR #1514, opened 2021-08-15
Various fixes for game popup in rating chart

GreenAsJade:
Fixes:
 - Lingering PieChart artifacts on resize (corner cases)
 - Display of game hover popup when pie chart won't fit (don't do that)
 - Overlapping textual win-loss information in by-game chart in phone form factor (don't display this info, for now)
 - Display of meaningless points when game is finished (don't do that)
 - Display result of game that is finished in the title (do do that)
 
## Proposed Changes

Got rid of renderWinLostResultsAsText in the by-game component, for now.   This is actually buggy in the original chart component: the render of this extends below the extent of the element (it looks like vertical clip was turned off to allow this). but due to fortunate spacing in the rating-by-time case, it looks almost OK.

I couldn't quickly work out how to solve this properly - the whole "calculation of the height of various elements" needs a big tidyup to be comprehendable, then this fix would probably be easy.  In the mean time, simply not showing it works.

The turning off and on of pie chart had some buggy cases during resize even in the original - that should be fixed now (I had thought that pushing it off to the right offscreen was enough, but it turns out sometimes it can sneak onscreen in thin form factors, so now it is completely un-rendered)

### DIFF
--- a/src/components/MiniGoban/MiniGoban.styl
+++ b/src/components/MiniGoban/MiniGoban.styl
@@ -40,7 +40,7 @@
         align-items: center;
         margin-bottom: 0.5rem;
 
-        .game-date {
+        .game-date, .game-result {
             font-size: smaller;
         }
     }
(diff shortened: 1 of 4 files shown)


anoek:
Thanks for the fixes!

PR #1517, opened 2021-08-19
OJE Throbber Spinner Light theme and Go Resources IGA

shinuito:
## Fixes 
In light theme while the OJE, joseki explorer is loading the throbber spinner is white like the background, hence is invisible, unlike in dark theme. Changed the background of spinner to use themed(background,shade0), which at least has some contrast to the theme background.

## Proposed Changes

  - Add a link to the Irish Go Association in the Other Go resources page, with Ireland flag beside the link.

### Comment/Question

I'm not sure why the yarn version is different from the main repository in package.json and yarn.lock. I'm using npm 6.14.14, node.js v14.17.5 . Is there another version I should use instead? The cache loader line just seems to have been moved automatically (alphabetically), but not changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@types/jest": "^26.0.15",
     "autoprefixer": "^10.2.5",
     "body-parser": "^1.16.1",
+    "cache-loader": "^4.1.0",
     "cssnano": "^5.0.2",
     "express": "^4.17.1",
     "express-http-proxy": "^1.6.0",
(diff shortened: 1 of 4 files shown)


anoek:
Thanks for the fix and the addition!

GreenAsJade:
@anoek  What was the deal with the changes to yarn.lock?   I saw you committed them ... curious what causes that to happen?

anoek:
Looks like just a version bump for yarn itself, pretty benign.

GreenAsJade:
Cool

PR #1522, opened 2021-08-27
Show date in private message chat

andrewShinjo:
Fixes #  1188 Added date in private message chat for first message of the day
image


anoek:
Thanks for the improvement!

PR #1523, opened 2021-09-04
Fix OJE caching root

GreenAsJade:
Fixes the problem where when you press double back arrow to get to the empty board, it re-fetches the result, even though we cached it.   That's because double back arrow requests "root" (because we don't know the empty board node-id) but we cached it by the resulting node-id.

## Proposed Changes

Cache all immediate fetches immediately, by the requested node-id (instead of by fetched-node-id via the pre-fetch result that comes next), ensuring there is an entry for "root".


anoek:
Thanks for the fix!

PR #1524, opened 2021-09-04
Optimize OJE prefetch

GreenAsJade:
Fixes some lagginess

## Proposed Changes

 - Keep track of what nodes have already been prefetched, don't do them again
 - Don't fire a second prefetch if one is in flight
 - Ensure that browser prioritisation funkiness doesn't result in prefetch hitting server before main fetch


anoek:
Nice, thanks for the optimization!

GreenAsJade:
(I'll take the logging out later - will be useful for profiling for now)

PR #1526, opened 2021-09-12
Add new mods

RubyMineshaft:
Added in the new mods and removed the hardcoded usernames since they are sometimes showing old usernames


BHydden:
Fight me 🤣

anoek:
Thanks for the PR!

PR #1527, opened 2021-09-14
update moderator language flags

RubyMineshaft:
Changed some language flags as requested in https://forums.online-go.com/t/language-skills-of-current-mod-team/39054/

PR #1528, opened 2021-09-19
Allow England flag as an alternative to GB

BHydden:
I don't have a testing environment set up so just hoping this is as anoek described in https://github.com/online-go/online-go.com/pull/1527#issuecomment-922269060

Fixes English flag becoming user's country / American flag


anoek:
Thanks!

PR #1531, opened 2021-09-25
in Tournament.tsx, handle the case where minimum and maximum players are equal, instead of showing it incorrectly as e.g. "10+"

pdg137:
This fixes a problem with display of tournaments that have equal numbers for minimum and maximum players , such as all the "start when full" tournaments in the [New Zealand Tournaments](https://online-go.com/group/3207) group. Here's an example of such a tournament on the beta site:

https://beta.online-go.com/tournament/138

Currently it shows up like this:

image

This change removes the "+" in the case where min and max are equal.

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -1400,7 +1400,13 @@ export class Tournament extends React.PureComponent<TournamentProperties, any> {
                                 {!editing
                                     ? <span>
                                          {tournament.players_start}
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for the fix :) I think the build error is a simple syntax error, using `{}` instead of `()`

pdg137:
Sorry, not sure why that worked for me before! Changed the `{}` to `()`, does that look better now?

anoek:
Thanks for the patch!

Issue #1536, opened 2021-10-18
Consider moving to ESLint instead of TSLint


benjaminpjones:
TSLint is [deprecated](https://github.com/palantir/tslint/issues/4534) in favor of ESLint.  Is there any interest in adding ESLint to the repo with the ultimate goal of removing TSLint?

I have an `.eslintrc.js` drafted that should be pretty close in spirit to the `tslint.json`: https://github.com/benjaminpjones/online-go.com/blob/eslint/.eslintrc.js

If there is interest in migrating, I can make a pull request.  After that, here are the steps that would need to be taken:

- [ ] Create roughly equivalent eslint config [DRAFT CREATED]
- [ ] Optionally import `eslint:recommended` for additional eslint rules.  `recommended` mostly flags potential problems rather than stylistic issues, see [list of available rules](https://eslint.org/docs/rules/) to see which are included.
- [ ] Remove tslint from package/CI and remove `tslint:disable` comments

anoek:
Not sure when you want to do it, but at some point during this if you could update/replace the gulp tasks associated with tslint with the eslint equivalent, that'd be swell :)

benjaminpjones:
Thanks for pointing that out - I'll add it to the list!

PR #1537, opened 2021-10-21
Eslint

benjaminpjones:
Addresses the first task in #1536

## Proposed Changes

  - Add eslint config that is as close to the tslint config as possible
  - Add `eslint-disable` comments for (a) errors that were caught by eslint, but not tslint, and (b) errors already marked with `tslint:disable`
  - Fix lint errors in tests (which were not being linted before.)
  - Add eslint to the GitHub workflow

## Changes _not_ done in this PR

  - I have not added either of the `recommended` rulesets
  - I have not run `lint --fix` on the code, or modified the application code beyond `eslint-disable` comments


src/lib/data.ts
```diff
@@ -217,6 +217,10 @@ export function watch(key, cb, call_on_undefined?: boolean, dont_call_immediatel
     event_emitter.on(key, cb);
 
     let val = get(key);
+
+    // The != can possibly be changed to !==, but I don't want to touch it
+    // without further investigation.
+    // eslint-disable-next-line eqeqeq
```
anoek, line 223: Yeah can and absolutely should

PR #1540, opened 2021-10-24
Notify of late chat on games

GreenAsJade:
Register a  notification from the server about late chat being received.

Of note

 - I hope I put Gulpfile back into a workable state ... that's worth checking
 - I'm not sure what emitNotification does ... it seems to be the pattern for handling notifications, but I couldn't see where the strings that I supplied (title and body) are used.
 - I tried to dig into whether this should (or does) generate an email, and whether that should be configurable, but found that hard to follow.  The translated strings at the bottom of Settings.tsx are very mysterious...

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -1,7 +1,7 @@
 'use strict';
 
 // Which backend server would you like to use today? ...
-let BACKEND = 'BETA';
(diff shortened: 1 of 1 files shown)


BHydden:
KoBa just reported that this notification is still triggered when shadowbanned users "comment" on games.

PR #1541, opened 2021-10-25
Eslint recommended

benjaminpjones:
Addresses task in #1536 

## Proposed Changes

  - `.eslintrc` to extend from `eslint:recommended` and `@typescript-eslint/recommended`
  - `eslint --fix`
    - Note: I actually ran this two times: once with only `prefer-const` turned on and a second time with the rest of the rules enabled.  The second run is arguably more interesting (and potentially more controversial). See commits 36d1a23 and 8fc7498.
    - Rules that appear to have been enforced: `prefer-const`, `no-array-constructor`, `no-extra-semi`, `no-inferrable-types`, `prefer-as-const`
    - Had to manually fix one instance of alignment getting wonky.
  - Suppress rules that still don't pass.
    - While I think it would be a great idea to fix and turn these on, I don't think this is within scope of the current project.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,6 @@ module.exports = {
         "no-prototype-builtins": "off", // 8 errors
         "no-self-assign": "off", // 2 errors
         "no-useless-escape": "off", // 24 errors
-        "prefer-const": "off", // 6 errors
         "prefer-spread": "off", // 7 errors
 
         //==============================================================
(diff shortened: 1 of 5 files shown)


benjaminpjones:
Okay I re-did the `--fix` run with a more permissive `no-inferrable-types` [3a4a6d2, 1fe1bc3].

In addition, I went ahead and fixed/silenced the remaining 6 `prefer-const` errors so that I could turn that rule on.  [e786278]

PR #1544, opened 2021-11-09
Fixed AI review chart's X axis when interactive variation goes past game's move count.

mmartin:
When the interactive AI review goes past the game's move count it gets cut off in the chart.

Example: game has 94 moves, created interactive AI variation from move 92 to 101.

Before:
image

After:
image

### DIFF
--- a/src/views/Game/AIReviewChart.tsx
+++ b/src/views/Game/AIReviewChart.tsx
@@ -143,12 +143,6 @@ export class AIReviewChart extends React.Component<AIReviewChartProperties, any>
 
 
         this.x_axis = this.prediction_graph.append("g");
-        this.x_axis
-            .attr("transform", "translate(0," + this.height + ")")
-            .call(d3.axisBottom(this.x))
-            .select(".domain")
-            .remove();
-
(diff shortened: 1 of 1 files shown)


mmartin:
Found the issue – not really related to clicking quickly and not caused by my patch. :-)

Line 311, array out of bounds:
https://github.com/online-go/online-go.com/blob/53c116ca2320c2bd43b78e2c023bfd1fda88293a/src/views/Game/AIReviewChart.tsx#L308-L315

I don't think we need any logic there at all, just use `x.score`, right?

anoek:
I'm inclined to agree looking at it now. I'm not sure what my reasoning was in attempting to fallback to `ai_review.scores` values. 

Thanks for finding my bug :)

mmartin:
Copy & paste from a few lines up, I guess?
https://github.com/online-go/online-go.com/blob/53c116ca2320c2bd43b78e2c023bfd1fda88293a/src/views/Game/AIReviewChart.tsx#L273-L280

Not sure why it's there, too.

anoek:
Thanks @mmartin !

PR #1548, opened 2021-11-20
Add types to swal

benjaminpjones:
## Proposed Changes

  - add `sweetalert2` to package
  - `import swal from 'sweetalert2'` (instead of `declare let swal`)
  - add sweetalert to webpack externals (so that the bundle size doesn't increase)
  - Add a bunch of `.catch(swal.noop)` to suppress no-floating-promises errors as well as swal console errors
  - remove `@types/sweetalert` from package.

## Notes

 - The version of `sweetalert2` I added to the package (v6.11.5) differs from the one that is used in [index.html](https://github.com/online-go/online-go.com/blob/5e0284ee37b1c8aebe2d513614c5dc5de84e02cb/src/index.html#L165) (v6.0.1).  This is because 6.0.1 does not have types, which was the purpose of this PR.  6.2.4 is the version that added types  I don't think it's necessary to bump the version that we get from the CDN, but it may be preferable.
 - If I understand correctly, webpack will leave out sweetalert2 from the bundle, but rather expect it to be present in the environment at runtime.  Triggering swal boxes seems to work in my local environment, but hard to be sure that means it works.  I can confirm in beta if this PR gets merged.


src/views/TournamentRecord/TournamentRecord.tsx
```diff
@@ -238,7 +238,7 @@ export class TournamentRecord extends React.PureComponent<TournamentRecordProper
             })
             .catch(errorAlerter);
         })
-        .cach(ignore);
```
benjaminpjones, line 241: This is the type of error that I hope this PR helps with :)

anoek:
Oh nice, yeah v6.0 was just whatever was the latest and greatest at the time and I haven't looked at it since. Will be nice to have that packaged.

v6.11 sounds great. I'd also be open to upgrading past it to whatever makes sense, it looks like they're all the way up to v11. I'm not sure what breaking changes there were along the way, or what nice features might have been added, but if it makes sense to jump some versions I'm fine with that too.

anoek:
Looks like they have some proper react integration too these days: https://github.com/sweetalert2/sweetalert2-react-content

Not suggesting we do anything with that in this PR, but as an aside note

benjaminpjones:
Yeah bumping to a newer version sounds cool.  I think the big breaking change there is they changed `swal()` to `swal.fire()` or something like that.  I wonder if that means we could put it in the bundle without too much bloat?

The other nice thing is that `catch(swal.noop)` isn't strictly necessary after 7.0 since they changed how rejections work.  But we'd still run into no-floating-promises so i dunno.

anoek:
Thanks @benjaminpjones !

PR #1552, opened 2021-11-26
Allow longer live games using Fischer time settings

### LINKED ISSUE #839: Increase maximum time in live fisher games
anoeks notes: The maximum time in fisher games is too low for EGF class A tournaments as per  http://europeangodatabase.eu/EGD/EGF_rating_system.php - we need to increase the maximum time for fischer games and look at increasing limits on the other timing systems as well.

Original issue body follows

---- 

### Summary

Currently, Fischer timing has a mandatory cap on OGS.  For live games, this is superfluous at best, but if it does have an effect, it is annoying.  Please add the option to select a Max Time of “none”, at least for non-correspondence games.

Related to this: please add a hint about the estimated game time, e. g. by appending (1:30 h/300 moves) after a 20/10 game, to both the Custom Game dialogue and the open games list.

### Environment

any

### Steps To Reproduce:

1. Go to https://online-go.com/play
2. Click “Custom Game”
3. Select Fischer timing, 1 hour basic time plus 20 sec per move
4. You are forced to select a Max Time of 1 hour
5. If you play a game like this and play your first move in less than 20 seconds, you do not get the normal 20 seconds bonus but less.
6. During the first few moves at least, you will run into this several times.  That is: you are being punished for playing too fast.

### Additional Information:

There was a discussion in the forum where people expressed the fear that games without a cap might take indefinitely long.  This is not the case, of course, because a game has a finite amount of moves and thus a finite amount of added time.

In fact, you can give a quite good upper bound for the duration of a game by adding the bonus time for 300 moves to the basic time of both players.  Since 300 seconds are 5 minutes, you can even do this in your head: 2×basic (in min) + 5×bonus (in sec) = estimated max game duration (in min).  Example: a 20/10 game would never take much longer than 90 minutes (1:30 h).  This is the second part of the request: add this estimate as a hint.

JonKo314:
Fixes #839

## Proposed Changes

  - Allow *initial time* of up to 4 hours instead of up to 1 hour
  - Allow *max time* of up to 4 hours or 28 days instead of up to 1 hour

## Notes
The lack of longer time settings where mentioned by European professional Ali Jabarin as an "OGS bug" during a European Go Championship stream. He thought players should gain time that exceeds the 1 hour *initial time*, which was not possible because the highest *max time* was (prior to this change) also 1 hour. It sounded like he was expecting no max time at all. @BHydden considered this also as "real Fischer" on the OGS forum.

I've chosen 28 days as highest *max time*, because I think (for live games) it is close enough to infinite time or no max time and can still nicely be represented in the select list without any further changes.

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -120,9 +120,9 @@ export const time_options = {
     },
     "live": {
         "fischer": {
-            "initial_time": gen(30, 3600),
+            "initial_time": gen(30, 3600 * 4),
             "time_increment": gen(10, 1800),
-            "max_time": gen(30, 3600),
+            "max_time": gen(30, 3600 * 4).concat(gen(86400 * 28, 86400 * 28)),
         },
         "simple": {
             "per_move": gen(10, 3600),
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -122,7 +122,7 @@ export const time_options = {
         "fischer": {
             "initial_time": gen(30, 3600 * 4),
             "time_increment": gen(10, 1800),
-            "max_time": gen(30, 3600 * 4).concat(gen(86400 * 28, 86400 * 28)),
+            "max_time": gen(30, 3600 * 4),
         },
         "simple": {
             "per_move": gen(10, 3600),


anoek:
28 days seems a bit excessive for a live game?

The point of having a limit is to help ensure the game doesn't turn into a multi-day correspondence style game, we have correspondence games for that. We also have pause for players if they want to go sleep and resume another day. Ideally we'd have a sealed move system for that, but I don't think it'd see that much use in practice.

Re https://europeangodatabase.eu/EGD/EGF_rating_system.php it looks like the EGF time settings don't really need crazy amounts of initial or max time, I think the main thing is raising the max time a bit so they can build up past 1 hour I reckon. I suspect 2 hours would be sufficient, however I don't see the harm in going all the way up to 4 hours max time, that should be more than enough without devolving into a correspondence game.

And I suppose I'm fine with allowing up to 4 hours for initial time too since Byo-Yomi/Canadian support that.

JonKo314:
I added the 28 days *max time* to allow for a *practically no max time* option. I guess 1 day would also work in practice. The EGF doesn't mention any max time on that site you linked, so I think it would be good to allow that as an option.

anoek:
Even a day seems a bit questionable for a live game, we have +30m increments so it could get out of hand pretty quickly I fear.

JonKo314:
Yes, but to me the +30min increment seems more excessive than a 1 day max time. For now probably even two hours max would be enough, but I'd rather give more options now then come back at a later point to add them. What do you think is okay for tournaments? Six hours, four, or three? I wouldn't go lower than that.

BHydden:
Yes to me as i understood it the main request is the ability to bank time with her early moves rather than having all of those get cut off at the cap. I suspect 4 hours should be sufficient depending on the chosen increment.

anoek:
Thanks @JonKo314 !

Issue #1362, opened 2021-01-22
Game results are not clear


Clossius:
**Is your feature request related to a problem? Please describe.**
When I'm in a review and I click game information it tells me the result type like res or +2.5 but it doesn't clearly state who won which is annoying in close games when I'm looking at a demo. I don't always know the result when I'm looking at reviews.

**Describe the solution you'd like**
I would like to see which color won in the result.
Like: w+res or b+2.5

benjaminpjones:
## Good First Issue

### Code Points

- [`GameInfoModal.render`](https://github.com/online-go/online-go.com/blob/259b6ae7e7584a9b4f7d938c9a7a645d3aab9beb/src/views/Game/GameInfoModal.tsx#L229-L233): this is where the game result is rendered.

### Prerequisites

- Some familiarity with [React](https://reactjs.org/) and [JSX](https://reactjs.org/docs/introducing-jsx.html).

### Additional info

- This issue is a little tricky because [`config.outcome`](https://github.com/online-go/goban/blob/840a03f7385497270179f3544ec8b23fd227e31b/src/GoEngine.ts#L123) is just a string.  Usually it holds a "2.5 points" or "Resignation", something like that, but in the case of demos, a user can actually enter a custom string.
- Additionally, there is the question of where to get the winning color.  Here is one possibility: [`config.winner`](https://github.com/online-go/goban/blob/840a03f7385497270179f3544ec8b23fd227e31b/src/GoEngine.ts#L124) is populated with a user ID for finished games.  This can be compared to [`config.white_player_id`](https://github.com/online-go/goban/blob/840a03f7385497270179f3544ec8b23fd227e31b/src/GoEngine.ts#L177) or [`config.black_player_id`](https://github.com/online-go/goban/blob/840a03f7385497270179f3544ec8b23fd227e31b/src/GoEngine.ts#L176) to determine the winner.
- Be sure to internationalize the result.  See #1586 for an example of how to do this.

maurelio1234:
Hi, I would like to work on this issue.
since it's the first time I try to contribute I just created a PR with a first PoC of a fix and some questions to gather some feedback.
Tell me what you think and how we could progress on this matter.

Thanks in advance,

PR #1555, opened 2021-11-27
Cleanup: GameInfoModal

benjaminpjones:
This should help with #1362 because the info that is needed is contained in config, for which this PR adds types.

Depends on https://github.com/online-go/goban/pull/45 (expect CI to fail)

## Proposed Changes

  - Eliminate `any` (implicit and explicit) from GameInfoModal.tsx
  - Move `yesno` to misc.ts
  - Remove unused imports and variables


anoek:
Thanks for the cleanup!

benjaminpjones:
Thanks for all the reviews @anoek!

PR #1557, opened 2021-12-02
GobanThemePicker: Cleanup & tooltips

benjaminpjones:
I think I was digging around `GobanCore`, and came here because I wanted to see how `preRender[white|black]` and `place[White|Black]Stones` were used externally.  Anyway, it always helps me to figure out what types are floating around..

## Proposed Changes

- Removed unused stuff (unused variables, imports, commented code)
- Added types to everything
- Added tooltip (screenshot below)
  
Screen Shot 2021-12-01 at 10 26 03 PM

If the tooltips are disagreeable, I can take that out. But the theme names are already translated, so I figured why not...


anoek:
Hooray for typings!

PR #1561, opened 2021-12-05
Some dev improvements

benjaminpjones:
Fixes some eslint-related issues that @GreenAsJade found.

## Proposed Changes

  - Add `'include': ['src']` to tsconfig.  This fixes the issue where new files in `src` would trigger lint errors before they were included elsewhere.
  - Run eslint once at start of gulp script.  Turns out `gulp.watch` has a setting for that!

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -75,7 +75,7 @@ function livereload_server(done) {
     done();
 }
 function watch_eslint(done) {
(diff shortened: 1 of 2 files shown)


anoek:
That's awesome, I wish I knew about that `ignoreInitial` thing, I'm going to have to trash half my rules now that do just that :sweat_smile:

PR #1564, opened 2021-12-08
data.get: deduce type from default_value

benjaminpjones:
This PR makes it so that the output of `data.get` is typed if a default value is provided.  For example:

```
// some_bool is now boolean instead of any
const some_bool = data.get("bool-prop", false);
```

## Proposed Changes

  - type data.get
  - type the automatch defaults in AutomatchSettings
      - This was prompted by the update to `data.get`.

## A motivating example

This is the code that prompted this change. Before, `challenge` was typed as `any`.  Now, it's pretty accurately typed, without having touched that area of code.

Screen Shot 2021-12-08 at 2 57 28 AM


anoek:
Awesome thanks for the typing :)

PR #1565, opened 2021-12-09
Make sure we come back to a known node after clicking unknown nodes

GreenAsJade:
Fixes issue returning to known nodes after clicking unknown ones, introduced by earlier fix to edit-mode description reset.

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -184,6 +184,7 @@ export class Joseki extends React.Component<JosekiProps, any> {
         this.state = {
             move_string: "",             // This is used for making sure we know what the current move is. It is the display value also.
             current_node_id: undefined as string,    // The server's ID for this node, so we can uniquely identify it and create our own route for it,
+            most_recent_known_node: undefined as string,  // the value of current_node_id when the person clicked on a node not in the db
             position_description: "",
             variation_label: '_',
             current_move_category: "",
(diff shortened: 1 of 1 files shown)


src/views/Joseki/Joseki.tsx
```diff
@@ -184,6 +184,7 @@ export class Joseki extends React.Component<JosekiProps, any> {
         this.state = {
             move_string: "",             // This is used for making sure we know what the current move is. It is the display value also.
             current_node_id: undefined as string,    // The server's ID for this node, so we can uniquely identify it and create our own route for it,
+            most_recent_known_node: undefined as string,  // the value of current_node_id when the person clicked on a node not in the db
```
benjaminpjones, line 187: Just a note: all these `as`s type annotations are getting lost because state, and therefore its members, is typed as `any`:

Screen Shot 2021-12-09 at 6 55 21 PM

If you want these annotations to be used by the compiler (and IDE tooling), the way to do that is to make an interface and pass it as a template arg to `React.Component`:

```
Interface JosekiState {
    current_move_id?: string;
    most_recent_known_node?: string;
    ...
}

React.Component<JosekiProps, JosekiState> {
    ...
```

GreenAsJade: I so totally have to start doing that :D

GreenAsJade: (although there's plenty of benefit in just recording that I intended it as a string :D :D )

PR #1566, opened 2021-12-10
hotfix

benjaminpjones:
Hi @anoek,

Maybe you're able to revert the build for a bit while I test this, but if not, I think this PR *should* fix https://forums.online-go.com/t/leela-ai-only-shows-nan-at-move-suggestions/40220/6

At least it won't be creating a scores array where there wasn't.


anoek:
Thanks for the hot fix!

benjaminpjones:
Not a problem! I checked a few of the games posted by @yewang this morning, and seems to be working as expected from what I can tell:

Screen Shot 2021-12-10 at 10 14 36 AM

PR #1568, opened 2021-12-13
Add State Types to all React.Components

benjaminpjones:
Pretty much all of these were marked as `any`.  Note, this is only `React.Component`s, there are plenty of `React.PureComponent`s that can possibly be updated in the same way.  I did not make logic changes in this PR, but left some comments about code that looks either unnecessary or mis-annotated and could be updated in the future.

## Proposed Changes

Add a `*State` interface for every `React.Component` that had state typed as `any`. In some cases, I just deleted `any` from the template arguments, since the default is `{}` and that is more appropriate for components that do not update state:

```diff
-class SomeComponent extends React.Component<SomeProps, any> {
+class SomeComponent extends React.Component<SomeProps, SomeState> {
    // this.state is now a SomeState

-class SomeComponent extends React.Component<SomeProps, any> {
+class SomeComponent extends React.Component<SomeProps> {
    // this.state is now a {}
```

### DIFF
--- a/src/components/GobanLineSummary/GobanLineSummary.tsx
+++ b/src/components/GobanLineSummary/GobanLineSummary.tsx
@@ -34,6 +34,9 @@ interface GobanLineSummaryProps {
     width?: number;
     height?: number;
 }
+
+// This state is very similar to MiniGobanState.
+// TODO (ben): Possibly pull shared members into a common type.
 interface GobanLineSummaryState {
     black_score: string;
     white_score: string;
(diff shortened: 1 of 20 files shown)


src/components/GobanLineSummary/GobanLineSummary.tsx
```diff
@@ -75,7 +100,13 @@ export class GobanLineSummary extends React.Component<GobanLineSummaryProps, any
             this.sync_state();
         });
 
-        this.goban.on("pause-text", (new_text) => this.setState({
+        // It appears "pause-text" is not part of the Goban event emitter.
```
anoek, line 103: I'm 99% sure this pause text is not used anymore

anoek:
This is awesome, thanks for the typing work!

src/components/GobanLineSummary/GobanLineSummary.tsx
benjaminpjones: Cool cool I can remove it in a follow up PR :)

Issue #1316, opened 2020-12-11
Full AI review API returns branches without moves


mmartin:
**Describe the bug**
Sometimes (usually at the end during passing) the full AI review API returns branches without moves, e.g.
```
"move-244": {
...
    "branches": [
...
        {
            "moves": [],
            "win_rate": 0.9987,
            "score": 8.52,
            "visits": 138,
            "lcb": 0.9976,
            "score_mean": 8.52,
            "score_stdev": 1.5283,
            "utility": 1.0349,
            "utility_lcb": 1.0271
        },
...
```

It breaks AI review if it is one of the first 6 branches:
https://github.com/online-go/online-go.com/blob/a90e0bcba6237554fda0fd7d3d46121d551a26c2/src/views/Game/AIReview.tsx#L391

Exception is thrown on line 432 (since `mv` is `undefined`):
https://github.com/online-go/online-go.com/blob/a90e0bcba6237554fda0fd7d3d46121d551a26c2/src/views/Game/AIReview.tsx#L428-L434

I guess, one of the solutions could be to replace line 391 with –
```ts
let branches = ai_review_move.branches.filter(branch => branch.moves.length > 0).slice(0, 6);
```

anoek:
Can you provide a link to a game with this bug by chance? You can send it to anoek@online-go.com if you don't want it here for whatever reason. Thanks!

anoek:
This was fixed in #1569

PR #1569, opened 2021-12-13
Ai summary table

shinuito:
Fixes:
When a game ends with consecutive passes, there's a console error which results from empty branches in the AI review having their (non existent) move coordinates checked. So added in a mv === undefined check.

## Proposed Changes
Adding in a new AISummaryTable component to the AIReview which tries to summarise the data according to score loss, and categorise the types of moves as Excellent = blue move, Great = one of the green moves, Good = some positive score change move or mistake <1 point, Inaccuracy <2 point mistake, Mistake <5 points loss, Blunder >=5 point loss.

  - It currently only shows for katago reviews (full or fast, but fast omits the Excellent and Great categories).
  - It currently only shows for moderators, in order to test and tune before showing it to everybody.
  - It should work on finished games and uploaded games, and now forked games.
  - Added a preference to remember if the user wants the table hidden or not and a toggle for that.

There's still an issue where if the user opens a game page and then navigates away quickly, there's a setState error ("Can't perform a React state update on an unmounted component."). A similar issues happens with PaginatedTable on profile pages (for moderators) and for the Tournaments list pages, but I don't know how to fix these issues yet.

### DIFF
--- a/src/views/Game/AIReview.tsx
+++ b/src/views/Game/AIReview.tsx
@@ -728,12 +728,16 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
 
         if (this.ai_review?.type === "fast" ){
             const scores = this.ai_review?.scores;
+            const is_uploaded = (this.props.game.goban.config.original_sgf !== undefined);
+            //one more ai review point than moves in the game, since initial board gets a score.
+            const check1 = !is_uploaded &&
+            (this.props.game.goban.config.moves.length !== (this.ai_review?.scores.length - 1));
+            // extra initial ! in all_moves which matches extra empty board score, except in handicap games for some reason.
+            // so subtract 1 if black goes second == bplayer
+            const check2 = is_uploaded &&
+            ((this.props.game.goban.config["all_moves"].split("!").length - bplayer) !== this.ai_review?.scores.length);
 
-            if (scores === undefined || ((this.props.game.goban.config.original_sgf === undefined) &&
-            (this.props.game.goban.config.moves.length !== (this.ai_review?.scores.length - 1))) ||
-            ((this.props.game.goban.config.original_sgf !== undefined) &&
-            ((this.props.game.goban.config["all_moves"].split("!").length - 1) !== this.ai_review?.scores.length))
-            ){
+            if (scores === undefined || check1 || check2){
                 return {"ai_table_rows" : default_table_rows, avg_score_loss};
             }
 
@@ -796,11 +800,16 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
         } else if (this.ai_review?.type === "full" ) {
             const num_rows = ai_table_rows.length;
             const movekeys = Object.keys( this.ai_review?.moves);
-            const checkmoves = ((this.props.game.goban.config.original_sgf === undefined) &&
-            (this.props.game.goban.config.moves.length !== (movekeys.length - 1))) ||
-            ((this.props.game.goban.config.original_sgf !== undefined) &&
-            (this.props.game.goban.config["all_moves"].split("!").length !== (movekeys.length - 1)));
-            if (this.state.loading || checkmoves ){
+            const is_uploaded = (this.props.game.goban.config.original_sgf !== undefined);
+            // should be one more ai review score and move branches for empty board.
+            const check1 = !is_uploaded &&
+            (this.props.game.goban.config.moves.length !== (movekeys.length - 1));
+            // extra initial ! in all_moves which matches extra empty board score, except in handicap games for some reason.
+            // so subtract 1 if black goes second == bplayer
+            const check2 = is_uploaded &&
+            ((this.props.game.goban.config["all_moves"].split("!").length - bplayer) !== movekeys.length);
+
+            if (this.state.loading || check1 || check2 ){
                 for (let j = 0; j < ai_table_rows.length; j++){
                     ai_table_rows[j] = ai_table_rows[j].concat(summary_moves_list[j]);
                 }
@@ -811,7 +820,6 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
             const othercounters = Array(2).fill(0);
             let wtotal = 0;
             let btotal = 0;
-            const scores = this.ai_review?.scores;
 
             for (let j = hoffset; j < Object.keys( this.ai_review?.moves ).length - 1; j++ ){
                 const playermove = this.ai_review?.moves[j + 1].move;
--- a/src/views/Game/AIReview.tsx
+++ b/src/views/Game/AIReview.tsx
@@ -1359,11 +1359,11 @@ class AiSummaryTable extends React.Component<AiSummaryTableProperties, AiSummary
                         }
                         {(this.props.pending_entries > 0) &&
                             <React.Fragment>
-                            <tr>
-                                <td colSpan={2}>{"Moves Pending"}</td><td colSpan={3}>{this.props.pending_entries}</td>
-                            </tr>
-                            <tr><td colSpan={5}><progress value = {this.props.max_entries - this.props.pending_entries} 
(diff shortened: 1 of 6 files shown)


mmartin:
Regarding empty AI branches,  #1316 can be closed when this PR gets merged.

BHydden:
> can be closed when this PR gets merged.

@mmartin FTR, GitHub will do this automatically if you type Fixes #1316 

- (`resolves` also works in place of `fixes`, and I think there might be some other key words that will trigger an auto close... `implents` maybe?)

src/views/Game/AIReview.tsx
```diff
@@ -687,6 +709,220 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
...
+            move_list.push(cur_move.player);
+            cur_move = cur_move.trunk_next;
+        }
+        return move_list;
+
+    }
+
+    private AiSummaryTableRowList(){
+        const summary_moves_list = [ ["", "", "", ""], ["", "", "", ""], ["", "", "", ""], ["", "", "", ""], ["", "", "", ""], ["", "", "", ""] ];
+        const ai_table_rows = [["Excellent"], ["Great"], ["Good"], ["Inaccuracy"], ["Mistake"], ["Blunder"]];
```
benjaminpjones, line 727: Would be good to internationalize these strings:

```
[[_("Excellent")], [_("Great")], [_("Good")], [_("Inaccuracy")], [_("Mistake")], [_("Blunder")]]
```

src/views/Game/AIReview.tsx
```diff
@@ -959,6 +1194,11 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
                         </span>
                     </div>
                 }
+                { (data.get("user").is_moderator && ((this.ai_review?.type === "full" ) || (this.ai_review?.type === "fast" )) && this.ai_review?.engine === "katago") &&
```
benjaminpjones, line 1197: Why check both 'full' and 'fast' here? I think those are the only two values `JGOFAIIReview.type` can have

src/views/Game/AIReview.tsx
```diff
@@ -959,6 +1194,11 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
...
+                <div>
+                    <AiSummaryTable headinglist = {["Type", "Black", "%", "White", "%"]} bodylist = {this.table_rows} avg_loss = {this.avg_score_loss} />
```
benjaminpjones, line 1199: Would be good to internationalize these headings too

src/views/Game/AIReview.tsx
shinuito: Sure, I assume that so they can be translated? Makes senes, I didn't know the syntax :)

BHydden: The two main ones are _("String") and [pgettext](https://github.com/online-go/online-go.com/blob/13fcc8b1e698eebb80b90790d6e8fe522c35dc8f/src/lib/translate.ts#L173) which allows you to add additional context for more complex strings

shinuito: Is there a simple way to just click to accept the suggested change?

src/views/Game/AIReview.tsx
shinuito: So it's essentially just throwing in a "true" in there I suppose. I think initially I wasn't going to show the table for fast reviews, which was also a comment I had received. 

That is just because the data won't be as accurate, but it's the same with the AI graph itself and we do show that also. 

I guess we can take out that part.

src/views/Game/AIReview.tsx
shinuito: I agree :)

src/views/Game/AIReview.tsx
shinuito:
```suggestion
        const ai_table_rows = [[_("Excellent")], [_("Great")], [_("Good")], [_("Inaccuracy")], [_("Mistake")], [_("Blunder")]];
```

src/views/Game/AIReview.tsx
shinuito:
```suggestion
                { (data.get("user").is_moderator && this.ai_review?.engine === "katago") &&
```

src/views/Game/AIReview.tsx
shinuito:
```suggestion
                    <AiSummaryTable headinglist = {[_("Type"), _("Black"), "%", _("White"), "%"]} bodylist = {this.table_rows} avg_loss = {this.avg_score_loss} />
```

src/views/Game/AIReview.tsx
benjaminpjones: Thanks for making that change!  Yeah basically there's a script that picks up all the _(...) and pgettext(...) and sends them to translate.online-go.com for translation.  Once the translators do their thing, the translations get added to the dictionary and these functions actually convert the text to the target language.

Pro tip: if you change your language to *Debug* in the settings, it shows you which strings have already been internationalized.  Strings wrapped in [...] are ready:

Screen Shot 2021-12-16 at 9 24 19 AM

shinuito: That sounds like a really convenient idea, to have a debug language like that :)

anoek:
This is great, thanks for making this happen @shinuito :)

For that toggle, how do y'all feel about moving it up onto the key moves line?

image

Maybe instead of the "Hide summary table" text we could just have a little table icon precede it, such as `<i className="fa fa-table" />` or something?

The other thing that might be nice is if the table populated while the results were being streamed in, is that something that could be done relatively easily, or would that be a pain?

shinuito:
> For that toggle, how do y'all feel about moving it up onto the key moves line?

That probably makes sense for saving space yes. Let me catch the branch up to devel and try that.

> Maybe instead of the "Hide summary table" text we could just have a little table icon precede it, such as `<i className="fa fa-table" />` or something?

I didn't actually think of using the fa icons, but that also seems like a good idea.

> The other thing that might be nice is if the table populated while the results were being streamed in, is that something that could be done relatively easily, or would that be a pain?

I was relying on the scores being the difference two consecutive entries so I'll have a think about it in an hour or so when I can look at it properly. Maybe something like looping through the existing entries by their keys for the full review, but I'll have to see how the scores array gets populated for the fast reviews.

It does seem like it would be nice to have something going on while the AI review is loading :)

anoek:
> I was relying on the scores being the difference two consecutive entries
That makes sense, I don't think you need to rethink that, but maybe you could do something like count them if you have both entries, and if you're missing the prior entry just count it in a "pending" number or something? It'd make the display more lively and folks definitely like that :)

shinuito:
@anoek I've updated that now. I've moved the toggle beside the "Key Moves" and used fa-table. I think it looks better.



I also have tried to make the table display in real time the data being populated, and it has an extra row called "Moves pending" that counts how many data points are missing and then disappears when it hits zero.

I suppose that could be styled into anything really, a loading bar etc.

Let me know what you think, or if you have any more suggestions :)

anoek:
Looking great! A loading bar would be awesome, not a requirement by any stretch but if you feel like it, that would be slick :)

shinuito:
So what do you think of something like this then?



That's just using the basic `<progress>` html element and passing the right values to it. I guess one thing is that left to itself it might be styled by the browser. I suppose one could override that with some custom styles but I'm not sure what would be the best.

I was hoping to let some of the mods try some things out with the table, suggest fixes and tweaks (like if they want to change any of the thresholds for good inaccuracy etc), then remove the `is_moderator` to let everyone see it once we're happy with it. I think there was some requests for other metrics that's might be interesting, but that's an ongoing discussion and maybe some future updates could be kept for mod usage so as not to overload the users with information.

anoek:
Awesome work @shinuito , i'll get it up in a sec

PR #1570, opened 2021-12-17
Rengo

GreenAsJade:
Adds rengo challenge management and game display.


GreenAsJade:
(There's a build-fail reported due to the goban dependency)

src/components/RengoAdminModal/RengoAdminModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+interface RengoAdminModalProperties {
+    challenge: any;
+    onAccept: (challenge) => void;
+    // id?: any,
+    // user?: any,
+    // callback?: ()=>any,
+}
+
+
+export class RengoAdminModal extends Modal<Events, RengoAdminModalProperties, {}> {
```
benjaminpjones, line 39: Does this class get used?

Totally possible I'm missing something, but I ctrl+F'd and couldn't find RengoAdminModal or openRengoAdminModal referenced outside of the module itself.

src/views/Play/Play.tsx
```diff
@@ -729,8 +813,270 @@ export class Play extends React.Component<PlayProperties, any> {
...
+            );
+        }
+
+        const user = data.get("user");
+
+        return this.state.rengo_list.map((C) => (
+            (((C.eligible || C.user_challenge || this.state.show_all_challenges)
+             && ((this.state.show_unranked_challenges && !C.ranked) || (this.state.show_ranked_challenges && C.ranked))
+             && ((this.state.show_19x19_challenges && C.width === 19 && C.height === 19) || (this.state.show_13x13_challenges && C.width === 13 && C.height === 13) || (this.state.show_9x9_challenges && C.width === 9 && C.height === 9) || (this.state.show_other_boardsize_challenges && (C.width !== C.height || (C.width !== 19 && C.width !== 13 && C.width !== 9)))) ) ?
+                    <div key={C.challenge_id} className={"challenge-row" + (C.rengo ? " rengo" : "")}>
```
benjaminpjones, line 885: if `C` is a member of `rengo_list`, isn't it a given that it's a rengo challenge?

src/views/Play/Play.tsx
```diff
@@ -619,38 +702,38 @@ export class Play extends React.Component<PlayProperties, any> {
...
-        const challengeList = isLive ? this.state.live_list : this.state.correspondence_list;
-
-        return challengeList.map((C) => (
+        return challenge_list.map((C) => (
             (((C.eligible || C.user_challenge || this.state.show_all_challenges)
              && ((this.state.show_unranked_challenges && !C.ranked) || (this.state.show_ranked_challenges && C.ranked))
              && ((this.state.show_19x19_challenges && C.width === 19 && C.height === 19) || (this.state.show_13x13_challenges && C.width === 13 && C.height === 13) || (this.state.show_9x9_challenges && C.width === 9 && C.height === 9) || (this.state.show_other_boardsize_challenges && (C.width !== C.height || (C.width !== 19 && C.width !== 13 && C.width !== 9)))) ) ?
-                    <div key={C.challenge_id} className="challenge-row">
-                        <span className="cell" style={{textAlign: "center"}}>
+                    <div key={C.challenge_id} className={"challenge-row" + (C.rengo ? " rengo" : "")}>
```
benjaminpjones, line 735: Likewise, if challenge list is guaranteed to be one of `live_list` or `challenge_list`, `C.rengo` will always be false

src/views/Play/Play.tsx
benjaminpjones: Also, I looked in the stylesheets, I'm not sure this class name gets used

src/components/RengoAdminModal/RengoAdminModal.tsx
GreenAsJade: Yeah, nah, not used ... should be axed, after we went to all the trouble of working out how to add a new one :D

src/views/Play/Play.tsx
```diff
@@ -729,8 +813,270 @@ export class Play extends React.Component<PlayProperties, any> {
...
+        .then(() => {
+            swal.close();
+        })
+        .catch((err) => {
+            swal.close();
+            errorAlerter(err);
+        });
+    };
+
+    rengoList = () => {
```
benjaminpjones, line 868: This seems *very* similar to the existing `challengeList()` does it really need to be its own function?

Or at least some duplicate parts be moved into a common function?  Pulling out `cheatAlert(C)` alone would save ~25 lines of JSX.

src/components/RengoAdminModal/RengoAdminModal.tsx
benjaminpjones: Haha at least ESLint got a fix and no poor souls will have to deal with that in the future >.<

src/views/Play/Play.tsx
BHydden: > save ~25 lines of JSX.

/FeelsGoodMan

src/views/Play/Play.tsx
GreenAsJade: I think that's a hangover from when I was toying with having Rengo games in amongst the other games

src/views/Play/Play.tsx
GreenAsJade: I'm not sure it's worth optimising this yet - it hasn't even been tried for real by anyone other than me.

People may hate having Rengo separately, for example...

GreenAsJade:
As of this comment, this PR works with the new game history solution.

src/views/Game/Game.tsx
```diff
@@ -2629,7 +2629,7 @@ export class Game extends React.PureComponent<GameProperties, GameState> {
                 }
                 {(this.state.phase === "finished" || null) &&  /* { */
                     <div className="analyze-mode-buttons">     {/* not really analyze mode, but equivalent button position and look*/}
-                        {(this.state.user_is_player && this.state.mode !== "score estimation" || null) &&
+                        {(this.state.user_is_player && this.state.mode !== "score estimation" && !this.goban.engine.rengo || null) &&
```
BHydden, line 2632: Why is it that people who have just completed a rengo game are less likely to want a rematch than those who complete a regular game? 

Or was the code just not playing nice and this feature is out of scope of an initial implementation?

GreenAsJade: To quote anoek:

"
* When sending a challenge to someone, either the rengo box should be hidden, or we should do something like navigate to the match making screen, add additional invites, or something of the sort. Currently, it just hangs even if the other player accepts.
"

... hence... ;)

BHydden: Thanks, I missed that. Makes sense.

GreenAsJade: Actaully, this is the anoek comment I was looking for, but similar sentiment:

"
* When the game is over all or none of players should see the "Rematch" button". Ideally "All", but it seems like that might be a bit challenging with rengo and expecatations and matchmaking, so maybe just hide it.
"

anoek (commented):
The "Jump to rengo games that start on seekgraph/gamestarted" commit didn't seem to sole the game starting on my end? when the game starts only the starting player navigates still

anoek:
Before release I reckon we should polish things up a bit, want to have good first impressions and all - below is what I observed. Not saying we *need* to do all of these, but we should consider them, and if we're going to do them, might as well do them now to maximize the wow factor / minimizing needless complaints :)

Polish Items
   - [ ] A joined player should see a withdraw button in the rengo pairing panel (I know it's a button down below but that's not obvious, should be in the match making part)
   - [ ] Polish the match making buttons
       - [ ] Let's try to avoid using red for things that aren't "dangerous" - specifically when users are unassigned, the assignment buttons shouldn't be red, and when users are assigned, the unassign button shouldn't be red
       - [ ] Consider replacing the circle x with a normal gray button with a label something like "Unassign", I thought the x would kick the player at first.
       - [ ] Consider replacing the arrows with "Move to black" / "Move to white" or something along those lines, we have plenty of horizontal space, it'd make it clearer what things do maybe?
   - [ ] Consider slapping in a chat box in tht rengo pairing panel for the specific challenge (should be easy, see https://github.com/online-go/online-go.com/blob/devel/src/views/Tournament/Tournament.tsx#L1742 - just change channel to something like `rengo-${challenge.id}` or whatever). Possibly replace the seek graph with that chat log?
   - [ ] Consider hiding the game list while doing pairing, making it a focused view for pairing for rengo. If they withdraw, everything comes back
   - [ ] Remove
image
   - [ ] Instead of 
image
 do what Anh Khoa does in the ios app: 
image


Gotta haves:
 - [ ] Players navigate to the game when the game starts
 - [ ] Player info boxes need to update when viewing past moves
 - [ ] Player info boxes need to update in reviews

src/views/Game/Game.tsx
anoek: Yeah to be clear, I have no problem with a *functional* rematch button, but the current rematch flow would need a good bit of work to make that work, and I didn't see it as critical to Rengo so I didn't want it to block merging.

GreenAsJade:
Status:

> Polish Items
> 
> * [x]  A joined player should see a withdraw button in the rengo pairing panel (I know it's a button down below but that's not obvious, should be in the match making part)
> * []  Polish the match making buttons
>   
>   * [x]  Let's try to avoid using red for things that aren't "dangerous" - specifically when users are unassigned, the assignment buttons shouldn't be red, and when users are assigned, the unassign button shouldn't be red

These buttons are so lightweight that I think it doesn't hurt to find out what they do by pressing them...

>   * [ ]  Consider replacing the circle x with a normal gray button with a label something like "Unassign", I thought the x would kick the player at first.
>   * [ ]  Consider replacing the arrows with "Move to black" / "Move to white" or something along those lines, we have plenty of horizontal space, it'd make it clearer what things do maybe?

Duly considering these:
> * [ ]  Consider slapping in a chat box in tht rengo pairing panel for the specific challenge (should be easy, see https://github.com/online-go/online-go.com/blob/devel/src/views/Tournament/Tournament.tsx#L1742 - just change channel to something like `rengo-${challenge.id}` or whatever). Possibly replace the seek graph with that chat log?
> * [ ]  Consider hiding the game list while doing pairing, making it a focused view for pairing for rengo. If they withdraw, everything comes back

I think this has a purpose, but I made it smaller to start with ...  won't do more unless forced :)
> * [ ]  Remove
>   image

This is a whole project  at my css level:      Don't plan to do before go-live, hopefully some other front-ender might get keen ...    :)
> * [ ]  Instead of
>   image
>   do what Anh Khoa does in the ios app:
>   image

> Gotta haves:
> 
> * [x]  Players navigate to the game when the game starts
> * [x]  Player info boxes need to update when viewing past moves
> * [x]  Player info boxes need to update in reviews

anoek:
> These buttons are so lightweight that I think it doesn't hurt to find out what they do by pressing them...

You and I know that, but someone looking at the interface for the first time doesn't know that. The objection isn't that the actions are dangerous, it's that the actions are *not* dangerous so shouldn't be red, as red discourages users from clicking them when we use it elsewhere to indicate "hey be careful with this button" kind of things.

GreenAsJade:
> > These buttons are so lightweight that I think it doesn't hurt to find out what they do by pressing them...
> 
> You and I know that, but someone looking at the interface for the first time doesn't know that. The objection isn't that the actions are dangerous, it's that the actions are _not_ dangerous so shouldn't be red, as red discourages users from clicking them when we use it elsewhere to indicate "hey be careful with this button" kind of things.

Oh, yeah, for sure, I got rid of red.    I was more responding to the idea of making them bigger with words.

Overall, I suspect the matchmaking will evolve a lot, which is another reason for not over-polishing now.

GreenAsJade:
> > These buttons are so lightweight that I think it doesn't hurt to find out what they do by pressing them...
> 
> You and I know that, but someone looking at the interface for the first time doesn't know that. The objection isn't that the actions are dangerous, it's that the actions are _not_ dangerous so shouldn't be red, as red discourages users from clicking them when we use it elsewhere to indicate "hey be careful with this button" kind of things.

Ah - I see why you said that: my comments in the above are about the thing that follows, not the thing you just read 😅

anoek:
Oh gotcha, at some point I figured that out, but I guess I didn't go back and re-apply, sorry for the confusion, carry on :)

GreenAsJade:
^^^ this is getting annoying.   My local npm run dev setup isn't catching these.

At first I wondered if I was simply missing the warning in that window, but I've just double-checked: this commit passes locally.

@benjaminpjones  thoughts? help?

benjaminpjones:
Does that commit really pass locally?  I just checked on my machine and lint does fail on that commit.

A couple of thoughts:

- the [ESLint plugin for VSCode](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) should let you know what's happening right away
- We've discussed integrating pre-commits for the linter/formatter before. Maybe that would help you?

GreenAsJade:
Huh.   So it turns out my vscode stopped eslinting.    All happy again now.

GreenAsJade:
OMG why didn't I do that before 😮    Thanks @benjaminpjones .

benjaminpjones:
Glad I could help!

BHydden:
Fun fact! Lint checkers work best when they're turned on 😉

anoek:
Great work on this @GreenAsJade !

PR #1571, opened 2021-12-18
Add state types for React.PureComponent

benjaminpjones:
## Proposed Changes

  - Removes a lot of `any`s
  - I made a point **not** to touch actual logic - hence the `TODO:` comments.
      - One exception: I did change a variable name, I will point that out in a review comment.

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -58,60 +58,60 @@ interface TournamentProperties {
 }
 
 interface TournamentState {
-        new_tournament_group_id: number;
-        tournament_id: number;
-        loading: boolean;
-        tournament: {
-            id: number;
-            name: string;
-            director: player_cache.PlayerCacheEntry;
-            time_start: string;
+    new_tournament_group_id: number;
+    tournament_id: number;
+    loading: boolean;
+    tournament: {
+        id: number;
+        name: string;
+        director: player_cache.PlayerCacheEntry;
+        time_start: string;
 
-            board_size: string|number;
-            rules: GoEngineRules;
-            description: string;
-            handicap: string;
-            time_control_parameters: TimeControl;
-            tournament_type: string;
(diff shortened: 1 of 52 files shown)


src/views/User/User.tsx
```diff
@@ -177,7 +259,7 @@ export class User extends React.PureComponent<UserProperties, any> {
     }
 
     isSpecialUser() {
-        if (this.state.user.supporter || this.state.user.is_moderator || this.state.user.is_superUser) {
```
benjaminpjones, line 180: I don't know what a "super user" is, but it seemed pretty clear this was a typo.  I did check the API call to verify this should be lowercase, but not sure where to verify in the UI.

anoek: It should indeed be lowercase, thanks for the fix!. And FTR a superuser is matburt or myself

src/views/Tournament/Tournament.tsx
```diff
@@ -57,6 +57,63 @@ interface TournamentProperties {
     };
 }
 
+interface TournamentState {
+        new_tournament_group_id: number;
```
anoek, line 61: Extra level of indent here

anoek:
This is fantastic, thanks for the typing work :)

src/views/User/User.tsx
benjaminpjones: Oh got it! That would explain the :spinning_smiley: OGS Developer :spinning_smiley:

src/views/Tournament/Tournament.tsx
benjaminpjones: Oops.  Sometimes I would copy the state initialization into the interface and modify so that's how that happened:

```
interface TournamentState {
    this.state = {  // cut this out after
        some_member: "Hi!",
    }
}
```

benjaminpjones:
Thanks for reviewing @anoek!

PR #1577, opened 2021-12-23
Move PaginatedTable to use the filter on props

GreenAsJade:
... get rid of PaginatedTable.filter_updated

Fixes #

Fixes the problem of wanting to update the filter, but needing to be able to have the URL param not included, but also wanting to use state changes -> prop changes to cause the table to update

## Proposed Changes

Use [this](https://amberley.dev/blog/2020-09-07-conditionally-add-to-array-or-obj/#conditionally-add-a-property-to-an-object) to let us put the necessary filter url param in place when we need to, but not have it there when it should not be, and do it based on the state of the instantiating component controlling the filter prop of the paginated table.

### DIFF
--- a/src/components/PlayerAutocomplete/PlayerAutocomplete.tsx
+++ b/src/components/PlayerAutocomplete/PlayerAutocomplete.tsx
@@ -124,7 +124,8 @@ export class PlayerAutocomplete extends React.PureComponent<PlayerAutocompletePr
                 }
             })
             .catch((err) => {
-                console.log(err);
+                // I'm not sure these logs are helpful: we get to here if they type ahead fast
+                // console.log(err);
             });
         } else {
             this.setState({
--- a/src/views/GroupList/GroupList.tsx
+++ b/src/views/GroupList/GroupList.tsx
@@ -24,8 +24,11 @@ import {PaginatedTable} from "PaginatedTable";
 import {SearchInput} from "misc-ui";
 import {navigateTo} from "misc";
 
+interface GroupListState {
+    name_contains_filter: string;
+}
 
-export class GroupList extends React.PureComponent {
+export class GroupList extends React.PureComponent<{}, GroupListState> {
     refs: {
         table;
     };
@@ -33,6 +36,7 @@ export class GroupList extends React.PureComponent {
     constructor(props) {
         super(props);
         this.state = {
+            name_contains_filter: "",
         };
     }
     componentDidMount() {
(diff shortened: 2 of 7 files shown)


src/components/PlayerAutocomplete/PlayerAutocomplete.tsx
```diff
@@ -124,7 +124,8 @@ export class PlayerAutocomplete extends React.PureComponent<PlayerAutocompletePr
                 }
             })
             .catch((err) => {
-                console.log(err);
```
benjaminpjones, line 127: Haha I deleted this on my local environment too. Console was very red.

Since aborting requests in flight is part of a normal user journey, I think suppressing the error is reasonable. Also, doesn't `requests.ts` post a duplicate error, with more info if it's not an abort?

src/views/GroupList/GroupList.tsx
```diff
@@ -66,7 +69,9 @@ export class GroupList extends React.PureComponent {
                             name="game-history"
                             source={`groups/`}
                             orderBy={["-member_count"]}
-                            filter={{ "name__icontains": "" }}
+                            filter={{
+                                ...(this.state.name_contains_filter !== "" && {'name__icontains': this.state.name_contains_filter})
```
benjaminpjones, line 73: This syntax  (`...(some_obj)` is cool I'd never seen that.

One note- it doesn't hurt to be explicit with the check, but `PaginatedTable *does* do some handling for empty strings in case you were wondering how the above line was possible before:

https://github.com/online-go/online-go.com/blob/52772889d53c97f038cea90020a6c9698bdee214/src/components/PaginatedTable/PaginatedTable.tsx#L132-L138

benjaminpjones (commented):
This is really great.  As part of this clean up effort (but not necessarily this PR), I think we can probably remove the refs for these too:

```
    refs: {
        vacation_left;
        bot_ai;
        game_table;  // no longer using this
        review_table; // no longer using this
    };
    ...
    <PaginatedTable
        className="game-history-table"
        ref="game_table"  // no longer used
    ...
    <PaginatedTable
        className="review-history-table"
        ref="review_table"  // no longer used
```

@anoek and @GreenAsJade does that track with what you know about `PaginatedTable`?  Or am I missing another usage of these refs?

anoek:
Yeah, I'm in the process of building on this patch, removing all the refs and fixing the inconsistencies as part of it. Should be done tomorrow morning.

anoek (commented):
The addition of the rengo channel is good, but should be part of the rengo stuff not here :)

anoek:
The non rengo related commit in this was merged in #1578 so closing this (I do want that channel, when rengo lands though :) )

GreenAsJade:
I had thought to add the Rengo channel into Main so people have a place to chat about Rengo and matchmaking while beta testing...

anoek:
Oh gotcha, yeah sure that's fine then I reckon

PR #1579, opened 2021-12-24
Tournament: show final results for finished tournaments

pdg137:
See this thread:

https://forums.online-go.com/t/tournaments-should-show-winners/33767/9

## Proposed Changes

  - Shows "Final results" with a list of the winners (top 3) of the tournament.
  - Each winner is shown with the corresponding trophy:

image

## Remaining issues

- I had to compute trophy filenames since these don't seem to be available in the API; they won't work for the special title tournaments or non-standard sizes.
- Someone should check how I did the styling.
- This introduces a new loop over players which could be slow for large tournaments.
- I didn't implement hover text on the trophies since it seemed redundant.
- Maybe should be a bit narrower for mobile.

### DIFF
--- a/src/views/Tournament/Tournament.styl
+++ b/src/views/Tournament/Tournament.styl
@@ -398,6 +398,7 @@
         border: 2px solid transparent;
         themed border-color fg
         width: 24rem;
+        max-width: 85vw;
         margin-left: auto;
         margin-right: auto;
         padding: 1rem;
@@ -417,11 +418,12 @@
             display: inline-flex;
             align-items: center;
             font-weight: bold;
-            width: 15rem;
             line-height: 2em;
         }
         .final-results-place {
             width: 9rem;
+            line-height: 1em;
+            padding-right: 0.5em;
             img {
               margin-right: 0.5em;
             }
--- a/src/views/Tournament/Tournament.styl
+++ b/src/views/Tournament/Tournament.styl
@@ -394,4 +394,38 @@
         }
     }
 
+    .final-results {
+        border: 2px solid transparent;
+        themed border-color fg
+        width: 24rem;
+        margin-left: auto;
+        margin-right: auto;
+        padding: 1rem;
+        margin-top: 0.5rem;
+
+        h2 {
+            margin-top: 0;
+            font-size: larger;
+            text-align: center;
+        }
+        > div {
+            display: flex;
+            align-items: center;
+        }
+        > div > span {
+            height: 2em;
+            display: inline-flex;
+            align-items: center;
+            font-weight: bold;
+            width: 15rem;
+            line-height: 2em;
+        }
+        .final-results-place {
+            width: 9rem;
+            img {
+              margin-right: 0.5em;
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


benjaminpjones:
`npm run lint:fix` should take care of the indentation issues for you.

benjaminpjones:
> Maybe should be a bit narrower for mobile.

`max-width: 85vw;` might help with the mobile width, although I did need to make the player names narrower to prevent "Second place" from splitting to the next line:

Current:
Screen Shot 2021-12-24 at 7 27 16 PM
With max-width:
Screen Shot 2021-12-24 at 7 26 42 PM
With `max-width=85vw` and players `width=12rem`:
Screen Shot 2021-12-24 at 7 26 56 PM

> This introduces a new loop over players which could be slow for large tournaments.

What are the largest tournaments? I checked this with recently finished Honinbo and Mingren Main Title tournaments (<300 players)... no performance issues.

Biggest tournament I can think of is [Nine*Nine, NineNineNine](https://online-go.com/tournament/72565) but even that is <1000, and while page load is generally a bit slow, `players` would have to be much bigger for a `filter`/`sort` to become perceptibly slow.

pdg137:
Thanks for the feedback @benjaminpjones! I cleaned up the indentation and played around with more mobile-friendly dimensions.  The `85vw` max-width helps, like you said, and I removed the explicit width on the second column so that it can start out wide, but then when it gets squeezed, the first one stays fixed as long as possible.  Also tweaked some of the settings so that it looks okay even if it ends up wrapping:

image

Does this work with your device?

> What are the largest tournaments? I checked this with recently finished Honinbo and Mingren Main Title tournaments (<300 players)... no performance issues.

Great! I guess I don't know of particularly large *finished* ones.

pdg137:
There's something weird with the OpenGotha tournaments onthe beta server:

http://localhost:8080/tournament/128

image

Is that four zeroth-place players or something?  Not sure that it's a real problem.

benjaminpjones:
>Does this work with your device?

Full disclosure, I'm just using the [Responsive Design](https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode) tools with Firefox (my screenshots were for the iPhone SE2 size).  But yes that looks good to me!

>Is that four zeroth-place players or something? Not sure that it's a real problem.

I don't know much about opengotha or why ranks would be 0, but it seems like it would be a problem in the sense that *every* player could be ranked 0, and that component could become unboundedly large.

larger example from online-go.com/tournament/80176:

Screen Shot 2021-12-25 at 4 18 36 PM

(Also an example of the tournament images missing.)

pdg137:
> might want to hide that component when tournament_type="opengotha"

Okay, I've hidden the final results for OpenGotha tournaments and just in case also excluded rank zero players from the list.

pdg137:
This reveals some surprising 3rd-place trophy winners:

image
https://online-go.com/tournament/84112

image
http://localhost:8080/tournament/84111

anoek:
This is great @pdg137 , thanks for taking this on :)

PR #1582, opened 2021-12-25
Tournaments: do not show wrong default data while loading

### LINKED ISSUE #1581: Tournaments display wrong information while loading
**Describe the bug**
While loading, there is a bunch of wrong default information at the top of the screen:

image

**To Reproduce**
1. Go to a tournament that's slow to load: https://online-go.com/tournament/80970
2. See wrong default information
3. Wait
4. Correct information appears after a while

**Expected behavior**
It should not show the wrong information

**Additional context**
I believe the defaults are also used for setting up new tournaments, and some (useless!) calculations are done with them while the page is loading, so it's not trivial to just blank them out.

pdg137:
## Proposed Changes

  - While a tournament is loading, do not show wrong default values

Fixes #1581, I think.  Not sure if this is the right place to put this line, but it seems to at least block anything from getting displayed while the tournament is loading. 

Of course it would be nicer if the OGS loading icon could continue bouncing up and down while the page is loading.

Merry Christmas OGS team! 🎅

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -1294,6 +1294,9 @@ export class Tournament extends React.PureComponent<TournamentProperties, Tourna
             null
             );
 
+            if (this.state.loading)
+                return "";
+
             return (
                 <div className="Tournament page-width">
                     <UIPush event="players-updated" channel={`tournament-${this.state.tournament_id}`} action={this.reloadTournament}/>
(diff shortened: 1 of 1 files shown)


anoek:
If you wanted the loading page, I think you can just do 

`import { LoadingPage } from 'Loading';` and then drop in `<LoadingPage />` instead of that ""

pdg137:
Okay, I made that change!

anoek:
Nice! Thanks for the improvements!

PR #1584, opened 2021-12-26
Tournament final results: do not bold player names since that messes up the emoji font

pdg137:
Example: https://online-go.com/tournament/16820

Before / after:

image
image


anoek:
Hah, that's kinda funny

pdg137:
There's a similar issue lower down on some of the tournament pages:

https://online-go.com/tournament/56272

image

PR #1585, opened 2021-12-27
How about a Rengo chat channel, for matchmaking in particular...

GreenAsJade:


BHydden:
New [global rengo](https://online-go.com/chat/global-rengo/rengo) channel works when I direct navigate to it, but if I change channels the rengo channel disappears. 

rengo
no-rengo

PR #1589, opened 2021-12-31
More key moves

philipkd:
Fixes this [forum comment](https://forums.online-go.com/t/top-3-moves-score-a-better-metric/32702/14):

> Those 6 moves are readily available, but I’m not exactly sure the best way to expose in the UI.")

## Proposed Changes

- shows six key moves, three from each player
- key-moves is now above win-score-toggler
- removed "Key moves:" label

before:

Screen Shot 2021-12-30 at 8 51 47 PM

after:

Screen Shot 2021-12-30 at 8 50 44 PM

## Background

When a game finishes, the AI review shows three buttons representing the moves that made the biggest difference in the game. A month ago, there was a non-functional "+n moves" label:

143656382-e12d5279-ab2d-4658-b5a6-94265fd2c1b6

Then, the label was removed as part of work towards creating an AI summary table.

### DIFF
--- a/src/views/Game/AIReview.styl
+++ b/src/views/Game/AIReview.styl
@@ -179,6 +179,7 @@
         user-select: none;
         display: inline-flex;
         width: 100%;
+        text-align: center;
 
         .move-list {
             flex: 1;
@@ -191,7 +192,6 @@
             border: 1px solid #888;
             width: 2.5rem;
             display: inline-block;
-            text-align: center;
             cursor: pointer;
         }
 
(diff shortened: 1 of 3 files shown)


anoek:
Thanks for the improvement!

Aesthetically I like this change. I do worry a little bit that their presence might be a little cryptic in this form, though I am not convinced that the "Key moves" label made them particularly less cryptic, so I'm inclined to merge anyways.

I'd like to keep this PR open for another day or two in case anyone has any ideas of ways to help users who first see this intuitively understand what these bubbles with letter/number combinations mean. If no one does though I'll just merge and we'll see shelve that thought/potential improvement for another time.

philipkd:
Great, thanks for the feedback!

anoek:
Thanks for the improvement @philipkd !

philipkd:
Great, thanks for the review and merge!

PR #1593, opened 2022-01-02
[DISCUSSION ONLY] Prettier

benjaminpjones:
Based on discussion from #1591 

I stripped down the prettier config to only settings that were relatively widespread in the codebase (e.g. 4 spaces, which was already enforced by eslint).  Then I ran it.

No plans to get this merged while #1570 is still on another branch, just wanted to open the discussion.  I guess these are the two main points I'd like to cover, but any thoughts welcome:
  - Is Prettier even desirable for this repo?
  - Any other options to set? (see [prettier options](https://prettier.io/docs/en/options.html))

## Proposed Changes

  - Strip down Prettier config
  - run Prettier
  - remove eslint formatting rules
  - (not done yet) add prettier to CI/gulp
      - Probably some discussion to have here re: dev workflow, but I see this working pretty similar to how eslint works now (gulp/CI/IDE complains, then run `prettier --write`)


benjaminpjones:
Oops looks like I had a couple of unrelated commits on my local `devel` branch.  Probably not worth rebasing, please ignore those!

src/components/ACLModal/ACLModal.tsx
```diff
@@ -16,23 +16,20 @@
...
+import { Player } from "Player";
+import { PlayerAutocomplete, PlayerAutocompleteRef } from "PlayerAutocomplete";
+import { GroupAutocomplete } from "GroupAutocomplete";
+import { errorAlerter, rulesText } from "misc";
+
+interface Events {}
 
 type ACLModalProperties =
-    { game_id: number }
+    | { game_id: number }
```
benjaminpjones, line 30: This one is *weird*. but WAI: https://github.com/prettier/prettier/issues/6299

src/components/AIReviewStream/AIReviewStream.tsx
```diff
@@ -41,30 +39,34 @@ export function AIReviewStream(props: AIReviewStreamProperties): JSX.Element {
...
             }
         }
 
         function onJwtChange() {
-            const user = data.get('config.user');
-            const user_jwt = data.get('config.user_jwt');
+            const user = data.get("config.user");
+            const user_jwt = data.get("config.user_jwt");
             if (!user.anonymous && user_jwt) {
-                ai_socket.send('authenticate', {jwt: user_jwt});
```
benjaminpjones, line 55: I removed single-quotes from the config because there are a lot of double quotes too so I figured let prettier take the wheel.

(I actually prefer single quotes, so wouldn't be mad if others wanted to re-add that setting)

src/components/Announcements/Announcements.tsx
```diff
@@ -158,40 +172,60 @@ export class Announcements extends React.PureComponent<{}, AnnouncementsState> {
...
                 {this.state.announcements.map((announcement, idx) => {
-                    const creator_blocked = getBlocks(announcement.creator.id).block_announcements;
+                    const creator_blocked = getBlocks(
+                        announcement.creator.id,
+                    ).block_announcements;
                     const type_muted = announcementTypeMuted(announcement);
 
                     if (creator_blocked || type_muted) {
-                        return (null);
+                        return null;
```
benjaminpjones, line 198: nice.

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -394,7 +468,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
         let player_id = 0;
         if (this.props.mode === "player") {
             player_id = this.props.playerId;
-            if (!player_id || (player_id === data.get("user").id)) {
+            if (!player_id || player_id === data.get("user").id) {
```
benjaminpjones: Things like this worry me. Clearly the author of this line thought order of operations are ambiguous (at least to a human), and the parentheses make it explicit.

Luckily it appears they keep parens for commonly confusing orders (`&&` mixed with `||`) https://github.com/prettier/prettier/issues/187

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -484,138 +573,160 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
...
                 }
-            }
-        })
-        .catch(errorAlerter);
+            })
+            .catch(errorAlerter);
     };
 
     /* update bindings  */
-    update_conf_bot_id          = (ev) => this.upstate("conf.bot_id", ev);
```
benjaminpjones: Goodbye alignment...

src/components/Announcements/Announcements.tsx
BHydden: this one is so pleasing

src/components/ChallengeModal/ChallengeModal.tsx
BHydden: yeah... new way looks tidier but you need to work harder to make sure you're reading it right, which goes against best coding practice of readability

src/lib/chat_manager.ts
```diff
@@ -209,36 +275,38 @@ data.watch("config.ogs", (settings) => {
...
         });
     } else if (channel.startsWith("game-")) {
-        return interpolate(_("Game {{number}}"), {"number": channel.substring(5)});
+        return interpolate(_("Game {{number}}"), {
+            number: channel.substring(5),  // eslint-disable-line
+        });
     } else if (channel.startsWith("review-")) {
-        return interpolate(_("Review {{number}}"), {"number": channel.substring(7)});
+        return interpolate(_("Review {{number}}"), {
+            number: channel.substring(7),  // eslint-disable-line
```
benjaminpjones: Here's a change that triggers eslint's `ban-names` rule.  ESLint wasn't detecting `"number": ...`, but it does detect `number: `

Not hard to work around (just use a name like `num_reviews` instead), but a change to be aware of.

src/components/ACLModal/ACLModal.tsx
anoek: I've been learning all kinds of new syntax and things allowed from y'all recently. I personally like this since it's visually consistent.

src/lib/chat_manager.ts
anoek: Interesting, yeah it's not that big of a deal I don't suppose, but modifying this causes translation churn as the whole string needs to be re-translated. It looks like there are only a few cases of "number" being used in translations though, so I reckon it's fine to just bite the bullet at some point and have some bad translations for a bit until they're patched.

src/components/AIReviewStream/AIReviewStream.tsx
anoek: I'm not really opinionated either way

prettier.config.js
```diff
@@ -1,12 +1,4 @@
-// prettier.config.js or .prettierrc.js
 module.exports = {
     trailingComma: "all",
     tabWidth: 4,
-    semi: true,
-    singleQuote: true,
-    printWidth: 140,
```
anoek, line 7: How do folks feel about `printWidth`? I generally find `80` to be harder to read and generally like `120` or so myself, but I'm flexible.

src/components/Steps/Steps.tsx
```diff
@@ -30,10 +30,16 @@ interface StepsProperties {
...
 
 export const Steps = (props: StepsProperties) => {
-    const children = (Array.isArray(props.children) ? props.children : (props.children ? [props.children] : [])).concat([]);
+    const children = (
+        Array.isArray(props.children)
+            ? props.children
+            : props.children
+            ? [props.children]
+            : []
+    ).concat([]);
```
anoek: The removal of the parens here is unfortunate, but good elsewhere so probably net positive I suppose

anoek:
I'm on board with doing this. 

As for dev workflow, are there notable downsides to having a pre-commit hook to start doing both `eslint --fix` and `prettier --write`? I'm not sure how many `lint` commits there have been but it's a lot, and is often a kind of annoying `x` that pops up when someone is first doing a PR to the repo, nagging them over a missing semicolon or whatever, if we could help automate fixing those seems like it'd be a good thing?

src/lib/chat_manager.ts
benjaminpjones: >modifying this causes translation churn as the whole string needs to be re-translated.

I should have clarified that the workaround I suggested was for future translations.  For existing translations, we could avoid churn by invoking `eslint-disable-line id-denylist` and maybe a clarifying comment.

prettier.config.js
benjaminpjones: I have an 80 column environment at my 9-5, so I'm pretty used to it.  `120` sounds reasonable to me though.  Just won't be able to use punch cards 😭

BHydden: I prefer 80 as an industry standard, but I don't think it's compatible with our current codebase... we have some really long lines that would be hard to read on 80 width but it would take a lot of work to find ALL of them and rewrite them into a tighter form factor (also you trade line width for total lines usually so it's a catch-22)

BHydden:
> nagging them over a missing semicolon or whatever

9/10 it's tabs over spaces or whitespace at the eol haha

yeah agreed some kind of auto-fix would clean up the commit tree a lot I think

benjaminpjones:
> As for dev workflow, are there notable downsides to having a pre-commit hook to start doing both eslint --fix and prettier --write?

I'm not aware of any notable downsides, but then again I've never used write operations in a pre-commit.  Maybe @maurelio1234 can speak to any downsides?

> I'm not sure how many lint commits there have been but it's a lot, and is often a kind of annoying x that pops up when someone is first doing a PR to the repo, nagging them over a missing semicolon or whatever, if we could help automate fixing those seems like it'd be a good thing?

Yeah it would be nice to be done with `lint fix` commits.  One thought: AFAIK, pre-commit tools need to be initialized by the developer ([at least that is the case with husky](https://blog.typicode.com/husky-git-hooks-autoinstall/)).  I don't know if that means we should add pre-commit initialization into the `Makefile` or if another instruction in `CONTRIBUTING.md` would suffice.  Either way, I think having the commits there if someone wants them would be a good thing.

src/components/AIReviewStream/AIReviewStream.tsx
GreenAsJade: I'm not fussed either.   Where context allows (I mean in the absence of anyone else's obvious standards and no semantic difference) I've used the "double quotes for things people are intended to see, single for internal strings" thing, FWIW, but that kind of thing only works if everyone knows and does ... unlikely in a shared codebase like this.   

And maybe that points us in the direction of "keep it simple, clear, set a standard and stick to it" -> turn it on.

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: And to be fair, as with `return (null)`  , the extra parens here create cognitive load, in the form of "wait, why did that put that there, am I reading it right?" so I support prettier here.

prettier.config.js
GreenAsJade: The natural margin of code can see in front of me is about 100 chars or so.

benjaminpjones: I'm going to change stance on this a bit.  [Prettier docs](https://prettier.io/docs/en/options.html#print-width) explicitly recommend against more than 80 chars.  The reasoning being that prettier will reformat short lines to be longer wherever possible, in contrast to ESLint's `max-len` which will only ever make lines shorter.

I opened a PR to compare 80 vs 120 printWidth: https://github.com/benjaminpjones/online-go.com/pull/1/files.  Beauty in the eye of the beholder and all that, but IMO too many lines are going into the gutter.  Therefore, if we can get closer to `80` and still be happy, I encourage that.

maurelio1234:
> I'm not aware of any notable downsides, but then again I've never used write operations in a pre-commit. Maybe @maurelio1234 can speak to any downsides?

The only downsides on my day to day use are that 
1. it's is easy to bypass (git commit --no-verify)
2. it takes time to run before each commit (but to me it's much better than waiting for a broken CI job).
3. you don't know how your code will look like before committing since it will be reformatted on the fly (but then... it's a different mindset... I'm fine about not caring about indent/formatting anymore)

anoek:
Could we do a pre-commit check of syntax, and just refuse until the errors are fixed? That way it's on the submitter to fix and they'll know what they're committing?

prettier.config.js
anoek: It's a shame we can't turn that off and let eslint enforce with it's maxlength. I don't like particularly long lines, I like how they read at `80` columns *IF* the code is indented only maybe 0-4 times, but looking at the JSX when it's 15-20 levels deep, 80 column wrapping starts really hurting as it explodes everything vertically, and I breaking out content into multiple sections to reduce indent just turns that content into spaghetti.  I think what I really want is something like `60` column wrapping ignoring indentation columns or something along those lines, that doesn't seem to be a thing though.

benjaminpjones:
>Could we do a pre-commit check of syntax, and just refuse until the errors are fixed? That way it's on the submitter to fix and they'll know what they're committing?

Yep, just need to run the `--check` versions of the commands (instead of `--write`/`--fix`).  If the command returns a non-zero error code, the commit will abort.  I actually used to do this with `tslint` and it saved me from a ton of `fix lint` commits.

prettier.config.js
benjaminpjones: I hear that.  Each option definitely seems to have its downsides...

I guess where does that leave us? Should I drop the idea of integrating prettier? Or if prettier is still desirable, 80 or 120 or something in between?

> breaking out content into multiple sections to reduce indent just turns that content into spaghetti.

FWIW, I think highly indented code could benefit from being broken out (as long as function names make sense).  But that doesn't really help with existing code that would be "prettified"...

GreenAsJade: I've started doing this in Play.tsx & User.tsx, where I'm messing around today.

Y'all  must detest my code, because I'm regularly out beyond column 100, even in logic.  With some refactoring this is down to 84:

```
    rengoReadyToStart = (challenge) => (
        challenge.rengo_black_team.length && challenge.rengo_white_team.length &&
        (challenge.rengo_black_team.length + challenge.rengo_white_team.length > 2)
```
        
I would really dislike having to put the onClick on the next line here:
```

                    <div className="rengo-challenge-buttons">
                        { (own_rengo_challenge || null) &&
                            <React.Fragment>
                                <div className='automatch-settings'>
                                    <button className='danger sm' onClick={this.cancelOwnChallenges.bind(self, this.state.rengo_list)}>
                                        {pgettext("Cancel challenge", "Cancel")}
                                    </button>
```


because then you get the content aligned with the continuation:

```
        <button className='danger sm' 
            onClick={this.cancelOwnChallenges.bind(self, this.state.rengo_list)}>
            {pgettext("Cancel challenge", "Cancel")}
       </button>
```
       
         
... wait, what, is that {pgettext part of the <button, or the content, ??

GreenAsJade: Note: I think my ongoing attempts to reduce line length are improving readability, but at the moment getting consistently near 80chars is a stretch...

BHydden: What about...

```
                                    <button 
                                        className='danger sm' 
                                        onClick={this.cancelOwnChallenges.bind(self, this.state.rengo_list)}
                                    >
                                        {pgettext("Cancel challenge", "Cancel")}
                                    </button>
```

GreenAsJade: Yeah - that's better than mine I think, but it still takes a human to work out that the > needs to go on the next line for readability

BHydden: If there was a rule saying something like `If the contents of any bracket ([{<>}]) covers multiple lines, indenting of contents and vertical alignment of bracket is appropriate? Feel like a robot could follow that fairly well, but I'm wondering if there is a situation we wouldn't want that to happen?

GreenAsJade: Right, but in any case try telling prettifier that ... ?

benjaminpjones: > it still takes a human to work out that the > needs to go on the next line for readability

> If there was a rule saying something like If the contents of any bracket ([{<>}]) covers multiple lines, indenting of contents and vertical alignment of the bracket is appropriate? Feel like a robot could follow that fairly well

> Right, but in any case try telling prettifier that ... ?

[If only there were a way to settle this...](https://prettier.io/docs/en/options.html#bracket-line)

BHydden: Lovely <3 I personally think the latter looks just a little bit nicer, but the 80pw is perfectly readable :)

benjaminpjones: >I personally think the latter looks just a little bit nicer, but the 80pw is perfectly readable

I think I feel about the same.  And I will point out that with 80pw, no side scrolling necessary!

<details> 
  <summary>screenshot</summary>
  Screen Shot 2022-01-05 at 12 44 30 PM
</details>

BHydden: I am a huge fan of no side scrolling! ❤️

GreenAsJade: Is this about side-scrolling?  Maybe you need a monitor upgrade? 

VSCode on my desktop has 2x 120 column windows _and_ the finder left pane open...

BHydden: He's talking about GitHub not VSCode

GreenAsJade: Ah hah - yes indeed, this github layout is curiously narrow isn't it :D  :D

GreenAsJade:
Easy to read:

    initialize() {
        this.goban = new Goban({
            "board_div": this.goban_div,
            "draw_top_labels": false,
            "draw_bottom_labels": false,
            "draw_left_labels": false,
            "draw_right_labels": false,
            "game_id": this.props.id,
            "display_width": this.props.displayWidth || (Math.min($("body").width() - 50, $("#em10").width() * 2)),
            "square_size": "auto",
            "width" : this.props.width || (this.props.json ? this.props.json.width : 19),
            "height" : this.props.height || (this.props.json ? this.props.json.height : 19)
        }, this.props.json);
```

Messy and confusing:

```    
    initialize() {
        this.goban = new Goban({
            "board_div": this.goban_div,
            "draw_top_labels": false,
            "draw_bottom_labels": false,
            "draw_left_labels": false,
            "draw_right_labels": false,
            "game_id": this.props.id,
            "display_width": this.props.displayWidth ||
                (Math.min($("body").width() - 50, $("#em10").width() * 2)),
            "square_size": "auto",
            "width" : this.props.width ||
                (this.props.json ? this.props.json.width : 19),
            "height" : this.props.height ||
                (this.props.json ? this.props.json.height : 19)
        }, this.props.json);
```

```

benjaminpjones:
How are you getting these examples?  Those snippets were most definitely not formatted by prettier...

GreenAsJade:
As I'm coding right now I'm trying to be considerate of my narrow-view colleagues  ( ;) :bzzz )  , and seeing if I can write shorter lines.

I encountered that initialization in my travels and tried to tidy it up to 80 chars.  The above is what I thought one would do.

GreenAsJade:
Here's another example.

I find this readily parseable because the most important thing is that it is a ternary statement:

```
                {black_name: this.goban.engine.rengo ?
                    this.goban.engine.rengo_teams.black[0].username + " +" +  (this.goban.engine.rengo_teams.black.length - 1) :
                    this.goban.engine.players.black.username }),
```

If I care about the truthy value, there it is all in one place, to dig into if needed

```
                {black_name: this.goban.engine.rengo ?
                    this.goban.engine.rengo_teams.black[0].username + " +" +
                        (this.goban.engine.rengo_teams.black.length - 1) :
                    this.goban.engine.players.black.username }),
```

is a bunch of lines blobbed together with +  and "  and - and : ...

Once again, just exploring in the spirit of "what really delivers the most readable result".

GreenAsJade:
That said, this is in the middle of a piece of code that needs a refactor, so that might be the right answer :)

benjaminpjones:
>I does raise a question in my mind: are we thinking that in a prettier regime you don't worry about formatting you just let prettier do it at some point? Or do you actually code with the view that if you get it within parameters then it won't mess with it?

I think that's one of the main selling points of prettier: you never have to think about formatting again.

After enough experience, one can probably guess what it's going to do, but there's really no need bc it will do all that for you!

> is a bunch of lines blobbed together with + and " and - and : ...

This is good feedback because I originally thought you were in the "more than 100 chars isn't readable" camp.  Personally, I don't really see much of a difference.. it's a lot of information to process either way. (side note: I couldn't find this passage in the code, but guessing it's for the game history or something?)

> That said, this is in the middle of a piece of code that needs a refactor, so that might be the right answer :)

Definitely! I mean hopefully whatever we do doesn't require a full refactor (I'm not too worried), but I can think of some ways to make this more concise regardless of the prettier config.

GreenAsJade:
> After enough experience, one can probably guess what it's going to do, but there's really no need bc it will do all that for you!


Hmm - really?   

I always have in the back of my mind "does this look readable?"  (believe it or not :p )

I can't quite picture how that will work if something is going to mess with it afterwards.   I had imagined I'd have the goal of coding so that it doesn't have to, much like with lint.

> (side note: I couldn't find this passage in the code, but guessing it's for the game history or something?)

https://github.com/online-go/online-go.com/blob/660423faa07381af0d4f8cdaa08168abf69309b8/src/components/MiniGoban/MiniGoban.tsx#L107

BHydden:
> I always have in the back of my mind "does this look readable?" (believe it or not :p )
> 
> I can't quite picture how that will work if something is going to mess with it afterwards. I had imagined I'd have the goal of coding so that it doesn't have to, much like with lint.

I have a very visual memory, so I would do the same purely out of self-interest haha, but I think bpj's point was that one COULD code that way if one had a mind to, not that one must. Certainly, we can code with minimal prettier changes in mind.

benjaminpjones:
>> (side note: I couldn't find this passage in the code, but guessing it's for the game history or something?)
> 
> online-go.com/src/components/MiniGoban/MiniGoban.tsx
> 
> Line 107 in 660423f
>  initialize() { 

No I found that one, couldn't find this:



> I can't quite picture how that will work if something is going to mess with it afterwards. I had imagined I'd have the goal of coding so that it doesn't have to, much like with lint.

@BHydden's right you can code however you want to :) I will note that prettier (and any other formatter really) has a lot less wiggle room by nature, so that's why I take the "I focus on logic, you focus on spacing" approach when I'm in an environment with a formatter.

benjaminpjones:
By the way, if you're worried about writing code one way, and submitting a PR that looks totally different, I'm a fan of [Format on Save](https://code.visualstudio.com/updates/v1_6#_format-on-save).  That way you're in control and you get to see how it looks as often as you `ctrl+S` (which for me is all the time haha)

GreenAsJade:
> By the way, if you're worried about writing code one way, and submitting a PR that looks totally different, I'm a fan of [Format on Save](https://code.visualstudio.com/updates/v1_6#_format-on-save). That way I'm in control and I get to see how it looks as often as I `ctrl+S` (which is all the time haha)

OMG but of course, right: that would be a must.

(I already have remove-trailing-spaces-on-save, but the assumption is if that if you're going to have a prettifier, then it does it as close as possible to the typing...  I assume that when I pull the commit when we introduce this, my vscode settings necssary will come with it, right ;)  )

benjaminpjones:
Alrighty, I think I got the feedback I needed so I can close this PR. Just want to be explicit that these are my takeaways (correct me if I misinterpreted!)

>    Is Prettier even desirable for this repo?

Yes

>    Any other options to set? (see prettier options)

`printWidth: 120`

I'll be back shortly with the real thing!

GreenAsJade:
Thanks for persisting through the feedback :)

GreenAsJade:
@benjaminpjones   At the moment vscode forces me to do this:

    if (comparison) {
        return comparison;
    } else {
        return challenge_sort(A, B);
    }

Wheras I would rather prefer to be allowed to do this

    if (comparison) {
        return comparison;
    } 
   else {
        return challenge_sort(A, B);
    }

What will our stance be when we have prettier?    If we'll be allowed to do the latter, I'd love help persuading vscode :)

benjaminpjones:
>At the moment vscode forces me to do this...

You can thank eslint for that. See: https://github.com/online-go/online-go.com/blob/c0c6093a0343041bb75ca8843cadc5121f3c3570/.eslintrc.js#L66-L70

> Wheras I would rather prefer to be allowed to do this...

Stroustrup fan, eh? That's an option in eslint too (just not in this repo)

>What will our stance be when we have prettier? If we'll be allowed to do the latter, I'd love help persuading vscode :)

We don't have a choice.  The former is what Prettier will do. (https://github.com/prettier/prettier/issues/840)

Is there a reason outside of personal preference to use the latter?

Side note: if you put a comment in between the closing bracket and else, Prettier is okay with that:

```
    if (comparison) {
        return comparison;
    }
    // do this instead
    else {
        return challenge_sort(A, B);
    }
```

BHydden:
bracket alignment though... 🤔 

```
if (comparison)
{
    return comparison;
}
else
{
    return challenge_sort(A, B);
}
```

...jk

GreenAsJade:
Fascinating how tastes change.  Once upon a time I thought that was a Good Thing.

Now I find myself deleting the empty line after

if (comparison) {


if I see it.

PR #1597, opened 2022-01-07
Bring back ranks in gamelists

GreenAsJade:
Fixes [the disappearance of ranks in GameList Minigobans](https://forums.online-go.com/t/rengo-status/40415/138?u=greenasjade
)

## Proposed Changes

Don't pass names to MiniGoban, that's a bad idea.

Let Minigoban work it all out.

Note: this solution currently doesn't display ranks for Rengo games.    Minigoban needs a refactor to achieve that.


GreenAsJade:
OMG, my `devel` fell behind, but the only change here is the last commit!

anoek:
Thanks!

PR #1600, opened 2022-01-07
Subtle highlight whos turn

GreenAsJade:
Fixes the problem that you have to think twice to recognise who's turn it is.

Who wants to be reading the words that say who's turn it is - a visual cue will help avoid that.

## Proposed Changes

Subtle box-shadow changes for the player-container who's turn it is

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -341,4 +341,4 @@ window['preferences'] = preferences;
 window['player_cache'] = player_cache;
 window['GoMath'] = GoMath;
 
-console.log("Finished rendering from main, rengo 1.2.1");
+console.log("Finished rendering from main, rengo 1.2.2");
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -94,6 +94,10 @@ goban-view-bar-width=400px
         font-size: smaller;
     }
 
+    .player-container.their-turn {
+        box-shadow: 0 0px 30px rgb(150 150 150) !important;
+    }
+
     .play-controls {
         flex-grow: 0;
         flex-shrink: 0;
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -94,8 +94,12 @@ goban-view-bar-width=400px
         font-size: smaller;
     }
 
-    .player-container.their-turn {
-        box-shadow: 0 0px 30px rgb(150 150 150) !important;
+    .black.player-container.their-turn {
+        box-shadow: 4px 6px 20px rgb(100 100 100) !important;
+    }
+
+    .white.player-container.their-turn {
+        box-shadow: -2px 6px 20px rgb(130 130 130) !important;
     }
 
     .play-controls {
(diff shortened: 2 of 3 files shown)


GreenAsJade:
@BHydden  if I'm going to get commit-credits for all those, I might as well make hay while the sun shines eh!   :D

benjaminpjones:
This doesn't feel that subtle in analyze mode 🎉🕺

https://user-images.githubusercontent.com/25233703/148596714-fdddfd95-0ebb-40c1-ab1e-f4eedd886e5a.mp4

GreenAsJade:
Yeah that really sucks lets turn it off in analyze mode eh?

BHydden:
Rather than turning it off, can we just pause the effect when it is called quickly? It's fine in analyse mode if you're looking at each move as you step through, it only looks bad if you're scrolling for a position... can we add like a delay on the call or something?

benjaminpjones:
I'm not sure I'd want a delay unless it was some sort of animated thing (fade in shadow, go back if quick change)

But honestly now that I've gotten used to it, it's not so bad.  Just had a bad case of change aversion when I was playing some conditional moves.

GreenAsJade:
I'm open to any suggestions - it seems odd in some ways to finesse such a small thing (this shadow)( with elements like delays etc, but if it would "make it work" it feels worth it ... it just (to me) looks really slick and makes it totally clear whos move it is, where before I kept having to consciously read "White To Move" and look up at the cards, if I wasn't watching the last move played.

A difference with Analysis is surely you are conscious of the last move "played" because you played it in every case...

benjaminpjones:
I'm not sure if this is related, but after playing a couple blitz rengo games, what I really want is to figure out when it's my turn. All the stuff on the right doesn't feel enough for me because (a) I need to avert my eyes and (b) I still have to do a bit of reading to figure out who the active player is.

A couple wacky ideas:

- different sound when it's my turn
- Highlighted border around the board when it's my turn (much like mini gobans on my profile turn green)
- Modified cursor or preview stone
   - related: I recently noticed that the preview stone doesn't show up on your turn until you move your mouse.  Therefore my current solution is to hover my mouse over an empty intersection and jiggle the mouse whenever a move is played.

GreenAsJade:
> * Highlighted border around the board when it's my turn (much like mini gobans on my profile turn green)

I think this is practical and functional.

I wonder how pervasive it should be?     I'm thinking that it might be the default for blitz rengo, but other than that maybe something you can turn on?

benjaminpjones:
I'm not sure! Haven't played enough correspondence or slow live yet...

GreenAsJade:
Upon reflection, looking at a page with a 19x19 board on it, I don't think a border highlight is the right answer.   At fullscreen, the border is almost in peripheral vision :)

Is it a mouse cursor thing?

What about mobile?   🤦‍♀️

benjaminpjones:
Is it bad if it's in the periphery? Since it's just a binary "my turn"/"not my turn", a colorful signal at the edge is enough for me (after all I should be looking at the A and T lines!)  I'd be more worried about the border highlight going off screen.

Mouse cursor could be good too, but as you say mobile.  Although I will say blitz mobile is a bad idea no matter how you cut it...

benjaminpjones:
A bad idea (I've got plenty): tengen starts blinking when it's your turn

PR #1602, opened 2022-01-07
Fix issue caused by last change: a fully fetched player object is needed...

GreenAsJade:
... to get and compute rating to display

Fixes 

Broken ratings on GameList Minigobans

## Proposed Changes

 - Still use engine.players.black instead of black_player_id
 - Fetch the player structure and compute the rank from that


GreenAsJade:
Note: I do think the real solution is to use <Player>  in MiniGoban, this remains interim.

PR #1607, opened 2022-01-08
Sort rengo challenges by timesetting first (and undebug ChallengeModal)

GreenAsJade:
Fixes the problem of finding live rengo challenges

## Proposed Changes

Sort rengo challenge list by isLive() (which happens to include Blitz, IIUC)

### DIFF
--- a/src/views/Play/Play.tsx
+++ b/src/views/Play/Play.tsx
@@ -188,7 +188,7 @@ export class Play extends React.Component<{}, PlayState> {
         }
         live.sort(challenge_sort);
         corr.sort(challenge_sort);
-        rengo.sort(timesetting_challenge_sort);
+        rengo.sort(time_per_move_challenge_sort);
 
         //console.log("list update...");
         this.setState({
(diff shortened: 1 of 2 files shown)


GreenAsJade:
Actually, this is not the best answer.  It should be sorting by time_per_move.

GreenAsJade:
Yeah, I think sorted by time_to_move is the most helpful.

src/views/Play/Play.tsx
```diff
@@ -180,15 +180,15 @@ export class Play extends React.Component<{}, PlayState> {
 
             if (C.rengo) {
                 rengo.push(C);
-            } else if (C.time_per_move > 0 && C.time_per_move < 3600) {
+            } else if (C.time_per_move > 0 && C.time_per_move < 3600) { // TBD: why aren't we using isLive here?
```
anoek, line 183: likely because it predated isLIve?

GreenAsJade: Makes sense - if we're happy with that, I'll clean next time through

PR #1608, opened 2022-01-08
Support "Join (nominate for) a Rengo Challenge from the SeekGraph

GreenAsJade:
Fixes the problem of people trying to do that and causing all sorts of havoc.

## Proposed Changes

Make SeekGraph know about the case where the dot is a rengo game, and whether the player has joined or not, and give them the join button if they have not already.

### DIFF
--- a/src/components/SeekGraph/SeekGraph.tsx
+++ b/src/components/SeekGraph/SeekGraph.tsx
@@ -32,6 +32,7 @@ import {kb_bind, kb_unbind} from "KBShortcut";
 import {Player} from "Player";
 import * as player_cache from "player_cache";
 import swal from 'sweetalert2';
+import { nominateForRengoChallenge } from "Play";
 
 interface Events {
     "challenges": Array<any>;
@@ -188,6 +189,9 @@ export class SeekGraph extends TypedEventEmitter<Events> {
                 }
             } else {
                 e.user_challenge = false;
+                e.joined_rengo = e.rengo && e.rengo_participants.includes(this_userid);
+
+                console.log(e);
                 if (data.get("user").anonymous) {
                     e.eligible = false;
                     e.ineligible_reason = _("Not logged in");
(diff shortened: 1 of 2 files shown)


GreenAsJade:
Note: the back end still should not have thrown in the way it did: needs to be tidied up.

GreenAsJade:
I kind of take this back.  If the client stupidly sends an `accept` for a rengo game, of course it is going to explode.

The only thing that does need consideration is whether we should be sending the raw error message back in the `error` property of the payload, resulting in that being displayed to the user.

Possibly this is _also_ a client problem: the catch of a 400 probably should just say "Ooops something went wrong".

I might do that...

GreenAsJade:
... or I might wait for advice :)    I see a fair bit of thought has gone into this (`errorAlerter` and friends) so I don't want to reinvent any wheels!

anoek:
Re `errorAlerter`, I think it's due for a bit of a freshen up. There's nothing friendly about an error box, but it could be friendlier than what it is now. I'd reckon some more generic "The server had an error doing what you want it to do" kind of message, and then stuff the more developer-useful details into a secondary `<pre>` block that's clear that the user doesn't need to care about this (But if they post a screenshot somewhere, someone who knows what they're looking at might be able to make sense of it or something)` kind of a thing might be a good general improvement.

IIRC most of the code in `errorAlerter` is just trying to guess how to extract useful information out of whatever blob it's passed, could be an exception, a json blob from the server, some string, etc - so most of that code could remain untouched, just gotta work out a better display, which likely involves doing something other than `swal`, but rather some new modal or something of the sort.

benjaminpjones:
>The only thing that does need consideration is whether we should be sending the raw error message back in the error property of the payload, resulting in that being displayed to the user.

Regardless of what happens to `errorAlerter` display, I think a more specific error would be helpful too.  `something something rengo accept/` seems easier to pin down than `something something NoneType`, at least later when rengo won't be the first thought

GreenAsJade:
Conclusion:

1) Clean up the back-end where it currently just sends the string of the thrown error 
     (These are catch-all clauses where we don't have any clue what is going, so the best that can be done is to report where we think we are and that something went wrong deeper inside)

2) Clean up the front ent (`errorAlerter` and friends).

BUT most of the time `errorAlerter` will actually have a helpful message, because most of the time the back-end is like this:

`return Response(
                {"error": "Your rank is too high to accept this challenge"}, status=status.HTTP_403_FORBIDDEN
            )
`

the trouble comes from this:

`       try:
            challenge.accept(request.data);

         except Exception as e:
            return Response({"error": str(e)}, status=status.HTTP_400_BAD_REQUEST)
`

... which seems to need the back-end solution: send a generic user friendly message in "error" and maybe even more detail than str(e) in some other field.

PR #1615, opened 2022-01-10
Remove two string refs from the Group News table.

### LINKED ISSUE #1614: Editing a news post in a group causes a Red bug page
**Describe the bug**
There is a bug with editing news posts in groups

**To Reproduce**
1. Be an admin in a group
2. Create or find a news post
3. Try to edit that news post

**Screenshots**
See https://forums.online-go.com/t/unable-to-edit-news-post-in-the-group/40755?u=shinuito

**Additional context**
The full message when following the link reads:

Element ref was specified as a string (editing_news_body) but no owner was set. This could happen for one of the following reasons: 1. You may be adding a ref to a function component 2. You may be adding a ref to a component that was not created inside a component's render method 3. You have multiple copies of React loaded See https://reactjs.org/link/refs-must-have-owner for more information.

benjaminpjones:
Fixes #1614 

## Proposed Changes

  - Remove two string refs (I could not find associated ref variables, so I don't think this will have a functional effect besides fixing the error)

Related: Perhaps the "Loading..." overlay should be suppressed for news editing.  It is displayed every time the user types:

Screen Shot 2022-01-10 at 8 26 18 AM

### DIFF
--- a/src/views/Group/Group.tsx
+++ b/src/views/Group/Group.tsx
@@ -655,7 +655,7 @@ export class Group extends React.PureComponent<GroupProperties, GroupState> {
                                     {header: _("News"), className: "none", render: (entry) =>
                                         <div>
                                             {this.state.editing_news && this.state.editing_news.id === entry.id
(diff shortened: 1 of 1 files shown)


anoek:
Awesome, thanks for the fix, I'll work on the loading thing

PR #1616, opened 2022-01-10
Try to fix display issue for old Apple devices

shinuito:
Old Apple devices and/or versions of Safari seemed to have issue with this, meaning buttons went under the player card, or the board went over the player card

Fixes https://forums.online-go.com/t/board-position-covers-names-rank-score/40679?u=shinuito

Tested with Xcode simulating iOS 12.4 below.

Further tests with updated browsers seem to indicate that it doesn't change the layout on recent Safari, Chrome, Firefox on Mac for example. Likely a remnant of css from rengo from before the team members were moved from below the player cards to inside them.

(images are a bit choppy, trying to remove backgrounds, poorly, from screenshots, and poorly scaling to not take up too much space in Github).

iPad Air

and in portrait mode

### DIFF
--- a/src/views/Game/Players.styl
+++ b/src/views/Game/Players.styl
@@ -38,8 +38,6 @@
 }
 
 .MainGobanView .players {
-    display: flex;
-    flex-direction: column;
 
     .rengo-team-members {
         display: flex;


anoek:
Thanks for hunting that down!

PR #1617, opened 2022-01-11
Re-order key moves in AI review by move_number

philipkd:
## Proposed Change

- re-orders the key moves in the AI Review to be in the move order, not based on the review's internal metric

before:

before-anno

after:

after-anno

## Context

@anoek had said in #1589 that he wanted to "help users who first see this intuitively understand what these bubbles with letter/number combinations mean."

This PR doesn't necessarily fix this, but having the key moves and the red dots in the same order further drives home the point that they refer to the same thing.

Players can also click through to see the story of the game unfold.

## Considered Alternatives

We don't show the underlying metric that the AI Review uses to score these moves to the user, nor would it be clear if we did. (I saw some healthy debate in the forums on ways to score these moves.)

## Code Notes

I had intended to get this into #1589, but I didn't want to change too much at once.


anoek:
Yeah this is probably good. The original metric they're sorted by is by worst move I thought, so biggest drops first - but that's not always really great, and it probably just makes more sense like this.

Thanks for thinking about it and working on it!

PR #1619, opened 2022-01-11
Alternative display for highlighting whose turn it is

anoek:
This patch replaces the drop shadow with an underscored blue line with a
small arrow as the turn indicator, up for aesthetic discussion. I'm not sure this is the best it can be, so if anyone has any ideas lets hear them :)

Relates to #1600

There's some places the current drop shadow looks a bit funky, so this PR is attempting to fix those with an alternative look.

## Old issues: 

1. Shadow overlaps neighboring player box (Might be fixable with a z index thing, I'm not sure)
image

2. Shadow gets clipped as it attempts to extend past the bounds of the containing box (I don't think this is fixable, not easily at anyrate). It also drops shadow deep into the "Black to Move" text, which could be fixed with margin, but that has some downsides too.
image

3. Similarly, on non rengo games the whose turn is it is not present, so there's not much room to see whose turn it is
image
image

## Proposed Changes:

This PR replaces the shadow with a blue line and triangle (triangle only in landscape mode, not portrait):

image

image

image

image

  -
  -
  -

### DIFF
--- a/src/views/Game/AIReview.styl
+++ b/src/views/Game/AIReview.styl
@@ -16,6 +16,7 @@
  */
 
 .AIReview {
+    margin-top: 0.7rem;
     .reviewing, .pending {
         text-align: center;
         width: 100%;
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -94,12 +94,34 @@ goban-view-bar-width=400px
         font-size: smaller;
     }
 
-    .black.player-container.their-turn {
-        box-shadow: 4px 6px 20px rgb(100 100 100) !important;
+
+    .black.player-container.their-turn, .white.player-container.their-turn {
+        position: relative;
+
+        &::after {
+            position: absolute;
+            content: "▲";
+            border-bottom: 0.15rem solid #000;
+            height: 0.15rem;;
+            bottom: -0.25rem;
+            themed border-color primary
+            themed color primary
+            font-size: 0.6rem;
+            display: block;
+            z-index: 1000;
+            width: 100%;
+            left: 0;
+            right: 0.1rem;
+        }
     }
 
-    .white.player-container.their-turn {
-        box-shadow: -2px 6px 20px rgb(130 130 130) !important;
+    &.portrait {
+        .black.player-container.their-turn, .white.player-container.their-turn {
+            &::after {
+                // in portrait mode, the arrow overlaps the board
+                content: "";
+            }
+        }
     }
 
     .play-controls {
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -96,32 +96,7 @@ goban-view-bar-width=400px
 
 
     .black.player-container.their-turn, .white.player-container.their-turn {
-        position: relative;
-
-        &::after {
-            position: absolute;
-            content: "▲";
-            border-bottom: 0.15rem solid #000;
-            height: 0.15rem;;
-            bottom: -0.25rem;
-            themed border-color primary
-            themed color primary
-            font-size: 0.6rem;
-            display: block;
-            z-index: 1000;
-            width: 100%;
-            left: 0;
-            right: 0.1rem;
-        }
-    }
-
-    &.portrait {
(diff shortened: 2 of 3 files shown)


BHydden:
I don't want to be that guy that criticises things without having a better option, but isn't that a little tacky?

benjaminpjones:
@BHydden that's not actionable criticism at all...

anoek:
> I don't want to be that guy that criticises things without having a better option, but isn't that a little tacky?

Perhaps so? Do you find the drop shadow tacky too?

BHydden:
@benjaminpjones exactly my point

@anoek I did like it initially, but I see now your examples of times where it clips and such and agree that's certainly not ideal either... UI is not my strong suit. Umm, how do I describe it... I guess the blue bar feels sort of stuck on, like it doesn't quite match our design... (it's also only 4am which might be part of why words are difficult to compose)

anoek:
Fair enough, yeah I'm not sure I'm super happy with either style at the moment, hence this feedback process.

BHydden:
Whenever I'm stuck on UI I ask myself "what did Lichess do?" but it's extra tricky in this instance.
Since they have fixed sides they can use top and bottom for their timers, freeing them up to use colour for turn...
but we need to use colour for identification since Go has no fixed sides... hmm... I wonder, what if the person whose turn it was was always, say, on the left? or something like that... and the whole player card could maybe switch sides depending on whose turn it was... could that maybe work?

anoek:
In general I'm not a fan of moving ui elements like that, especially during play as it's kind of a big jolt. Also, there's mobile to consider.

BHydden:
What if we combine the two? A shadow bar instead of a shadow box? softer than the blue line, less clipping than the full box around the whole border...

anoek:
Something like this?

image

BHydden:
Yeah... I think that looks pretty decent... Shade, gradient, and thickness I think could maybe all be tweaked a bit but that's splitting hairs. Overall if this avoids the clipping issues you raised above I feel this solution fits better with our overall design theme/motif ❤

benjaminpjones:
@anoek what does the most recent example look like in light mode?  The reason I ask is that light mode already has shadows as part of the styling, so it may be hard to tell the difference between style and information (which is one reason I like the blue).

Screen Shot 2022-01-11 at 4 49 15 PM

Both those options are improvements.  However, I guess I'm in the camp of "do we need an indicator on the player cards at all?"  In that general area we already have "Black to play" and "White to play", and I've never felt lost as to which color is up next because we always have the last move indicator:

Screen Shot 2022-01-11 at 4 52 42 PM

BHydden:
> light mode already has shadows as part of the styling, so it may be hard to tell the difference between style and information

Easy fix, just remove the light theme and make dark accessible the only theme available 😁😉

> we always have the last move indicator
Yeah this is a good point, but as long as it's not too much trouble I think some kind of tracking on the player card is a nice touch 🙂

anoek:
image

image

Pretty hard to tell for black.. hmm.

shinuito:
I do find it hard to come up with alternate suggestions. I can't say UI is my strong suit either.

I'm not particularly against the blue underline. I thought the current shadow was a bit jarring on mobile. That is it stood out to me and was obvious it wasn't there before. Maybe mildly distracting to see something changing at the top of the mobile screen while trying to focus on the board (I suppose that might apply to a blue underline either, not sure without trying it out.)

In any case though, I wasn't against the current shading method, more that it stood out. 

One thing I will say about it is that if you cycle through the moves, pause to look at the board or look away and come back, it's not clear whose move it is just from the shading. That is I can't tell from the shadow whose move it is because it's a bit vague and the UI has shadows all over the place anyway. I think it only works well while you're changing moves (looking back through the game in analysis etc)

I probably prefer the underline with the arrow to say a  full blue border for example (at least I can't think of a way to make it look nice :))

anoek:
Alrighty, I switched to the shadow, I don't know if it's enough for those that want it or too distracting for those that don't, at least it doesn't clip though. Future improvements and alterations are welcome, it might even be something we want a preference option for in settings.

PR #1620, opened 2022-01-11
Prettier actual

benjaminpjones:
I'm realizing there may never be a clear window to put this change in.  But that's okay!  The following commands can be used to step over merge conflicts:

```
 git fetch <OFFICIAL REPO>
 git merge <OFFICIAL REPO>/devel --strategy-option ours
 npm run prettier
 git add -p
 git commit
```

Edit: I guess if multiple people are working on the same files `--strategy-option ours` could be a problem.  I don't think that is the case right now, but the above commands should be used with caution.

## Proposed Changes

  - Update prettier config
  - Run prettier globally
  - Remove formatting rules from ESLint
  - Add prettier to CI and lint-staged

Note: I haven't done any gulp stuff, but [gulp-prettier](https://www.npmjs.com/package/gulp-prettier) does exist.


anoek:
Awesome. Let me complete my billing system switchover (I'm waiting on one thing from the new payment processor, it should happen today or tomorrow though I think), and sync up with @GreenAsJade to make sure he's not in the middle of some rengo UI stuff, and then we can pull the trigger.

Instead of doing merges post-write, which seems like it would get ugly fast, does it make sense to make a PR that has everything except the `--write`? Then when it's time to merge, we merge all the other pending PRs, merge that PR, do a `--write` and commit?

benjaminpjones:
I think that works!

anoek:
It'll fail but that's ok, I can still hit merge :)

anoek:
oh because it blocks you from committing huh, noted - well I reckon I can cherry-pick from this commit ? Or you can, if you want to do the actual merge commit

benjaminpjones:
>It'll fail but that's ok, I can still hit merge :)

Oh huh good point.  Well anyway, I opened #1621 before I saw these so probably easiest to work from there...

>oh because it blocks you from committing huh, noted - well I reckon I can cherry-pick from this commit ? Or you can, if you want to do the actual write commit

precommit can be bypassed with `--no-verify` so that's not a problem either

As far as who does the actual write commit, that's fine if you run it!  I have no strong desire to be all over the git history haha

PR #1621, opened 2022-01-12
Prettier everything but write

benjaminpjones:
This PR should make it possible to run prettier globally when we have an open window.  To perform this global write, this command should work:

```
    npm run prettier
```

## Proposed Changes

  - Remove unused and un-recommended rules from the Prettier config.
  - Remove ESLint formatting rules (and fix the duplicate `extends` field)
  - Disable eslint on lines that will trigger `id-denylist`
  - Add prettier to `lint-staged` and a (commented out) step in the GitHub workflow

## Things I haven't done
  - Turn on the prettier check in GitHub workflow
  - Run prettier --write (obviously 🙂)
  - Added IDE specific configs

### DIFF
--- a/src/views/Ladder/Ladder.tsx
+++ b/src/views/Ladder/Ladder.tsx
@@ -538,14 +538,34 @@ export class LadderRow extends React.Component<LadderRowProperties, LadderRowSta
 function canChallengeTooltip(obj: any): string {
     if (obj.reason_code) {
         switch (obj.reason_code) {
-            case 0x001: return pgettext("Can't challenge player in ladder because: ", "Can't challenge yourself");
-            case 0x002: return pgettext("Can't challenge player in ladder because: ", "Player is a lower rank than you");
-            case 0x003: return pgettext("Can't challenge player in ladder because: ", "Player is not in the ladder");
-            case 0x004: return pgettext("Can't challenge player in ladder because: ", "Player's rank is too high");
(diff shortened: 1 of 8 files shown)


anoek (approved):
Thanks for doing all the legwork for this!

src/views/Ladder/Ladder.tsx
```diff
@@ -538,14 +538,34 @@ export class LadderRow extends React.Component<LadderRowProperties, LadderRowSta
...
+                return pgettext("Can't challenge player in ladder because: ", "Player is not in the ladder");
+            case 0x004:
+                return pgettext("Can't challenge player in ladder because: ", "Player's rank is too high");
+            case 0x005:
+                return interpolate(
+                    pgettext(
+                        "Can't challenge player in ladder because: ",
+                        "Already playing {{number}} games you've initiated",
+                    ),
+                    { number: obj.reason_parameter },  // eslint-disable-line id-denylist
```
benjaminpjones, line 555: Here (and elsewhere) I ran `Format Selection`, then added the `eslint-disable` comment.  `Format Selection` caused the whole switch block to be formatted.  Hopefully that explains why the delta on this hunk is so high, even though the salient part is the `// eslint-disable...`

benjaminpjones:
My pleasure @anoek!  Thanks for working with me on such a large scale change :)

PR #1623, opened 2022-01-12
Bugfix: you should always be able to turn off rengo, and ranked

GreenAsJade:
... in the Challenge modal

Fixes a lockup Sofia found due to a previous bug.

IE she had found the way to get ranked and rengo checked, and this was saved in her challenge modal data.

## Proposed Changes

Don't disable ranked or rengo if they are checked.

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -688,7 +688,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                     <div className="checkbox">
                         <input type="checkbox"
                             id="rengo-option"
-                            disabled={this.state.challenge.game.private || this.state.challenge.game.ranked}
+                            disabled={!this.state.challenge.game.rengo && (this.state.challenge.game.private || this.state.challenge.game.ranked) }
                             checked={this.state.challenge.game.rengo} onChange={this.update_rengo}/>
                     </div>
                 </div>
@@ -714,7 +714,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                             <div className="checkbox">
                                 <input type="checkbox"
                                     id="challenge-ranked"
-                                    disabled={this.state.challenge.game.private || this.state.challenge.game.rengo}
+                                    disabled={!this.state.challenge.game.ranked && (this.state.challenge.game.private || this.state.challenge.game.rengo)}
                                     checked={this.state.challenge.game.ranked} onChange={this.update_ranked}/>
                             </div>
                         </div>


benjaminpjones:
Not disabling these means that someone could (attempt to) create a ranked rengo match right?

Wouldn't it make more sense to either un-rank the game or un-rengo the game when the saved config is mismatched like this?

P.S. how in the world 🤔

benjaminpjones:
>you cant turn ranked on if you have rengo turned on

If this is true, why allow an invalid state at all?  Granted this is a weird edge case, but why not "fix" the error for them instead of leaving both on and allowing the user to send a broken API call?

Here is the journey I'm thinking of:

1. User *somehow* has both `challenge.game.rengo` and `challenge.game.ranked` set to true.
2. OGS doesn't fix it for them (or mention there's a problem)
3. User changes some settings (but doesn't fix the ranked/rengo conflict) and submits a challenge
4. Server responds "you can't do that."
5. Return to step one (hopefully user knows what to do now)

ALTERNATIVELY

1. User *somehow* has both `challenge.game.rengo` and `challenge.game.ranked` set to true.
2. OGS *does* fix it for them.
3. User changes some settings and submits a challenge.
4. Server responds with success and a challenge is created.

If the server is going to tell them no anyway, why not tell them earlier?

GreenAsJade:
Ah true.   OK maybe I'll do that instead - or you could submit the alternative, if you're on it?

(Independent of what we do front end, the server has to reject it, because we can't control what other clients do.   I think the server in fact accepts but ignores the ranked, but I have become nervous about that, and will "do something" shortly)

BHydden:
If they're mutually exclusive, perhaps they could be on a
ranked/neither/rengo trinery switch?
(First time replying through email, see if this works)

GreenAsJade:
(reply from email, to see if it works)

There's some fancy logic in ChallengeModal to support this sort of thing.
 That's what I initially tried, but the evidence was that I didn't get it
right, since Sofia achieved ranked rengo.

Rather than dig further into that, I saw that this disable solution
appeared to be in place for private/ranked so I copied it.   But of course
that didn't suffer from prior testing by Sofia :D

benjaminpjones:
> OK maybe I'll do that instead - or you could submit the alternative, if you're on it?

@GreenAsJade something like [this](https://github.com/online-go/online-go.com/pull/1626)?

GreenAsJade:
Looking more closely how ranked and private were handled, I think that in my original submission I overlooked the interaction between this part:

https://github.com/online-go/online-go.com/blob/a5ddd80704098be38e7f8b3119912ec3da6def90/src/components/ChallengeModal/ChallengeModal.tsx#L563

and the actual button disabling part.  These have to work together.

Looking again now, I think the logic in this PR still makes sense, because it sends a clear message about what is possible and what is not _in addition_ to #1626 ... the latter makes sure we don't leave a challenge in a bad state.

benjaminpjones:
Oh huh, yeah I didn't really look at those.

One thing I am noticing (and this has probably been like this for a while): there is some asymmetry in the **Private**/**Ranked** interaction.  I'm not sure if there is a good reason for that, but here is what I'm talking about:

- Checking **Private** disables **Ranked**
- Checking **Ranked** does *not* disable **Private**
    - If someone checks **Private** when the game is **Ranked**, **Ranked** will become unchecked

GreenAsJade:
That's the sort of thing that was making my head spin :)  

FWIW private and ranked are asymmetrical for one reason being that setRanked is special (lots of side effects take care of).   With that being the case, I could see where the asymmetry started.

anoek:
Thanks!

PR #1624, opened 2022-01-13
Re-enable some eslint rules

dsaxton:
Thanks @anoek for the excellent project, huge fan.

This PR just re-enables a few eslint rules, partly by disabling by line instead of project-wide. Let me know if you'd like any changes (e.g., is it fine to blanket allow empty catch clauses? That was almost all the no-empty errors.).

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -733,6 +733,9 @@ class NotificationEntry extends React.Component<{notification}, any> {
 
     renderNotification() {
         const notification = this.props.notification;
+        let outcome: any;
+        let now: number;
+        let left: number;
 
         switch (notification.type) {
             case "test":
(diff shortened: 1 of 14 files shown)


src/components/Notifications/Notifications.tsx
```diff
@@ -733,6 +734,9 @@ class NotificationEntry extends React.Component<{notification}, any> {
 
     renderNotification() {
         const notification = this.props.notification;
+        let outcome: any;
+        let now: number;
```
dsaxton, line 738: TypeScript newb here, but is this technically correct? Or should it be something like `number | undefined` (and same with below)?

anoek: Typescript understands that it's initially undefined. With member fields of a class, you have to be more specific if you want them to be able to be undefined (otherwise it expects them to be defined within the constructor), but for locally scoped variables such as these, they're undefined if you don't define them, simple enough.

src/components/Notifications/Notifications.tsx
```diff
@@ -796,8 +800,8 @@ class NotificationEntry extends React.Component<{notification}, any> {
                 );
 
             case "timecop":
-                const now = (Date.now()) / 1000;
-                const left = Math.floor(notification.time / 1000 - now);
+                now = (Date.now()) / 1000;
```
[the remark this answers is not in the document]
dsaxton: Nice, it works. :+1:  I agree the other way was a tad awkward.

anoek:
Awesome, thanks for the improvements!

PR #1626, opened 2022-01-13
Don't allow challenges to be rengo if they are ranked or private.

benjaminpjones:
The alternative discussed in #1623

## Proposed changes

There is already some handling in the constructor of ChallengeModal for bogus configs.  This just adds the constraint "rengo must not be private or ranked."


anoek:
Indeed, thanks!

PR #1627, opened 2022-01-13
Support opening concurrent rengo management panes

GreenAsJade:
## Proposed Changes

Make show_rengo_management_pane an array, and open the pane for each entry in the array

Screen Shot 2022-01-13 at 6 09 02 pm (3)


GreenAsJade:
As discussed, this needs a decision about what to do about `<colspan>` needs `<table>`

anoek:
I don't mind table. Ideally this page would be great on mobile but the current stuff doesn't do anything fancy either for mobile, and that's quite the undertaking, so for now, whatever.

I pushed up some changes to fix up the warnings and errors, but there's an unexplainable gap on the right side of the table, I'm not sure where it's coming from:
image

src/views/Play/Play.tsx
```diff
@@ -956,55 +968,50 @@ export class Play extends React.Component<{}, PlayState> {
...
     };
 
-    stopManagingRengoChallenge = () => {
-        this.setState({show_in_rengo_management_pane: 0});
-    };
-
     rengoManageListItem = (C, user) => {
         return (
-            <tr key={C.challenge_id} className={"challenge-row rengo-management-row"}>
+            <tr key={`manage-list-item-${C.challenge_id}`} className={"challenge-row rengo-management-row"}>
```
benjaminpjones, line 1007: Might be my own React ignorance, but is this key (`manage-list-item-<id>`, `rengo-list-item-<id>`) necessary? It seems like since the `React.Fragment` that wraps this is keyed on Challenge ID, there is no reason to key its children...

benjaminpjones: Looking at the commit history, I realize this question might be for @anoek

src/views/Play/Play.tsx
```diff
@@ -545,9 +551,16 @@ export class Play extends React.Component<{}, PlayState> {
                                 {this.cellBreaks(7)}
```
benjaminpjones, line 551: It looks like these cell breaks are pushing everything to the left

Screen Shot 2022-01-13 at 8 43 16 PM

src/views/Play/Play.tsx
GreenAsJade: I think you're correct that the conversion from `<>` to `<React.Fragment>` does obviate the need for the inside keys.

GreenAsJade: That said ... initially each of those two components were an alternate in that slot, which is why they were originally keyed by challenge id.    anoek made them both show at the same time  (very good move, why didn't I think of that!?), ay which point keying at the top became possible.

The thing is, each component is clearly designed to be instantiated in a list.   From a reuse and design perspective, doesn't this mean that they should key themselves anyhow?   I'm not sure it feels right to be putting in and taking out keys based on where something is being used...

anoek: Yeah it's no longer necessary. Though it doesn't hurt if they are there and not needed, no need to take them out and put them back if you move stuff around or whatever.

benjaminpjones: Perhaps this helps (from [Lists and Keys: Extracting Components with Keys](https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys)):

>Keys only make sense in the context of the surrounding array.
>
>For example, if you extract a `ListItem` component, you should keep the key on the <ListItem /> elements in the array rather than on the <li> element in the ListItem itself.

While the code in question isn't a component (although it could be!), what I get from this is that keys should generally be directly in the `map` function if possible

GreenAsJade: Lightglobe: you can put a key prop on anything, and React consumes it, you don't have to declare it on the component, so put the key in at the map level.   Thanks for pointing that out.   If either of us (or anyone) are in there fixing up the space thing you found, lets tidy that up too.

GreenAsJade: .... and if those things _had_ been pulled out as proper components, instead of inserted as functions, it all would have been a lot clearer what is going on.     ` {  this.renderSometing(whatever) }` probably should be deprecated in favour of <Something whatever={whatever}>  eh?   Is that the right conclusion?

GreenAsJade: ... which is easy to say, but somethings renderSomething just wants to call many functions in the component instantiating it :s

GreenAsJade: ... I just discovered something that I can't decide whether it's hideous or wonderful:


```            corre_list.map((C) => (
                (this.visibleInChallengeList(C) || null) &&
                    <React.Fragment key={C.challenge_id}>
                        <this.rengoListItem C={C} user={user} key={`rengo-list-{C.challenge_id}`}/>
                        {(this.state.show_in_rengo_management_pane.includes(C.challenge_id) || null) &&
                            <this.rengoManageListItem C={C} user={user} key={`rengo-manage-list-{C.challenge_id}`}/>
                        }
                    </React.Fragment>
            ))
```

... local components!?   They have access to `this.someFunction()`  😮

GreenAsJade: ... so you can do this:

```
            live_list.map((C) => (
                this.visibleInChallengeList(C) ?
                    <this.rengoListItem C={C} user={user} key={`rengo-list-item-{C.challenge_id}`}/>
                : null
            )),
```

... keyed as and when needed at the map level

benjaminpjones: Woah that's pretty funky... I like it :)

Those functions don't access any instance variables right? Maybe `static` is the way to go?  Then it would be `Play.RengoListItem` instead of `this.RengoListItem`

benjaminpjones: Ah nvm I see you already put this in.. cool stuff!

GreenAsJade: `Play.RengoListItem`  is nicer aesthetically - but I think it has to be `this.RengoListItem` because the reason for doing is is that you want to access functions that themselves update `state`, like `closePane` or `startChallenge` ...

anoek:
Nice, thanks!

PR #1630, opened 2022-01-13
Add stale action

dsaxton:
Adding an initial GitHub action for marking issues as stale. This is in "dry-run" mode right now (`debug-only: true`) so it won't actually modify issues yet (that way we can settle on what the best parameters would be beforehand).

Also, I think the `stale` label would need to already exist in order for this to work.

For reference: https://github.com/actions/stale


anoek:
Thanks for putting that together!

Issue #1629, opened 2022-01-13
Enable React.StrictMode and fix the complaints that ensue


anoek:
Ideally, we should be running under `React.StrictMode` ( https://reactjs.org/docs/strict-mode.html ) which will help identify deprecated react stuff and in general keep the code base "cleaner".

This can be turned on for the whole app by wrapping our `{routes}` in `<React.StrictMode>{routes}</React.StrictMode>` here: https://github.com/online-go/online-go.com/blob/devel/src/main.tsx#L337 and navigate around. 

However doing so will generate a number of warnings about deprecated string ref usage, use of `UNSAFE_` methods (which would probably be best fixed by porting those classes to functional react components), and various other things that aren't currently broken, but are considered deprecated or bad practice by the latest and greatest React system, so we should probably update as time and interest permits.

Note that only warnings about what components are activate in whatever you're currently viewing will be generated, so to see all warnings we'll eventually have to click and view everything to hunt every warning down.

Those interested in contributing to this effort need not take on fixing the whole code base in one PR, just turn it on for your PR, fix a thing or two that you notice and want to fix, and submit a PR for those fixes, reference this issue, and in time we'll whittle down the warnings and eventually be able to turn it on for the whole app.

PR #1631, opened 2022-01-14
Remove UNSAFE_componentWillMount

dsaxton:
It seems some of these may have a simple fix of moving into the constructor according to https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

xref: https://github.com/online-go/online-go.com/issues/1629

### DIFF
--- a/src/components/ACLModal/ACLModal.tsx
+++ b/src/components/ACLModal/ACLModal.tsx
@@ -60,11 +60,9 @@ export class ACLModal extends Modal<Events, ACLModalProperties, any> {
         } else {
             throw new Error(`ACLModal created with invalid parameters`);
         }
-    }
-
-    UNSAFE_componentWillMount() {
         this.refresh();
     }
+
     componentWillUnmount() {
     }
     refresh = () => {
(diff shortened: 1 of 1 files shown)


benjaminpjones:
Since `this.refresh()` makes call to the REST API, I would recommend putting that into the `componentDidMount` lifecycle hook.  From [the linked article](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data):

> The recommended upgrade path for most use cases is to move data-fetching into `componentDidMount`

This is because calling `setState` before initial render can lead to unexpected behavior.

benjaminpjones:
On a related note, best practice would be to abort the fetch in `componentWillUnmount`.  But that is out of scope for this PR, and I'm not sure [`abort_requests_in_flight()`](https://github.com/online-go/online-go.com/blob/01b152b88538a46dbba2e6eec1ad2185e78e6b48/src/lib/requests.ts#L196) provides the right support for this (ideally there would be a way to abort only requests that *this* component created).

dsaxton:
Thanks @benjaminpjones, I think that makes sense. Let me know if this looks better.

benjaminpjones:
Thanks @dsaxton! That looks good to me.

anoek:
Thanks for the fix!

PR #1633, opened 2022-01-15
Fix more strict mode warnings

dsaxton:
A few more of these that seem to work fine from what I can tell


anoek:
Awesome thanks for doing these :)

src/components/KBShortcut/KBShortcut.tsx
```diff
@@ -49,7 +49,7 @@ export class KBShortcut extends React.Component<KBProps> {
     shouldComponentUpdate() {
         return false;
     }
-    UNSAFE_componentWillReceiveProps(next_props) {
+    componentDidUpdate(next_props) {
```
anoek, line 52: A difference between `componentWillReceiveProps` and `componentDidUpdate` is that the first is passed `nextProps` and the second is passed `prevProps, prevState`, so it seems to me like maybe the rest of this function needs to switch to using `this.props` instead of the now mislabeled `next_props` (since it's really `prevProps`?) I think this might cause subtle bugs as it is if the component is reused (perhaps such as when changing between games or something of the sort, I'm not sure though on a place where it'd reliably be reused for testing)

anoek (changes requested):

src/components/KBShortcut/KBShortcut.tsx
dsaxton: Ah okay. Does the updated version look right?

src/components/Markdown/Markdown.tsx
```diff
@@ -207,7 +207,7 @@ export class Markdown extends React.PureComponent<MarkdownProps, { html }> {
         return source;
     }
 
-    UNSAFE_componentWillReceiveProps(next_props) {
+    componentDidUpate(next_props) {
         if (next_props.source !== this.props.source) {
```
dsaxton, line 211: Hmm, this is a weird one then. If we update `next_props` to become `this.props` as above then we end up with `this.props.source !== this.props.source`, which seems unlikely to be correct :) Maybe I need to better understand how this is working.

anoek: The comparison is still valid I think, it's just comparing with `this.props` as being the current value instead of the old value, and the parameter being passed in is now the old value instead of the value to be.

src/components/KBShortcut/KBShortcut.tsx
anoek: I think it does, but I'll want to bang on that a little before a merge

src/components/StarRating/StarRating.tsx
```diff
@@ -56,7 +56,7 @@ export class StarRating extends React.PureComponent<any, { rating }> {
         }
     }
 
-    UNSAFE_componentWillReceiveProps(nextProps) {
+    componentDidUpdate(nextProps) {
```
anoek, line 59: Lets change the variable name here to be `oldProps` or whatever, for the sake of future editors :)

anoek (approved):

src/components/Markdown/Markdown.tsx
```diff
@@ -207,11 +207,11 @@ export class Markdown extends React.PureComponent<MarkdownProps, { html }> {
         return source;
     }
 
-    UNSAFE_componentWillReceiveProps(next_props) {
-        if (next_props.source !== this.props.source) {
+    componentDidUpate(oldProps) {
```
benjaminpjones, line 210: Typo: `componentDidUpate` => `componentDidUpdate`

anoek: Good catch 👍

anoek:
Thanks for the fixes!

anoek:
#1629

PR #1634, opened 2022-01-15
Set up unused import eslint rule

dsaxton:
Setting up a rule to clean up some unused imports (in case there's interest in doing this) as well as removing a few. I didn't want to do too much to keep the diff small, but if we want it's easy to get rid of the rest.

Seems there's some overlap between this plugin and `@typescript-eslint/no-unused-vars` so the unused variable part can probably just be left off.

For reference: https://github.com/sweepline/eslint-plugin-unused-imports (for whatever reason it seems to work fine ignoring the part about React)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,8 @@ module.exports = {
     "plugins": [
         "eslint-plugin-jsdoc",
         "@typescript-eslint",
-        "@typescript-eslint/tslint"
+        "@typescript-eslint/tslint",
+        "unused-imports"
     ],
     "extends": [
         "eslint:recommended",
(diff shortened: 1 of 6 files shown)


.eslintrc.js
```diff
@@ -40,6 +41,8 @@ module.exports = {
         "no-self-assign": "off", // 2 errors
         "no-useless-escape": "off", // 24 errors
         "prefer-spread": "off", // 7 errors
+        "unused-imports/no-unused-imports": "off", // 301 errors
+        "unused-imports/no-unused-vars": "off",
```
anoek, line 45: What does `unused-imports` do for us when these are both off? It looks like these are the only two rules according to https://www.npmjs.com/package/eslint-plugin-unused-imports ?

anoek:
Oh I get it, this is testing the waters - Yeah I'm on board with getting rid of unused imports :)

dsaxton:
> Oh I get it, this is testing the waters - Yeah I'm on board with getting rid of unused imports :)

Right :smile: I just wanted to kind of propose this along with setting things up first, and then leave fixing them all as a follow-up potentially

PR #1635, opened 2022-01-15
Lint imports

dsaxton:
Updated `.eslintrc` to turn these into errors and then ran `npm run lint -- --fix`. As a sanity check everything appeared to compile / run afterwards.


anoek:
Awesome, thanks for the cleanup!

PR #1637, opened 2022-01-16
Enable prettier check in GitHub workflows.

benjaminpjones:
Follow up to 98bc1c8


benjaminpjones:
Heh I thought now would be a good time to completely refresh my local repo.  Forgot to `husky install`.  Hopefully I won't be making prettier/lint commits after this!

PR #1639, opened 2022-01-16
Have eslint run our prettier checks

anoek:
Having eslint run our prettier checks gets us some prettier output in
our gulp build loop window so we know when prettier is failing before a
commit.  The gulp-prettier plugin seemed to do the same but with less
information (no line numbers for instance, although I assume I just
didn't have it configured right), and overall it was slower, probably
because it had to run a full prettier check after a full eslint check,
and either eslint's caching is applied to the prettier checks when
eslint does it, or it's run in parallel, or both, I'm not sure - but
this seemed to be the better option, but I'm open to being corrected.

@benjaminpjones @dsaxton

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,8 @@ module.exports = {
         "eslint-plugin-jsdoc",
         "@typescript-eslint",
         "@typescript-eslint/tslint",
-        "unused-imports"
+        "unused-imports",
+        "prettier"
     ],
     "extends": [
         "eslint:recommended",
@@ -120,6 +121,7 @@ module.exports = {
             "error",
             { "allowNamedFunctions": true }
         ],
+        "prettier/prettier": "error",
         "unused-imports/no-unused-imports": "error",
         "use-isnan": "error",
         "@typescript-eslint/tslint/config": [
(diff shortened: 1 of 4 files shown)


dsaxton:
Nice :+1:

benjaminpjones:
I think this means slightly more red squigglies for me, but I just tried editing with this turned on and didn't find it bothersome.

If performance is still an issue (full eslint takes about 14 secs on my machine), one thing I can look into is (re)implementing incremental build support.  The old TSLint gulp code had something like this:

```
        gulp.src(ts_sources, {base: '.'})
        .pipe(gulpTsLint({
            //formatter: "prose",
            configuration: "./tslint.json",
            formatter: "stylish",
            program: program,
        }))
        .pipe(gulpTsLint.report({
            emitError: false,
            reportLimit: 0,
            summarizeFailureOutput: false
        }))
```

I didn't copy that over because I didn't understand what it was doing , but it may have been the thing that enabled these heavier tasks to run quickly.

anoek:
Oh interesting, yeah if there's an incremental system that would be awesome

benjaminpjones:
I don't know exactly how to do it, so don't let it be a blocker for this change. Just thought it was worth mentioning as an avenue for speeding gulp up. I shall investigate!

PR #1640, opened 2022-01-16
Remove unused imports plugin

dsaxton:
@anoek I just realized that `@typescript-eslint/no-unused-vars` actually covers unused imports as well (hence why the number of errors from that plugin went down so much after cleaning those up) so actually I think we can remove `unused-imports/no-unused-imports`. :man_facepalming: 

(Adding an extra import and running with both turned on shows output below)

```
   21:26  error  'api1ify' is defined but never used                     @typescript-eslint/no-unused-vars
   21:26  error  'api1ify' is defined but never used                     unused-imports/no-unused-imports
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,8 +11,7 @@ module.exports = {
     "plugins": [
         "eslint-plugin-jsdoc",
         "@typescript-eslint",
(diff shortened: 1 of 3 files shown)


anoek:
Awesome, thanks for the de-dup :)

PR #1641, opened 2022-01-16
Remove unused vars

dsaxton:
This one definitely requires a closer look from someone with more context than me, but in any case this gets rid of almost all the remaining unused variables errors and ignores the rest (we can use an underscore prefix if for whatever reason a variable needs to be there, like if we need an argument in a certain position within an argument list or something like that).

In some cases I wasn't sure if I was glossing over possible bugs, for example situations where state was being intentionally set in more than one place but seemingly never used, or arguments passed and then ignored. Anyways everything seems to be compiling / running for me locally, but there are a lot of changes here and easy for a mistake to creep in.

Edit: @anoek if you'd prefer I break this into smaller pieces to reduce risk and make things easier to review we can do that as well.

image


src/components/Chat/ChatIndicator.tsx
```diff
@@ -52,16 +52,12 @@ function onChatSubscribeGroupMentionsChange(pref) {
     chat_subscribe_new_group_chat_mentioned = pref;
     event_emiter.emit("subscription_changed");
 }
-let chat_subscribe_new_tournament_chat_messages = false;
 preferences.watch("chat-subscribe-tournament-chat-unread", onChatSubscribeTournamentMessageChange);
-function onChatSubscribeTournamentMessageChange(pref) {
-    chat_subscribe_new_tournament_chat_messages = pref;
```
benjaminpjones, line 58: This one actually does look like a bug to me.  See:

https://github.com/online-go/online-go.com/blob/29b7a1160d6634fa66c9fa0f3b93d201f7ee8b79/src/components/Chat/ChatIndicator.tsx#L68-L79

`chat_subscribe_new_tournament_chat_messages` is returned twice!

benjaminpjones (commented):
Seeing all the dead code going away brings major joy 😄 

It's hard for me to say whether the whole PR is 100% safe or addressing existing bugs, but if there are hidden bugs, it seems like all the more reason to turn the lint rule on so we can catch the bugs early.

src/components/Chat/ChatIndicator.tsx
benjaminpjones: Tagging @flovo for more context.

src/components/Chat/ChatIndicator.tsx
```diff
@@ -52,16 +52,12 @@ function onChatSubscribeGroupMentionsChange(pref) {
...
+function onChatSubscribeTournamentMessageChange() {
     event_emiter.emit("subscription_changed");
 }
-let chat_subscribe_new_tournament_chat_mentioned = false;
 preferences.watch("chat-subscribe-tournament-mentions", onChatSubscribeTournamentMentionsChange);
-function onChatSubscribeTournamentMentionsChange(pref) {
-    chat_subscribe_new_tournament_chat_mentioned = pref;
```
benjaminpjones, line 64: same here (see `getMentionedChatPreference()`)

src/components/PaginatedTable/PaginatedTable.tsx
```diff
@@ -35,18 +35,15 @@ interface Filter {
     [key: string]: string | number | boolean;
 }
 
-interface PagedResults<EntryT = any> {
```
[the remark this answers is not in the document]
dsaxton: Ah that would make more sense. Would you recommend updating in this way rather than removing the `EntryT` type?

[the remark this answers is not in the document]
benjaminpjones: Only if it's not too much trouble. That way `SourceFunction` would not lose its template argument as well 😀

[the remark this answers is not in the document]
dsaxton: Good point, got a bit aggressive with the delete button

src/components/PaginatedTable/PaginatedTable.tsx
```diff
@@ -37,7 +37,7 @@ interface Filter {
 
 interface PagedResults<EntryT = any> {
     count: number; // total results
-    results: Array<any>;
+    results: Array<EntryT>;
 }
 
 type SourceFunction<EntryT> = (
```
dsaxton, line 43: Just guessing here, but should `sorting` also have type `Array<EntryT>` below or is it correct as is?

benjaminpjones: I think it's correct as-is.  `order_by` (a `string[]`) is passed as `sorting` when `props.source` is called. (`order_by` specifies the ordering for the table. example: on the [**Puzzles**](https://online-go.com/puzzles) page, if I sort on the rank column, `order_by = ['min_rank', 'max_rank']`)

https://github.com/online-go/online-go.com/blob/29b7a1160d6634fa66c9fa0f3b93d201f7ee8b79/src/components/PaginatedTable/PaginatedTable.tsx#L134

The funny thing though is looking into it, I couldn't find any examples of `SourceFunction` in action.  It seems like all existing `PaginatedTable`s have `props.source` set as a string, not a function.  Perhaps the function option is just used for testing.  @anoek may be able to give insight.

anoek: Yeah it doesn't look like we ever use source as a function. Could have been that we did once upon a time, or that I thought we might when I was originally implementing that - but we don't, and it doesn't seem particularly useful so I'd say it's safe to remove and just use strings if that makes things easier.

src/components/Chat/ChatIndicator.tsx
anoek: I'm 99% sure that's a bug, so yeah let's not remove that variable and instead fix the usage.

src/components/Resizable/Resizable.tsx
```diff
@@ -17,15 +17,6 @@
 
 import * as React from "react";
 
-interface ResizableProperties {
```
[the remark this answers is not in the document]
anoek: Seconded, let's definitely swap that in instead of removing it.

anoek (changes requested):
Thanks a ton for the effort @dsaxton , and thanks a ton for the review @benjaminpjones , I always feel better about big changes when I see you've gone over it, you're a better reviewer than I :)

Shall we go all in and turn on `noUnusedLocals` in `tsconfig.json` too? https://github.com/online-go/online-go.com/blob/devel/tsconfig.json#L36

It's mostly redundant but it did catch one private class field as being unused

src/components/Chat/ChatIndicator.tsx
dsaxton: Another good catch, went ahead and updated

src/components/PaginatedTable/PaginatedTable.tsx
dsaxton: Is this what you had in mind: https://github.com/online-go/online-go.com/pull/1641/commits/d80902e70f28ad4c1c40e612bc3abe69fdac73cf? If so looks a lot simpler to me.

src/components/Resizable/Resizable.tsx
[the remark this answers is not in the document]
dsaxton: Updated. Can I remove the commented code though? :smiley:

anoek (approved):

.eslintrc.js
```diff
@@ -54,6 +53,7 @@ module.exports = {
                 "ignoreProperties": true
             }
         ],
+        "@typescript-eslint/no-unused-vars": ["error", { "varsIgnorePattern": "^_", "argsIgnorePattern": "^_" }],
```
anoek, line 56: It occurs to me that we use `_` as the string translate function (due to xgettext history), so I'm thinking we should avoid having a pattern that might unintentionally encourage the use of `_` exactly as a "ignore this" variable.  Specifically I'm thinking we should have our pattern make sure at least one character follows the leading `_` so we don't inadvertently shadow `_`. If I understand the pattern usage correctly, I think if we change the patterns to something like `^_[a-zA-Z_]` would do the trick?

anoek (changes requested):
Almost there on this PR I think :) Found a couple of things that broke on the settings page, but seemed everything else was in working order still.

.eslintrc.js
dsaxton: That makes sense. For the arguments modified in this commit, would you prefer we just drop them rather than using the underscore names: https://github.com/online-go/online-go.com/pull/1641/commits/e91f6ff2032a3319012d280e1a6210cbe9da40c1?

anoek: Yeah lets just get rid of those, not sure why I felt the need to have parameters in there that I just ignored.

dsaxton: Updated. Here's a weird thing I just noticed: the underscore at this line is actually unused but didn't get flagged as such even though it shouldn't match the pattern: https://github.com/dsaxton/online-go.com/blob/e1f61597eb93124d0f85f0b70e9be5db3317ebfe/src/lib/translate.ts#L376. But apparently that's correct according to JavaScript regex, or am I missing something here?

```javascript
> /_/.test("^_[a-zA-Z_]")
true
```

I think this should be false according to the Python `re` library at least.

dsaxton: Updated, I wonder why this undescore doesn't get flagged as unused though: https://github.com/dsaxton/online-go.com/blob/e1f61597eb93124d0f85f0b70e9be5db3317ebfe/src/lib/translate.ts#L376 :thinking:

src/views/Settings/Settings.tsx
```diff
@@ -1122,12 +1116,6 @@ function GamePreferences(props: SettingGroupProps): JSX.Element {
                 preferences.set(`one-click-submit-${speed}` as ValidPreference, false);
                 break;
         }
-        if (speed === "live") {
-            _setLiveSubmitMode(getSubmitMode(speed));
-        }
-        if (speed === "correspondence") {
-            _setCorrSubmitMode(getSubmitMode(speed));
-        }
```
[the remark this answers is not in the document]
dsaxton: Ok, let me just add it back to be on the safe side

.eslintrc.js
benjaminpjones: By default, yes (but it's configurable)

>### args
>
>The args option has three settings:
>
>    after-used - unused positional arguments that occur before the last used argument will not be checked, but all named arguments and all positional arguments after the last used argument will be checked.
>    all - all named arguments must be used.
>    none - do not check arguments.

https://eslint.org/docs/rules/no-unused-vars#args

anoek (approved):

anoek:
Thanks again for your work on this @dsaxton!

PR #1644, opened 2022-01-17
Bugfix <Markdown>

GreenAsJade:
Fixes issue where Markdown was not responding to changes in props

## Proposed Changes

Instead of using lifecycle functions to protect against sanitizeHtml being repeatedly called, use React.memo


anoek:
Fixes bug introduced in #1633 

Related to #1629

anoek:
Sorry you had to hunt that down, thanks for doing so and getting it fixed so fast though!

PR #1646, opened 2022-01-17
Use createRef for omnisearch_input

dsaxton:
Strict mode doesn't seem to like this: `Warning: A string ref, "omnisearch_input", has been found within a strict mode tree. String refs are a source of potential bugs and should be avoided.`

xref: https://github.com/online-go/online-go.com/issues/1629


anoek:
Thanks for the cleanup!

PR #1651, opened 2022-01-18
Construction of the rengo challenge list needed a couple more keys.

GreenAsJade:
Fixes a "need unique keys" error in console.

## Proposed Changes

 ~~Give all items in the constructed list keys.~~

Don't use an array in that place, it's not an array, it's just a list of elements.


anoek:
Thanks for the cleanup!

PR #1652, opened 2022-01-19
Remove rengo team list from Game Info Modal

GreenAsJade:
Fixes the broken layout people are reporting, with large rengo teams overlapping the subsequent information in <920px screen height

## Proposed Changes

~~Let the rengo-team-container scroll.~~

Remove that list.  They are on the Game page anyhow.  Instead, tell them the count of participants.

### DIFF
--- a/src/views/Game/GameInfoModal.tsx
+++ b/src/views/Game/GameInfoModal.tsx
@@ -156,18 +156,7 @@ export class GameInfoModal extends Modal<Events, GameInfoModalProperties, {}> {
 
     rengoTeamList = () => (
         <div className="rengo-teams-container">
-            <div className="rengo-team-list">
-                <span className="team-title">Black Team</span>
-                {this.props.config.rengo_teams.black.map((player) => (
-                    <Player disableCacheUpdate icon rank user={player} />
-                ))}
-            </div>
-            <div className="rengo-team-list">
-                <span className="team-title">White Team</span>
-                {this.props.config.rengo_teams.white.map((player) => (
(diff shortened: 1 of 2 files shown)


GreenAsJade:
... but it needs a min height, coming up...

GreenAsJade:
... nah, I'm stuck: how to fit all that stuff in a Modal on a small screen??

src/views/Game/GameInfoModal.tsx
```diff
@@ -156,18 +156,7 @@ export class GameInfoModal extends Modal<Events, GameInfoModalProperties, {}> {
...
-                    <Player disableCacheUpdate icon rank user={player} />
-                ))}
-            </div>
-            <div className="rengo-team-list">
-                <span className="team-title">White Team</span>
-                {this.props.config.rengo_teams.white.map((player) => (
-                    <Player disableCacheUpdate icon rank user={player} />
-                ))}
-            </div>
+            {/* this space intentionally left blank: it doesn't work out to try to show rengo teams in this small modal */}
```
benjaminpjones, line 159: Why not just delete the whole function?  Getting rid of that would simplify the caller a bit too:

```diff
- {this.props.config.rengo ? this.rengoTeamList() : this.twoPlayerTeamList()}
+ {this.props.config.rengo || this.twoPlayerTeamList()}
```

GreenAsJade:
I considered that, but "it may be that this is coming back as soon as we can figure out how to lay it out, so might as well have the foundations there" was the thought process.

(Can you really render "true" in that way?)

PR #1653, opened 2022-01-20
Disable fork and auto-handicap for rengo games

GreenAsJade:
Fixes people trying to use features that don't work in rengo

## Proposed Changes

Disable fork and auto handicap in the client

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -621,6 +621,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
         this.upstate([
             ["challenge.game.rengo", ev],
             ["challenge.game.ranked", false],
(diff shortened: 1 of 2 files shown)


anoek:
Sounds good. For future record if we want to support game forking of rengo games it shouldn't be that bad to implement, I think all we need to do is override the `rengo` field to be `false` right? Then it'd be a two player game with whoever was playing and we'd just ignore all the other rengo fields that might be floating around?

GreenAsJade:
Yeah that would be a quick solution.   I had thought that the rengoers were dreaming of forking a rengo game that ended prematurely and re-rengoing!

anoek:
I think support for that would be fairly natural to add if we get around to doing challenge/invite rengo staging, then it'd be fork a game, *set* the rengo flag, etc.

I reckon we can wait and see when folks start complaining about fork missing before worrying about either case too much  anyways.

PR #1654, opened 2022-01-20
The rest of the RouteComponentProps types

benjaminpjones:
## Proposed Changes

Replace instances of:

```
interface SomeComponentProps {
    match: {
        params: {
            id: string;
        };
    };
}
```

with:

```
type SomeComponentProps = RouteComponentProps<{ id: string }>
```

### Motivation

It's a little more succinct, but more importantly it becomes clear that the structure of props is dictated by React Router.

### Note

~A lot of `params` are treated as numbers even though they are strings.  It doesn't seem to matter, but TS complains, so I made those `any` with an explanatory comment.~

`IdType = string | number` has been introduced to deal with these numbery strings.

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -18,7 +18,7 @@
 /* A page for looking up and playing against josekis stored in the OGS OJE*/
 
 import * as React from "react";
-import { Link } from "react-router-dom";
+import { Link, RouteComponentProps } from "react-router-dom";
 import ReactResizeDetector from "react-resize-detector";
 import * as queryString from "query-string";
 
@@ -145,12 +145,7 @@ const ColorMap = {
 
 type MoveType = "bad" | "good" | "computer" | "complete";
 
-interface JosekiProps {
-    match: {
-        params: any;
-    };
-    location: any;
-}
+type JosekiProps = RouteComponentProps<{ pos: string }>;
 
 interface MoveTypeWithComment {
     type: MoveType;
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -39,6 +39,7 @@ import { JosekiSourceModal } from "JosekiSourceModal";
 import { JosekiVariationFilter } from "JosekiVariationFilter";
 import { JosekiTagSelector } from "JosekiTagSelector";
 import { Throbber } from "Throbber";
+import { IdType } from "src/lib/types";
 
 const server_url = data.get("joseki-url", "/oje/");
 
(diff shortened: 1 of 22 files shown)


src/views/Joseki/Joseki.tsx
```diff
@@ -145,12 +145,7 @@ const ColorMap = {
 
 type MoveType = "bad" | "good" | "computer" | "complete";
 
-interface JosekiProps {
-    match: {
-        params: any;
-    };
-    location: any;
```
anoek, line 152: `props.location` is used down below, I think we need to extend `RouteComponentProps<{ pos: string }>` to add that in right?

src/views/PuzzleCollection/PuzzleCollection.tsx
```diff
@@ -23,14 +23,15 @@ import { get, del, put, abort_requests_in_flight } from "requests";
...
 
 export function PuzzleCollection({
     match: {
         params: { collection_id },
     },
-}: {
-    match: { params: { collection_id: number } };
-}): JSX.Element {
+}: RouteComponentProps<{
+    collection_id: any; // string treated as a number
```
anoek, line 33: That's always going to be `string` I think, and we should `parseInt` to convert it to a number if we need a number I reckon?

anoek (changes requested):
This is great @benjaminpjones , that's a lot better than the match/params crap I was doing before.

src/views/PuzzleCollection/PuzzleCollection.tsx
benjaminpjones: Yep we should, but we don't currently.  I was trying to keep functional changes separate from type changes, but it's simple enough that I can make those updates in this PR.

The thing I need to investigate further is whether is whether conversion to number is necessary at all.  For example, `collection_id` gets passed into `ACLModal`, who says it should be a `number` (that's why TS complains about string).  But then I look at `ACLModal` and it just formats the number back into a string for a URL.  So is the right answer to change `ACLModal`'s `props.puzzle_collection_id` into a `string|number` instead of converting with `parseInt()`?

benjaminpjones:
Thanks for reviewing! I changed the PR to a draft, I think I'll go ahead and try to turn those `any`s to `string`s.  I'll try modifying types on the components that consume these params, and fallback on `parseInt` if that doesn't work.  Does that sound good?

benjaminpjones:
Okay, I added a new type for these number-but-actually-okay-if-it's-string IDs.  This suppressed most of the errors, although there was one `parseInt` I still had to modify.

Let me know if that works!

anoek:
Thanks @benjaminpjones - `IdType` is probably a good solution. I feel like all `_id`'s *should* always be numbers, but the fact that they come as strings from the router, and a lot of times strings work just as well as numbers for us, `IdType` is probably a good convenience and enforces we `parseInt` if we're really expecting it to be a number.

anoek (approved):

PR #1658, opened 2022-01-22
Use forceUpdate to re-render

dsaxton:
Referring to these comments I think having this as a named function should be a bit more clear (assuming it works, I'm actually not entirely sure how to reproduce the original issue):

https://github.com/online-go/online-go.com/pull/1641#discussion_r788721193
https://github.com/online-go/online-go.com/pull/1641#discussion_r788721773

Technique is basically a copy / paste of this: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate


anoek:
I dunno, I kinda agree with the comment on that page you linked: `Try to avoid this pattern if possible.`.. while the current verbose solution might seem superfluous, I feel like it's a little bit clearer, and if we come along and want to make changes to how the settings page looks and feels, it'll be easier to work with as is.

PR #1661, opened 2022-01-23
go to beta for ai in development environments

GreenAsJade:
Fixes ongoing web socket messages in the consle.

## Proposed Changes

~~If we're pointing at localhost, use beta ai server.~~

If we have NODE_ENV=development, then go to the beta ai server


GreenAsJade:
I couldn't see what is the intent    `window.location.hostname.indexOf("dev") >= 0 || `

My npm run dev says:

## Development server started on port 8080
##  ( http://localhost:8080 )

so that's what I point at...

anoek:
Thanks!

PR #1665, opened 2022-01-23
Add npm test script

dsaxton:
Could be a good way to document how to run the tests


benjaminpjones:
Yeahhh I should have done that back when I added tests.  I didn't realize npm had a standard name for that back then!

anoek:
Thanks @dsaxton !

PR #1669, opened 2022-01-24
NavBar.tsx: migrate notification_list to createRef

### LINKED ISSUE #1628: Players can create games with null x 10s byo-yomi
**Describe the bug**

**To Reproduce**

**Expected behavior**

E.g.

### LINKED ISSUE #1629: Enable React.StrictMode and fix the complaints that ensue

maurelio1234:
Partially fixes #1629

Tell me if you are open to these kinds of PRs and how to coordinate on them

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -87,9 +87,9 @@ export function logout() {
 export class NavBar extends React.PureComponent<{}, any> {
     refs: {
         input: any;
-        notification_list: NotificationList;
     };
     omnisearch_input_ref = React.createRef<HTMLInputElement>();
+    notification_list = React.createRef<NotificationList>();
 
     constructor(props) {
         super(props);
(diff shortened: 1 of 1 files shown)


anoek:
Awesome, thanks @maurelio1234 !

anoek:
And yes, definitely open to these kinds of PRs. As for coordination I wouldn't expect you'd run into many conflicts, @dsaxton has been doing similar things at times too so maybe a bump here and there. If you're doing a massive sweep or something, maybe let folks know on slack  but otherwise I wouldn't worry about it too much, just submit and I'll try and I'll get them merged asap :)

PR #1671, opened 2022-01-26
Rengo casual mode

GreenAsJade:
Fixes instant game end when a rengo player bails

## Proposed Changes

Add "casual mode" for rengo
 - Resignation and timeout each just remove the player, if there are other players left
 - The "next player in line" is given the turn for that team, with a fresh clock
 - Enforcing "Simple mode" timing mimics the "clock per player" that is needed for this mode to make sense


GreenAsJade:
Note: the build fails because this requires goban 0.5.74

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -769,6 +789,40 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
...
+                        <div
+                            className={
+                                "casual-time-warning" +
+                                (this.state.challenge.game.rengo &&
+                                this.state.challenge.game.rengo_casual_mode
+                                    ? ""
+                                    : " hide")
+                            }
+                        >
+                            *{_("use Simple time")}
```
anoek, line 822: We need to enforce this by removing the option, no one reads :sweat_smile:

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -595,12 +601,26 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
...
             ["challenge.game.rengo", ev],
             ["challenge.game.ranked", false],
             ["challenge.game.handicap", 0],
         ]);
+    };
+    update_rengo_casual = (ev) => {
+        this.upstate("challenge.game.rengo_casual_mode", ev);
+        // brute force ensure that they are using simple time in casual mode
+
+        this.setState({
```
anoek, line 615: We should probably default the speed to whatever the current speed is, so if they have correspodence set, we set simple correspondence with whatever time, etc. Bonus points for remembering their last rengo time settings :)

src/views/Game/Game.tsx
```diff
@@ -2350,6 +2366,17 @@ export class Game extends React.PureComponent<GameProperties, GameState> {
...
+
+        if (this.goban.engine.rengo && this.goban.engine.rengo_casual_mode) {
+            const team = this.goban.engine.rengo_teams.black.find(
+                (p) => p.id === data.get("user").id,
+            )
+                ? "black"
+                : "white";
+            ratbag = this.goban.engine.rengo_teams[team].length > 1;
+        }
+        console.log("ratbag: ", ratbag);
```
anoek, line 2379: Might should remove this before production :grin: might should change the variable name too, it's amusing, but boring code is more maintainable and we'll get less curses put upon us by future developers.

anoek (changes requested):

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: Note: the server rejects casual mode challenges with non-simple time, so if they don't read, they get a user friendly worded error dialog to this effect when they hit "Create Challenge".

(Not saying it couldn't be better, just that its not dire 😁  )

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: Yeah, indeed we should.  Do we have to?  The full solution may include fixing the current bug with saving the different speed settings, which is daunting...

GreenAsJade: ... note that it's only a something you experience the first time, or if you swap between casual and strict rengo Otherwise it just remembers that you like casual rengo at the speed you set last time.

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: Done.

GreenAsJade:
Hmm - that last commit has a bug.

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: Yay, with the other fix, this is no longer a thing

GreenAsJade:
This thing can totally look like its working when it's not :D   In the more recent instance, it looked like it was forcing as I wanted, but it was just remembering 🤦‍♀️

src/components/TimeControl/TimeControlPicker.tsx
```diff
@@ -18,54 +18,66 @@
...
-// TODO: add TimeControl as the state for this component.
-// However, it will require some extra TLC. There are many checks like this:
-//     if (this.state.system === "fischer") { doSomething(this.state.initial_time; }
-// that TypeScript would have a better time with if we were a little more
-// precise in what properties we're looking for:
-//     if ("initial_time" in this.state) { doSomething(this.state.initial_time; }
 export class TimeControlPicker extends React.PureComponent<
     TimeControlPickerProperties,
-    any /*TimeControl*/
+    TimeControl
```
benjaminpjones, line 36: ❤️

src/components/TimeControl/TimeControlPicker.tsx
```diff
@@ -18,54 +18,66 @@
...
+            this.props.value !== next_props.value ||
+            this.props?.force_system !== this.state.system
+        ) {
+            if (next_props.force_system) {
+                update = Object.assign(
+                    this.state,
+                    recallTimeControlSettings(this.state.speed, next_props.force_system),
+                );
+            } else if (next_props.value) {
+                update = { ...makeTimeControlParameters(next_props.value) };
```
benjaminpjones, line 70: Also here... we *don't* want to update from `next_props.value` if `force_system` exists?

src/models/challenges.d.ts
```diff
@@ -22,6 +22,11 @@ declare namespace socket_api {
          *
          * This is a work in progress. Trust these values at your own risk.
          * */
+
+        export interface RengoGameFlags {
```
benjaminpjones, line 26: Is this still `RengoGameFlags` used?  If it is necessary, can you move the `seekgraph` comment below it?  Otherwise intellisense won't know that it's describing `Challenge`

Screen Shot 2022-01-28 at 9 41 21 PM

GreenAsJade: No, RengoGameFlags turned out to be a bad idea, flags on a game need to be individual, not grouped in that way, for ease of translation into DB schema

src/components/TimeControl/TimeControlPicker.tsx
```diff
@@ -18,54 +18,66 @@
...
-            update = Object.assign(update, makeTimeControlParameters(next_props.value));
-        }
-        if (Object.keys(update).length) {
-            this.time_control = makeTimeControlParameters(update);
-            this.setState(update);
+        let update: TimeControl;
+        // console.log("Timepicker props", next_props);
+        if (
+            this.props.value !== next_props.value ||
+            this.props?.force_system !== this.state.system
```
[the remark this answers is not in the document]
GreenAsJade: http://blogs.aalto.fi/lighting/files/2020/04/codingmeme.jpg

That doesn't look good does it!  Investigating...

src/components/TimeControl/TimeControlPicker.tsx
GreenAsJade: Pfff... you are right, that is not the intent, force_system should be letting through next_props.  I think this only works by luck of timing.    Ay yah yay yah!

GreenAsJade: Actually, the way it "works" is luck in the sense that `value` and `force` were never asserted together in the scenarios I tried.   However, there is nothing that rules this out, so a fix is needed.

GreenAsJade: OK - fix coming.  I can see now why it seemed unlikely that `value` and `force` would ever be applied together, because value is used in direct challenges and tournies and `force` is used in rengo which doesn't support direct or tournies... today!

BHydden: Sounds like tomorrow's problem ;) haha

anoek (approved):

src/components/TimeControl/TimeControlPicker.tsx
benjaminpjones: Ahh got it, the way it's used currently, it's only one or the other.  Well thanks for the fix, that looks good and will probably help if someone tries to do both down the line!

src/components/RengoManagementPane/RengoManagementPane.tsx
```diff
@@ -20,7 +20,7 @@ import * as React from "react";
 import { _, pgettext } from "translate";
 
 interface RengoManagementPaneProperties {
-    user_id: number;
+    user: any;
```
benjaminpjones, line 23: Hated to see a new `any` added, so made a PR for an interface!

https://github.com/GreenAsJade/online-go.com/pull/4

PR #1672, opened 2022-01-26
Cleanup: remove unused parameters from LadderComponent

benjaminpjones:
Despite all the removed code, this change should be a no-op.  There was a lot of code in there about a "Full View".  Must have been replaced by `Ladder.tsx` or something.

Tested: Compared beta and localhost, no discernable changes.

View on beta:
Screen Shot 2022-01-25 at 11 09 22 PM

View on localhost:
Screen Shot 2022-01-25 at 11 08 41 PM

### DIFF
--- a/src/components/LadderComponent/LadderComponent.tsx
+++ b/src/components/LadderComponent/LadderComponent.tsx
@@ -18,12 +18,11 @@
 import * as React from "react";
 import ReactResizeDetector from "react-resize-detector";
 import { _, interpolate } from "translate";
-import { post, get } from "requests";
+import { get } from "requests";
 import { errorAlerter } from "misc";
 import { Player } from "Player";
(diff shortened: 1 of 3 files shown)


benjaminpjones:
Okay I should stop removing stuff for now, but the more I remove, the more I see that can be removed.

I *think* we can remove `componentDidMount` and `componentDidUpdate` and actually save ourselves a call to the REST API (yay network optimization!).

What I really wanted to remove was that `ref`, but with the `UIPush` calling `updatePlayers()`, I don't know if that's possible.

BHydden:
nek minute bph reduces the entire OGS front-end down to one carefully worded line of code 😅 turns out that's all we needed all along 🤣

anoek:
Awesome thanks for the cleanup :)

PR #1676, opened 2022-01-28
Remove some more table refs in favor of uiPushProps

benjaminpjones:
Follow-up to #1675 

## Proposed Changes

  - Update any PaginatedTable + UIPush combos to just use uiPushProps
  - remove refs where possible

Tested: players table on the Groups page

Given that everything else was mod stuff, I didn't test that, but the changes are pretty straight-forward so I don't think there's a whole lot that can go wrong.


anoek:
Awesome, yeah nice cleanup, thanks!

PR #1679, opened 2022-01-29
User.tsx: Extract Game History tables, mod tools into separate components.

benjaminpjones:
I believe the User class has some natural component boundaries where modularization would be beneficial (even though the components are not used elsewhere).  Below is a visual representation of where these boundaries are.

Fixes [one block of JSX](https://github.com/online-go/online-go.com/blob/devel/src/views/User/User.tsx#L877-L2162) being over 1000 lines long.

Screenshot 2022-01-29 at 10-40-30 matburt

### DIFF
--- a/src/views/User/GameHistoryTable.tsx
+++ b/src/views/User/GameHistoryTable.tsx
@@ -1,0 +1,383 @@
+/*
+ * Copyright (C) 2012-2022  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as React from "react";
+import { _ } from "translate";
+import { PlayerAutocomplete } from "PlayerAutocomplete";
+import { Card } from "material";
+import { PaginatedTable } from "PaginatedTable";
+import { effective_outcome } from "rank_utils";
+import { getGameResultText } from "misc";
+import { openUrlIfALinkWasNotClicked, maskedRank } from "./common";
+import * as moment from "moment";
+import { rankString } from "rank_utils";
+import { Player } from "Player";
+import { Link } from "react-router-dom";
+import { interpolate } from "translate";
+import * as preferences from "preferences";
+
+interface GameHistoryProps {
+    user_id: number;
+}
+
+export function GameHistoryTable(props: GameHistoryProps) {
+    const [player_filter, setPlayerFilter] = React.useState<number>(null);
+    const [show_rengo, setShowRengo] = React.useState<boolean>(false);
+
+    function game_history_groomer(results) {
+        const ret = [];
+        for (let i = 0; i < results.length; ++i) {
+            const r = results[i];
+            const item: any = {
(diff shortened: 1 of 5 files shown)


src/views/User/ModTools.tsx
```diff
@@ -0,0 +1,177 @@
...
+            .then(() => {})
+            .catch(errorAlerter);
+
+        moderator_note.current.value = "";
+    };
+
+    const moderator_log_anchor = React.useRef<HTMLDivElement>(null);
+    React.useEffect(() => {
+        if (props.show_mod_log && moderator_log_anchor.current !== null) {
+            moderator_log_anchor.current.scrollIntoView();
```
benjaminpjones, line 58: @GreenAsJade checking the commit history, I think this belongs to you.  Can you verify my update doesn't adversely affect the behavior?  I wasn't sure what scenarios scroll-to-mod-log is desired.  I at least verified it works on page load.

benjaminpjones: Logic dealing w/ refs is the hardest to extract 😅

GreenAsJade: scroll to mod log is invoked when the moderator clicks on the add player note button that's in the header of the PM dialog bar - it causes the underlying page from that modal to show the most recent mod notes, so the mdoerator doesn't type a note only to find (after visiting the person's page) that there is more relevant information they were not aware of.

I just tested it on this branch, it works (do you have a Mod account on Beta?  Can't see why you would not...)

benjaminpjones: Yay! Thanks for checking that!  So it sounds like "on load" is the only time this *needs* to happen.  I think the code also scrolls to mod log whenever the component updates, but it doesn't seem like the component updates too much for that to be an issue.

And no, I don't have a mod account, but it might be useful. Are there any community mod accounts on beta?

On second thought... do I *really* want to see how the sausage gets made? 😬

BHydden: Guess that's a question for @anoek to ponder, whether devs who aren't otherwise mods should have access to mod testing on beta 🤷‍♂️

GreenAsJade: Yes, it's an anoek question - I'm not sure it'd be a general thing, but benjito's work is so extensive it may make sense (and not that any determined developer can see these things anyhow in the code)

src/views/User/GameHistoryTable.tsx
```diff
@@ -349,9 +349,8 @@ export function GameHistoryTable(props: GameHistoryProps) {
                                       },
                                       {
                                           header: _("Result"),
-                                          className: () => " annulled",
-                                          /* a nice background color for this... */ render: () =>
-                                              _("fun was had"),
+                                          className: (X) => (X && X.annulled ? " annulled" : ""),
+                                          render: (X) => X.result,
```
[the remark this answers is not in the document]
GreenAsJade: It's strange - somewhere I got the idea that the "result" was not being calculated properly, but maybe this got fixed along the way as a side effect of making something else rengo-aware!   It does seem to work now.

GreenAsJade (approved):

src/views/User/GameHistoryTable.tsx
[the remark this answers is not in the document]
benjaminpjones: Ha! Well glad it's working now! Guess I just checked the API at the right time 😄

src/views/User/ModTools.tsx
anoek: @benjaminpjones You are now a moderator on beta, FYI :)

anoek:
Thanks for the re-org!

src/views/User/ModTools.tsx
benjaminpjones: oh cool, thanks!

benjaminpjones:
Thanks @anoek and @GreenAsJade for reviewing!

src/views/User/ModTools.tsx
BHydden: don't share any of our dark secrets 🤣

PR #1680, opened 2022-01-30
Add link to API to Moderator Center

ManchurianChicken:
add link to API

### DIFF
--- a/src/views/Moderator/Moderator.tsx
+++ b/src/views/Moderator/Moderator.tsx
@@ -278,11 +278,11 @@ export class Moderator extends React.PureComponent<{}, ModeratorState> {
                                         }
                                     >
                                         <a
-+                                            href={
-+                                                "https://online-go.com/api/v1/moderation/recent_users?id=" +
-+                                                X.id
-+                                            }
-+                                            target={"_blank"}
+                                            href={
+                                                "https://online-go.com/api/v1/moderation/recent_users?id=" +
+                                                X.id
+                                            }
+                                            target={"_blank"}
                                          >
                                             {X.last_ip}
                                         </a>
--- a/src/views/Moderator/Moderator.tsx
+++ b/src/views/Moderator/Moderator.tsx
@@ -277,10 +277,13 @@ export class Moderator extends React.PureComponent<{}, ModeratorState> {
                                             (X.should_ban ? "should-ban" : "")
                                         }
                                     >
-                                        <a 
-                                            href={ "https://online-go.com/api/v1/moderation/recent_users?id=" + X.id }
-                                            target={ "_blank" }
-                                        >
+                                        <a
++                                            href={
++                                                "https://online-go.com/api/v1/moderation/recent_users?id=" +
++                                                X.id
++                                            }
++                                            target={"_blank"}
+                                         >
                                             {X.last_ip}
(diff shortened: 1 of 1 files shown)


GreenAsJade:
This works, and doesn't provide the info if I'm not a moderator.

It fails lint though.

If you run `npm run dev` it will tell you what the problem is.

I _think_ that if you run `npm run prettier` it will fix it up for you, though I have VSCode set up to do it each time I save, so that's not something I do regularly.

ManchurianChicken:
To be honest, I have not the slightest clue what you're saying...
Or was this directed at anoek?

GreenAsJade:
Heh heh - directed at whomever is interested in getting this PR merged ;)

> This works, and doesn't provide the info if I'm not a moderator.

I tested that when I clicked the link as a moderator I could see what I was supposed to see, and that when I hit that route as not-a-moderator I did not see anything.

> It fails lint though.

One of the checks done before accepting a PR is a "lint check".   This checks that the code passes a whole range of non-functional requirements  ... coding standards.

It is very fussy about layout - really the only way to comply is to run "prettier" to get it laid out the way it likes it.

I'm reasonably sure that all developers test their code by running the `npm run dev` command which brings up a local server on your machine that points to the OGS beta back end.   

The output of `npm run dev` has got the information from the lint check in it.   Your code will run even if this check fails, but your PR will not be able to be accepted if the lint check is failing: that yellow triangle above will turn red after anoek enables the workflow, it will say "build check failed" or similar

> I think that if you run `npm run prettier` it will fix it up for you

I'm reasonably confident, actually, that if you run the command `npm run prettier` it will fix up the layout in your code that it doesn't like.

>  though I have VSCode set up to do it each time I save, so that's not something I do regularly.

Developers have various options for dealing with this problem ongoing.   I use Visual Studio Code for editing OGS files, and this has extensions that run the check continually, and fix any layout errors every time I save the file.

Broadly, the layout rule is something like "no lines longer than 100 chars", but it is also fussy about where you split the line, so it's a heck of a lot easier to just let it do it the way it likes it.

benjaminpjones:
FYI, here's how [prettier](https://prettier.io/) wants that to look:

```diff
                                             (X.should_ban ? "should-ban" : "")
                                         }
                                     >
-                                        <a
-                                            href={ "https://online-go.com/api/v1/moderation/recent_users?id=" + X.id }
-                                            target={ "_blank" }
+                                        <a
+                                            href={
+                                                "https://online-go.com/api/v1/moderation/recent_users?id=" +
+                                                X.id
+                                            }
+                                            target={"_blank"}
                                         >
                                             {X.last_ip}
                                         </a>
```

Although I might suggest using a format string if you wanted all that on the same line 🙂:

```
<a
    href={`https://online-go.com/api/v1/moderation/recent_users?id=${X.id}`}
    target={"_blank"}
>
```

BHydden:
> here's how prettier wants that to look

One of the few times I've strongly disagreed with prettier... that is abhorrent.

> template literal

ohhhhhh neat hack... I like it :+1:

ManchurianChicken:
After searching for a bit, I think I managed to edit the file and fix the layout.

anoek:
Thanks @ManchurianChicken !

PR #1681, opened 2022-01-31
A little more type safety for data.ts

benjaminpjones:
`data` seems to be responsible for a lot of the missing types in the code base.  #1564 was an attempt to fix this by inferring types from the provided default value.  There are three problems with this approach:

1) Types were widened in some cases when they shouldn't be:

```
const tc = data.get("time_control.system", "fischer" as TimeControlSystem);
// tc now has type string, not TimeControlSystem
```

This could be fixed by removing type widening, but then the behavior could also be confusing:

```
const tc = data.get("time_control.system", "fischer");
// tc now has type "fischer", not TimeControlSystem or string
```

2) Types are not checked for consistency across separate calls:

```
data.set("foo", "bar");
const foo: number = data.get("foo", 8); // not good, but legal
```

3) A default value or template argument may not be provided at all!

```
const user = data.get("user"); 
// user now has type any
```

This PR is a proposal to make `data` more typesafe.

## Proposed Changes

  - Add `data_schema.ts` to document and enforce the types that get saved in `data.ts`.
  - Update the function signatures in `data.ts` to use the schema.
  - Changed some concatenated strings to template literals, because TS had a hard time picking up prefixes in the concatenated strings.

### DIFF
--- a/src/lib/cached.ts
+++ b/src/lib/cached.ts
@@ -188,7 +188,7 @@ export const cached = {
                 });
         },
     },
-};
+} as const;
 
 let current_user_id = 0;
 let refresh_debounce = setTimeout(refresh_all, 10);
--- a/src/lib/data_schema.ts
+++ b/src/lib/data_schema.ts
@@ -23,15 +23,32 @@
 import { GroupList, ActiveTournamentList } from "./types";
 import { Announcement } from "src/components/Announcements";
 
-export interface DataSchema {
+interface CachedSchema {
+    groups: GroupList;
+    active_tournaments: ActiveTournamentList;
+
+    // TODO: examine these endpoints and write interfaces for these.
+    config: any /* /login */;
+    ladders: any /* /me/ladders */;
+    challenge_list: any /* /me/challenges */;
+    blocks: any /* /me/blocks */;
+    friends: any /* ui/friends */;
+    group_invitations: any /* me/groups/invitations */;
+}
+
+type Prefixed<T, P extends string> = {
+    [K in keyof T as K extends string ? `${P}.${K}` : never]: T[K];
+};
+
+export interface DataSchema extends Prefixed<CachedSchema, "cached"> {
     user: any;
     bid: string;
     theme: string;
     debug: boolean;
 
+    cached: CachedSchema;
+
     // TODO: make a types for each of these that list the keys explicitly
-    cached: any;
-    [cached_key: `cached.${string}`]: any;
     config: any;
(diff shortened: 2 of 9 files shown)


src/lib/data_schema.ts
```diff
@@ -0,0 +1,117 @@
...
+        Prefixed<ConfigSchema, "config"> {
+    user: any;
+    bid: string;
+    theme: string;
+    debug: boolean;
+
+    cached: Partial<CachedSchema>;
+    config: Partial<ConfigSchema>;
+
+    // TODO: make a types for each of these that list the keys explicitly
```
benjaminpjones, line 85: Note: even if this PR is accepted, `data` still has a good deal of work before it would be type safe. Most of the prefixed keys in data have been added in a very permissive way to avoid compiler errors.  But I want to get confirmation that this PR is the right direction before adding more schema interfaces.

src/lib/cached.ts
```diff
@@ -188,7 +188,7 @@ export const cached = {
                 });
         },
     },
-};
+} as const;
```
anoek, line 191: Since this was declared const, does this `as const` do some magic I don't know about or is it redundant?

anoek (approved):
This is fantastic, thanks for doing this :)

src/lib/cached.ts
benjaminpjones: Yep a bit of magic :)

The key difference with the declarations is that the type is not widened when using `as const`[^1]:

```
const obj = { foo: "foo"; bar: 28 } // type is { foo: string; bar: number }
const obj_as_const = { foo: "foo"; bar: 28 } // type is { foo: "foo"; bar: 28 }
```

This is important because data now only accepts keys that have been listed in the schema, so lines like this would fail if `cached.challenge_list` were an arbitrary `string`:

https://github.com/online-go/online-go.com/blob/48efeca3e5120f5fb63a8a9af2e901d1fe6bdf37/src/lib/cached.ts#L72

[^1]: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

benjaminpjones: BTW we *could* allow data to accept arbitrary strings, but I'm not sure that's desirable.  I'm thinking about potential mistakes like:

```
data.get("confgi.user");
              ^^
             typo
```

strict key checking could be quite useful!

anoek: Fancy, yeah strict sounds good :)

PR #1685, opened 2022-02-02
Make it so you can't have pause on weekend with Simple time.

GreenAsJade:
Fixes games hanging because the clock stops ticking forever in Simple time games that have pause on weekends set.

## Proposed Changes

Disable the setting of pause on weekends for Simple time.

### DIFF
--- a/src/components/TimeControl/TimeControlPicker.tsx
+++ b/src/components/TimeControl/TimeControlPicker.tsx
@@ -79,8 +79,7 @@ export class TimeControlPicker extends React.PureComponent<
 
             if (update) {
                 //console.log("Updating time control:", update);
-                this.time_control = makeTimeControlParameters(update);
-                this.setState(update);
+                this.syncTimeControl(update);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Note: this has become important because currently any rengo casual mode game with simple time and pause on weekend is pretty doomed to get stuck when one of the players decides to stop playing.

eg https://online-go.com/game/40362506

(It ended up in this state because " a moderator" started it, which converted it to Simple time ... but it had the pause on weekend flag set, and this stayed that way.)

anoek:
Oye, yeah we should probably fix that bug so it doesn't happen in the first place. I'm going to leave this open as a nag for either of us to evaluate the difficulty of fixing the simple timer resetting after weekend (or I think any pause?). It might not even be that bad to fix.

GreenAsJade:
I do have it on my "todo list".  I think it'd be worth pushing this into prod though, because otherwise we are likely to end up with more games where there is no way to rescue them...

GreenAsJade:
... interestingly it's hard to guard against, because as far as I can see, a moderator can't tell if an active challenge is pause on weekends or not!

anoek:
OK I think I found and fixed the issue with simple time, obviating the need for this patch. Here's to hoping I don't screw up more than 1,000 with the fix. :crossed_fingers:

GreenAsJade:
🎉  🎉  🎉   🤞  🤞

PR #1686, opened 2022-02-02
Moderator rengo team management

GreenAsJade:
Fixes moderators being unable to start games that inadvertently got the same player on both teams, and being unable in general to assist with team makeup.

## Proposed Changes

Make team management controls available to moderators.


anoek:
Nice, thanks :)

PR #1688, opened 2022-02-03
Make it so that live rengo chat appears only in the automatch container

GreenAsJade:
Fixes [apparent sync delay between chat panes](https://forums.online-go.com/t/rengo-status/40415/452?u=greenasjade)

## Proposed Changes

 Don't show the same chat pane in two places :)


anoek:
Thanks for the fix!

PR #1689, opened 2022-02-04
Open correspondence rengo management pane on challenge creation

GreenAsJade:
Fixes possibility of newbies not even realising they need to open that pane.

## Proposed Changes

  "Challenge created" callback in ChallengeModal, allowing Play to open the management pan.=e

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -33,7 +33,14 @@ import {
 } from "rank_utils";
 import { errorLogger, errorAlerter, rulesText, dup, ignore } from "misc";
 import { PlayerIcon } from "PlayerIcon";
-import { timeControlText, shortShortTimeControl, isLiveGame, TimeControlPicker } from "TimeControl";
+import {
+    timeControlText,
+    shortShortTimeControl,
+    isLiveGame,
+    TimeControlPicker,
+    TimeControlTypes,
+} from "TimeControl";
+import { ChallengeDetails, IdType } from "types";
 import { sfx } from "sfx";
 import * as preferences from "preferences";
 import { notification_manager } from "Notifications";
@@ -55,6 +62,33 @@ interface ChallengeModalProperties {
     playersList?: Array<{ name: string; rank: number }>;
     tournamentRecordId?: number;
     tournamentRecordRoundId?: number;
+    created?: (c: ChallengeDetails) => void;
+}
+
+interface ChallengeState {
+    live?: boolean; // computed during creation of challenge
+    initialized: boolean;
+    min_ranking: number;
+    max_ranking: number;
+    challenger_color: string; // "automatic"
+    game: {
+        name: string; //"",
+        rules: string; // "japanese",
+        ranked: boolean;
+        width: IdType;
+        height: IdType;
+        handicap: number; // 0,
+        komi_auto: string; //"automatic",
+        komi: number; // 5.5,
+        disable_analysis: boolean;
+        initial_state: any; // TBD
+        private: boolean;
+        rengo: boolean;
+        rengo_casual_mode: boolean;
+        pause_on_weekends: boolean;
+        time_control: TimeControlTypes.TimeControlSystem;
+        time_control_parameters: TimeControlPicker;
+    };
 }
 
 export const username_to_id = {};
@@ -110,7 +144,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
 
         const speed = data.get("challenge.speed", "live");
 
-        const challenge = data.get(`challenge.challenge.${speed}`, {
+        const challenge: ChallengeState = data.get(`challenge.challenge.${speed}`, {
             initialized: false,
             min_ranking: 5,
             max_ranking: 36,
@@ -418,7 +452,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
             next.challenge.game.komi = null;
         }
 
-        const challenge: any = Object.assign({}, next.challenge);
+        const challenge: ChallengeState = Object.assign({}, next.challenge);
         challenge.game = Object.assign({}, next.challenge.game);
 
         let player_id = 0;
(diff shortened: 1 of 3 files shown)


src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -55,6 +62,33 @@ interface ChallengeModalProperties {
...
+    live?: boolean; // computed during creation of challenge
+    initialized: boolean;
+    min_ranking: number;
+    max_ranking: number;
+    challenger_color: string; // "automatic"
+    game: {
+        name: string; //"",
+        rules: string; // "japanese",
+        ranked: boolean;
+        width: IdType;
```
benjaminpjones, line 78: Not sure I'd use `IdType` here.  While the underlying `number | string` is correct, semantically `IdType` says "this is an ID", right?

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -55,6 +62,33 @@ interface ChallengeModalProperties {
...
+
+interface ChallengeState {
+    live?: boolean; // computed during creation of challenge
+    initialized: boolean;
+    min_ranking: number;
+    max_ranking: number;
+    challenger_color: string; // "automatic"
+    game: {
+        name: string; //"",
+        rules: string; // "japanese",
```
benjaminpjones, line 76: [`Ruleset`](https://github.com/online-go/online-go.com/blob/d533e3501ea25e61b58634b93af8f8f1d5463749/src/lib/types.ts#L88)

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -55,6 +62,33 @@ interface ChallengeModalProperties {
...
+        ranked: boolean;
+        width: IdType;
+        height: IdType;
+        handicap: number; // 0,
+        komi_auto: string; //"automatic",
```
benjaminpjones, line 81: "custom"|"automatic"

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -55,6 +62,33 @@ interface ChallengeModalProperties {
...
+        komi: number; // 5.5,
+        disable_analysis: boolean;
+        initial_state: any; // TBD
+        private: boolean;
+        rengo: boolean;
+        rengo_casual_mode: boolean;
+        pause_on_weekends: boolean;
+        time_control: TimeControlTypes.TimeControlSystem;
+        time_control_parameters: TimeControlPicker;
```
benjaminpjones, line 90: Shouldn't `time_control_parameters` be a `TimeControl`?

benjaminpjones (commented):
`ChallengeState` is a great addition, thanks! Leaving some comments on how some of the string members can be tightened into string literal types

src/components/ChallengeModal/ChallengeModal.tsx
benjaminpjones: I'm basing that off this line:

https://github.com/online-go/online-go.com/blob/6a5c8a15f6a570d0c128dd9d153fb6299c550aa8/src/components/ChallengeModal/ChallengeModal.tsx#L493

`this.refs.time_control_picker` should be the `TimeControlPicker`, so `this.refs.time_control_picker.time_control` should be a `TimeControl`.

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: True, I originally had number | string, and plonked that in as a last minute "oh, we have this type".

I actually don't even like that it has to be number or string, so that's another darn thing to fix

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: RuleSet :p

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: Dad gang.  Another "so.... how does this even work then?!"

BHydden: JS helpfully guessing what the monkey meant to do, even if he didn't do it ❤️

GreenAsJade: Kinda more like JS doing what it always does, regardless of what we're telling the compiler...

GreenAsJade: So I've totally refactored everything again.  Standby :)

src/models/challenges.d.ts
```diff
@@ -66,3 +66,31 @@ declare namespace socket_api {
...
     }
 }
+
+declare namespace rest_api {
+    // Payload of challenge POST
+    interface ChallengeDetails {
+        initialized: boolean;
+        min_ranking: number;
+        max_ranking: number;
+        challenger_color: string; // “automatic”
```
benjaminpjones, line 76: Are the comments for default values still meaningful in this context?

benjaminpjones: Also, `challenger_color: "black"|"white"|"automatic"|"random";`

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -662,8 +628,22 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
     update_board_size = (ev) => {
         this.syncBoardSize((ev.target as HTMLSelectElement).value);
     };
-    update_board_width = (ev) => this.upstate("challenge.game.width", ev);
-    update_board_height = (ev) => this.upstate("challenge.game.height", ev);
+
+    update_board_width = (ev) => {
+        this.setState({
+            challenge: Object.assign({}, this.state.challenge, {
+                board_width: parseInt(ev.target.value),
```
benjaminpjones, line 635: Nice! (Also, I had no idea that `upstate` and `setState` were basically the same thing - I finally looked up the definition 😅)

src/models/challenges.d.ts
GreenAsJade: If one was going to do that, one would define a type like ColourSelectionOptions eh....

GreenAsJade: Then one would realise one has to call it ColorSelectionOptions I guess :cry:

src/components/ChallengeModal/ChallengeModal.tsx
GreenAsJade: I've been nervous about that.  My understanding is that upstate handles various types of `ev` seamlessly.

I stuck with using it as long as I could, not knowing if that was all, but that pattern is copied from pre-existing below, so "ought to be OK".

benjaminpjones: Might be related to https://github.com/online-go/online-go.com/issues/1697

PR #1690, opened 2022-02-04
Casual rengo auto start

GreenAsJade:
Gives people a way to not have to manage the rengo challenge.

## Proposed Changes

Casual-mode rengo challenges can specift an auto-start number, and when that many people join, they start.

### DIFF
--- a/src/components/RengoManagementPane/RengoManagementPane.tsx
+++ b/src/components/RengoManagementPane/RengoManagementPane.tsx
@@ -15,6 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { interpolate } from "goban";
 import * as React from "react";
 
 import { _, pgettext } from "translate";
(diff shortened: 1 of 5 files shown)


GreenAsJade:
Note this is built on, and therefore includes, PR-1689 (relies on the type refactor)

src/components/RengoManagementPane/RengoManagementPane.tsx
```diff
@@ -15,6 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { interpolate } from "goban";
```
benjaminpjones, line 18: I think this needs to come from the translate module.  No sure what happens if you use goban's version...

GreenAsJade: LOL VSCode put that there for me.  It "seems" to work   :D

src/models/challenges.d.ts
```diff
@@ -45,6 +45,7 @@ declare namespace socket_api {
             time_per_move: number;
             rengo: boolean;
             rengo_casual_mode: boolean;
+            rengo_auto_start: boolean;
```
benjaminpjones, line 48: Is this really boolean?  If that's the case, this line won't render:

https://github.com/online-go/online-go.com/pull/1690/commits/928c708694ba79b5355ab2a1a2012f6e4f610c8f#diff-0213bf671b75d32ee5d114fa9778328a33dff730ace5addd93bc948dff2f0230R1338

However, I suspect this is a number, like the other one :)

GreenAsJade: Eyes like a hawk, this man!

GreenAsJade: ... but why didn't things complain?!

benjaminpjones: It's all valid code.  Keep in mind we're just making up types for the API.  No way to actually verify it (unless the backend were written in TS and we used the same .d.ts files)

benjaminpjones: Also, if you wanted a complaint, annotate the variable as string:


Screen Shot 2022-02-03 at 10 11 38 PM

GreenAsJade: ^^ I know it's just compiler hints, it's the complaining I was missing.   I think that's an excellent explanation though: if you want types to help you have to type your darn variables darn it.

GreenAsJade: ... although then it's a hassle because auto_start is a number 🤣

benjaminpjones: Oh wait yeah hmm.. well I guess explicit `number|string` is better than implicit `boolean|string`...

GreenAsJade: ... and if you make it as string then shortcut || doesn't work eh...

GreenAsJade: Ah - I was thinking  `const rengo_auto_start_text: string = C.rengo_auto_start.toString() || "No"`

but you meant `const rengo_auto_start_text: string | number = C.rengo_auto_start || "No".

Right, wake up me.

benjaminpjones: Yep that haha

PR #1691, opened 2022-02-05
Game history rengo merge

benjaminpjones:
Implements https://forums.online-go.com/t/rengo-status/40415/483

## Proposed Changes

  - Refactor game_history_groomer: this function was really long (120 lines!), so I broke out some logical functions.  It's still pretty long but now it's half the length and fits on one "page".
  - Make the normal columns handle rengo games (sort of)
  - Remove the rengo-specific table
  - Add some types to the Game History table

## Notes

I could use some API support:

- player lists or at least number of players for rengo matches
- ~~Fix the game result fields (black_lost, white_lost)~~ i hear a fix is in flight :)
- include the player whose page we are on in the `game.players` and `game.historical_ratings` (currently it just picks a player at random from each team?)

Screen Shot 2022-02-05 at 12 58 20 AM

(screenshot from [Sophiam's page](https://beta.online-go.com/player/1775))

### DIFF
--- a/src/lib/rank_utils.ts
+++ b/src/lib/rank_utils.ts
@@ -308,6 +308,9 @@ export interface EffectiveOutcome {
     white_effective_stronger: boolean;
 }
 
+/**
+ * @returns a ratings object containing ratings adjusted for the handicap.
+ */
 export function effective_outcome(
     black_rating: number,
     white_rating: number,
--- a/src/models/games.d.ts
+++ b/src/models/games.d.ts
@@ -15,6 +15,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 declare namespace rest_api {
+    /**
+     * One element of `results` from `player/%player_id%/games`
+     *
+     * This is a work in progress. Trust these values at your own risk.
+     * */
     interface Game {
         related: {
             detail: string; // route to full game info
@@ -38,7 +43,7 @@ declare namespace rest_api {
         komi: string; // floating point number
         time_control: import("../components/TimeControl").TimeControlTypes.TimeControlSystem;
 
-        // these don't appear to be populated with userful data
+        // these don't appear to be populated with useful data
         black_player_rank: number;
         black_player_rating: string; // floating point number
         white_player_rank: number;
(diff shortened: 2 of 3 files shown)


GreenAsJade:
> I could use some API support:
>  
>  * player lists or at least number of players for rengo matches

I submitted a fix for this.

> include the player whose page we are on in the game.players and game.historical_ratings (currently it just picks a player at random from each team?)

Pseudo random of course.  I _think_ it is the first players on each team :) 

This looks to need some thought from anoek, due to gotchas (we only have those two player's full details at the time this is being serialised)   

We could easily make it send nothing instead?

benjaminpjones:
>> - player lists or at least number of players for rengo matches
>I submitted a fix for this.

Awesome, thanks!

> We could easily make it send nothing instead?

If there are significant technical barriers, I suppose nothing is fine, but if we want to populate the rank column consistently, I think the player-of-interest needs to show up in the `historical_rankings` somehow.

Screen Shot 2022-02-05 at 4 12 15 PM

GreenAsJade:
I have a path forward to deliver the historical rating for the person who's page we are on.

The weird thing is that this api also returns the ratings for the player of the "other" colour.    In that case, I think I continue to return the "random" player in the "opposite" field - what else could it be?

benjaminpjones:
>I have a path forward to deliver the historical rating for the person who's page we are on.

Woo awesome! Thanks for the help with this, I didn't realize there was backend support needed when I started, but I think the changes you're adding will really make this sparkle :)

> The weird thing is that this API also returns the ratings for the player of the "other" colour. In that case, I think I continue to return the "random" player in the "opposite" field - what else could it be?

For normal games, the ratings are shown in the opponent column so I think that makes sense.  For rengo though, yeah I don't think I can logically show an "opponent" so random player (or no player) in that field is fine as far as the game history goes.

anoek:
@benjaminpjones I just pushed up @GreenAsJade 's API support for you to beta so you can validate things are behaving as you expect and whatnot. I'm gonna mark this as draft, when you've banged on it against the beta API and things are working as expected mark as ready for merge

benjaminpjones:
@anoek Sounds good!  Just so I understand the backend changes, past games haven't been corrected right?  I'll need to examine new rengo matches to see the changes?

Also, possibly related: when I go to beta.online-go.com I get a 500 error.  The site still functions but I see the pop up when I visit the homepage.

anoek:
Sorry, beta should be fixed now, I forgot to run a migration.

I *think* the way @GreenAsJade wrote it, it should work for past games too.. but don't hold me to that :)

benjaminpjones:
Awesome, thanks!  API looks good, I'll need to make some minor changes to get the new information to show up.

GreenAsJade:
... it's definitely the case that _lost is computed at game completion, so won't be fixed by the code change, but ought to be correct now for new games.

Nice table :)    I kinda liked "fun was had", in sending the right message, but I guess ... my fun was had :D

GreenAsJade:
^^ confirmed, it is "end_game_task", which runs once, that calculates and saves _lost.

anoek:
Great work again @benjaminpjones :)

PR #1698, opened 2022-02-08
Bugfix: incorrect result class

benjaminpjones:
Fixes: https://forums.online-go.com/t/possible-rating-bug/41395

## Proposed Changes

  - Remove a not where a not should not be.

<details>
<summary>Screenshots</summary>

BEFORE:
Screen Shot 2022-02-08 at 12 07 13 PM

AFTER:
Screen Shot 2022-02-08 at 12 02 22 PM
</details>

### DIFF
--- a/src/views/User/GameHistoryTable.tsx
+++ b/src/views/User/GameHistoryTable.tsx
@@ -290,7 +290,7 @@ function getResultClass(game: rest_api.Game, user_id: number): ResultClass {
         );
         const player_effective_stronger =
             (black_effective_stronger && played_black) ||
-            (!white_effective_stronger && !played_black);
+            (white_effective_stronger && !played_black);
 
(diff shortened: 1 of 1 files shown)


src/views/User/GameHistoryTable.tsx
```diff
@@ -290,7 +290,7 @@ function getResultClass(game: rest_api.Game, user_id: number): ResultClass {
         );
         const player_effective_stronger =
             (black_effective_stronger && played_black) ||
-            (!white_effective_stronger && !played_black);
+            (white_effective_stronger && !played_black);
```
benjaminpjones, line 293: `black_effective_stronger === played_black` probably also works but 🤷

anoek:
Thanks for the fix!

PR #1702, opened 2022-02-10
Remove AGA ranking checkbox

benjaminpjones:
This PR is more of a question than anything else.. The AGA ranking stuff was hidden behind a flag (`aga_rankings_enabled`) that doesn't appear to be settable via the UI.  Is this used in a way that I am missing?

## Proposed Changes

  - Remove (hidden) AGA Ranked checkbox and associated code from the Challenge Modal.

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -259,17 +259,6 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
 
         this.setState({ conf: conf, challenge: challenge });
     }
-    setAGARanked(tf) {
-        const next = this.nextState();
-
-        next.challenge.aga_ranked = tf;
-        if (tf && this.state.challenge && data.get("user")) {
-            next.challenge.game.ranked = true;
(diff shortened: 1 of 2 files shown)


src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -259,17 +259,6 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
 
         this.setState({ conf: conf, challenge: challenge });
     }
-    setAGARanked(tf) {
-        const next = this.nextState();
-
-        next.challenge.aga_ranked = tf;
-        if (tf && this.state.challenge && data.get("user")) {
-            next.challenge.game.ranked = true;
-            this.state.setRanked(true);
```
benjaminpjones, line 268: Also, looks like a bug - should be

```diff
- this.state.setRanked(true);
+ this.setRanked(true);
```

anoek:
Yeah I reckon it's safe to remove. Many moons ago the AGA was working towards doing online ratings and so we had that prepared for them, but their efforts seem to be largely abandoned now. It'd be easy enough to add back in should they ever get around to it I reckon.

PR #1703, opened 2022-02-11
Make rengo challenge chat consistent

GreenAsJade:
Fixes people complaining that the chat has gone, and that they can't view it before they join

## Proposed Changes

Put rengo chat window in the challenge list instead of the automatch pane


anoek:
Thanks for fielding the UI feedback :) I suspect this will be a bit confusing too as it puts the chat in a different place than where you're looking for live matches, but I'm not sure I have any great ideas short of re-envisioning this page all together for both rengo and non rengo games.

GreenAsJade:
Yes, there's no perfect solution here, other than working out why we can't have two on the same page and fixing that 😮 

At least this way, non-participants can check the chat 😊

PR #1708, opened 2022-02-13
Add rest_api.GameDetails

benjaminpjones:
Adds a type to describe the `/api/v1/games/%game_id%` endpoint.


anoek:
Awesome, thanks for all this typing work :)

PR #1711, opened 2022-02-15
Update tournaments to show details as soon as they are loaded.

benjaminpjones:
So I was looking at [Through the Years: Long Correspondence](https://online-go.com/tournament/59567), and thinking to myself "I'm sure we could show *something* before the whole thing loads".  Well turns out there's 3 API calls and only one of them is insanely slow - all players.

There are a lot of components that don't depend on `players`, only the tournament details, so we can load all of those first.  Between the description, the info, and the chat, it makes up almost a full page of content!

## Proposed changes

- Render description, info and chat when the details are loaded (no longer wait for `rounds`/`players`)

I made a side-by-side demo (new stuff on the left):

https://user-images.githubusercontent.com/25233703/154016273-b3fdfbbc-1091-4d11-ada5-6c6cd4d8e79b.mov

Unfortunately running both at once caused the page to freeze up (the loading icon normally keeps spinning) but already one can see that the left side shows meaningful information earlier.

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -60,6 +60,7 @@ interface TournamentState {
     new_tournament_group_id: number;
     tournament_id: number;
     loading: boolean;
+    info_loaded: boolean;
     tournament: {
         id: number;
         name: string;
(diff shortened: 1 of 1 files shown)


BHydden:
I think the same improvement could be done with profile pages, if we delayed loading games (both active and historical) until after the rest of the page loads?

anoek:
Awesome, good improvement :) I'll have to get used to the position of the chat box, but I reckon it's for the better.

benjaminpjones:
@anoek uh oh did I move the chat box? I moved it out of the div so that I could load it separately, but I didn't mean to change the location!

anoek:
Yeah it used to be on the left (although apparently not for all tournament types perhaps?)

image


and now it spans on top of the results and whatnot

image

For double elimination I guess it's always been spanning the top though, so this kinda normalizes things, maybe a good thing.

PR #1712, opened 2022-02-15
Check if player is in rengo black team before saying white paused.

GreenAsJade:
Fixes https://forums.online-go.com/t/rengo-status/40415/685?u=greenasjade

## Proposed Changes

  Check if player is in rengo black team before saying white did the pause


anoek:
Thanks for the fix!

PR #1714, opened 2022-02-16
TimeControlPicker.tsx: sanitize value of numeric inputs

### LINKED ISSUE #1628: Players can create games with null x 10s byo-yomi

maurelio1234:
Fixes #1628 

## Proposed Changes
 
  - Sanitize values of numeric input fields so that they match the min value if the stored value is falsy.
 
In fact the bug exists on both Canadian and byo-yomi rules. Hopefully the back-end blocks invalid Canadian challenges.

The point here is also starting a discussion on what is the best fix. 
Sanitizing all values in the localstorage before use seems to be a much better fix, but it is a much bigger change and we could also argue that its content is valid since we wrote it... (except when in this precise case).

### DIFF
--- a/src/components/TimeControl/TimeControlPicker.tsx
+++ b/src/components/TimeControl/TimeControlPicker.tsx
@@ -484,7 +484,7 @@ export class TimeControlPicker extends React.PureComponent<
                                     min="1"
                                     max="300"
                                     className="challenge-dropdown form-control"
-                    value={(this.state as TimeControlTypes.ByoYomi).periods || 1}
+                                    value={(this.state as TimeControlTypes.ByoYomi).periods || 1}
                                     onChange={this.update_periods}
                                 />
                             </div>
@@ -545,7 +545,8 @@ export class TimeControlPicker extends React.PureComponent<
                                     max="50"
                                     className="challenge-dropdown form-control"
                                     value={
-                                        (this.state as TimeControlTypes.Canadian).stones_per_period || 1
+                                        (this.state as TimeControlTypes.Canadian)
+                                            .stones_per_period || 1
                                     }
                                     onChange={this.update_stones_per_period}
                                 />
--- a/src/components/TimeControl/TimeControlPicker.tsx
+++ b/src/components/TimeControl/TimeControlPicker.tsx
@@ -484,7 +484,7 @@ export class TimeControlPicker extends React.PureComponent<
                                     min="1"
                                     max="300"
                                     className="challenge-dropdown form-control"
-                                    value={(this.state as TimeControlTypes.ByoYomi).periods}
+                                    value={(this.state as TimeControlTypes.ByoYomi).periods || 1}
                                     onChange={this.update_periods}
                                 />
                             </div>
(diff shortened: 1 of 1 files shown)


benjaminpjones:
>Sanitizing all values in the localstorage before use seems to be a much better fix, but it is a much bigger change and we could also argue that its content is valid since we wrote it... 

Yeah this is a good question. I'm hoping now that `data.get()` and `data.set()` are typed in a way that they have to match, it will be harder to save bad data... but it's still possible especially if the data scheme changes.

maurelio1234:
> @maurelio1234 if you run `npx husky install`, you will be saved from "fix lint" commits grinning

anything against it being in the postinstall so that it is installed automatically? 
https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts

benjaminpjones:
No qualms on my end about making installation automatic.  `prepare` seems to be recommended over `postinstall` though.

https://typicode.github.io/husky/#/?id=install
https://github.com/typicode/husky/issues/884

maurelio1234:
> No qualms on my end about making installation automatic.  `prepare` seems to be recommended over `postinstall` though.
> 
> https://typicode.github.io/husky/#/?id=install
> https://github.com/typicode/husky/issues/884

LGTM. 

I just want to avoid having to think about it consciously. 

When I have time I will create another PR with that and see what happens 😏.

anoek:
Thanks for the improvement/fix @maurelio1234 !

src/components/TimeControl/TimeControlPicker.tsx
```diff
@@ -194,7 +194,7 @@ export class TimeControlPicker extends React.PureComponent<
     update_period_time = (ev) => this.syncTimeControl({ period_time: parseInt(ev.target.value) });
     update_periods = (ev) =>
         this.syncTimeControl({
-            periods: Math.max(1, Math.min(300, parseInt(ev.target.value) || 1)),
```
anoek, line 197: I hit merge on this too quick, this is still needed as folks can put in whatever they want in those input fields, so folks can create a game with 3248987 periods or whatever. (I should probably validate that on the server too.. but would be good to enforce that here as well)

maurelio1234: I will check again later. But as far as I remember the value comes from the numeric input with a Max and min correctly set. Users cannot change them as they want we just need to make sure we give it a sane initial value. Otherwise it just passes the value forward (like here) without changing it.

anoek:
They're set correctly there but input fields only use that for enforcing it when folks are using the up/down arrows or scroll wheel or whatever. If they type in a number, they can put in whatever they want and it'll accept it, which is the problem and the need for second check.

src/components/TimeControl/TimeControlPicker.tsx
anoek: They're set correctly there but input fields only use that for enforcing it when folks are using the up/down arrows or scroll wheel or whatever. If they type in a number, they can put in whatever they want and it'll accept it, which is the problem and the need for second check.

maurelio1234: Oh I see! I will propose something later tonight.

maurelio1234: @anoek what do you think about this approach: https://github.com/online-go/online-go.com/pull/1721 ?

maurelio1234: hmm there's still something missing here... a null value on the storage will still be sent to the backend even though the UI looks sane ...

shinuito: Yes, people can still create null byo-yomi games, presumably if it’s in local storage.

PR #1716, opened 2022-02-18
Rengo team balance button in rengo management pane

turtoise0:
## Proposed Changes

  - Adds button for organizer to unassign players from rengo teams
  - Clicking the button again reassigns players in approximately balanced teams

### DIFF
--- a/src/lib/rengo_balancer.ts
+++ b/src/lib/rengo_balancer.ts
@@ -25,15 +25,19 @@ type Participant = {
 };
 
 // Convert player IDs to Participants with ratings.
-async function toParticipants(playerIDs: number[]): Promise<Participant[]> {
-    const participants = [];
+function toParticipants(playerIDs: number[]): Promise<Participant[]> {
     const required_fields = ["rating"];
+    const promises = [];
     for (const id of playerIDs) {
-        // TODO is there a way to fetch multiple players at once?
-        const entry = await player_cache.fetch(id, required_fields);
-        participants.push({ user_id: id, rating: entry.rating });
+        const promise = player_cache.fetch(id, required_fields).then((entry) => {
+            return {
+                user_id: id,
+                rating: entry.rating,
+            };
+        });
+        promises.push(promise);
     }
-    return participants;
+    return Promise.all(promises);
 }
 
 function sortParticipants(participants: Participant[]) {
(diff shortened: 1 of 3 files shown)


src/lib/rengo_balancer.ts
```diff
@@ -0,0 +1,154 @@
...
+    user_id: number;
+    rating: number;
+};
+
+// Convert player IDs to Participants with ratings.
+async function toParticipants(playerIDs: number[]): Promise<Participant[]> {
+    const participants = [];
+    const required_fields = ["rating"];
+    for (const id of playerIDs) {
+        // TODO is there a way to fetch multiple players at once?
```
benjaminpjones, line 32: I think `termination-api/players` supports this with the `ids` field, but `PlayerCache` may be doing this under the hood for you already with the `Batcher` utility:

https://github.com/online-go/online-go.com/blob/46b0852579d02da7035e6685f9b3cbf532395f93/src/lib/player_cache.ts#L241

You can confirm by watching the network tab in developer tools.

Screen Shot 2022-02-18 at 12 58 51 AM

BHydden: What is this witchcraft? Such nimble navigation 😮 starting to think you know the code better than anoek 😅

src/lib/rengo_balancer.ts
```diff
@@ -0,0 +1,154 @@
...
+        const entry = await player_cache.fetch(id, required_fields);
```
benjaminpjones, line 33: Oh I see what the comment was about - `await` may cause batching not to happen.  In order to parallelize the async calls, it is good practice to collect a list of promises, and then combine them with `Promise.all()`. Something like:

```
const promises = [];
for (const id of ids) {
    const participant_promise = fetch(id).then((entry) => {
        return {user_id: entry.id, rating: entry.rating};
    };
    promises.push(participant_promise);
}
return Promise.all(promises);
```

(irrelevant parts omitted for brevity)

turtoise0: Thanks, I've added these changes.

anoek:
Nice improvement, thanks @turtoise0 !

Sofia-mal:
https://forums.online-go.com/t/rengo-status/40415/808?u=_sofiam 
Benjito brought me here :)

First of all, thank you for that!

Using the "auto-balance" button before starting [this game](https://online-go.com/game/41546691) we obtained the following teams: (Rating(Nick))

1604 (He Who Walks in Shadows)
1642 (Vladimir Lukic)
1929 (Decros)
Total: 5175

1578 (PopcornCzar)
1737 (🐬Sofiam🐬)
2042 (NateC89)
Total: 5457

5457-5175= 282

There are several ways (e.g. swapping NateC89 and Decros, or 🐬Sofiam🐬 and He Who Walks in Shadows) to get a difference of less than 282.
Not a huge difference in this case, but the obtained teams are not the best. 
No idea what went wrong, or if it's just ok 😅

PR #1718, opened 2022-02-18
Paginated table types

benjaminpjones:
`forwardRef` kind of eats up template parameters (or rather turned them into `any`).  This casts the type of the modified component so that callers can benefit from type-checking. The type check goes like so: the return value of `groom` must be the same as the argument of `className()` and `render()`

## Proposed Changes

  - Add strong type to `PaginatedTable`
  - Remove inferrable (and repetitive) type annotations in `GameHistory`

Screen Shot 2022-02-18 at 12 29 01 AM

Now, the compiler knows what `X` is whether or not we annotate explicitly :)


anoek:
Awesome :)

PR #1720, opened 2022-02-18
Add faster simple correspondence timings.

benjaminpjones:
Rengo has awakened interest in more simple time settings (see forums: https://forums.online-go.com/t/rengo-status/40415/468).

## Proposed Changes

  - Allow simple correspondence to be as fast as 8hr per move.

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -209,7 +209,7 @@ export const time_options = {
             max_time: gen(86400, 86400 * 28),
         },
         simple: {
-            per_move: gen(3600 * 8, 86400 * 28),
+            per_move: gen(3600 * 12, 86400 * 28),
         },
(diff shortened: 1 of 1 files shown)


anoek:
I don't know about that, 8 hour simple time is pretty brutal, kinda means the last thing you do before going to bed is make sure it's not your turn and you'd better not get a full nights sleep...

benjaminpjones:
That's fair. Actually I chose 8 hrs because one could get a full(ish) night's sleep 😅

Would 12hr or 16hr minimum work instead? I figure it's not for everybody, but for people who do like to play in the morning and at night it would be a nice option to have

anoek:
I bumped it to 12 hours.. I still think that might result in a lot of timeouts but I reckon we can see, not like folks are forced to accept

PR #1721, opened 2022-02-18
TimeControlPicker.tsx: only accept valid numeric values

### LINKED ISSUE #1628: Players can create games with null x 10s byo-yomi

maurelio1234:
Fixes #1628

## Proposed Changes

  - Only accept valid values on numeric inputs on `update_periods` and `update_stones_per_period`
  - Fix invalid numbers in `[makeTimeControlParameters]`

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -225,6 +225,11 @@ export const time_options = {
     },
 };
 
+function parseIntWithDefaultValue(value: string, defaultValue: number): number {
+    const parsedInt = parseInt(value, 10);
+    return isNaN(parsedInt) ? defaultValue : parsedInt;
+}
+
 export function makeTimeControlParameters(tc: any): TimeControl {
     const tpm = computeAverageMoveTime(tc);
     const speed: TimeControlTypes.TimeControlSpeed =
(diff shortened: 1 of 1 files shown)


maurelio1234:
What do you think about this approach? 

I can go even further and move the  `default_time_options` to `TimeControlPicker` and use the default values for all ints in `makeTimeControlParameters`  if they return NaNs.

src/components/TimeControl/TimeControlPicker.tsx
```diff
@@ -193,11 +193,13 @@ export class TimeControlPicker extends React.PureComponent<
     //update_main_time            = (ev)=>this.syncTimeControl({main_time: ev.target.value});
     update_period_time = (ev) => this.syncTimeControl({ period_time: parseInt(ev.target.value) });
     update_periods = (ev) =>
```
maurelio1234, line 195: Please check https://github.com/online-go/online-go.com/issues/1727

anoek:
I like how you're enforcing it, *but* doing this in `onChange` makes it hard to edit the field. For instance, if folks try and clear out the field to type in `5`, it won't let you, it'll force a 1 in there, so you have to do something like select it all and press `5` or end up with `15` and then delete the `1`, either way it's a bit hard to use.

What about accepting whatever text they want to enter in `onChange`, and then correct it when `onBlur` is triggered, or something like that?

Issue #1713, opened 2022-02-15
Blank tournament page


gezabohus:
Link to the 3rd International Mandelbrot Tournament leads to a blank page. It's been like this for a day or two. Link: https://online-go.com/tournament/85113

anoek:
Can you provide a link to the tournament please?

gezabohus:
Sorry, there now.

BHydden:
> There is a flash of what appears to be the correct page, but it is replaced with the blank page within a second.

I see the same thing, these are the console errors I get.

![image](https://user-images.githubusercontent.com/28991649/154766376-9035e20d-4b6d-46e3-95fa-a99de775db8d.png)


```
Uncaught TypeError: Cannot read properties of undefined (reading 'notes')
    at wm (ogs.5.1-4075-g87407d2b-8c5b6ad03f2844c9c4785ce85007af68.js:2:770638)
    at Array.sort (<anonymous>)
    at ogs.5.1-4075-g87407d2b-8c5b6ad03f2844c9c4785ce85007af68.js:2:810306
    at Array.map (<anonymous>)
    at km.render (ogs.5.1-4075-g87407d2b-8c5b6ad03f2844c9c4785ce85007af68.js:2:810270)
    at es (vendor.f7abdec63796a339d922bf4e7590325e.js:2:927270)
    at Qa (vendor.f7abdec63796a339d922bf4e7590325e.js:2:927052)
    at nc (vendor.f7abdec63796a339d922bf4e7590325e.js:2:971434)
    at Ic (vendor.f7abdec63796a339d922bf4e7590325e.js:2:955158)
    at qc (vendor.f7abdec63796a339d922bf4e7590325e.js:2:954980)
    at Wc (vendor.f7abdec63796a339d922bf4e7590325e.js:2:954813)
    at zc (vendor.f7abdec63796a339d922bf4e7590325e.js:2:951646)
    at vendor.f7abdec63796a339d922bf4e7590325e.js:2:900396
    at t.unstable_runWithPriority (vendor.f7abdec63796a339d922bf4e7590325e.js:2:1395222)
    at Ki (vendor.f7abdec63796a339d922bf4e7590325e.js:2:900173)
    at Zi (vendor.f7abdec63796a339d922bf4e7590325e.js:2:900341)
    at $i (vendor.f7abdec63796a339d922bf4e7590325e.js:2:900276)
[...]

PR #1722, opened 2022-02-19
Make sure sorted_players doesn't contain undefined values.

benjaminpjones:
Addresses #1713

I can see the tournament now, but not sure if this is the "right" fix.  The issue is that a couple of the players are no longer returned from the players/%%/players/all endpoint and therefore places that expect a player may get `undefined` instead.  The real question is why aren't these players populated?  They don't appear to be deleted players or anything...

`DavidHall` (id: 527014)  and `DQuad` (id: 371) are the missing players in the tournament from the bug: https://online-go.com/termination-api/tournament/85113/players

Sharing this patch in case it helps debug. (also better than showing nothing 😄)

## Proposed Changes

  - For round robin tournaments, filter out undefined values from the players array.

<details><summary>screenshot</summary>
Screen Shot 2022-02-18 at 6 21 33 PM
</details>


anoek:
Interesting, wonder why that's happening in the first place.. Thanks for the client side fix regardless!

PR #1724, opened 2022-02-20
User.tsx: Move vs stuff into its own component

benjaminpjones:
I thought it would make sense to gather the `vs`-related code together in one component.  Currently, the logic is split across `update()` and `render()`.  Now, there is no `vs` logic in `update`, and the JSX in `User.render()` is simpler.

## Proposed Changes

  - Create `VersusCard` component
  - Add `PlayerDetails` to `rest_api` types


anoek:
Much better, thanks!

PR #1726, opened 2022-02-21
Overview.tsx cleanups

benjaminpjones:
This (hopefully) doesn't have any visible changes, but now Overview.tsx is compliant with TS strict mode.

## Proposed Changes

  - Remove some unused variables/code
  - Add types
  - Remove some `abort_requests_in_flight()` calls
      - This may have a functional effect, but I don't think we should be unsubscribing from these fetches from the Overview components now that the fetches are initiated by cached.ts

### DIFF
--- a/src/lib/data_schema.ts
+++ b/src/lib/data_schema.ts
@@ -41,7 +41,7 @@ interface CachedSchema {
 }
 
 export interface ConfigSchema {
-    user: any;
+    user: rest_api.UserConfig;
     cdn: string;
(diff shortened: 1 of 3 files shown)


anoek:
Thanks for the cleanup :) 

Re the removal of `abort_requests_in_flight`, I don't think it'll hurt anything in practice, and I'm generally not opposed to removing them as they don't actually stop any work being done on the server, so no big deal.

PR #1730, opened 2022-02-25
User.tsx: move Avatar into its own component

benjaminpjones:
## Proposed Changes

  - Move Avatar into its own component
  - Some additional code cleanups
      - Remove unused timeout_provisional code
      - Strict mode compliance
      - added react-dropzone type (bumping to current version would also work, but I couldn't test this so left that for another day.)
  - Default to https for websites where the URL is not specified.
  - Fix bug where vacation time takes a second to show up
  - Fix bug where missing bot_owner caused page to crash (thanks for reporting, @shinuito!)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   ],
   "devDependencies": {
     "@types/jest": "^26.0.15",
+    "@types/react-dropzone": "^3.13.4",
     "@typescript-eslint/eslint-plugin": "^5.9.1",
     "@typescript-eslint/eslint-plugin-tslint": "^5.9.1",
     "@typescript-eslint/parser": "^5.9.1",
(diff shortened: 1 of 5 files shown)


BHydden:
Sorry if this is a silly question, but why is user.tsx a bad place to have avatar details kept?

BHydden:
Thank you for your detailed and thought-out response. I really appreciate it. As a programming novice, I can find specific answers to code well enough, but not always these higher-level practices and their pros/cons in an easily digestible form.

My thought was that avatar and user were so strongly linked that it seemed silly for avatar to have its own file. I do agree with the single responsibility principle and feel that often our components here in OGS are bloated and weighed down with many tasks. I did not consider the benefit you mentioned of break-out components having their own file makes seeing dependencies for that file easier. 

Obviously, there is some balance between each function having its own file and every component being in one massive file, and where to draw that line will necessarily cause conflict on best practice. I don't think I know enough about the specifics to comment on your choice here, but I appreciate you sharing your reasoning and giving me something to think about.

benjaminpjones:
@BHydden, no problem!  By the way, that last bugfix commit makes me think of another reason to break out components: the stack trace is much more specific when you break the component hierarchy up.

Originally:

**User.tsx > Player.tsx** hmm... there's lots of places `<Player>` pops up on the User page, the bug could be anywhere!

Now:
**User.tsx > Avatar.tsx > Player.tsx**: okay now we've narrowed the problem down quite a bit.  There are only 2 instances of `<Player>` in Avatar.tsx.

Future (just kidding, but illustrating a point):
**User.tsx > Avatar.tsx > BotInfo.tsx > Player.tsx**: I wouldn't even have to open an IDE to debug!

anoek:
Thanks @benjaminpjones !

PR #1731, opened 2022-02-27
Show console logs from rengo team balancer

turtoise0:
## Proposed Changes

  1. Remove unnecessary props from RengoManagementPane
  2. Show console logs from rengo team balancer
  
## Context for (2)

The rengo balancer [doesn't seem to be working correctly](https://forums.online-go.com/t/rengo-status/40415/828). It's hard to test on the beta site, because there's not a lot of players with stable ranks there. The console logs would make it easier to find the problem.


anoek:
Thanks for the continued work on this!

PR #1735, opened 2022-03-06
ChallengeModal.tsx: set komi to 5.5 if it is not defined

### LINKED ISSUE #1727: Server Error (500) when trying to challenge
I am getting a 500 Server Error when trying to send a challenge.

**To Reproduce**
- I am https://online-go.com/player/4928/
- I am in the group https://online-go.com/group/9655 and on that group page
- I click yeonwoo in the Admins frame and choose Challenge (german „Herausfordern“)
- I set
  - Game name: Awesome teaching game
  - Private: no
  - Ranked: no
  - Board size: 19×19
  - Rules: Japanese
  - Speed: correspondence
  - Time: Fischer
  - Starting time: 3 d
  - Time increment: 1 d
  - Time cap: 3 d
  - Pause on weekends: no
  - Handicap: 3
  - Komi: custom, 0
  - Colour: Black
  - Deactivate analysis: yes
- Then I click send challenge

**Expected behavior**
A challenge is sent.

**Instead**
A modal window appears saying »(X) Server Error (500) [OK]«

**Screenshots**
ogs-error

 - OS: Arch Linux
 - Browser Firefox 97.0 (64-Bit)

maurelio1234:
Fixes #1727 

In the end of the day the problem seems to be unrelated to https://github.com/online-go/online-go.com/pull/1721
It seems that there's a confusion on the `komi` property.
In `challenges.d.ts` it is defined as a `number` whereas we clearly send a string to the backend.

The constructor of `ChallengeModal` contains code to "fix dirty data", so I guess it's the right place to check if the `komi` is "valid". 

Maybe we could even parse it and check if it is a `float`. What do you think?

The only thing I don't understand (yet) is why the default komi is set to the empty string, as I had this value on my local environment already, even though I cannot see traces of it on the code.

## Proposed Changes

  - Replace `null` check by truthy check, so that the empty string is replaced by the valid default handicap

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -161,7 +161,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
         }
 
         challenge.game.initial_state = null;
-        if (!challenge.game.komi) {
+        if (typeof challenge.game.komi !== "number" && !challenge.game.komi) {
             challenge.game.komi = 5.5;
         }
 
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -161,7 +161,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
         }
 
         challenge.game.initial_state = null;
-        if (challenge.game.komi == null) {
+        if (!challenge.game.komi) {
             challenge.game.komi = 5.5;
         }
 


benjaminpjones:
> Replace `null` check by truthy check

I think 0.0 is a valid setting, and a truthy check would change it to 5.5, right?

maurelio1234:
> I think 0.0 is a valid setting, and a truthy check would change it to 5.5, right?

that's what's weird about this code... 0.0 (the number) is not valid, "0" or "0.0" are.
i will do some tests, I'd prefer dealing with actual numbers then stringifyed numbers.
i will check if changing the type in `challenges.d.ts` works...

BHydden:
IIRC, numbers are (annoyingly) stringified to help the translator, as not all languages handle numbers the same (position within a sentence, etc.) and the translator code takes strings as input... but I could be wrong 😛

benjaminpjones:
@maurelio1234 Oh got it that makes sense, thanks. I really would have expected that to be a number!

anoek (changes requested):
Can we add a check to see if it's not a number as well? Otherwise if/when we get around to fixing this component so `komi` is a number like it should be, this code will start replacing `0` with `5.5` as noted above.

anoek:
Thanks for the fix @maurelio1234 !

PR #1737, opened 2022-03-08
Hide rengo balance button from non-organisers

turtoise0:
In my previous pull request (#1732), I forgot to hide the balance button from non-organizers. 

## Proposed Changes

  - Only show the button to owner of rengo challenge.


anoek:
Thanks for the improvement!

anoek:
Thanks for the improvement!

PR #1743, opened 2022-03-12
Bugfix and cleanup: AvatarCard

### LINKED ISSUE #1742: Old user icon rendered after saving on the profile page
**Describe the bug**
If you change the icon on the profile page, it looks like it updates.  But when you hit save, it reverts to the old icon.  This issue was introduced in #1730 (sorry!)

I have a fix drafted, but creating a bug for tracking purposes.

**To Reproduce**
1. Go to profile page for logged in user
2. Click on 'edit'
3. Click on the icon, select an image to upload.
4. Click 'Save'
5. Observe that the old icon is shown (both on the profile page, and in the icon in the top right)

**Expected behavior**
The new icon should be shown.

benjaminpjones:
Fixes #1742.  I'm not entirely sure what changed in #1730 that caused the breakage, but basically, we are calling `player_cache.update()` with `User.state.user` which has `user.icon` set to the outdated icon from the initial fetch.  Setting `user.icon` when modifying the icon seems to fix this.

Full disclosure: there is no good way to test the "updateIcon" codepath since I get a 500 error when saving a new icon from `localhost`.  However, the "clearIcon" path is basically the same and I was able to test that successfully.

## Proposed Changes

  - Fix a bug where the old icon got rendered when the user saves a new icon.
  - Move editing state variable into props.
  - Delete a couple of unused state properties in User.tsx.

### DIFF
--- a/src/views/User/AvatarCard.tsx
+++ b/src/views/User/AvatarCard.tsx
@@ -62,6 +62,7 @@ interface AvatarCardUserType extends AvatarCardEditableFields {
 interface AvatarCardProps {
     user: AvatarCardUserType;
     force_show_ratings: boolean;
+    editing: boolean;
 
     /** called when the edit button is clicked */
     onEdit: () => void;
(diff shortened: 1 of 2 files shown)


anoek:
Awesome thanks for the fix!

anoek:
Awesome thanks for the fix!

PR #1748, opened 2022-03-18
Cleanup: Remove unused properties and methods.

benjaminpjones:
## Proposed Changes

  - Remove class properties that were either unused, or set but not read
  - Remove class methods that weren't used
      - Many of these functions were moved into PlayerDetails popup, but never removed from User.tsx
      - I'm guessing the superUser stuff was moved into ModerateUser modal (ps ~no more~ less jQuery!!)


anoek:
Thanks for the cleanup!

PR #1751, opened 2022-03-27
Don't show 'Support OGS' if not logged in.

benjaminpjones:
The new support OGS page doesn't show for anonymous users.  This makes sense because there are no more one-off donations.  But if that's the case, I don't think that page should be linked from the menu.

Screen Shot 2022-03-27 at 1 01 25 AM

Forum thread: https://forums.online-go.com/t/support-page-unavailable-without-login/41438

## Proposed Changes

  - Remove 'Support OGS' from the side menu if user is not logged in.


anoek:
Yikes, thanks for bringing it up. I fixed the page instead so something shows up

PR #1753, opened 2022-03-28
More typing in Chat.tsx

benjaminpjones:
This is a types only change, so should have no effect on functionality (or even generated code)


anoek:
Hooray for types, thanks!

PR #1755, opened 2022-03-29
Rename termination_socket and comm_socket to socket.

benjaminpjones:
## Proposed Changes

 Rename `termination_socket` and `comm_socket` to `socket`.  Since they are the same:

https://github.com/online-go/online-go.com/blob/052321f1415044e6c2cc59e07b66f981f346942b/src/lib/sockets.ts#L46

this does not constitute a functional change, but it will be more clear to anyone not familiar with the sockets.

Tested: while I'm sure I can't test every socket call on the site, it looks like I get the usual socket traffic:

Screen Shot 2022-03-29 at 12 39 58 AM

And I tried a few socket-y things like playing a move in a game and sending a PM.


anoek:
Awesome thanks, I've been meaning to do that for awhile :)

PR #1756, opened 2022-03-31
Reinstate position and user joseki admin filters 

GreenAsJade:
## Proposed Changes

Tweak the user and position audit filters in joseki admin to work with the new backend.

(New backend still doesn't support 'audit type' filter)


anoek:
Thanks!

PR #1757, opened 2022-03-31
Popup a player card for who played a clicked rengo move in the tree

GreenAsJade:
Fixes wondering who played a move, and squinting at the player cards

## Proposed Changes

Popup a player card when told by Goban that a move with a played_by was clicked on.

Requires https://github.com/online-go/goban/pull/65

### DIFF
--- a/src/lib/popover.ts
+++ b/src/lib/popover.ts
@@ -30,7 +30,7 @@ interface PopupCoordinates {
 
 interface PopoverConfig {
     elt: React.ReactElement<any>;
-    at?: PopupCoordinates;
+    at?: PopupCoordinates; // Relative to window origin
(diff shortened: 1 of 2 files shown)


GreenAsJade:
I reckon this would build cleanly now that the goban PR is merged, but I can't see how to retrigger that (short of pushing another checkin)

anoek:
Yep works good.

I feel like it's a little unexpected when clicking the move in the tree to have that player box pop up, I think ultimately we'll want a different way to indicate who made the move, but I don't currently have any better idea's so might as well merge and see if the community has some feedback or ideas there.

GreenAsJade:
This was exactly my thinking.   Agile and all that :D

PR #1762, opened 2022-04-03
Enables no-empty-function eslint rule

cacieprins:
Fixes #

## Proposed Changes

  - Enables no-empty-function eslint rule
  - Removes empty componentWillUnmount lifecycle methods
  - Changes promise `.then` chains to async/await when `then` was noop but catch was significant
  - removed a no-op `catch` from `Supporter.tsx`
  - `submitNote` in `ModNoteModal.tsx` changed: `close` event emitter moved to before promise is invoked via `await` - this preserves the ux of closing the modal while waiting for xhr to complete. Was a little worried about preserving a callback after a component had been unmounted, but it doesn't look like `errorAlerter` relies on `ModNoteModal` itself
  - removed no-op `save` method on `<Report />`; commented out the `Save` button, since clicking it is a no-op. 
  - autoformatter caught `.eslintrc.js` - I hope these superficial changes are okay to include in this PR; if not, I'll open a separate one

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,31 +1,21 @@
 module.exports = {
-    "env": {
-        "browser": true,
-        "es6": true
+    env: {
+        browser: true,
+        es6: true,
     },
-    "parser": "@typescript-eslint/parser",
-    "parserOptions": {
-        "project": "tsconfig.json",
-        "sourceType": "module"
+    parser: "@typescript-eslint/parser",
+    parserOptions: {
+        project: "tsconfig.json",
+        sourceType: "module",
     },
-    "plugins": [
-        "eslint-plugin-jsdoc",
-        "@typescript-eslint",
-        "@typescript-eslint/tslint",
-        "prettier"
-    ],
-    "extends": [
-        "eslint:recommended",
-        "plugin:@typescript-eslint/recommended",
-        "prettier"
-    ],
-    "rules": {
+    plugins: ["eslint-plugin-jsdoc", "@typescript-eslint", "@typescript-eslint/tslint", "prettier"],
+    extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"],
+    rules: {
         // Recommended rules with errors
         // TODO: Fix these and re-enable them
 
         "@typescript-eslint/ban-types": "off", // 35 errors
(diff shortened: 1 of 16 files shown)


anoek (commented):
Awesome, thanks for the cleanup!

.eslintrc.js
```diff
@@ -41,36 +31,33 @@ module.exports = {
 
         //==============================================================
 
-
         "@typescript-eslint/adjacent-overload-signatures": "error",
         "@typescript-eslint/consistent-type-assertions": "error",
         "@typescript-eslint/member-delimiter-style": "error",
+        "@typescript-eslint/no-empty-function": "error",
```
benjaminpjones, line 37: fwiw, I don't think you have to re-add the "recommended" rules below, they are included automatically by `extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"...]`

src/components/ModNoteModal/ModNoteModal.tsx
```diff
@@ -40,11 +40,14 @@ export class ModNoteModal extends Modal<Events, ModNoteModalProperties, any> {
...
-
+    submitNote = async () => {
+        try {
+            await put(`players/${this.props.player_id}/moderate`, {
+                moderation_note: this.state.current_draft,
+            });
+        } catch (e) {
+            errorAlerter(e);
+        }
         this.close();
```
benjaminpjones, line 51: I think the description said close was moved before the xhr call, is there a commit missing?

src/views/User/ModTools.tsx
```diff
@@ -44,9 +44,13 @@ export function ModTools(props: ModToolsProps): JSX.Element {
...
-        }).catch(errorAlerter);
+        try {
+            await put(`players/${props.user_id}/moderate`, {
+                moderation_note: txt,
+            });
+        } catch (e) {
+            errorAlerter(e);
+        }
 
         moderator_note.current.value = "";
```
benjaminpjones, line 55: Not sure if it matters, but I think the behavior changes here in a similar way. (`moderator_note.current.value` will be set to `""` after the xhr completes)

anoek:
Awesome, thanks @cacieprins !

PR #1766, opened 2022-04-03
Remove user switch for OJE backend

GreenAsJade:
and  rename data-based selector so old… one is not used for anyone

Fixes anyone using the old one, which they aren't supposed to anymore.

## Proposed Changes

Took backend toggle out of UI, and renamed the key in `data` so that any old user settings of that are no longer used.


anoek:
Awesome, thanks for the continued effort on the Joseki system :)

PR #1767, opened 2022-04-04
User.tsx refactorings

benjaminpjones:
This one is a little bigger than previous ones.  I was trying to do everything that would make incrementally loading the profile page easier. Please let me know if I should break this up into separate PRs.

## Proposed Changes

- Remove the update function: most of it was not necessary except for the `setState` call
- Factor out a few more components:
    - Activities
    - Bot Management
- deleted `syncRating()` as it was unused
- add types for selected size and speed
- add UserProperties and UserState as parameter for the props functions
- fix ts "strict" errors.  The biggest one here was strictNullChecks

### DIFF
--- a/src/models/user.d.ts
+++ b/src/models/user.d.ts
@@ -14,6 +14,8 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+
+type Server = "kgs" | "igs" | "dgs" | "golem" | "wbaduk" | "tygem" | "fox" | "yike" | "goquest";
 declare namespace rest_api {
     interface RatingsConfig {
         rating: number;
@@ -56,6 +58,21 @@ declare namespace rest_api {
         is_announcer: boolean;
     }
 
+    type AccountLinks = {
+        [org in `org${1 | 2 | 3}`]?: string;
+    } & {
+        [org_id in `org${1 | 2 | 3}_id`]?: string;
+    } & {
+        [org_rank in `org${1 | 2 | 3}_rank`]?: string;
+    } & {
+        [server_username in `${Server}_username`]?: string;
+    } & {
+        [server_rank in `${Server}_rank`]?: string;
+    } & {
+        hidden: boolean;
+        hidden_ids: boolean;
+    };
+
     interface PlayerDetails {
         user: {
             id: number;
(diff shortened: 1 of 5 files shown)


src/models/user.d.ts
```diff
@@ -95,6 +95,13 @@ declare namespace rest_api {
             registration_date: "2018-04-09T13:04:44.987830Z";
             vacation_left: number;
             on_vacation: boolean;
+            self_reported_account_linkages?: {
+                org1: string;
+                org1_id: string;
+                org1_rank: string;
+                last_updated: number;
+            };
```
anoek, line 103: Should/could we use https://github.com/online-go/online-go.com/pull/1767/files#diff-3634b4c362041964a17503f2ba9d107297f02376af3788b4648644d880de5887R546-R556 here too? Also FYI there's a `hidden: boolean` as well.

(That link isn't loading initially because User.tsx isn't being rendered by github, but in that file you have a nice explicit list of those that goes something like

```
type Server = "kgs" | "igs" | "dgs" | "golem" | "wbaduk" | "tygem" | "fox" | "yike" | "goquest";

type AccountLinks = {
    [org in `org${1 | 2 | 3}`]?: string;
} & {
    [org_id in `org${1 | 2 | 3}_id`]?: string;
} & {
    [org_rank in `org${1 | 2 | 3}_rank`]?: string;
} & {
    [server_username in `${Server}_username`]?: string;
} & {
    [server_rank in `${Server}_rank`]?: string;
};
```

src/models/user.d.ts
```diff
@@ -105,6 +112,7 @@ declare namespace rest_api {
         tournaments: Array<{
             id: number;
             name: string;
+            icon: string; // URL
```
anoek, line 115: This is good, I also added `icon` to `ladders` up above if you want to add that in. 

Down below we still have `groups: any[]` - this can be array of `                {
                    "id": number;
                    "name": string;
                    "icon": string;
                }
`
too

and while we're at it `titles` is `
                {
                    "icon": string;
                    "title": string;
                }
`

anoek (commented):
Awesome, as always thanks for the cleanup :)

src/models/user.d.ts
benjaminpjones: Added! (and `hidden_ids` too!)

src/models/user.d.ts
benjaminpjones: Added.

benjaminpjones:
Thanks for reviewing!

PR #1768, opened 2022-04-04
Api documentation

shinuito:
Add the current version of the api docs file, so that other can also begin to work on it and fill in the gaps as in https://github.com/online-go/online-go.com/discussions/1761

The current version lists most of the get endpoints that I can think of, but lacks examples and schema for the endpoints. I've tested most of them with the swagger editor that I could, although I couldn't figure out so far how to do authentication to test for example the /me endpoints for now (although I can test them in the rest api directly).

There's a couple of commented out endpoints, one is a `post` one for challenges which again I haven't tested as it related to the authentication above. The other commented ones are images like banners and icons, where the endpoint redirects to a url where the file is hosted, and I'm not sure how or if one can parse that within swagger. If not the commented sections can just be removed entirely.

There's still a few more to add I believe, for example the apng ones for the recent animated gifs of boards, and possibly the ai review ones.


anoek:
Awesome, thanks for spearheading that :)

PR #1770, opened 2022-04-06
Table check engine and Median score loss

shinuito:
The name of the engine at  `/api/v1/games/{id}/ai_reviews` recently changed from `katago` to `katago:fast` which stopped the Review table from populating automatically as it looked for `katago`. So

## Proposed Changes

  - check if the engine contains `katago` instead of `katago` exactly.
  - added in new table rows to compute the median score loss of the players, and a function to compute the median score loss of a list of numbers (private to AIReview for the moment).

### DIFF
--- a/src/views/Game/AIReview.tsx
+++ b/src/views/Game/AIReview.tsx
@@ -77,6 +77,7 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
     ai_review?: JGOFAIReview;
     table_rows: string[][];
     avg_score_loss: number[];
+    median_score_loss: number[];
     moves_pending: number;
     max_entries: number;
 
@@ -103,6 +104,7 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
         const ai_table_out = this.AiSummaryTableRowList();
         this.table_rows = ai_table_out.ai_table_rows;
         this.avg_score_loss = ai_table_out.avg_score_loss;
+        this.median_score_loss = ai_table_out.median_score_loss;
         this.moves_pending = ai_table_out.moves_pending;
         this.max_entries = ai_table_out.max_entries;
         if (!data.get("user").is_moderator) {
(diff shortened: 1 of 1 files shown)


src/views/Game/AIReview.tsx
```diff
@@ -883,6 +904,22 @@ export class AIReview extends React.Component<AIReviewProperties, AIReviewState>
...
+                return a - b;
+            });
+
+            median_score_loss[0] =
+                this.medianList(score_loss_list[0]) !== undefined
+                    ? Math.round(10 * this.medianList(score_loss_list[0])) / 10
+                    : 0;
+            median_score_loss[1] =
+                this.medianList(score_loss_list[1]) !== undefined
+                    ? Math.round(10 * this.medianList(score_loss_list[1])) / 10
```
benjaminpjones, line 920: `round(x*10)/10` just truncates this number to the nearest 10th right?  You might be interested in the `toFixed()` function:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

shinuito: I think at the moment it might be the case that the median returns undefined when the list is empty, so I'd still need some conditional operator in that table row. 
The alternative would be to return zero in that case I suppose. Defaulting to zero might make things easier I suppose.

shinuito: I've updated to use toFixed now for the mean, median and when calculating a percentage of the moves in a particular category.

shinuito:
I need to run prettier again, but I also want to try to figure out why the table data for fast reviews is slightly different on load to if it's manually selected from the dropdown list. I haven't noticed that behaviour with full reviews as of yet, but I want to look into it anyway.

shinuito:
So because I had the table set state when it was populated fully, it turned out for fast reviews, that this was happening before the ai_review worst moves scores ended up in `ai_review.scores` which seems to happen a good bit later.

So when there's 3 worst moves in ai_review.moves I use those scores instead, which seems to happen when there's at least 4 moves in the game. If the game has 3 or less moves I just use what's available so it might still recategorise a move or two if selected again from the menu and forced to update.

If anyone has suggestions to avoid this type of workaround I can do that instead, it's mainly because I have trouble with the asynchrony of state being updated and the ai review data being received and updated.

anoek:
Awesome, thanks @shinuito !

PR #1773, opened 2022-04-08
Load some parts of the profile page early.

benjaminpjones:
Improves the UX for slower-loading pages

## Proposed Changes

  - get user data from **api/v1/players/%%** and **termination-api/player/%%**.
  - display the available data before the slower **api/v1/players/%%/full** fetch returns.
  - Clean up/add some types in user.d.ts

Demo:

https://user-images.githubusercontent.com/25233703/162365605-2238d853-866a-47c0-8da2-d0262ff28417.mov

cc: @Sofia-mal

### DIFF
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -61,7 +61,7 @@ interface UserProperties {
 type RatingsSpeed = "overall" | "blitz" | "live" | "correspondence";
 type RatingsSize = 0 | 9 | 13 | 19;
 
-interface UserState extends Partial<rest_api.PlayerDetails> {
+interface UserState extends Partial<rest_api.FullPlayerDetail> {
     editing: boolean;
     selected_speed: RatingsSpeed;
     selected_size: RatingsSize;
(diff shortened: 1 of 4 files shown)


src/views/User/User.tsx
```diff
@@ -119,8 +119,37 @@ export class User extends React.PureComponent<UserProperties, UserState> {
...
+                    is_tournament_moderator: false,
+                    is_watched: false,
+                    on_vacation: false,
+                    vacation_left: 0,
+                    deviation: responses[0].ratings.overall.deviation,
+                    ranking: responses[1].ranking,
+                    rating: responses[1].rating,
+                    ratings: responses[1].ratings,
+                    first_name: null,
+                    last_name: null,
```
anoek, line 143: If `players/%%/full` completes before `players/%%` and `termination-api/player/%%`, won't these constants override the good values?

benjaminpjones: If `full` returns first, `state.resolved` should be `true` right?

I check that at the beginning of this callback and exit early if that's the case.

anoek: Right you are, I overlooked that, thanks!

anoek:
Awesome work, thanks!

benjaminpjones:
Of course, it's been a fun problem to solve 😀 Thanks for reviewing!

PR #1780, opened 2022-04-17
Remove cache-loader

GreenAsJade:
Fixes cache-loader dependency issues brought with webpack 5.0.

(Removing react-select types snuck in here too: this is deprecated because react-select apparently comes with types now)

## Proposed Changes

Dump cache-loader.  

I _think_ that webpack 5 is doing what we want anyhow, though this definitely should be checked.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
         "@typescript-eslint/parser": "^5.14.0",
         "autoprefixer": "^10.4.2",
         "body-parser": "^1.16.1",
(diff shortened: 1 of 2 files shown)


anoek:
Huh nice, glad it just does that these days, thanks!

I'm gonna add back in `"@types/react-select": "^5.0.1",` though since it doesn't cause troubles on my end, so probably want those types and then we can work out what we need to do to make the build happy for everyone

PR #1781, opened 2022-04-17
Pulled types for react-select and react-virtualized out of package.json

GreenAsJade:
Fixes type errors and warnings

## Proposed Changes

 Remove @types for react-select (because the message says to do that) and react-virtualized (because there are type errors due to react now defining the needful).


anoek:
It looks like `@types/react-select` is indeed no longer needed as `react-select` ships with types these days, but it looks like we still need  `@types/react-virtualized`, so maybe keep that?

GreenAsJade:
Will sort this out afresh with the right version of npm

PR #1782, opened 2022-04-17
OJE update react-charts

GreenAsJade:
Fixes dependency issues with React 18

## Proposed Changes

Make OJE use react-charts 3.x


GreenAsJade:
Pff, not intended for devel

PR #1783, opened 2022-04-17
Update react-charts to 3.0 beta (gets rid of dependency issues)

GreenAsJade:
Fixes dependency issues between react-charts and React 18

## Proposed Changes

Make OJE use react-charts 3.x


GreenAsJade:
I'm not sure what the deal with the build failure is - I didn't get a new package.lock, I think because we're using npm now eh?

GreenAsJade:
So - I generated package-lock.json.   It's rather bigger than I remember, but it seems to have quieted build here?

anoek:
Looks like you lost some dark theme label colors, not sure if you care at this point or not:

image

vs old:

image

FWIW with the latest npm I dont have any dependency issues over here, so the old version compiles just fine - though I'm happy to upgrade to 3.0 beta too, and if you don't care about the color issue I'll just click merge :)

GreenAsJade:
I'm going to redo the PR without regenerating package-lock, with latest npm, and see how we go.

I think I'm the only person who looks at this graph, so losing the label-theme for a while is no biggie - worth it to get on a leading instead of trailing version of the package :)

PR #1787, opened 2022-04-18
Put <React.Strict> inside <Router> so that links work.

GreenAsJade:
Fixes "Links won't navigate (until I press reload)"

## Proposed Changes

 https://github.com/remix-run/react-router/issues/7870#issuecomment-1099884642


GreenAsJade:
Pff - I didn't mean to close it...

anoek:
Nice, thanks for digging into that, lack of working links was quite annoying!

PR #1791, opened 2022-04-20
Factored out NotificationManager

GreenAsJade:
Fixes big file.

## Proposed Changes

Put NotificationManager class into its own file.


GreenAsJade:
(I hope to look into Clossius request in the devel branch, will be good not to let react18 get too different!)

anoek:
Timing wise, I hope to be able to merge react18 by the end of the week FYI. The Game thing is the last major overhaul I wanted to do, there's a few more fixes that need doing here and there to finish getting rid of the UNSAFE's, but I think after that we should be g2g

PR #1792, opened 2022-04-20
React18 notification click algorithm update

GreenAsJade:
Fixes people wanting to get to the most urgent game

## Proposed Changes

        // If we are looking at a board with our move, we want to chose the next lowest-expiration board
        // from the one we chose last time.

        // If we're not looking at a board with move of ours, and we have a move to play, then we want to chose the
        // most urgent move to jump to.

### DIFF
--- a/src/components/Notifications/NotificationManager.tsx
+++ b/src/components/Notifications/NotificationManager.tsx
@@ -172,7 +172,7 @@ export class NotificationManager {
     unread_notification_count;
     boards_to_move_on;
     active_boards;
-    turn_offset;
+    advances;
     auth;
     event_emitter: TypedEventEmitter<Events>;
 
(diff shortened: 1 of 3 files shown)


GreenAsJade:
Pushed this up now to see if you hate any of the tidyups.

GreenAsJade:
(I'm not sure why this build fails - it works locally, I can't see why it doesn't like Goban there: is it a react version thing?)

GreenAsJade:
This seems to basically work in react18 branch.   I haven't gone nuts on corner case testing till it gets a look over.

GreenAsJade:
Requires https://github.com/online-go/ogs-node/pull/14

anoek:
It tends to work better to commit to one branch, then merge that branch into another, as opposed to committing the same changes into two branches and trying to merge them - so that's what I've done here, I reverted the notification manager commit into react18, merged devel into react 18, and so at this point I **think** we're in a good state with this and this PR can be closed? If there were any other changes that happened that were react18 specific though, then those haven't made it in and we'll want to patch for those

GreenAsJade:
Yeah, I typically would have done that, but I was expecting the branches to diverge for react and devel.

PR #1794, opened 2022-04-20
React18 cleanup rating chart

GreenAsJade:
Fixes rating chart issues under Strict

## Proposed Changes
 
 - Move updating the label text into didMount (fixes UNSAFE)
 - Use createRef to make & maintain the `this.container` ref, don't do sideffects on ref-callback function
 -- Allows us to not unassign `this.container` on `deinitialize`, which is important because Strict calls willUnmount even though the component doesn't unmount.


src/components/RatingsChart/RatingsChart.tsx
```diff
@@ -235,6 +235,9 @@ export class RatingsChart extends React.Component<RatingsChartProperties, Rating
     }
 
     setRanges = () => {
+        if (!this.container) {
```
anoek, line 238: Should this be `!this.container.current` ?

anoek:
Thanks for the fix! Seems to work good, I'm not sure if the above check is required these days or when it would trigger, so thought I'd check with you on that before clicking merge.

src/components/RatingsChart/RatingsChart.tsx
GreenAsJade: Ah - I think that probably isn't needed anymore, once I got rid of `this.container = null` from `deinitialize`.

I'll take it out and check things out.

PR #1795, opened 2022-04-20
Notifications refactor, algorithm update

GreenAsJade:
Fixes "want to get to most urgent game"

## Proposed Changes

    // If we are looking at a board with our move, we want to chose the next lowest-expiration board
    // from the one we chose last time.

    // If we're not looking at a board with move of ours, and we have a move to play, then we want to chose the
    // most urgent move to jump to.

... the same as https://github.com/online-go/online-go.com/pull/1792

Requires https://github.com/online-go/ogs-node/pull/14


GreenAsJade:
If #1792 passes muster, then maybe we just wack this into Prod, before finishing React18 ?

src/lib/misc.ts
```diff
@@ -22,8 +22,47 @@ import { browserHistory } from "ogsHistory";
 import * as preferences from "preferences";
 import swal from "sweetalert2";
 
+import { Goban } from "Goban";
```
anoek, line 25: `from "goban"` instead of `from "Goban"` should fix the build issue :)

GreenAsJade: Hah!   Will fix.   But ... why does it work locally?

anoek: I fixed it already, I reckon it works for you because your filesystem isn't case sensitive ☺️

GreenAsJade: Wah???   (runs away, hopes never to hear about this again).

PR #1797, opened 2022-04-23
Give the balance buttons to moderators

GreenAsJade:
Fixes moderators finding it hard to deal with abandonned or otherwise un-admined rengo challenges

## Proposed Changes

Simply make the buttons available - the back-end doesn't mind.   (It turns out the backend already assumed moderators would be allowed to do /team actions)


anoek:
Nice, thanks!

PR #1798, opened 2022-04-24
Moderator rengo kick

GreenAsJade:
Fixes moderators needing a way to eject someone from all rengo challenges

## Proposed Changes

GIve moderators a boot-someone-out-of-all-rengo-challenges button.

(Note: the positioning of the button doesn't make it entirely clear that it will eject the person from _all_ rengo challenges, but I couldn't think of a way to make that clear, and really moderators kind of know what they are doing, plus it's not the end of the world if someone is accidentally kicked from rengo challenges)

Requires https://github.com/online-go/ogs/pull/1734

Screen Shot 2022-04-24 at 5 03 21 pm (2)


anoek:
Nice, thanks!

PR #1802, opened 2022-04-26
Made it so that correspondence games still go to the next least urgent

GreenAsJade:
... after a move is played

Fixes people (rightly) [complaining that the behaviour changed](https://forums.online-go.com/t/changed-behavior-of-black-circle-your-move-button/42937/14?u=greenasjade).

## Proposed Changes

Make it so that if we played a move in a correspondence game and auto-advance, then we advance to the next least urgent.

In live, the new behaviour remains, which is that after playing a move with auto-advance turned on, the most urgent move will be selected (as per forum, there is some debate about this).

### DIFF
--- a/src/components/Notifications/NotificationManager.tsx
+++ b/src/components/Notifications/NotificationManager.tsx
@@ -200,14 +200,21 @@ export class NotificationManager {
         this.connect();
     }
 
-    advanceToNextBoard(ev?) {
+    // This function supports the idea of advancing to the next most relevant board either because the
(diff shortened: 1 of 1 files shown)


GreenAsJade:
... still needs some testing not-just-before-bed-time, but a review at this point would also be good.,

GreenAsJade:
I tested some more: it seems to work 🤞

anoek:
Awesome, thanks for the followup and fix :)

PR #1805, opened 2022-04-29
Make OJE position saving work again, for extended sequences

GreenAsJade:
Fixes "Save" not working properly

## Proposed Changes

Instead of using the updated State from the position creation after "processNewPosition", just use the raw node id received from the server.

Goodness knows how this ever worked - it was highly dependent on when State is actually updated.


anoek:
Ahh race conditions, thanks for the fix!

PR #1808, opened 2022-04-30
Oje turn on tag filter by default

GreenAsJade:
From chat, as background:

 I'm leaning in the direction of having the "Joseki: settled" filter active when you arrive at OJE
 This will have the benefits of:
   - Making people more aware that the whole filtering feature is there
   - Not exposing everyone to every variation that isn't actually Joseki (in theory non-Joseki paths filtered out by this should be a very small subset of all the paths, because OJE is supposed to be "validated Joseki only")

## Proposed Changes

TL:DR: set the Filter to "Joseki" Tag by default, when the Joseki page loads.

Involved:
 - ripping the filter state out of the filter select components, to allow the initial value to be set
       -  fixing various initialisation sequence issues that result
 - converting them to Functional, because why not
 - typing to suit

### DIFF
--- a/src/components/JosekiTagSelector/JosekiTagSelector.tsx
+++ b/src/components/JosekiTagSelector/JosekiTagSelector.tsx
@@ -17,7 +17,7 @@
 
 import * as React from "react";
 
-import Select from "react-select";
+import Select, { MultiValue } from "react-select";
 
 export interface JosekiTag {
     value: string;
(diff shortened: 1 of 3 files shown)


anoek:
Awesome, and thanks for the refactor, seems like a lot of third party libs are starting to expect you're using hooks.. so I reckon in time we'll have to convert a lot of our stuff over, might as well be 100% eventually.

PR #1812, opened 2022-05-02
Turn off the "undo" button if it is a rengo game.

GreenAsJade:
Fixes people complaining that undo doesn't work.

## Proposed Changes

Detect rengo on the goban engine and turn off Undo button.

Things for reviewer attention:

A one line change, but is this a legit way of defending against goban and its ref not being ready?   And is there anywhere else that would care?

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -1338,6 +1338,7 @@ export function Game(): JSX.Element {
             <span className="play-buttons">
                 <span>
                     {((cur_move_number >= 1 &&
+                        !goban?.current?.engine.rengo &&
                         player_not_to_move === data.get("user").id &&
                         !(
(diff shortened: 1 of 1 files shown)


src/views/Game/Game.tsx
```diff
@@ -1342,6 +1342,7 @@ export function Game(): JSX.Element {
             <span className="play-buttons">
                 <span>
                     {((cur_move_number >= 1 &&
+                        !goban?.current?.engine.rengo &&
```
anoek, line 1345: Technically you don't need the first `?.` since `goban` is always initialized, but `.current` can be null so you do need the second `?.`, so we would only need `!goban.current?.engine.rengo`.  To answer your question, yep this is a fine way to guard against the goban ref not being initialized yet.

PR #1813, opened 2022-05-02
Extract PlayerCards into its own file.

benjaminpjones:
## Proposed Changes

  - Remove commented code from Game.tsx (except for the one that has a date on it)
  - Extract the player cards
  - turn on ts strict checks for that file


src/views/Game/Game.tsx
```diff
@@ -65,30 +61,21 @@ import { openACLModal } from "ACLModal";
...
-    move_number?: string;
-}
-*/
-
 const win = $(window);
 
 export function Game(): JSX.Element {
-    //const params = useParams<GameParams>();
-    const params = useParams();
+    const params = useParams<"game_id" | "review_id" | "move_number">();
```
anoek, line 78: Ah-hah, thanks for figuring out how to type that properly

anoek:
Awesome, thanks for the refactor!

PR #1814, opened 2022-05-03
GoResources: Replace common patterns with components.

benjaminpjones:
Since formatting for most entries is the same, the `Book` and `BasicResource` components handle formatting, and allow the caller to focus on content.

## Proposed Changes

  - Add `Book` and `BasicResource` components
  - Search and replace
  - output HTML is virtually identical, but a couple non-rendered spaces removed: https://www.diffchecker.com/8OBPOgk7
  - tsconfig compliance (there was just one strictNullChecks error.)


anoek:
Nice, thanks!

PR #1815, opened 2022-05-03
Game

benjaminpjones:
This moves about 500 lines of code from Game.tsx into its own file.

## Proposed Changes

  - Create GameDock component.
  - Fix TS Strict errors in GameDock code.


anoek:
Thanks!

PR #1818, opened 2022-05-03
Don't default to "un" flag on Game page

### LINKED ISSUE #1816: Flag display errors on game page
**Describe the bug**
Flag for all users reset on game page, and misaligned on mobile

**To Reproduce**
1. Go to game page
2. See flags wrong

**Expected behavior**
Flags displaying correctly.

**Screenshots**

**Additional context**
[Forum Discussion Thread](https://forums.online-go.com/t/country-displaying-unspecified-in-game-for-every-player-right-now/43053?u=bhydden)

benjaminpjones:
Fixes #1816 

See also #1819.

This PR returns to previous behavior. If this is the PR we go with, further investigation is needed into why we aren't seeing country data in non-rengo games.

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -301,7 +301,7 @@ function PlayerCard({
                             </div>
                         )}
                         <div className="player-flag">
-                            <Flag country={player.country ?? "un"} />
+                            <Flag country={player.country} />
                         </div>
                         <ChatPresenceIndicator channel={chat_channel} userId={player.id} />
                     </div>


src/views/Game/PlayerCards.tsx
```diff
@@ -301,7 +301,7 @@ function PlayerCard({
                             </div>
                         )}
                         <div className="player-flag">
-                            <Flag country={player.country ?? "un"} />
+                            <Flag country={player.country} />
```
benjaminpjones, line 304: Looks like if we want to get this working for non-rengo, we can just pull `country` off historical:

```
<Flag country={player.country || historical?.country} />
```

So just one more option..

anoek:
Closing in favor of https://github.com/online-go/online-go.com/commit/f41f357ac58aa2e438fe1aea108de8948174ba6d which does the fetching of the flag if need be and fixes the mobile layout thing. Thanks for working on it though!

benjaminpjones:
Oh yep, that looks good. Thanks @anoek!

PR #1819, opened 2022-05-03
Remove `Flag` from PlayerCards altogether

### LINKED ISSUE #1816: Flag display errors on game page

benjaminpjones:
Fixes #1816 

See also #1818 

Nothing against displaying the flag, but recent behavior was that the Flag was *not* shown in normal games, but shows up in rengo.  I think I prefer removing the flag altogether, due to consistency.

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -19,7 +19,6 @@ import * as React from "react";
 import { Goban, Score, PlayerScore } from "goban";
 import { icon_size_url } from "PlayerIcon";
 import { CountDown } from "./CountDown";
-import { Flag } from "Flag";
 import { ChatPresenceIndicator } from "ChatPresenceIndicator";
 import { Clock } from "Clock";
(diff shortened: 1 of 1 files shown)


pdg137:
I am in favor of removing it altogether to keep the game display simpler and less political, and to not cover part of the avatar.

I believe that the only other place we show flags on top of avatars is in the player card pop-up.

anoek:
If there's a general sentiment towards this on the site I'm open to that, otherwise I've always had the impression that the flags are either a positive or neutral thing for most people? I'm closing this in favor of keeping the flags until we have some community evidence that they should be removed.

PR #1822, opened 2022-05-05
Add board size and ranked filters to game history

rmcat:
Previous discussion about filtering games in the Game History can be found in this [forum thread](https://forums.online-go.com/t/feature-request-filter-game-history/38657/8). I decided to give this feature a go and add it in a way I found useful.

In the Game History table, I added two button sets for a board size filter and a ranked games filter. The top of the table looks like:
[ Player name search ] [ 9x9 | 13x13 | 19x19 ] [ Ranked | Unranked ]

For each filter, the filter is only active if a button is pressed, and only one button may be pressed at a time.

One peculiarity in this change is that annulled games are filtered out when either the Ranked or Unranked button is pressed. I don't know whether this is a good idea or not, but I found it to be more practical than adding another button.

## Proposed Changes

  - Added a board size filter and a ranked filter to the Game History table to the right of the Player filter in the user profile page.
  - Added preferences for the board size filter and ranked filter.
  - Changed the css so that the Player Search filter is aligned to the top when the page is narrower.

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -113,6 +113,9 @@ export const defaults = {
     "moderator.join-games-anonymously": true,
 
     "table-color-default-on": false,
+
+    "game-history-size-filter": "all",
+    "game-history-ranked-filter": "all",
 };
 
 defaults["profanity-filter"][current_language] = true;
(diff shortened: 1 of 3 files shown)


anoek:
Nice, thanks!

> One peculiarity in this change is that annulled games are filtered out when either the Ranked or Unranked button is pressed. I don't know whether this is a good idea or not, but I found it to be more practical than adding another button.

I agree, I like it as chose

PR #1825, opened 2022-05-05
Extract PlayControls and ReviewControls from Game.tsx

benjaminpjones:
This extracts over 1k LOC from Game.tsx.  In addition, I notice there are some commonalities between PlayControls and ReviewControls.  If we can unify the two, I think the interface can become even cleaner.

## Proposed Changes

  - Extract `PlayControls` and `ReviewControls` from Game.tsx
  - Also extract `EstimateScore`, `PlayButtons` and `AnalyzeButtonBar` components.


anoek:
Very cool, thanks!

PR #1826, opened 2022-05-06
Update Team.tsx

Sofia-mal:
Kosh neither confirms nor denies that he is Australian and he prefers to avoid misunderstandings. I wanted to try to send a pull request.

### DIFF
--- a/src/views/docs/Team.tsx
+++ b/src/views/docs/Team.tsx
@@ -67,6 +67,7 @@ export class Team extends React.PureComponent<{}, any> {
         const jp = getLanguageFlag("japan", country, "jp");
         const fi = getLanguageFlag("finnish", country, "fi");
         const it = getLanguageFlag("italian", country, "it");
+        const un = getLanguageFlag("un", country, "un");
         const eng = "_England";
 
         const moderators = [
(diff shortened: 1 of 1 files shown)


Sofia-mal:
Maybe I've done something wrong, it says "Error:   70:15  error  'un' is assigned a value but never used. Allowed unused vars must match /^_[a-zA-Z_]/u  @typescript-eslint/no-unused-vars"
I saw "un" [here](https://github.com/online-go/online-go.com/blob/devel/src/views/docs/GoResources.tsx), line 49. But I don't understand much

koshogs:
Just in case someone else tackles this, my first preference would have been "Oceania" for location on the 'Team' page but that doesn't seem to be an available option because the flag is the same as "un" and therefor no tooltip available. None of this is actually important FTR.

src/views/docs/Team.tsx
```diff
@@ -67,6 +67,7 @@ export class Team extends React.PureComponent<{}, any> {
         const jp = getLanguageFlag("japan", country, "jp");
         const fi = getLanguageFlag("finnish", country, "fi");
         const it = getLanguageFlag("italian", country, "it");
+        const un = getLanguageFlag("un", country, "un");
```
benjaminpjones, line 70: This variable (`un`) isn't used since you just use a string literal (`"un"`, with quotes) below.  You can just delete this line and I think it will work as expected 😃

Sofia-mal: Thank you, I'll try it now :)

BHydden:
I thought I double checked you wanted Australian before proposing the original implementation. My sincerest apologies if I failed to do so.

anoek:
Thanks!

PR #1827, opened 2022-05-06
Create usePreference hook

benjaminpjones:
I noticed a common pattern is to create a state variable that is always in sync with preference variables.

```
    const [value, setValue] = React.useState(
        preferences.get("some-key")
    );

   ... later ...

    setValue(new_val);
    preferences.set("some-key", new_val);
```

IMO this is hard to maintain because one always has to remember to set the preference as well as the state variable, and it's a bit verbose.

This PR introduces a custom react hook to wrap the two setters up in one:

```
    const [value, setValue] = usePreference("some-key");

    // sets both preference and state variable
    setValue(new_val);
```

## Proposed Changes

  - Create the `usePreference` hook.
  - Add stricter typing to preferences, and fix type errors.
      - The Puzzle type was part of an investigation into how to approach type fixes.
  - Use `usePreference` in Settings.tsx

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -17,6 +17,7 @@
 
 import * as data from "data";
 import { GobanSelectedThemes, GoThemes } from "goban";
+import React from "react";
 import { current_language } from "translate";
 import { DataSchema } from "./data_schema";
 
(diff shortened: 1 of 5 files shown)


benjaminpjones:
Ha, oops I didn't realize I had edited some Goban files in node_modules.

```
interface PuzzleConfig {
        ...
    puzzle_rank?: number | string;
    puzzle_description?: string;
    puzzle_collection?: number | string;
        ...
}
```

The problem is that PuzzleConfig says that `puzzle_rank` and `puzzle_collection` are `number`s (makes enough sense) but the API passes `string`s.

I think I will just edit the `rest_api` interface to pretend it's a number, even though that's not true.

src/lib/preferences.ts
```diff
@@ -201,3 +215,29 @@ export function watchSelectedThemes(cb) {
...
+
+    const setStateAndPreference = (v: PreferencesType[KeyT]) => {
+        stateSetter(v);
+        set(key, v);
+    };
+
+    // TODO: maybe add a watcher here for if a different component updates the value?
+    // Something like...
+    //
+    // React.useEffect(() => {
```
anoek, line 238: I think this would be great. I had just implemented a `useUser` earlier today which does the watch thing and it's great. I think I'll probably ditch that and do a `useData` based off of this though as a more generic way of doing things.

anoek:
Awesome, definitely the direction we should be moving. Thanks!

src/lib/preferences.ts
benjaminpjones: Awesome, `useUser`/`useData` sounds really useful too.  FWIW I think a dedicated `useUser` function could be nice even if it just wraps the data function, since it's used everywhere and sort of special as data goes.

One notable difference is that `data.get("user")` is always defined(?) whereas most other data entries can return `undefined`.  If we eventually turn `strictNullChecks` on, it will be much more ergonomic to have a user function that guarantees a non-null return value.

PR #1829, opened 2022-05-07
Don't reset page title unnecessarily.

mmartin:
Page title is reset on every history event (as a fallback mechanism for pages which do not have a custom title, I guess?).
This breaks in such cases as:

1. click on “Home” in the navbar;
2. homepage opens and the title shows number of games where it's my move;
3. click “Home” again;
4. title is reset, i.e. it becomes just “OGS”.

## Proposed Changes
  - Reset page title if history event's action is not “Replace”.


anoek:
Huh, indeed, looks like there's a number of things that need improving in this regard. Thanks for the patch!

Issue #1374, opened 2021-01-29
Top menu items get cropped on narrower screens


kevincoleman:
**Describe the bug**
The top menu items get hidden without much explanation as the window narrows. Especially for users who don’t know anything has been cropped this could be confusing.

**To Reproduce**
1. Go to online-go.com
2. Resize window
3. Observe items disappearing

**Expected behavior**
Menu items should either be space-efficient enough to persist down to the smallest expected window sizes, or they should be decried as hidden with some form of “••• More” menu.

**Screenshots**
https://user-images.githubusercontent.com/445707/106336210-467f2280-6243-11eb-990a-76046b81e094.mov

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Any
 - Version: Any

**Smartphone (please complete the following information):**
 - Device: Any modern one, also tablets

**Additional context**
This was discussed as a side item on [the forum post](https://forums.online-go.com/t/feedback-requested-on-proposed-nav-bar-change/33600/20) about making the top menu slightly bigger and styled differently.

# Proposed Fixes (action items)
- [ ] Consolidate current top menu items into left menu. Proposed menu structure:
```
Home
Play (takes you to My Games / Home on 2nd click)
  My Games
  All Games
  Tournaments
  Ladders
Study (takes you to Learn to Play (?) on 2nd click)
  Learn to play Go
  Demo Board
  Puzzles
  Joseki
  SGF Library
  Other Go Resources
Community (takes you to Leaderboards (?) on 2nd click)
  Leaderboards
  Chat
  Groups
  Forums
  (Moderator Center)
  (Announcement Center)
About (takes you to About page on 2nd click)
  Documentation & FAQ
  Support OGS
Profile (takes you to your user profile on 2nd click)
  Settings
  (Admin)
  Logout
```

Optional followup items:
- [ ] Move search (to top right?)
- [ ] Add board theming options to Settings > Game Preferences
- [ ] Add board theming options to appropriate contextual location (like on a single game view or 'My Games')
- [ ] Revisit right-bar content to make it collapsible

kevincoleman:
Some top suggestions that have come to mind:
1. From forum user MooToYouToo: Gather links into categorized drop-downs to save space. Example:

    Play
        Create a Game
        Tournaments
        Ladders

    Study
        Learn to play Go
        Joseki
        Puzzles
        Other Go Resources

    Community
        Forums
        Chat
        Leaderboards
        Groups

[...]

PR #1830, opened 2022-05-07
Toggleable UI experiments, and an experimental new nav bar

anoek:
This PR provides a new `<Experiment>` tag that picks a variant of a component to render based off of a user preference, presumably set somewhere in settings. Currently there is an "Enable experimental interface changes" toggle now under General Preferences that will toggle experiments flagged under the `v6` name.

Also included in this PR is an attempt at overhauling our nav system, which relates to #1374 . To view the new update, flip the toggle under settings.

### DIFF
--- a/src/components/Announcements/Announcements.styl
+++ b/src/components/Announcements/Announcements.styl
@@ -17,10 +17,7 @@
 
 .Announcements {
     position: absolute;
-    top: v5-navbar-height;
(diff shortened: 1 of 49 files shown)


BHydden:
Silly me tried to find the toggle in settings and this PR isn't even merged yet  🤣 I think a lot of people will love that this is getting worked on

PR #1831, opened 2022-05-07
Remove portrait_tab state variable.

benjaminpjones:
`portrait_tab` was always set to `"game"`, so the `portrait_tab === "chat"` and
`portrait_tab === "dock"` code paths were unreachable.

## Proposed Changes

  - Remove `[portrait_tab, set_portrait_tab]` state variables and all related conditional logic.
  - Remove `togglePortraitTab()` callback


anoek:
Thanks for the cleanup! I noticed it too during the FC refactor, but I was scared to touch it at that time.

benjaminpjones:
Yeah I think being able to say "no" to random fixes during a big refactor is a good thing.  I try my best, but still get bitten sometimes!

PR #1832, opened 2022-05-07
Joseki: replace fetch with requests functions

benjaminpjones:
Now that OJE is part of the Django API, it doesn't seem to need custom headers. Using the requests.ts functions seems to simplify the code quite a bit because we are no longer manually passing headers and pulling json from the response.

## Proposed Changes

  - JSDoc for requests.ts
      - I considered adding tests too, but mocking `$.ajax` was tricky.  At some point I think we will want to switch to the `fetch` API there instead any way.
  - Replace all `fetch` calls with the equivalent requests.ts function.
  - Fix the `no-empty-pattern` error in `Joseki.tsx`

### DIFF
--- a/src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
+++ b/src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
@@ -103,7 +103,6 @@ export function JosekiVariationFilter(props: JosekiVariationFilterProps) {
         //const tags = (e === null || e.length === 0) ? null : e.map(t => typeof(t) === 'number' ? t : t.value);
         const new_filter = { ...props.current_filter, tags };
 
-        // console.log("new tag filter", new_filter);
         props.set_variation_filter(new_filter); // tell parent the fiter changed, so the view needs to change
     };
 
--- a/src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
+++ b/src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
@@ -21,11 +21,11 @@ import { _ } from "translate";
 import * as player_cache from "player_cache";
 import { JosekiTagSelector, JosekiTag } from "../JosekiTagSelector";
 import { PlayerCacheEntry } from "player_cache";
+import { get } from "requests";
 
 export type JosekiFilter = { contributor: number; tags: JosekiTag[]; source: number };
 
 interface JosekiVariationFilterProps {
-    oje_headers: HeadersInit;
     contributor_list_url: string;
     source_list_url: string;
     set_variation_filter: any;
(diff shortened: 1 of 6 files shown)


anoek:
Awesome, I'll wait for @GreenAsJade before merging but lgtm I think.

Re switching to fetch, yes I think that's what we want to do - I was reluctant to do so for awhile as there wasn't a mechanism to abort fetch's in flight like there was for xhr's, but it looks like since I last looked `AbortController` has been added to address that, so I reckon we could do that whenever we wanted now.

benjaminpjones:
Awesome, thanks for reviewing!

src/components/JosekiVariationFilter/JosekiVariationFilter.tsx
```diff
@@ -79,16 +75,9 @@ export function JosekiVariationFilter(props: JosekiVariationFilterProps) {
...
-            headers: props.oje_headers,
-        })
-            .then((res) => res.json())
+        get(props.source_list_url)
             .then((body) => {
-                // This can possibly be changed to !== undefined or != null,
-                // But I don't want to touch this right now -BPJ
-                // eslint-disable-next-line eqeqeq
-                if (body.sources != undefined) {
+                if (body.sources != null) {
```
GreenAsJade, line 80: The comment in here was telling us that the sentry logs shows that sometimes `body.sources` was uyndefined ... it would have been some message like `trying to read thingy, undefined`.

I have no idea why I used `!=` and not `!==`  ... but I guess since null == undefined, this is all good.

src/views/Joseki/Joseki.tsx
```diff
@@ -43,7 +43,7 @@ import { JosekiTagSelector, JosekiTag } from "JosekiTagSelector";
 import { Throbber } from "Throbber";
 import { IdType } from "src/lib/types";
 
-const server_url = data.get("oje-url", "/oje/");
+const server_url = "/oje/";
```
GreenAsJade, line 46: Why are we doing this?

The reason for being able to change the url is in case you want to bring up a different joseki server... it certainly turned out  handy when migrating....

GreenAsJade (commented):
I think I'd be inclined to leave the server able to be set via data, unless this considered a Bad Thing now.   When debugging joseki, it was handy to uncomment all those logs without having to think about re-instrumenting.   I guess we might consider OJE as "working now" and thus hope to not need that anymore!

src/views/Joseki/Joseki.tsx
benjaminpjones: Ah I didn't see a corresponding `data.set` call so I figured it was just obsolete. 

How do you use this generally? Mainly I'm wondering what is the benefit to checking localstorage rather than just modifying this line of code next time you need to talk to a different server..

benjaminpjones: At any rate, added that back in!

PR #1833, opened 2022-05-08
Persist player OJE-variation-filter settings

GreenAsJade:
Fixes irritatingly having to  set the filter each time you arrive at OJE.

## Proposed Changes

  Store the filter in `data`


anoek:
Nice, thanks!

PR #1836, opened 2022-05-08
Fix resize, undo bugs

### LINKED ISSUE #1834: "Undo" button display broken.
**Describe the bug**
"Undo" button is displayed for games where I'm not even playing.

**To Reproduce**
1. Go to someone else's ongoing game;
2. See "Undo" button.

This was broken in #1825, I think. @benjaminpjones 

https://github.com/online-go/online-go.com/blob/e817825048eb108c266e492bccaad2d551ea80ae/src/views/Game/PlayControls.tsx#L763

### LINKED ISSUE #1835: Resizing makes everything a tie again
**Describe the bug**
Related to https://forums.online-go.com/t/everything-is-a-tie/43102/5

This was semi-fixed by #1828, but resizing causes it to happen again because the PlayControls are destroyed and state is lost.

**To Reproduce**
1. Go to any finished game
2. resize until it becomes "portrait" mode
3. See that it will say "Tie by ..." instead of the correct text

**Expected behavior**
Show the correct result.

benjaminpjones:
Fixes #1834 
Fixes #1835
Fixes [Missing grant undo button](https://forums.online-go.com/t/cant-find-grant-undo-button/43124/7)

## Proposed Changes

  - Don't show `PlayButtons` if user is not a participant.
  - Set initial state from Goban in `useState` calls.

### DIFF
--- a/src/views/Game/PlayControls.tsx
+++ b/src/views/Game/PlayControls.tsx
@@ -219,7 +219,6 @@ export function PlayControls({
 
     const [winner, set_winner] = React.useState(goban.engine.winner);
     React.useEffect(() => {
-        console.log("new goban");
         goban.on("load", () => set_winner(goban.engine.winner));
         goban.on("winner", set_winner);
     }, [goban]);
--- a/src/views/Game/PlayControls.tsx
+++ b/src/views/Game/PlayControls.tsx
@@ -325,7 +325,7 @@ export function PlayControls({
     return (
         <div className="play-controls">
             <div className="game-action-buttons">
-                {((mode === "play" && phase === "play") || null) && (
+                {mode === "play" && phase === "play" && user_is_player && (
                     <PlayButtons
                         resign_text={resign_text}
                         show_undo_requested={show_undo_requested}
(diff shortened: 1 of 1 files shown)


anoek:
I'm going to merge this because it's better than it was, however I just noticed that file is doing 

https://github.com/online-go/online-go.com/blob/devel/src/views/Game/PlayControls.tsx#L121-L123

which as I understand it might not be legal as it's an early return before other `useState`/`useRef` stuff. I'm wondering if it's possibly allowed if it's null and before everything else like you have it, but I don't know for sure, if it's not allowed we'll need to move that down to after the other hook usages and null-guard the various initial values you just added.

anoek:
It looks like we guard against it ever being null from `Game` https://github.com/online-go/online-go.com/blob/devel/src/views/Game/Game.tsx#L1567-L1569 which I reckon why it hasn't exploded.

benjaminpjones:
Yeah you're right about removing the early return, I think the hooks do some kind of counting behind the scenes to make the magic work - I should have removed that when I broke out the component!

PR #1837, opened 2022-05-08
Move side effects from translate.ts to main.tsx.

benjaminpjones:
For some reason, the tests don't play nicely with Goban, especially when functions are called in a module.  One of the hacks I had in every test suite was this:

```
// workaround for setGobanTranslations not found error
// This can probably be fixed by removing sideeffects from translate.ts
jest.mock("goban", () => ({
    setGobanTranslations: jest.fn(),
}));
```

By removing the side effects from translate.ts (and putting them in main.tsx), I was able to remove these mocks from the tests.

At some point, I need to figure out how to actually deal with goban code in tests, but this makes writing tests just a tiny bit easier.

## Proposed Changes

  - Call `setCurrentLanguage` in main.tsx, not translate.tsx.
  - Remove goban mock from tests that didn't have anything to do with Goban.

### DIFF
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -34,12 +34,6 @@ import {
     effective_outcome,
 } from "./rank_utils";
 
-// workaround for setGobanTranslations not found error
-// This can probably be fixed by removing sideeffects from translate.ts
-jest.mock("goban", () => ({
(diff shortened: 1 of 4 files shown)


anoek:
That's a better place for it anyways, thanks!

PR #1838, opened 2022-05-09
Add react-hooks/rules-of-hooks lint check.

benjaminpjones:
reacthooksmeme

## Proposed Changes

  - Add the `react-hooks` eslint plugin, and turn on the `rules-of-hooks` rule.
  - Fix the resulting errors.

## Additional Context

Discussion on #1836

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -125,7 +125,7 @@ module.exports = {
                 },
             },
         ],
-        "react-hooks/rules-of-hooks": "warn",
+        "react-hooks/rules-of-hooks": "error",
     },
     overrides: [
         {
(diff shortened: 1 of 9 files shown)


benjaminpjones:
So I've tested the pages basically by going to each page and checking that they load.  Unfortunately I get the billing configuration error on the Supporter Page:

Screen Shot 2022-05-09 at 12 44 01 PM

 but that appears to be a problem (in a dev environment) even when I check out `devel` again.  Any thoughts on how to approach this?

anoek:
Now we're talking, that's awesome, thanks for getting that wired up and fixing my bugs :grin:

anoek:
@benjaminpjones Re the supporter page not working, I fixed it up - i hadn't added added the proxy setting for the new billing stuff, so should be fixed now - also though it looks like it works great :)

benjaminpjones:
Awesome, thanks for looking into that!

PR #1839, opened 2022-05-10
Oje tag i18n

GreenAsJade:
Fixes tags not being able to be translated

## Proposed Changes

Ask the translators to translate using `try { _("tag")}` trick from @benjaminpjones , use these translations.

(This PR coming to you from the Queensland Hinterland, after a day of dodging leeches snakes and poisonous plants :D )

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -1135,7 +1135,7 @@ class _Joseki extends React.Component<JosekiProps, JosekiState> {
                     .filter((t) => t.count > 0)
                     .map((t, idx) => (
                         <div className="variation-count-item" key={idx}>
-                            <span>{t.tagname}:</span>
+                            <span>{pgettext("This is a Joseki Tag", t.tagname)}:</span>
                             <span>{t.count}</span>
                         </div>
                     ))}
@@ -1150,7 +1150,7 @@ class _Joseki extends React.Component<JosekiProps, JosekiState> {
                   .sort((a, b) => Math.sign(a.group - b.group))
                   .map((tag, idx) => (
                       <div className="position-tag" key={idx}>
-                          <span>{tag["description"]}</span>
+                          <span>{pgettext("This is a Joseki Tag", tag["description"])}</span>
                       </div>
                   ));
 
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -24,8 +24,8 @@ import ReactResizeDetector from "react-resize-detector";
 import * as queryString from "query-string";
 
 import * as data from "data";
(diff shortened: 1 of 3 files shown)


src/views/Joseki/Joseki.tsx
```diff
@@ -259,6 +259,47 @@ class _Joseki extends React.Component<JosekiProps, JosekiState> {
...
+        // arange translation of known joseki tags
+        // (these values come from the Joseki Tags table on the server)
+        try {
+            pgettext("This is a Joseki Tag", "Joseki: Done");
+            pgettext("This is a Joseki Tag", "Fuseki: Done");
+            pgettext("This is a Joseki Tag", "Current");
+            pgettext("This is a Joseki Tag", "Basic");
+            pgettext("This is a Joseki Tag", "Position is settled");
+            pgettext("This is a Joseki Tag", "Fighting pattern");
+            pgettext("This is a Joseki Tag", "Black gets the corner");
```
benjaminpjones, line 272: I suppose reducing translator workload might be a reason to prefer the static solution:

```
function getTagText(tag_id: number) {
    switch (tag_id) {
        ...
    case 0x06:
        return interpolate(pgettext("Joseki Tag", "{{color}} gets {{result}}"), { color: _("Black"), result: pgettext("Joseki Tag Result", "the corner"));
    case 0x07:
        return interpolate(pgettext("Joseki Tag", "{{color}} gets {{result}}"), { color: _("Black"), result: pgettext("Joseki Tag Result", "the corner and top"));
        ...
}
```

That only cuts the translations roughly in half, but given that OGS supports 64 languages and there are 34 Joseki tags, I thought it was worth mentioning.

benjaminpjones: On the other hand, interpolation adds complexity for translators too, so perhaps the costs outweight the benefits...

anoek: Single words like color can work sometimes, more complicated multi word pastes like that though usually don't translate so well due to grammatical differences

GreenAsJade: Where does this leave us?

anoek: We're good, was just on my phone when typing that so wasn't the time to merge - doing so now, thanks for the improvement!

GreenAsJade: Woo hoo and phew.

Now I just have to do the "Categories" and OJE should be just about fully done (Also Empty Board and Tenuki button).

GreenAsJade: No drama at alll - it's great to have review & options :)     Alex's app does support OJE, so that was one reason (in addition to inertia) why I stuck with what we have :D

PR #1840, opened 2022-05-10
Demo: PlayControls tests

benjaminpjones:
This is a demo for the `goban` change https://github.com/online-go/goban/pull/69

## Proposed Changes

  - Import [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/)
  - Write a couple tests for PlayControls
      - Verify "Resign" is present if user is a participant.
      - Verify "Resign" is not present if user is a participant.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
         "defaults"
     ],
     "devDependencies": {
+        "@testing-library/react": "^13.2.0",
         "@types/jest": "^26.0.15",
         "@typescript-eslint/eslint-plugin": "^5.14.0",
         "@typescript-eslint/eslint-plugin-tslint": "^5.14.0",
--- a/setup-jest.js
+++ b/setup-jest.js
@@ -1,2 +1,8 @@
 global.window = window
 global.$ = require('jquery');
+
+// Goban stuff
+HTMLCanvasElement.prototype.getContext = () => {
+    return {};
+};
+global.CLIENT = true;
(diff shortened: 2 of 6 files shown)


anoek:
lgtm, though I'm seeing a warning when I run jest:

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.
```

and running with `--detectOpenHandles` hangs - any ideas?

benjaminpjones:
I remember seeing that warning, but I'm not sure why it's popping up.  I do not see it on my machine, and running with `--detectOpenHandles` passes with no warning messages.

The last time this popped up was in https://github.com/online-go/online-go.com/pull/1661 where @GreenAsJade was changing the way the AI server url gets set.  You don't happen to have modifications in `sockets.ts` right now do you?

At any rate, for this test all network calls should be mocked.  I don't pass a server_socket into `goban` and no XHR requests in PlayControls either so I'm a little stumped how a similar error could be introduced in this PR.

anoek:
Hmm interesting, no I don't have any modifications in sockets.ts or anywhere else :confused:

benjaminpjones:
Darn, was hoping it'd be something simple.  A couple things:

- I re-triggered CI, and while it does take a little while, it didn't fail.
- I looked into whether I need to be calling [`cleanup`](https://testing-library.com/docs/react-testing-library/api/#cleanup) from the React Testing Library, but documentation is pretty clear that it will do this automatically after each test in jest.

I'm not sure if there's anything else I can dig into for now since I can't reproduce..

anoek:
I'm good with merging, if you're ready?

PR #1842, opened 2022-05-12
Add more tests for the play buttons.

benjaminpjones:
This gives us a bit of coverage for the bugs addressed in #1836

In a previous PR, I only tested existence/absence of the "Resign" button.  These additions test the Undo and Pass buttons as well.


anoek:
Neat, I'm going to have to come up to speed on how to work with this, very cool we're moving in the direction of having testable components

PR #1843, opened 2022-05-13
Translated `MoveCategory`s, and a bunch of missed/bugged other translations

GreenAsJade:
Fixes patchy translation of OJE

## Proposed Changes

Force-translate `MoveCategory`s, and apply translations to everything that needs to be translated.

### DIFF
--- a/src/components/JosekiAdmin/JosekiAdmin.tsx
+++ b/src/components/JosekiAdmin/JosekiAdmin.tsx
@@ -16,7 +16,7 @@
  */
 
 import * as React from "react";
-import { _ } from "translate";
+import { _, interpolate, pgettext } from "translate";
 import { get, post, put } from "requests";
 
 import ReactTable from "react-table";
@@ -408,10 +408,15 @@ export class JosekiAdmin extends React.PureComponent<JosekiAdminProps, JosekiAdm
                 />
                 <div className="explorer-stats">
                     <span>
-                        {_("Page visits")}: {this.state.page_visits || "..."}
+                        {interpolate(_("Page visits: {{count}}"), {
+                            count: this.state.page_visits || "...",
+                        })}
                     </span>
                     <button className="btn s" onClick={this.showVisitStats}>
-                        {_("details")}
+                        {pgettext(
+                            "A button that shows details of joseki visit statistics",
+                            "details",
+                        )}
                     </button>
                 </div>
 
--- a/src/components/JosekiAdmin/JosekiAdmin.tsx
+++ b/src/components/JosekiAdmin/JosekiAdmin.tsx
@@ -407,9 +407,11 @@ export class JosekiAdmin extends React.PureComponent<JosekiAdminProps, JosekiAdm
                     ]}
                 />
                 <div className="explorer-stats">
-                    <span>Page visits: {this.state.page_visits || "..."}</span>
+                    <span>
+                        {_("Page visits")}: {this.state.page_visits || "..."}
+                    </span>
                     <button className="btn s" onClick={this.showVisitStats}>
-                        details
+                        {_("details")}
                     </button>
                 </div>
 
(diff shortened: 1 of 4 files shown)


src/components/JosekiAdmin/JosekiAdmin.tsx
```diff
@@ -407,9 +407,11 @@ export class JosekiAdmin extends React.PureComponent<JosekiAdminProps, JosekiAdm
                     ]}
                 />
                 <div className="explorer-stats">
-                    <span>Page visits: {this.state.page_visits || "..."}</span>
```
anoek, line 410: You probably want to use interpolate here, 

`interpolate(_("Page visits: {{count}}"), {"count": this.state.page_visits || "..."})` 

that way translators can move that count around for grammatical correctness. Sometimes they come back and want us to use things like `ngettext` singular/plural changes too, but hopefully in this case it doesn't matter/make sense?

src/components/JosekiAdmin/JosekiAdmin.tsx
```diff
@@ -407,9 +407,11 @@ export class JosekiAdmin extends React.PureComponent<JosekiAdminProps, JosekiAdm
...
+                    <span>
+                        {_("Page visits")}: {this.state.page_visits || "..."}
+                    </span>
                     <button className="btn s" onClick={this.showVisitStats}>
-                        details
```
anoek, line 412: We should use `pgettext` to provide some context,

`pgettext("Details about visit statistics", "details")` 

kind of thing

GreenAsJade: Yeah I looked at that one and was lazy, mea culpa.

src/components/JosekiAdmin/JosekiAdmin.tsx
GreenAsJade: Will look into it.

src/components/JosekiAdmin/JosekiAdmin.tsx
anoek: Yeah, I've regretted being lazy about that in the past, the only reason why I'm being more pedantic about it now :) translators used to (and probably still do) have a hard time translating some things I was lazy about because they lacked context.. now I try to provide context for anything that's kind of generic like this.

src/views/Joseki/Joseki.tsx
```diff
@@ -297,6 +299,11 @@ class _Joseki extends React.Component<JosekiProps, JosekiState> {
             pgettext("This is a Joseki Tag", "White gets the centre and right");
             pgettext("This is a Joseki Tag", "Black gets sente");
             pgettext("This is a Joseki Tag", "White gets sente");
+
+            // and move categories
+            Object.keys(MoveCategory).forEach((category) => {
+                pgettext("Joseki move category", category);
```
benjaminpjones, line 305: I'm fairly certain you need to specify all the string literals here (or elsewhere) in order for the scripts to know what phrases to hand off to the translators.

src/views/Joseki/Joseki.tsx
```diff
@@ -103,16 +103,16 @@ const applyJosekiMarkdown = (markdown: string): string => {
...
+    // conceivably, should fetch these from the backend - the string value is used in comparisons :(
     IDEAL = "Ideal",
     GOOD = "Good",
     MISTAKE = "Mistake",
     TRICK = "Trick",
     QUESTION = "Question",
 }
 
-const bad_moves = ["MISTAKE", "QUESTION"] as any; // moves the player is not allowed to play in Play mode
+const bad_moves = ["MISTAKE", "QUESTION"]; // moves the player is not allowed to play in Play mode
```
benjaminpjones, line 115: ❤️

benjaminpjones (changes requested):

anoek:
Thanks!

PR #1844, opened 2022-05-13
Make target language "debug" a special case for autotranslate

GreenAsJade:
that does not got out to get translated, but behaves as if it did.

Fixes hitting external translation service with retries for "debug" language

## Proposed Changes

Interept at the point we go to request the translation and supply a "debug" translation instead, if debug is the target.


anoek:
Thanks!

PR #1846, opened 2022-05-14
PlayButtons: Shorten props list and derive state from goban

benjaminpjones:
Following along with the strategy described in [Goban Centric Components](https://docs.google.com/document/d/1viy_xeQbEZ4_N6SOQll2R4vMRDjxhDz5-wMdG7GtnHI/preview), I have replaced most props for `PlayButtons` (and its subcomponent `CancelButton`) with custom hooks:

Before:
```
<PlayButtons
    resign_text={resign_text}
    show_undo_requested={show_undo_requested}
    cur_move_number={cur_move_number}
    player_to_move={player_to_move}
    onCancel={onCancel}
    goban={goban}
    show_cancel={show_cancel}
    view_mode={view_mode}
/>
```
After:
```
<PlayButtons goban={goban} show_cancel={show_cancel} />
```

## Proposed Changes

  - Add GameHooks.ts (a collection of hooks that subscribe to goban events).
  - Reduce PlayButtons and CancelButton to 1.5 props each.
  - Write tests.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -58,7 +58,6 @@ import { goban_view_mode, goban_view_squashed, ViewMode, shared_ip_with_player_m
 import { game_control } from "./game_control";
 import { PlayerCards } from "./PlayerCards";
 import {
-    CancelButton,
     EstimateScore,
     PlayControls,
     AnalyzeButtonBar,
@@ -67,6 +66,7 @@ import {
     deleteBranch,
     ReviewControls,
 } from "./PlayControls";
+import { CancelButton } from "./PlayButtons";
 import { GameDock } from "./GameDock";
 import swal from "sweetalert2";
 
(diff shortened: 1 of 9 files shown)


anoek:
Awesome, I think this is going to help with that page a lot.

I do see an issue currently where if you have one player request an undo, the other player will see "Accept Undo" - but if they go back a few moves, every time it's "their turn" again in the history they will see that "Accept Undo" button again, it should be hidden if they are viewing historical moves.  This happens when analysis mode is disabled, I haven't checked the case when it is enabled yet.

Similarly the 'undo' button itself is shown every other move when exploring historical moves, though that is true on devel too - so pre-existing bug but it might be a related quick fix.

benjaminpjones:
Ooh that's a sneaky one - I actually didn't realize you could look at historical moves with analysis disabled.  I'll try and get a fix in there for both!

benjaminpjones:
Okay I'm actually not sure what change caused that reversion, but the issues you mentioned are both fixed now!

anoek:
Awesome, thanks!

PR #1848, opened 2022-05-14
Made is so that the auto-translated text appears as the primary text

GreenAsJade:
… with the original as the "FYI" afterwards.

Fixes?   Well, it's a proposal intended to fix the feeling of the translation being an afterthought, and move in the direction of the site feeling more "natural" for a non-English visitor.

## Proposed Changes

If there's a translation, then display it in the primary formatting, with the original text as an FYI.

This commit also changes the way that the "Debug" language is displayed in translated form, so that it doesn't break Markdown.   If this isn't merged, I'll resubmit that bit :D


anoek:
Yeah that's better, thanks!

PR #1849, opened 2022-05-14
Update GameHistoryTable.tsx

Sofia-mal:
https://forums.online-go.com/t/number-of-games-in-list/43186

Some people said they wanted 50 back. I wonder if this is enough and if there were important reasons behind the decision to change it

### DIFF
--- a/src/views/User/GameHistoryTable.tsx
+++ b/src/views/User/GameHistoryTable.tsx
@@ -246,7 +246,7 @@ export function GameHistoryTable(props: GameHistoryProps) {
                         }}
                         orderBy={["-ended"]}
                         groom={game_history_groomer}
-                        pageSizeOptions={[10, 15, 25]}
+                        pageSizeOptions={[10, 25, 50]}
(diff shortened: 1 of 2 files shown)


koshogs:
50 is a good amount of past games to examine for moderating purposes. I really don't see the point in having 15 as an option when there's already 10 and 25. Not much functionality in that.

benjaminpjones:
Whatever ends up happening, probably should do it to the Reviews and Demos table as well?

(Going off this commit: https://github.com/online-go/online-go.com/commit/67c4343e689a00a3c0bdb55af5b241064a815b5b)

anoek:
Lets just add `50` to the list so we have 4 options, the reason why 15 was added was I was contacted by a user requesting 15 as he indicated he did the math and that's the number of games that dominates computing someones rank. I'm not completely sure how true that is or not, but it's plausible given how Glicko 2 works and how fast it adapts, anyways they wanted to be able to look at that specific number of games, and that seemed kinda reasonable.

I don't have a good reason for removing 50, we should add it back in both places.

anoek:
Thanks @Sofia-mal !

PR #1856, opened 2022-05-16
Initialize our state if we start off with an existing goban that's already been loaded

### LINKED ISSUE #1855: Resizing a review results in a "Review by [NULL USER]"
Likely introduced by #1825

**To Reproduce**
1. Go to any review
2. Resize the screen until the layout changes (portrait vs wide)
3. Observe "Review by [NULL USER]

**Expected behavior**
Resizing should not affect the "review by" user.

**Screenshots**

Screen Shot 2022-05-15 at 10 03 18 PM

anoek:
Without initializaing we were counting on the `load` event firing to set
these correctly, which doesn't happen if the goban is already loaded.
(Though maybe it should?)

Fixes #1855

Ping @benjaminpjones to see if you have any better ideas on how to
reduce code duplication.

### DIFF
--- a/src/views/Game/PlayControls.tsx
+++ b/src/views/Game/PlayControls.tsx
@@ -1070,9 +1070,15 @@ export function ReviewControls({
 }: ReviewControlsProps) {
     const user = data.get("user");
 
-    const [review_owner_id, set_review_owner_id] = React.useState<number>();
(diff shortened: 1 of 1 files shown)


anoek:
That does seem slick, I'll close this and make way for your approach

PR #1857, opened 2022-05-16
Abort game load requests if we bail from the Game or Review page

### LINKED ISSUE #1854: Player info from the wrong game displayed
**Describe the bug**
This happens rarely, but seems to happen when sequencing through games quickly.  I believe there is a race condition when there are two outgoing requests to the `games/` endpoint.  It appears the Game Info modal may also show the wrong info when this occurs.

**To Reproduce**
It's very difficult to reproduce this on-demand, but @Sofia-mal was able to reproduce on beta by clicking through the TurnIndicator repeatedly:

> If I repeatedly click on the 0 of the correspondence moves, sometimes in the game against benjito it shows me Venezia instead of benjito. The opposite has never happened.
Other times, I see Venezia for a very short time (after that the game has changed) and then it is changed to benjito.

**Expected behavior**
The player cards should always show info for the correct game.

**Additional context**
[Forums: Some bugs report](https://forums.online-go.com/t/some-bugs-report/40254)

anoek:
Should fix #1854

As per @benjaminpjones suggestion, which I agree with, I think this should fix the loading info from the wrong game issue we were having when flipping through games fast

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -23,7 +23,7 @@ import { useParams } from "react-router-dom";
 import { browserHistory } from "ogsHistory";
 import { _, interpolate, current_language } from "translate";
 import { popover } from "popover";
-import { post, get } from "requests";
+import { post, get, abort_requests_in_flight } from "requests";
 import { KBShortcut } from "KBShortcut";
 import { UIPush } from "UIPush";
 import { errorAlerter, ignore } from "misc";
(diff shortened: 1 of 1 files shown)


anoek:
@Sofia-mal this patch is now up on beta, since you were able to reproduce it, would you mind trying again to see if it's fixed or not before we merge?

Sofia-mal:
I wouldn't mind :) I will try now

Sofia-mal:
I tried clicking quite a bit and it never happened, I think it's fixed. Thank you!

anoek:
Thanks for checking!

PR #1858, opened 2022-05-16
Fix timecop notification

turtoise0:
Fixes the time in the notification when you're about to timeout.

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -29,6 +29,7 @@ import { deepEqual } from "misc";
 import { isLiveGame, durationString } from "TimeControl";
 
 import { NotificationManager } from "./NotificationManager";
+import { guessResolution } from "./util";
 
 export const notification_manager: NotificationManager = new NotificationManager();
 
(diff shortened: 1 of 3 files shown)


anoek:
Are we seeing `notification.expiration` in both seconds and milliseconds?

turtoise0:
I think it's in milliseconds, but I wasn't sure.

anoek:
Do you happen to recall what kind of clock was being used in the game(s) you saw the problem in? I think I want to fix the underlying bug rather than do the guessing in this patch, otherwise the mobile devs are probably going to be even more frustrated with me :sweat_smile:

turtoise0:
I was looking at https://beta.online-go.com/game/12170 and https://beta.online-go.com/game/12161.

anoek:
Looks like this has been broken for a *long* time, I think we broke it with  https://github.com/online-go/online-go.com/pull/287 

I reverted that and it looks like it's fixed up, I'm going to close this in favor of that revert to avoid confusion about needing to determine if expiration is in ms/s, it's always in ms

PR #1861, opened 2022-05-17
Remove strict seki mode.

benjaminpjones:
This checkbox has been hidden since before the open source release.

Screen Shot 2022-05-17 at 3 20 47 PM

## Proposed Changes

  - Remove state variables `strict_seki_mode` and `rules`
  - remove the Strict Scoring checkbox JSX


anoek:
Hah, yep I reckon that's long enough to prove that people don't care about that. Thanks for the cleanup!

PR #1863, opened 2022-05-19
New report category

Sofia-mal:
Another thing that seemed easy to try. 
Usually people who leave games are reported on "game stalling" but I think they are two different things.

### DIFF
--- a/src/components/Report/Report.tsx
+++ b/src/components/Report/Report.tsx
@@ -113,7 +113,7 @@ export const report_categories: Array<ReportDescription> = [
         ),
         game_id_required: true,
     },
-     {
+    {
         type: "escaping",
         title: pgettext("Report user for escaping", "Escaping"),
         description: pgettext(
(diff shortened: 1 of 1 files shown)


Sofia-mal:
And instead I did something wrong, but I don't understand what 😬

anoek:
Ahh yeah that one is a bit cryptic, it want's you to remove a space before the `{` bracket - you should also be able to run

`npm run prettier`

and it should automatically fix it for you - depending on what editor you're using, it might also have the ability (probably through a plugin) to run prettier when you save a file, that's what I do that way it doesn't yell at me.

The other thing you might want to do is run

`npx husky install`

which will install some "hooks", which just run these kinds of tests when you are making a commit, so it'll complain at you then as opposed to waiting until after the commits are pushed up and a pull request is made, since that's kind of a slow way of checking to see if our pedantic style checkers are happy yet or not.

Sofia-mal:
You talk in a complicated way 🤣 (my fault of course!)
I should have removed the space

anoek:
Thanks for the improvement!

PR #1869, opened 2022-05-21
announcements: show tournament announcement up to 35 minutes in advance, so clock skew does not hide it

### LINKED ISSUE #1399: Tournament indicator only appears on page reload (mobile)
**Describe the bug**
The tournament indicator (`a.TournamentIndicator`), which displays a trophy and a countdown timer until the start of the next automatic live tournament, is not visible on pages loaded from xx:30-xx:59, even after crossing into the next hour.  Because of the way the application is served, you can be actively using the site for the full 30 minutes and never see the indicator.

**To Reproduce**
1. Open any page in a mobile browser at, say 5:59.
2. Wait until 6:00.
3. Observe that no tournament indicator appears.
4. Reload and immediately see the indicator.

**Expected behavior**
The tournament indicator should appear on the hour.

**Screenshots**
Before reload (note clock time):

before

After reload:

after

 - Device: Pixel 3 XL
 - OS: Android
 - Browser: Brave
 - Version 1.20.103 (latest)

**Additional context**

It works fine on the desktop.

pdg137:
Fixes #1399.

You will now see a trophy indicator every time instead of with 50% probability depending on the direction of your device's clock skew.  Allowing up to 5 minutes of skew since I don't think there's any downside, but I have not tested this much.

### DIFF
--- a/src/components/Announcements/Announcements.tsx
+++ b/src/components/Announcements/Announcements.tsx
@@ -128,7 +128,9 @@ export class Announcements extends React.PureComponent<{}, AnnouncementsState> {
             }, moment(announcement.expiration).toDate().getTime() - Date.now());
         } else {
             const t = moment(announcement.expiration).toDate().getTime() - Date.now();
-            if (t > 0 && t < 30 * 60 * 1000) {
+            // Tournaments are announced 30 minutes prior, but allow
+            // up to 5 minutes of clock skew.
+            if (t > 0 && t < 35 * 60 * 1000) {
                 data.set("active-tournament", announcement);
             }
         }


anoek:
That's interesting, can't say I understand why the issue was originally happening or why this fixes it, but since you're both the reporter and fixer I'll assume it works and I don't see a particular downside. Thanks for hunting that down!

pdg137:
Thanks for merging. I hope it really works!

For the record, what I think was going on is that the live/blitz tournaments are announced by the server 30 minutes before the start time. Bur the client was rejecting any announcement coming more than 30 minutes before its start time. So if the client's clock was slow at all (even milliseconds!) it would reject all of these announcements.

There's probably still an issue with correspondence tournaments, since I think these get announced way earlier. Maybe on the server you could change to announcing (or re-announcing) those ones also 30 minutes before they start?

anoek:
Oooh got it, cool yeah this should work - thanks for the fix!

PR #1870, opened 2022-05-21
popover: remove max-height of 100vh, fixing click events on usernames in tournaments

### LINKED ISSUE #1604: Tournament page player names have conflicting features bound to a tap/click
**Describe the bug**
On a tournament page, there are multiple events bound to a tap or click on a player name, and they don't work together.  They don't all always work, either, so it's often a surprise what will happen if you tap a player in a tournament.

**To Reproduce**
1. Go to a double-elimination tournament on mobile: https://online-go.com/tournament/78957
2. Tap on a player name
3. See three things happen simultaneously: the player is highlighted in green, the player card is shown, and you are quickly directed to the player's page so you don't get to look at the green highlights or player card for more than an instant.

In a Simultaneous McMahon tournament, it used to only *sometimes* perform the redirect to the player page, but while testing just now it seems to always happen for me.

**Expected behavior**
A tap/click should not do multiple things that don't work together.  Probably best if it just stays on the tournament page, since going back and forth doesn't work well with big tournaments anyway, and opening the player card is the normal behavior on the rest of the site.

**Screenshots**

Slightly before the redirect happens:

image

You only get to see this for a fraction of a second.

pdg137:
Fixes #1604.  BUT mobile users will no longer get to see the green highlight in D. E. tournaments, which they could only see for a fraction of a second before getting redirected.

### DIFF
--- a/src/lib/popover.styl
+++ b/src/lib/popover.styl
@@ -20,7 +20,6 @@
     display: flex;
     //min-height: 100%;
     //height: 100%;
-    max-height: 100vh;
     top: 0;
     left: 0;
     right: 0;


anoek:
I couldn't replicate the original bug, but everything seems to work after this patch too so seems ok, and glad it fixed it for you. Thanks for finding, hunting, and fixing!

anoek:
I couldn't replicate the original bug, but everything seems to work after this patch too so seems ok, and glad it fixed it for you. Thanks for finding, hunting, and fixing!

PR #1871, opened 2022-05-21
right-col: use margin+padding instead of just padding

### LINKED ISSUE #1807: rengo scrollbar hidden behind popout menu
The main scrollbar in rengo games is in a strange place, obscured by the sidebar menu widget:

image

I believe (this is from my notes from a while ago) that

>  `.right-col` has padding 2.3rem - if you make that margin 2 rem and padding 0.3 rem it's more reasonable.
> 
image

So this might be a one-line fix for someone who has time to try it and make sure it doesn't mess anything else up.

pdg137:
Fixes #1807.

When a scrollbar appears (in e.g. rengo games), it is visible instead of behind the right menu.

Before:

image

After:
image

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -416,7 +416,8 @@ goban-view-bar-width=400px
     .right-col {
         flex-basis: 25%;
         padding-left: 0.3rem;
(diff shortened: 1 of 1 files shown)


anoek:
Thanks!

PR #1873, opened 2022-05-21
Move resize logic into GobanContainer.

benjaminpjones:
More for the [Goban Centric Components](https://docs.google.com/document/d/1viy_xeQbEZ4_N6SOQll2R4vMRDjxhDz5-wMdG7GtnHI/preview) refactor.

Goban, Joseki and Puzzles all contain similar logic that handles the resizing of the goban.  My hope is that this class can hide all this away, and make it easier to create Goban-centric views in the future.  While the resize logic was similar, they all differed slightly.  I tried to keep the implementation closest to the `Game` implementation as it seems to be the most mature (specifically it has some de-bounce logic to prevent unnecessary re-renders.)

## Proposed Changes

  - Create a component called `GobanContainer` that handles resize logic for the goban.
  - Move all this logic out of `Game`, `Joseki` and `Puzzle` components.
  - use the `setCoordinates` function to ensure toggling coordinates triggers a recalculation of square size (requires https://github.com/online-go/goban/pull/72).

### DIFF
--- a/src/components/GobanContainer/GobanContainer.tsx
+++ b/src/components/GobanContainer/GobanContainer.tsx
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2012-2022  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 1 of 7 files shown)


anoek:
Nice, thanks!

benjaminpjones:
Arg, I missed small size (e.g. 13x13) gobans in testing. I can take a look at resolving this evening, but may need to rollback this PR for now.

PR #1877, opened 2022-05-23
Quick fix for AI based off of @mmartin's patch

### LINKED ISSUE #1876: AI analysis broken
**Describe the bug**
AI analysis got broken in #1868. @benjaminpjones 

**To Reproduce**
1. Go to a game which has AI analysis.
2. Navigate moves back and forth
3. See that analysis is not working

**Screenshots**
image

**Additional context**
I think this line was forgotten during the refactoring in the PR:
https://github.com/online-go/online-go.com/blob/4965a3a53aeaa81821841ee1b6f1dcb1e79860f7/src/views/Game/Game.tsx#L1144

This fixes the issue:
```diff
diff --git a/src/views/Game/Game.tsx b/src/views/Game/Game.tsx
index cd40cf7d..4cc09d7d 100644
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -28,6 +28,7 @@ import { UIPush } from "UIPush";
 import { errorAlerter, ignore } from "misc";
 import {
     Goban,
+    GobanCore,
     GobanCanvas,
     GobanCanvasConfig,
     GoMath,
@@ -64,7 +65,7 @@ import {
 import { CancelButton } from "./PlayButtons";
 import { GameDock } from "./GameDock";
 import swal from "sweetalert2";
-import { useShowTitle, useTitle, useUserIsParticipant } from "./GameHooks";
+import { generateGobanHook, useShowTitle, useTitle, useUserIsParticipant } from "./GameHooks";
 import { GobanContainer } from "GobanContainer";

 export function Game(): JSX.Element {
@@ -118,6 +119,7 @@ export function Game(): JSX.Element {
     const [show_game_timing, set_show_game_timing] = React.useState(false);

     const title = useTitle(goban.current);
+    const score = useScore(goban.current);

     const [mode, set_mode] = React.useState<GobanModes>("play");
     const [score_estimate_winner, set_score_estimate_winner] = React.useState<string>();
@@ -1911,3 +1913,11 @@ function bindAudioEvents(goban: Goban): void {
         }
     });
 }
+const useScore = generateGobanHook(
+    (goban: GobanCore) => {
+        if (!goban) return;
+        const engine = goban.engine;
+        return engine.computeScore(true);
+    },
+    ["cur_move"],
+);
```

anoek:
Fixes #1876

This is a quick fix that should be revisited in the morning, it's not clear to me why recomputing the score fixes #1876 but it does seem to work around the issue if nothing else. 

Tagging @benjaminpjones in case you have any ideas before I get to diving deeper into it.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -64,7 +64,7 @@ import {
 import { CancelButton } from "./PlayButtons";
 import { GameDock } from "./GameDock";
 import swal from "sweetalert2";
-import { useShowTitle, useTitle, useUserIsParticipant } from "./GameHooks";
+import { useScore, useShowTitle, useTitle, useUserIsParticipant } from "./GameHooks";
 import { GobanContainer } from "GobanContainer";
 
 export function Game(): JSX.Element {
@@ -118,6 +118,7 @@ export function Game(): JSX.Element {
     const [show_game_timing, set_show_game_timing] = React.useState(false);
 
     const title = useTitle(goban.current);
+    useScore(goban.current);
 
     const [mode, set_mode] = React.useState<GobanModes>("play");
     const [score_estimate_winner, set_score_estimate_winner] = React.useState<string>();
--- a/src/views/Game/GameHooks.ts
+++ b/src/views/Game/GameHooks.ts
@@ -124,3 +124,14 @@ export const useShowTitle = generateGobanHook(
 
 /** React hook that returns the title text (e.g. "Black to move"). */
 export const useTitle = generateGobanHook((goban: GobanCore | null) => goban?.title, ["title"]);
+
+export const useScore = generateGobanHook(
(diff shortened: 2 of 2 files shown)


benjaminpjones:
Thanks for fixing that. I don't really understand the patch since `score` isn't being used, but I can take a look later as well.

anoek:
Yeah, it *shouldn't* fix it, but it does so deploying for now..

benjaminpjones:
Okay, I looked a bit last night.  It's not so much about the `engine.computeScore()` (that function has no side-effects, thank goodness!) as much as we need to trigger a re-render of the AIReview on a "cur_move" event.  I'm not really sure the best way to do that yet, but it's worth noting that replacing `return engine.computeScore()` in this patch with something like `return Math.random()` has a similar effect!

PR #1879, opened 2022-05-26
Added option to set auth_host for social login in config

GreenAsJade:
(plus routes comment and data schema de-dupe)

Fixes not being able to use social_auth on dev frontend server


GreenAsJade:
Hmm - this turns out to be not so easy, I had tricked myself into thinking it was working.

PR #1881, opened 2022-05-27
Make `GameAcceptModal` pass the right type `user` to `Player`

GreenAsJade:
Fixes confusion when reading a `challenge` being passed as a `user` to a `Player`

## Proposed Changes

Create the correctly typed object for the props of `Player` and pass that in.

### DIFF
--- a/src/components/GameAcceptModal/GameAcceptModal.tsx
+++ b/src/components/GameAcceptModal/GameAcceptModal.tsx
@@ -76,9 +76,10 @@ export class GameAcceptModal extends Modal<Events, GameAcceptModalProperties, {}
         }
 
         const challenger_details: PlayerObjectType = {
+            id: challenge.user_id,
+            username: challenge.username,
             pro: challenge.pro,
             rank: challenge.rank,
-            username: challenge.username,
         };
 
         return (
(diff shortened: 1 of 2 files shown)


src/components/GameAcceptModal/GameAcceptModal.tsx
```diff
@@ -75,12 +75,18 @@ export class GameAcceptModal extends Modal<Events, GameAcceptModalProperties, {}
             player_color = _("Random");
         }
 
+        const challenger_details: PlayerObjectType = {
+            pro: challenge.pro,
+            rank: challenge.rank,
+            username: challenge.username,
```
anoek, line 81: Let's add: `id: challenge.user_id` too, having that user id makes lookup quicker and nicer on the system if we need to fetch details (which we do for this case since we need the player icon).

anoek (changes requested):
Good change. I think this was one of the last places we were still using `user_id` as one of the places we look for a user id. All this typing is great - slowly converging on consistency and sanity :)

src/components/GameAcceptModal/GameAcceptModal.tsx
GreenAsJade: 👍

anoek:
Awesome, thanks!

PR #1884, opened 2022-05-28
Add link to Ukrainian Go Federation

lykahb:
Verified locally that the changes work.
image


anoek:
Thanks for adding that!

PR #1887, opened 2022-05-29
Visually separate the "create an account by signing up" fields 

GreenAsJade:
... from the other ways of getting in.

Fixes slight visual confusion about what that `Card` has on it, prepares hopefully for extending the component to support ChallengeLink flow.

## Proposed Changes

Add the word "Or..." after "Sign In" (which means Sign Up) and make it clearer that the OGS option is not an email based registration form (in fact, it goes to OGS login).

### DIFF
--- a/src/views/Register/Register.tsx
+++ b/src/views/Register/Register.tsx
@@ -18,7 +18,7 @@
 import * as React from "react";
 import { Link } from "react-router-dom";
 import * as data from "data";
-import { _ } from "translate";
+import { _, pgettext } from "translate";
 import { Card } from "material";
 import { errorAlerter } from "misc";
 import { post } from "requests";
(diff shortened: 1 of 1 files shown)


anoek:
Looking at this page again, I agree it needs an update, what about the following though:

Instead of "or..." - what about re-using what we have on the sign in page there instead? https://github.com/online-go/online-go.com/blob/devel/src/views/SignIn/SignIn.tsx#L217-L223 - might be a little clearer and visually consistent?

And then, how would y'all feel about removing the "Sign in with Email" / "Sign in direct" and instead, at the bottom of that card have something like "Have an account?" [Sign in] - basically same visual pattern that we have on the sign in page that directs people to the registration page.

GreenAsJade:
I like those suggestions, I'll do that.

We'll definitely need a different button for Sign In and Sign Up in this case, since they will be on the same page!

benjaminpjones:
> We'll definitely need a different button for Sign In and Sign Up in this case, since they will be on the same page!

As I understood it, they would still be on separate pages, but they would just point to each other in visually similar ways:

Sign in pointers

P.S. It seems like these two screens are starting to diverge stylistically (grey/white SSO buttons), I wonder if there's something we can do about that?

P.P.S. I *do* think we should choose more obviously different words for the Sign In/Sign Up button.  Perhaps **Register** or **Create account** v.s. **Log in** (I'm seeing a lot of good advice in this article btw: [15 Tips for Better Signup / Login UX](https://learnui.design/blog/tips-signup-login-ux.html))

GreenAsJade:
There's another version of this now too - the "Challenge Link Landing Page" view.     So far, it's basically the same as the right hand one above, but with a different header.

Putting them next to each other like that was very handy :)

I'll converge them if I can.

BHydden:
`/ears prick up`

ChallengeLink you say... 🥰

GreenAsJade:
It's been good to get feedback.

The changes proposed a big enough, and entangled enough with ChallengeLinks flow, that I'm going to do what's proposed in ChallengeLinks and we can thrash out the details as part of that.

PR #1888, opened 2022-05-30
Game.tsx: Replace useScore with useCurrentMove

benjaminpjones:
Though there is no functional change, I think this makes it more clear why the hook is necessary: "cur_move" events trigger a rerender of `AIReview`, which depends on the value of `goban.engine.cur_move`.

Related: #1877

## Proposed Changes

  - Replace `useScore` with `useCurrentMove`
  - Pass the result in as `props.move`

Note: I considered moving `move` from `AIReview.props` into `AIReview.state` because it can be derived from `goban`, but that ended up being way more verbose because we can't make use of the custom hook inside a class component. See https://github.com/online-go/online-go.com/commit/0200d664c601735aa7c162480499cafaed221b0b.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -64,7 +64,7 @@ import {
 import { CancelButton } from "./PlayButtons";
 import { GameDock } from "./GameDock";
 import swal from "sweetalert2";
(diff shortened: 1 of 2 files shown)


anoek:
Awesome thanks :)

PR #1892, opened 2022-05-31
Converge signin and registration

GreenAsJade:
Fixes inconsistent appearance and flow.

## Proposed Changes

Factored out SocialButtons, aligned styles and layout.


GreenAsJade:
Note: I just noticed that this change inadvertently positions the Card vertically in the middle of the screen on larger screens.

I hadn't intended that, but I don't actually mind it.  

It appears that a `min-height: 100%` crept in, which causes this.

If we hate it, that's what will have to be weeded out.

anoek:
Thanks for the improvement!

PR #1894, opened 2022-06-05
Added a component template

GreenAsJade:
Fixes "darn, now how do I do this again?"

## Proposed Changes

Add a 'templates' directory and wack an example component in it with various things that we always need.

### DIFF
--- a/src/templates/component/TemplateComponent/TemplateComponent.tsx
+++ b/src/templates/component/TemplateComponent/TemplateComponent.tsx
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2012-2022  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as React from "react";
+
+import * as data from "data";
+
+import { pgettext } from "translate";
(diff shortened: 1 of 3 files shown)


src/templates/component/TemplateComponent/index.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2012-2017  Online-Go.com
```
anoek, line 2: geeze where did you dig that year from? Guess I missed the source for that a few times :sweat_smile:

GreenAsJade: LOL yeah ... if you grep it you'll find plenty of hits, it turns out!  :o

src/templates/component/TemplateComponent/TemplateComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+type TemplateComponentProps = {
+    render_twice?: boolean;
+};
+
+export function TemplateComponent(props: TemplateComponentProps): JSX.Element {
+    const [username, setUsername] = React.useState("");
+
+    const hangs_around = React.useRef("forever");
+
+    const user = data.get("config.user"); // user.anonymous will be set if they aren't logged in
```
anoek, line 33: better to use `const user = useUser();` these days, comes from `import { useUser } from "hooks";`

src/templates/component/TemplateComponent/index.ts
GreenAsJade: ... yet another reason for having a template that is correct :D

src/templates/component/TemplateComponent/TemplateComponent.tsx
GreenAsJade: 😮    ... and another reason to have learnings in a template!

anoek: (useUser will automatically refresh when the user object is updated, as in when the username/icon/ui_class/rank changes kinda stuff)

anoek: Yeah it's pretty new, figured if we're making a template though, that should be in there :)

GreenAsJade: Template's been there <24 hours, already winning :D  :D

PR #1905, opened 2022-06-14
Factor out ChallengeDetailsReviewPane from GameAcceptModal 

GreenAsJade:
Fixes need for a pane to display challenge details in different places

## Proposed Changes

Factor out challenge details review pane from GameAcceptModal.


anoek:
Thanks for the refactoring!

PR #1906, opened 2022-06-15
Add OgsResizeDetector, which does not throw on older browsers.

benjaminpjones:
Fixes https://forums.online-go.com/t/website-crashing-on-all-of-my-games/43524

## Proposed Changes

  - Create a wrapper that replaces `ReactResizeDetector` with `React.Fragment` if `window.ResizeObserver` is undefined.

This is an imperfect solution, but makes the site usable for folks with older browsers.

## Alternatives considered

- `ErrorBoundary`
- polyfills - this would actually support resizes, but also adds a dependency.
    - react-resize-detector/build/withPolyfill removed in January
    - [resize-observer-polyfill](https://www.npmjs.com/package/resize-observer-polyfill) last updated 4 years ago
    - [@juggle/resize-observer](https://www.npmjs.com/package/@juggle/resize-observer) last updated 10 months ago

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -18,7 +18,7 @@
 import * as data from "data";
 import * as player_cache from "player_cache";
 import * as React from "react";
-import ReactResizeDetector from "react-resize-detector";
+import { OgsResizeDetector } from "OgsResizeDetector";
 import { browserHistory } from "ogsHistory";
 import { _, pgettext, interpolate } from "translate";
 import { post, del } from "requests";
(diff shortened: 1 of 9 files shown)


anoek:
Thanks for the fix, I added a hacky fallback system to support orientation changes on those devices here: 563ee372b585f8db95f0469d8011a49f15dc7b5b , though I don't have an old iOS device laying around to test it beyond what I can do by settting ResizeObserver to null, so hopefully it works for them - or at least doesn't break.

benjaminpjones:
Awesome! Yeah I think window resizes are responsible for a large portion (if not all) of the resize events, so that probably fixes just about everything.

PR #1908, opened 2022-06-17
Modify iframe embed code to play nicely with Discourse

benjaminpjones:
Discourse doesn't respect certain iframe attributes such as style. Therefore,
style="width: ..; height: .." was replaced with width=".." and height="..".

Screen Shot 2022-06-16 at 8 09 42 PM

Note: I also took width and height down to 345px as that is closer to the native elements on the [Observe Games](https://online-go.com/observe-games) page.


anoek:
Huh cool, thanks!

PR #1909, opened 2022-06-18
Hide Announcements (and PMs etc.) for embeds

benjaminpjones:
Fixes

>Site notifications show up inside the embed :smile: 

https://forums.online-go.com/t/embed-live-game/42216/10

## Proposed Changes

  - Just tell OGS we are in zen mode in embed panes
  - Remove NavBar hiding logic since it respects zen mode.

This feels like a bit of a hack since this isn't technically zen mode, but zen mode does so much, and embeds needs it all so why re-invent the wheel? :)

### DIFF
--- a/src/components/Announcements/Announcements.tsx
+++ b/src/components/Announcements/Announcements.tsx
@@ -175,13 +175,6 @@ export class Announcements extends React.PureComponent<{}, AnnouncementsState> {
                     return (
                         <div className="announcement" key={idx}>
                             <i className="fa fa-times-circle" onClick={announcement.clear} />
-                            {/*
-                    {(announcement.type === 'tournament' || null) &&
-                        <span className='expiration'>
-                            {moment(announcement.expiration).fromNow(true)}:
-                        </span>
-                    }
-                    */}
                             {announcement.link ? (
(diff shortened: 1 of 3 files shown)


benjaminpjones:
To explain that last commit - OldNavBar gets hidden a bit differently than everything else.  The Game component was actually positioned to cover up the nav bar, instead of setting `display: none` as is done with most other components that get hidden during zen mode.

https://github.com/online-go/online-go.com/blob/229d2c84c1d1ff1f0438d160d0eefc9084932063/src/views/Game/Game.styl#L34-L36

https://github.com/online-go/online-go.com/blob/324314f609c8056b7a9bf29312720fd5da2e7d1b/src/components/Announcements/Announcements.styl#L46-L48

https://github.com/online-go/online-go.com/blob/b2846e0c9d79442e35d4367e377405df7c388091/src/components/NavBar/EXV6NavBar.styl#L23-L25

https://github.com/online-go/online-go.com/blob/324314f609c8056b7a9bf29312720fd5da2e7d1b/src/components/PrivateChat/PrivateChat.styl#L116-L122

anoek:
Awesome thanks!

PR #1910, opened 2022-06-18
Upgrade to sweetalert2 v11.x

GreenAsJade:
Fixes out of date dependency.

## Proposed Changes

 - Change version of sweetalert2 from 6.x to 11.x
 - Factor out configuration from main.tsx into separate file
 - Convert from no-longer-supported `default` to `mixin`
 - Convert all uses of `swal` to use the newly defined `alert`  with  the mixed-in defaults
 - Handle alert-dismissal the new way (alert acceptance and dismissal both happen inside promise-accept - it's necessary to test the promise payload to see what happened).
    - https://github.com/sweetalert2/sweetalert2/releases/tag/v7.0.0 
 -  Use `void` to indicate where we are ignoring promise-reject results from the alert, instead of various forms of "do nothing" in the `catch`.

### DIFF
--- a/src/components/Announcements/ActiveAnnouncements.tsx
+++ b/src/components/Announcements/ActiveAnnouncements.tsx
@@ -31,7 +31,7 @@ import { getBlocks, setAnnouncementBlock } from "../BlockPlayer";
 import * as data from "data";
 import * as preferences from "preferences";
 
-import swal from "sweetalert2";
+import { alert } from "swal_config";
 
 // Holds the expirations dates of cleared announcements
 const hard_cleared_announcements: { [id: number]: number } = data.get(
@@ -115,20 +115,23 @@ export class ActiveAnnouncements extends React.PureComponent {
                                 username: announcement.creator.username,
                             }),
                             onClick: () => {
-                                swal({
-                                    text: interpolate(
-                                        _(
-                                            "Are you sure you want to hide all announcements from {{name}}?",
+                                alert
+                                    .fire({
+                                        text: interpolate(
+                                            _(
+                                                "Are you sure you want to hide all announcements from {{name}}?",
+                                            ),
+                                            { name: announcement.creator.username },
                                         ),
-                                        { name: announcement.creator.username },
-                                    ),
-                                    showCancelButton: true,
-                                    confirmButtonText: _("Yes"),
-                                    cancelButtonText: _("Cancel"),
-                                })
-                                    .then(() => {
-                                        setAnnouncementBlock(announcement.creator.id, true);
-                                        this.forceUpdate();
+                                        showCancelButton: true,
+                                        confirmButtonText: _("Yes"),
+                                        cancelButtonText: _("Cancel"),
+                                    })
+                                    .then(({ value: yes }) => {
+                                        if (yes) {
+                                            setAnnouncementBlock(announcement.creator.id, true);
+                                            this.forceUpdate();
+                                        }
                                     })
                                     .catch(() => 0);
                                 return;
@@ -140,17 +143,20 @@ export class ActiveAnnouncements extends React.PureComponent {
                         announcement_actions.push({
                             title: _("Hide stream announcements"),
                             onClick: () => {
-                                swal({
-                                    text: _(
-                                        "Are you sure you want to hide all announcements for streamers?",
-                                    ),
-                                    showCancelButton: true,
-                                    confirmButtonText: _("Yes"),
-                                    cancelButtonText: _("Cancel"),
-                                })
-                                    .then(() => {
-                                        preferences.set("mute-stream-announcements", true);
-                                        this.forceUpdate();
+                                alert
+                                    .fire({
+                                        text: _(
+                                            "Are you sure you want to hide all announcements for streamers?",
+                                        ),
+                                        showCancelButton: true,
+                                        confirmButtonText: _("Yes"),
+                                        cancelButtonText: _("Cancel"),
+                                    })
+                                    .then(({ value: yes }) => {
+                                        if (yes) {
+                                            preferences.set("mute-stream-announcements", true);
+                                            this.forceUpdate();
+                                        }
                                     })
                                     .catch(() => 0);
                                 return;
@@ -162,17 +168,20 @@ export class ActiveAnnouncements extends React.PureComponent {
                         announcement_actions.push({
                             title: _("Hide event announcements"),
                             onClick: () => {
-                                swal({
-                                    text: _(
-                                        "Are you sure you want to hide all event announcements?",
-                                    ),
-                                    showCancelButton: true,
-                                    confirmButtonText: _("Yes"),
-                                    cancelButtonText: _("Cancel"),
-                                })
-                                    .then(() => {
-                                        preferences.set("mute-event-announcements", true);
-                                        this.forceUpdate();
+                                alert
+                                    .fire({
+                                        text: _(
+                                            "Are you sure you want to hide all event announcements?",
+                                        ),
+                                        showCancelButton: true,
+                                        confirmButtonText: _("Yes"),
+                                        cancelButtonText: _("Cancel"),
+                                    })
+                                    .then(({ value: yes }) => {
+                                        if (yes) {
+                                            preferences.set("mute-event-announcements", true);
+                                            this.forceUpdate();
+                                        }
                                     })
                                     .catch(() => 0);
                                 return;
@@ -184,17 +193,20 @@ export class ActiveAnnouncements extends React.PureComponent {
                         announcement_actions.push({
                             title: _("Hide go service advertisements"),
                             onClick: () => {
-                                swal({
-                                    text: _(
-                                        "Are you sure you want to hide all go related advertisements?",
-                                    ),
-                                    showCancelButton: true,
-                                    confirmButtonText: _("Yes"),
-                                    cancelButtonText: _("Cancel"),
-                                })
-                                    .then(() => {
-                                        preferences.set("mute-event-announcements", true);
-                                        this.forceUpdate();
+                                alert
+                                    .fire({
+                                        text: _(
+                                            "Are you sure you want to hide all go related advertisements?",
+                                        ),
+                                        showCancelButton: true,
+                                        confirmButtonText: _("Yes"),
+                                        cancelButtonText: _("Cancel"),
+                                    })
+                                    .then(({ value: yes }) => {
+                                        if (yes) {
+                                            preferences.set("mute-event-announcements", true);
+                                            this.forceUpdate();
+                                        }
                                     })
                                     .catch(() => 0);
                                 return;
--- a/src/components/ModerateUser/ModerateUser.tsx
+++ b/src/components/ModerateUser/ModerateUser.tsx
@@ -19,7 +19,7 @@ import * as React from "react";
 import * as data from "data";
 import { _ } from "translate";
 import { put, get, del } from "requests";
-import { errorAlerter, ignore } from "misc";
+import { errorAlerter } from "misc";
 import { proRankList } from "rank_utils";
 import { Modal, openModal } from "Modal";
 import { lookup } from "player_cache";
@@ -30,7 +30,7 @@ interface ModerateUserProperties {
     playerId?: number;
 }
 
-import swal from "sweetalert2";
+import { alert } from "swal_config";
 
 const pro_ranks = proRankList(false);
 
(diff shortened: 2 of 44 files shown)


GreenAsJade:
I've tested instances of the changes noted above, but it seems infeasible to test all of them, or even all the ones that have the semantic change.

benjaminpjones (approved):
This looks great! Seems like v11 has a lot to offer :)

src/components/Announcements/ActiveAnnouncements.tsx
```diff
@@ -162,17 +168,20 @@ export class ActiveAnnouncements extends React.PureComponent {
...
+                                alert
+                                    .fire({
+                                        text: _(
+                                            "Are you sure you want to hide all event announcements?",
+                                        ),
+                                        showCancelButton: true,
+                                        confirmButtonText: _("Yes"),
+                                        cancelButtonText: _("Cancel"),
+                                    })
+                                    .then(({ value: yes }) => {
```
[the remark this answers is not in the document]
GreenAsJade: I think I chose `yes` where the coder changed the text of the button to `yes`.   It means that they chose what we asked them.

GreenAsJade:
I noticed that the icons are animated now - a style choice.  I left that turned on, but I think it's worth consciously deciding if we want that or not.   I'd go with whatever.

src/views/Firewall/Firewall.tsx
```diff
@@ -232,19 +230,18 @@ function FirewallRuleRow({
...
-            cancelButtonText: "Cancel",
-        })
+        alert
+            .fire({
+                text: "Are you sure you want to delete this row?",
+                showCancelButton: true,
+                confirmButtonText: "Yes",
+                cancelButtonText: "Cancel",
+            })
             .then(() => {
```
anoek, line 240: I think we need a yes/no check here too?

src/components/ModerateUser/ModerateUser.tsx
```diff
@@ -120,18 +120,18 @@ export class ModerateUser extends Modal<Events, ModerateUserProperties, any> {
...
-        swal({
-            text: `Are you sure you want to delete the account "${username}" (${user_id})? This cannot be undone.`,
-            showCancelButton: true,
-        })
+        void alert
+            .fire({
+                text: `Are you sure you want to delete the account "${username}" (${user_id})? This cannot be undone.`,
+                showCancelButton: true,
+            })
             .then(() => {
```
anoek, line 128: I think we need a yes/no check here?

src/views/Ladder/Ladder.tsx
```diff
@@ -100,15 +100,16 @@ class _Ladder extends React.PureComponent<LadderProperties, LadderState> {
...
+            .fire({
+                text: _(
+                    "Are you sure you want to withdraw from the ladder? If you decide to rejoin the ladder in the future you will have to start from the bottom!",
+                ),
+                showCancelButton: true,
+                confirmButtonText: _("Yes"),
+                cancelButtonText: _("No"),
+                focusCancel: true,
+            })
             .then(() => {
```
anoek, line 113: I think we need a yes/no check here too?

src/views/Ladder/Ladder.tsx
```diff
@@ -596,17 +598,18 @@ export class LadderRow extends React.Component<LadderRowProperties, LadderRowSta
...
+                    _(
+                        "Are you ready to start your game with {{player_name}}?",
+                    ) /* translators: ladder challenge */,
+                    { player_name: ladder_player.player.username },
+                ),
+                showCancelButton: true,
+                confirmButtonText: _("Yes!"),
+                cancelButtonText: _("No"),
+            })
             .then(() => {
```
anoek, line 613: I think we need a yes/no check here too?

src/views/LearningHub/LearningHub.tsx
```diff
@@ -259,10 +259,11 @@ class SectionNav extends React.Component<{}, any> {
...
-        swal({
-            text: _("Are you sure you wish to reset your tutorial progress?"),
-            showCancelButton: true,
-        })
+        alert
+            .fire({
+                text: _("Are you sure you wish to reset your tutorial progress?"),
+                showCancelButton: true,
+            })
             .then(() => {
```
anoek, line 267: I think we need a yes/no check here too?

anoek (commented):
Awesome, thanks for taking on the upgrade :)

src/views/Ladder/Ladder.tsx
```diff
@@ -596,17 +598,18 @@ export class LadderRow extends React.Component<LadderRowProperties, LadderRowSta
...
-                    "Are you ready to start your game with {{player_name}}?",
-                ) /* translators: ladder challenge */,
-                { player_name: ladder_player.player.username },
-            ),
-            showCancelButton: true,
-            confirmButtonText: _("Yes!"),
-            cancelButtonText: _("No"),
-        })
+        alert
+            .fire({
```
benjaminpjones, line 602: y/n check

src/views/Ladder/Ladder.tsx
```diff
@@ -100,15 +100,16 @@ class _Ladder extends React.PureComponent<LadderProperties, LadderState> {
...
-        swal({
-            text: _(
-                "Are you sure you want to withdraw from the ladder? If you decide to rejoin the ladder in the future you will have to start from the bottom!",
-            ),
-            showCancelButton: true,
-            confirmButtonText: _("Yes"),
-            cancelButtonText: _("No"),
-            focusCancel: true,
-        })
+        alert
```
benjaminpjones, line 103: y/n check

src/views/Ladder/Ladder.tsx
```diff
@@ -461,11 +462,12 @@ export class LadderRow extends React.Component<LadderRowProperties, LadderRowSta
     adjustLadderPosition(player) {
         console.log(player);
 
-        swal({
-            text: "New ladder position for player " + player.username,
-            input: "number",
-            showCancelButton: true,
-        })
+        alert
+            .fire({
```
benjaminpjones, line 466: `new_rank` needs to be handled differently here.

```
({value: new_rank}) => {
    if (new_rank === undefined) { return; }
    ...
}
```

src/views/LearningHub/LearningHub.tsx
```diff
@@ -259,10 +259,11 @@ class SectionNav extends React.Component<{}, any> {
     }
 
     resetProgress = () => {
-        swal({
-            text: _("Are you sure you wish to reset your tutorial progress?"),
-            showCancelButton: true,
-        })
+        alert
+            .fire({
```
benjaminpjones, line 263: y/n check

src/views/Moderator/ban_functions.tsx
```diff
@@ -19,21 +19,23 @@ import * as React from "react";
...
-                .then(resolve)
-                .catch(reject);
-        }, reject);
+        alert
+            .fire({
+                text: prompt,
+                input: "text",
+                showCancelButton: true,
+            })
+            .then((reason) => {
```
benjaminpjones, line 32: `reason` => `{value: reason}` (and null check)

src/views/PuzzleCollection/PuzzleCollectionList.tsx
```diff
@@ -151,16 +151,18 @@ export function PuzzleCollectionList(): JSX.Element {
...
-        })
-            .then((name) => {
+        void alert
+            .fire({
+                text: _("Collection name"),
+                input: "text",
+                showCancelButton: true,
+            })
+            .then((res) => {
+                const name = res.value;
```
benjaminpjones, line 161: nullcheck `name`

src/views/SignIn/SignIn.tsx
```diff
@@ -159,18 +159,19 @@ export class SignIn extends React.PureComponent<{}, any> {
...
-            showCancelButton: true,
-        })
+        alert
+            .fire({
+                text: _("What is your username?"),
+                input: "text",
+                showCancelButton: true,
+            })
             .then((username) => {
                 post("/api/v0/reset", { username: username })
```
benjaminpjones, line 169: `{value: username}` (and null check!)

src/views/User/AvatarCard.tsx
```diff
@@ -92,27 +92,29 @@ export function AvatarCard({
 
     const toggleEdit = () => {
         if (editing) {
-            let promise: Promise<void>;
+            let promise: Promise<any>;
```
benjaminpjones, line 95: Nit: `Promise<{ value: boolean }>`

benjaminpjones (changes requested):
My apologies, I took another look and found some stuff I think needs to be changed.

src/views/User/AvatarCard.tsx
GreenAsJade: I tried really hard to type this, getting as far as 

`let promise: Promise<SweetAlertResult<boolean>> | Promise<{ value: boolean }>;`

however, the thing that Sweetalert returns is just not amenable to clean faking.

Or - if it is, I'd be fascinated to see how.

benjaminpjones: Oh huh, I didn't think optionality would be an issue since `strictNullChecks` is turned off.  Looks like this works instead:

```
Promise<{ value?: boolean }>
               ^
```

`SweetAlertResult<boolean>` looks like the following, so the `value?` needs to match:
```
{
    isConfirmed: boolean
    isDenied: boolean
    isDismissed: boolean
    value?: boolean
    dismiss?: Swal.DismissReason
}
```

src/views/Moderator/ban_functions.tsx
```diff
@@ -29,13 +29,18 @@ function moderate(player_id, prompt, obj) {
...
             })
-            .then((reason) => {
-                obj.moderation_note = reason;
-                console.log(obj);
-                put("players/" + player_id + "/moderate", obj)
-                    .then(resolve)
-                    .catch(reject);
-            }, reject);
+            .then(({ value: reason }) => {
+                if (reason) {
```
benjaminpjones, line 33: What behavior do we want here if `reason === ""`?

Previously, it would still send a request to the server and the server would respond with a Bad Request message.  Now the box closes with no error.

Screen Shot 2022-06-19 at 4 31 44 PM

src/views/PuzzleCollection/PuzzleCollectionList.tsx
```diff
@@ -157,24 +157,25 @@ export function PuzzleCollectionList(): JSX.Element {
...
-            .then((res) => {
-                const name = res.value;
-                if (!name || name.length < 5) {
-                    void alert.fire({
-                        text: _("Please provide a longer name for your new puzzle collection"),
-                    });
-                    return;
-                }
+            .then(({ value: name }) => {
+                if (name) {
```
benjaminpjones, line 161: I think the same issue here, the empty string no longer gets the same treatment (alert: "Please provide a longer name for your new puzzle collection").  I would recommend explicitly null checking (`=== undefined`) or swal's `isConfirmed` property which I believe is more readable:

```
.then(({ value: name, isConfirmed}) => {
    if (isConfirmed) { /* do stuff */ }
});
```

src/views/User/AvatarCard.tsx
GreenAsJade: All hail @benjito, the typing guru 🙇

src/views/SignIn/SignIn.tsx
```diff
@@ -159,27 +159,28 @@ export class SignIn extends React.PureComponent<{}, any> {
...
-                        } else {
-                            console.error(res);
-                            errorAlerter(res);
-                        }
-                    })
-                    .catch(errorAlerter);
-            })
-            .catch(ignore);
+            .then(({ value: username }) => {
+                if (username) {
```
benjaminpjones, line 169: same empty string feedback: previously we saw the server error ("username too short"), now the box closes as if the request had completed successfully.

src/views/Moderator/ban_functions.tsx
GreenAsJade: As discussed, I put in isConfirmed handlers where appropriate for all `input: "text"` alerts.

As I was typing a description of what I did, I realised I could do better, hence the second commit.

Sweetalert's input validation helps here.

Where before it made sense to defer validation to the server (for example, and empty moderation note might be OK, who knows?) it does that still.

Where actually some validation of the input text is obviously needed, I've done that.

GreenAsJade:
I've marked this as draft to indicate I think it's best if we test this with challenge_links (via PR #1912) on beta first.

PR #1911, opened 2022-06-19
Fix for #1240

### LINKED ISSUE #1240: Komi behaving weirdly in certain games
**Describe the bug**
Komi seems to be behaving weirdly in a small subset of 9x9 games

**Screenshots**

https://online-go.com/game/27578385
Screen Shot 2020-10-16 at 11 53 46 PM

https://online-go.com/game/27578346
Screen Shot 2020-10-16 at 11 58 22 PM
Screen Shot 2020-10-16 at 11 58 38 PM

alexanderrae:
Fixes #1240 

## Proposed Changes

  -  Game outcome, when a number is passed, will be rounded to nearest half. 
Solution from: https://stackoverflow.com/questions/6137986/javascript-roundoff-number-to-nearest-0-5

Tested with custom komi of 0.1 and 0.4, which would normally produce a score like 4.999999999999. 
This is my first pull request, so I hope I did everything right

### DIFF
--- a/src/lib/misc.ts
+++ b/src/lib/misc.ts
@@ -201,7 +201,8 @@ export function getOutcomeTranslation(outcome: string) {
     }
 
     if (/[0-9.]+/.test(outcome)) {
-        const num = outcome.match(/([0-9.]+)/)[1];
+        const num = Math.round(outcome.match(/([0-9.]+)/)[1]*2)/2;
+
         return interpolate(pgettext("Game outcome", "{{number}} points"), { number: num }); // eslint-disable-line id-denylist
     }
 
--- a/src/lib/misc.ts
+++ b/src/lib/misc.ts
@@ -203,7 +203,7 @@ export function getOutcomeTranslation(outcome: string) {
     if (/[0-9.]+/.test(outcome)) {
         const num: number = +outcome.match(/([0-9.]+)/)[1];
         const rounded_num = Math.round(num * 2) / 2;
-        return interpolate(pgettext("Game outcome", "{{number}} points"), { number: rounded_num }); // eslint-disable-line id-denylist
+        return interpolate(pgettext("Game outcome", "{{score}} points"), { score: rounded_num });
     }
 
     return outcome;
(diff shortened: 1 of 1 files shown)


GreenAsJade:
At first I was thinking (about the original code, not this change) "why the heck do we need to disable eslint for this?"

`return interpolate(pgettext("Game outcome", "{{number}} points"), { number: num }); // eslint-disable-line id-denylist`

... but actually "number" is a pretty good word to provide to the translators there, I have to admit.

I wonder if "score" would arguably be even better?  Afterall, it's not just any "number", it is in fact the score, isn't it?

GreenAsJade:
@alexanderrae  Welcome and thanks for contributing!

I tried it out, it works for me :)

I do wonder whether we should as well validate the komi input ... is there any reason to allow wierd komi?

(Note that your fix is still needed, because other clients could supply wierd komi as well, unless we validate on the backend also!)

GreenAsJade:
Hey @alexanderrae  if you'd like to chat more with other OGS developers, you can find us in slack, here: https://join.slack.com/t/online-go/shared_invite/zt-xuhusnyf-ovDkTTNrRvDdJKamDua9iQ

src/lib/misc.ts
```diff
@@ -201,8 +201,9 @@ export function getOutcomeTranslation(outcome: string) {
     }
 
     if (/[0-9.]+/.test(outcome)) {
-        const num = outcome.match(/([0-9.]+)/)[1];
-        return interpolate(pgettext("Game outcome", "{{number}} points"), { number: num }); // eslint-disable-line id-denylist
+        const num: number = +outcome.match(/([0-9.]+)/)[1];
+        const rounded_num = Math.round(num * 2) / 2;
+        return interpolate(pgettext("Game outcome", "{{score}} points"), { score: rounded_num });
```
anoek, line 206: Let's change this back to "{{number}} points", since that string is already translated - replacing it with "{{score}} points" will cause it to show up in english until translators come along and re-translate it.

anoek (changes requested):
Very cool, thanks for the fix! Just a quick string fix so we don't break translations and we're g2g I think.

anoek:
Just read the conversation about number vs score.. I totally agree it *should* have been score, that's a better name. But I don't think changing it is worth the translation churn?

src/lib/misc.ts
GreenAsJade: Hah that was my fault for suggesting the change 😊

anoek (approved):

anoek:
Thanks for the fix!

PR #1912, opened 2022-06-20
Challenge links with swal v11

GreenAsJade:
Fixes challenge_links not having swal v11 :)

## Proposed Changes

  - Merges PR #1910 onto #1893 
  - Uses new swal to put a link in the footer of invite-only challenge acceptance


src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -633,6 +633,9 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                             const footer_text = _("View your invite-only challenges");
                             void alert.fire({
                                 text: _("Invite-only Challenge created!"),
+                                // It could be better if this were a <Link>
+                                // The problem with this is discussed here: https://stackoverflow.com/a/72690830/554807
+                                // This can be fixed when HistoryRouter is properly supported, if we can be bothered.
```
anoek, line 638: FTR I think we can actually remove HistoryBrowser and use the normal `BrowserRouter`, I didn't at the time because I didn't want to break things, but I've since experimented with it a bit and I don't think we're doing anything with HistoryBrowser that isn't directly supported (with some small code refactoring) by BrowserRouter.

anoek: (Not that we should change things in this PR or anything, just making note)

GreenAsJade: Right - but even if we were to do that, the problem would still exist (as I understand it, though I haven't tested this assertion).   HistoryRouter-with-global-history is the solution to "how to render links outside a react-router-dom ReactTree" independent of whether the original ReactTree is using it (again AIUI).

anoek:
Thanks for the big feature!

PR #1914, opened 2022-06-22
Note in eslintrc that we should investigate our exhaustive-deps warnings

GreenAsJade:
Fixes wondering whether we have this ommitted on purpose or by ommission.

## Proposed Changes

 - Add the rule into the "off" section of eslintrc for now, because don't want the noise now, but we should investigate


anoek:
That seems like a really great thing to turn on, thanks for the find

PR #1917, opened 2022-06-24
Fix URL validator regex

mmartin:
URL validator regex has a superfluous `*` character, which makes the regex engine go completely bananas for certain invalid inputs.

For example with `https://en.wikipedia.org/wiki/Colonoscopy@`:
```
$ time node test.js
node test.js  41.91s user 0.34s system 99% cpu 42.387 total
```


anoek:
Teach me to copy/paste some random regex.. Also, nice example test.

Thanks for the fix!

PR #1919, opened 2022-06-25
Disable touch actions when pen tool is selected

### LINKED ISSUE #1913: Pen in analysis mode is impossible to use on mobile
Because the screen keeps moving.

**Expected behavior**
Maybe scrolling should be disabled when using the pen on mobile?

 - OS: android 12
 - Chrome 102 and firefox daylight 101.2.0

turtoise0:
Fixes #1913

## Proposed Changes

  - Sets `touch-action: none` when pen tool is selected
  - Fires an alert when user selects pen tool on a touch device

### DIFF
--- a/src/global_styl/global.styl
+++ b/src/global_styl/global.styl
@@ -632,6 +632,5 @@ display: inline-block;
 }
 
 .noscroll {
-    height: 100%;
-    overflow: hidden;
+    touch-action: none;
 }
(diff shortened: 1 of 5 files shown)


benjaminpjones:
I just checked how this currently works on mobile - disabling touch scrolling will make the drawing tool 1000x more usable :)

One suggestion - I think it may be more intuitive to leave the touch actions enabled on the elements outside the Goban.  That way the user can still scroll without switching tools.

This can be achieved by applying the no-touch-action class to the goban-container instead of the document element, possibly like:

```
$(".goban-container").addClass("no-touch-action");
```

What do you think?

turtoise0:
I also removed the alert, your idea is more intuitive.

PR #1920, opened 2022-06-26
Don't autocapitalize username SignIn field

turtoise0:
**Proposed changes**

Turn off autocapitalize on the username field. Chrome Android for example has it on by default.


anoek:
Thanks!

Issue #1918, opened 2022-06-25
Invite only shows up on demo creation modal


benjaminpjones:
<img width="799" alt="Screen Shot 2022-06-24 at 9 10 04 PM" src="https://user-images.githubusercontent.com/25233703/175757542-537d5fab-88fb-4492-bb8b-1a420125ea11.png">

@GreenAsJade

mmartin:
This issue is fixed and can be closed, right?

PR #1926, opened 2022-06-28
Remove "Invite only" from demo creation modal

Sofia-mal:
See https://github.com/online-go/online-go.com/issues/1918, it should work.

There are also three commits that cancel each other out. I realized late that "demo board" was on "tool" and couldn't delete the commits and I couldn't find a way to send just one. 

(If those 3 commits are a problem tell me how I can solve it please. Maybe with a new repository?)

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -1629,7 +1629,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                         </button>
                     )}
                 </div>
-                {!(this.props.playerId || null) && (mode !== "demo" || null) && (
+                {!(this.props.playerId || null) && (mode === "open" || null) && (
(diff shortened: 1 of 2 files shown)


benjaminpjones:
Hi @Sofia-mal,

I don't think the extra commits are an issue, but it may be easier in the future if you work on a branch instead of directly on `devel`. [CONTRIBUTING.md: How to Make Changes](https://github.com/online-go/online-go.com/blob/devel/CONTRIBUTING.md#how-to-make-changes) has some guidance on how to do that.  That way, you can easily start a new branch if you aren't happy with the commit history :)

Also, one note now that I am looking at the code: It looks like mode can be any of "open", "player", "computer", and "demo". I don't think we want the Invite-only message showing up in "computer" challenges either, so it may make sense to just check `mode === "open"`.

<details>
    <summary>Screenshot of computer challenge</summary>
  
Screenshot_20220628-090426_Firefox
</details>

anoek:
Thanks for the fix!

PR #1927, opened 2022-06-29
Fix list of active games

### LINKED ISSUE #1925: Active games list: clock and opponent's clock are switched sometimes
**Description**

Active games show incorrect entries for rengo games.

**Screenshots**

From here it looks like I have 16 hours remaining.

screenshot

But in fact, I have less than 9.

screenshot (1)

**Additional context**
This bug only affects rengo games where I'm playing as white, so my guess is that the game list incorrectly calculates the player's color.

turtoise0:
Fixes #1925. Also fixes how rengo games are displayed in the GameList.

## Proposed Changes

- Fix GameData type
- Add `rengo_teams` to `GobanLineSummaryProps`
- Fix calculation of player color in GobanLineSummary


anoek:
Awesome, thanks for the fix!

PR #1930, opened 2022-07-02
Ai review chart changes

mmartin:
If I create a variation while waiting for AI analysis to process, the wiggly lines in `AIReviewChart` breaks a bit.

image

In a separate commit I also simplified min/max score calculation.


anoek:
Awesome, thanks for the fix!

PR #1931, opened 2022-07-04
Update source-map-loader to 4.0.0

mmartin:
After updating to React 18 in #1779 I've often had issues with source maps, updating `source-map-loader` fixes it.

## Proposed Changes
  - Update to `source-map-loader-4.0.0`


anoek:
Awesome, thanks!

PR #1934, opened 2022-07-07
Gave the variant container divs an id

GreenAsJade:
… so it's easy to see where they come from when debugging.

Fixes wondering what created that div :)


anoek:
Thanks, and sorry for the delay!

PR #1938, opened 2022-07-14
Fix resign from tournament

mmartin:


anoek:
I knew we'd miss one with that sweetalert upgrade :sweat_smile: Thanks for hunting it down!

PR #1939, opened 2022-07-14
Update util.ts

Sofia-mal:
This I believe (and hope, otherwise forgive me 😅) makes games with 10s per move blitz, resolving the inconsistency with the quick match finder. 
I don't know what it would imply and if it is wanted, but it is easy to propose :)

### DIFF
--- a/src/components/TimeControl/util.ts
+++ b/src/components/TimeControl/util.ts
@@ -243,7 +243,7 @@ export function parseIntWithDefaultValue(
 export function makeTimeControlParameters(tc: any): TimeControl {
     const tpm = computeAverageMoveTime(tc);
     const speed: TimeControlTypes.TimeControlSpeed =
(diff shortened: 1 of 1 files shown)


koshogs:
iirc currently games that are set up as blitz through some channels then get categorised as live in game history. 

From a moderation perspective it would be good to have fixed because when you are looking through a user's game history in relation to 'escaping', blitz timeouts are not usually of interest.

anoek:
It does need to be fixed, but yeah unfortunately there's a lot to touch on the backend to ensure things are consistent.

Sofia-mal:
Excuse me. I didn't know and I tried, but maybe I should have realized that it couldn't be that easy.

anoek:
No problem and thanks for trying! But yeah it's mostly a backend thing.

PR #1946, opened 2022-07-28
Try to clarify which props are for what.

GreenAsJade:
Fixes reader confusion about what the type `SettingsGroupProps` is for, and the variable `props` in the `<Settings>` component.

## Proposed Changes

props -> child_props

SettingsGroupProps -> SettingsGroupPageProps

(allowing for later creation of the interface `SettingsGroupProps` for the `SettingsGroup` component, which may be needed to add ref forwarding into it)


anoek:
Thanks for the cleanup!

PR #1947, opened 2022-07-29
Incorporate react dynamic help - guest flow

GreenAsJade:
Fixes guests not knowing that they should update their password, etc

## Proposed Changes

 - Integrate [react-dynamic-help](https://www.npmjs.com/package/react-dynamic-help), instantiating `HelpProvider` and `HelpFlows` at the top level
 - Add `HelpItems` for a guest-user flow that shows how to set password and profile appearance.

This PR is draft, because the guest-user flow is turned on by default, rather than being triggered by arriving on a challenge link, so that it's easy to test.

It's only implemented in EXV6.

Debug is also "turned on", which provides a "reset button" in the bottom left, which is needed to get the help flow turned on again once you have consumed it.

A full implementation should also provide the means to turn this off and on via settings (and with it, persistence using OGS `data` rather than inbuilt `storage`).

Upon loading, you should see:

Screen Shot 2022-07-29 at 1 26 13 pm (2)

... and if you follow the prompts you should end up at:

Screen Shot 2022-07-29 at 1 26 48 pm (2)

You can also dismiss the flow at any time using the popup's dismiss button.

You  can get back to the beginning by pressing the reset button (and possibly reloading the page and pressing it again grrr ... something about strict mode I think).

The `HelpItems` in this flow range from "the simplest possible" (`profile-edit-link`)  to "more tricky situations" (`account-settings-button`).

The latter is tricky because the target element on the page is generated inside a loop, so the ref for it has to be injected as part of that.   In addition, the target is not a refable item, so it had to have ref forwarding added to it ... this shows that this can be done.

Note that this PR is on top of #1946, which makes it appear bigger (in diff line count) than it actually is.

The way to approach the code for the help flow is probably to look in `HelpFlows`, see a `HelpItem`, note it's `target`, then find where that is registered and see how it is used from there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
+        "react-dynamic-help": "latest",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "^0.15.0",
+        "react-dynamic-help": "^0.15.1",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "latest",
+        "react-dynamic-help": "^0.15.0",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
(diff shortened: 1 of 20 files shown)


GreenAsJade:
rdh is now attempting to put its popups onto the screen if the screensize is too small for the app-requested position.

(aka mobile support)

anoek:
Very cool! I like how things are declared, nice work!

I noticed a couple of things to fix up before merge

* The `help-flow-container` causes a huge page height div at the end of all pages

image

* The reset rdh thing should be removed before merge, it might be good to have that somewhere in settings though. 

* The 'x' thing is a bit small to click, I'd either make that bigger or make it a click-the-bubble-to-dismiss kind of thing (or keep it small and make the bubble or target tap area a little bigger?)

* I suspect it'd look a notch better if the background color of the bubbles was lightened so that the contrast between the black text and the background was increased (bonus points for darkening it / switching to light text for dark themes). The other possible improvements there would be to increase the padding, especially the left/right padding, and possibly add a drop shadow if that looks good.

* The automatic positioning on the settings page is a little wonky:

image

image

GreenAsJade:
That huge div turns out to be totally unnecessary, so I'll remove that. 

I'll see what I can do about the bubbles.

I've realised that my goal of keeping PR diffs manageable still isn't really realised unless we have a feature branch to merge "done" ones into.   I'm almost tempted to do that now.

The remainder of your comments are addressed in the subsequent PRs :)

GreenAsJade:
This PR will be replaced by one for the feature branch 🤞

benjaminpjones:
Hey, sorry for the wait on the review.  Really great work on the RDH @GreenAsJade!  I don't really have much to add - y'all covered the two pieces of feedback I had (contrast, layout on small screens).

For the contrast thing, here's a tool I like to use to get an objective check: https://webaim.org/resources/contrastchecker/

PR #1957, opened 2022-08-04
New nav bar updates and fallout from the changes

anoek:
This PR addresses a number of feedback points that came back from the community. Some of the feedback items, such as removing the tournament, group, and ladders from the nav bar have some fallout (namely that we now need a place to store those, since we're removing them from the home screen).


BHydden:
Awesome changes ❤️ I think putting personal ladders on ladders page was something I tried to implement in the past so exciting to see it coming in now ❤️

PR #1958, opened 2022-08-05
Use words instead of 'x' to dismiss popups

### LINKED ISSUE #1955: Have "Don't show these" and "Skip" options for RDH helps
Interactive help like we're adding is one of those things some folks find really helpful and some folks really hate. The two options we should probably add in before releasing is to be able to click a "Don't show these", and "Skip" - where "Skip" skips that entire chain, so wont' show progression of the set password help for instance.

image

### LINKED ISSUE #1956: Blank RDH help settings lines are appearing
I'm seeing this on my help settings page:

image

A quick look at the code looked like we're iterating over some some stuff that might be pulled from settings or something, so it might be that I just have stale settings, but we should probably figure it out so that as we muck with identifiers and help strings as time goes on, we don't get ghost settings like this showing up.

GreenAsJade:
Fixes #1955, #1956

## Proposed Changes

 - Put "links" into the popup instead of 'x', as per #1955 
 - Allow app to provide translations for the words, via props on HelpProvider
 - Move to react-dynamic-help 1.1.2, to fix #1956

As a consequence of "turning off help" being an event that comes from an RDH popup, the settings page logic for displaying that is updated as well.   The previous implementation was a wrong direction, duplicating the preference in OGS preferences and RDH preference storage.  So that's fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "1.1.1",
+        "react-dynamic-help": "1.1.2",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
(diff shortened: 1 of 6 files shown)


anoek:
Yay, thanks :)

Issue #247, opened 2017-04-26
Inviting people to groups should be better


Nazgand:
When you invite the first person, you get feedback in that it says 'invited', which is a noticeable change.
But when you invite the second person, there is no noticeable change. I propose that it should instead say 'invited username', so that it is obvious that the button did not break after the first invite. Low priority, of course.

benjaminpjones:
## Good First Issue

### Code Points:
- [Group.invitePlayer](https://github.com/online-go/online-go.com/blob/01b152b88538a46dbba2e6eec1ad2185e78e6b48/src/views/Group/Group.tsx#L437)

### Prerequisite knowledge:
- Some familiarity with [asyncronous JavaScript](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous)
- Familiarity with JavaScript functions, since you will need to pull in the invited player's username into the callback.

### Things to check:
- Make sure new code is internationalized.  [`translate._()`](https://github.com/online-go/online-go.com/blob/c0c6093a0343041bb75ca8843cadc5121f3c3570/src/lib/translate.ts#L350) translates the text in the code right now.  You may need to use [`translate.interpolate()`](https://github.com/online-go/online-go.com/blob/c0c6093a0343041bb75ca8843cadc5121f3c3570/src/lib/translate.ts#L324) to handle arbitrary usernames.
- Data races (if you invite two players in quick succession, will the correct user be displayed?)

If you would like to tackle this issue and have questions, feel free to leave a comment here!

pdg137:
Tournaments have the same issue.

PR #1960, opened 2022-08-07
Show username of invited player

turtoise0:
Previously, when inviting multiple players, it's hard to tell which player "Player invited" refers to.

## Proposed Changes

  - Show username of invited player in group and tournament invite result message.


benjaminpjones:
Awesome stuff. I believe this closes #247

anoek:
Thanks for the patch!

PR #1961, opened 2022-08-07
Don't autocapitalize username in player invite input

turtoise0:
## Proposed Changes

  - Turn off autocapitalize in PlayerAutocomplete input.


anoek:
Thanks!

PR #1963, opened 2022-08-09
Add profile fields to Account Settings page

anoek:
This patch finally adds the username, icon, country, website, name fields to the "Account Settings" page.

The fields are still editable on the profile pages as well, at least for the time being.

---

@GreenAsJade I removed a line

```
const { ref: profileLink } = registerTargetItem("profile-edit-link"); // cleared on Profile page
```
that was related to the help stuff you've been working on. I *think* this was not being used anymore, and would be obviated anyways, but I wanted to let you know in case I missed something.

### DIFF
--- a/src/components/InfoBall/InfoBall.styl
+++ b/src/components/InfoBall/InfoBall.styl
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2012-2022  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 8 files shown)


benjaminpjones:
I think we lost the ["You can only change your username every 30 days" warning](https://github.com/online-go/online-go.com/blob/0f88621b0abd0ef0d2efa68793f334f6b9c3d8d1/src/views/User/AvatarCard.tsx#L111).

Overall, I really like this addition to settings, makes a lot of sense :)

anoek:
Good catch!

GreenAsJade:
I agree that the ref to the profile link is obviated now.

I think that both places where the icon appears, it should be half the size.

This would allow for some left margin on the right div where the profile appearance editing is now.    I put 10rem and it still was barely enough to get it away from the text to the left - I wonder if a kind of "space between" layout would work better>?


Screen Shot 2022-08-10 at 8 40 23 am (2)


(Obviously there are margins/alignment to be fixed after resizing the avatars, this is just a brute force shrink to test it out)

GreenAsJade:
Also, in the screenshot above, there is something odd about "name" ... it has "First" above it, as if there was going to be a field for "Last" ?

anoek:
Those are implemented as placeholders, so if you don't have a last name filled in, it looks like this:

image

which may not be the best, it's carryover from how it was implemented on the profile page.. we do have room for field names here though I suppose

GreenAsJade:
Ah makes sense - mine only looks odd because I put a name there that isn't in fact my last name :D   

Although - I think it also looks odd because it is not at all clear that the "First" is an empty input field.    Unlike your screenshot, in mine, there is no visual indication that I am supposed to type where it says "First" ...

PR #1965, opened 2022-08-10
Reset help flows when the person logs out

GreenAsJade:
Fixes [activated flows from a  logged out user flowing into next user to log in](https://forums.online-go.com/t/preview-beta-test-popup-help-for-new-users-challenge-link-guests/44310/3?u=greenasjade)

## Proposed Changes

Reset the help flows upon logout (specifically, when it's detected that the config.user is anonymous).

Needs rdh 2.1.5, for lots of stuff.

Needs https://github.com/online-go/ogs-node/pull/17 for "remote_storage_sync_complete"

### DIFF
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -115,6 +115,8 @@ const defaults: Partial<DataSchema> = {};
 const store: Partial<DataSchema> = {};
 const event_emitter = new TypedEventEmitter<Events>();
 
+//  Note that as well as "without emit", this is "without remote storage" as well.
+// (you cant set-remote-storage-without-emit)
 export function setWithoutEmit<KeyT extends Extract<keyof DataSchema, string>>(
     key: KeyT,
     value: DataSchema[KeyT] | undefined,
@@ -380,7 +382,7 @@ interface RemoteKV {
     modified?: string;
 }
 
-let remote_store: { [key: string]: RemoteKV } = {};
+let remote_store: { [key in keyof DataSchema]?: RemoteKV } = {};
 let wal: { [key: string]: { key: string; value?: any; replication: Replication } } = {};
 let wal_currently_processing: { [k: string]: boolean } = {};
 let last_modified = "2000-01-01T00:00:00.000Z";
@@ -490,6 +492,7 @@ function _process_write_ahead_log(user_id: number): void {
                 // just dump this attempt from our wal so the client doesn't
                 // keep trying to send updates to the server which will never
                 // succeed.
+                console.error("... couldn't retry!");
             }
 
             if (wal[data_key].value !== kv.value || wal[data_key].replication !== kv.replication) {
@@ -574,6 +577,7 @@ ITC.register("remote_storage/sync_needed", () => {
 // for each key that's been updated since the timestamp we sent
 socket.on("remote_storage/update", (row: RemoteKV) => {
     const user = store["config.user"];
+
     if (!user || user.anonymous) {
         console.error(
             "User is not logged in but received remote_storage/update for some reason, ignoring",
@@ -645,6 +649,8 @@ function load_from_local_storage_and_sync() {
                 const key = full_key.substr(store_prefix.length);
                 try {
                     remote_store[key] = JSON.parse(localStorage.getItem(full_key)) as RemoteKV;
+                    store[key] = remote_store[key].value;
+                    emitForKey(key as keyof DataSchema);
                 } catch (e) {
                     console.error(`Error loading remote storage key ${full_key}, removing`, e);
                     localStorage.removeItem(full_key);
@@ -675,5 +681,4 @@ function load_from_local_storage_and_sync() {
     }
 }
 
-load_from_local_storage_and_sync();
 watch("config.user", load_from_local_storage_and_sync);
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -527,6 +527,7 @@ let currently_synchronizing = false;
 let need_another_synchronization_call = false;
 
 function remote_sync() {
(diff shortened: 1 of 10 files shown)


anoek:
This works for the guest specific help, but if we have help that is enabled for non guests, I think we might want to remember what users have seen it perhaps?

Some folks habitually log out of everything, or have a shared computer, so when they login the next day, we don't want them to have to go through whatever help stuff again I reckon.

GreenAsJade:
Yeah I think that is a problem, and this isn't the solution darn :)

I have to separate out the difference between "has seen it" and "is actively visible now".

GreenAsJade:
Latest commit does:

 - Update `storage` as well as `remote_storage` during `reload_from_local_storage_and_sync()`
 - Emit for those updates
 - Don't call `r_f_l_s_a_s` twice at the beginning
 - type `remote_storage` more strictly
 - move to RDH 2.1.0, which properly supports user-specific client-supplied storage
 - supply storage to RDH 
 - Add "seen" status to HelpSettings
 - Make HelpFlows properly update RDH when the user changes
 - Tidy up exv6 help flow in line with recent OGS change

src/lib/data.ts
```diff
@@ -645,6 +649,8 @@ function load_from_local_storage_and_sync() {
                 const key = full_key.substr(store_prefix.length);
                 try {
                     remote_store[key] = JSON.parse(localStorage.getItem(full_key)) as RemoteKV;
+                    store[key] = remote_store[key].value;
```
anoek, line 652: Does this break `LOCAL_OVERWRITES_REMOTE` ? At a glance it seems like it might, but I'm just waking up over here.

src/lib/data.ts
```diff
@@ -675,5 +681,4 @@ function load_from_local_storage_and_sync() {
     }
 }
 
-load_from_local_storage_and_sync();
```
anoek, line 678: I want to say this was here for a reason, likely to solve a race condition. I probably should have put a comment..

src/lib/data.ts
GreenAsJade: Hmm - yeah, maybe it should not be done in that case: definitely needs thought

GreenAsJade: ( but then - no-one is using LOCAL_OVERWRITES_REMOTE, so....  ;) :o )

GreenAsJade: ... yeah, it definitely should not be done if LOCAL_OVERWRITRS_REMOTE.

src/lib/data.ts
GreenAsJade: I have logged the heck out of it, I think that actually it is not needed, but it doesn't hurt because it does nothing.

In fact, because it does nothing, the race that I thought I found before does not exist: although this is called here, it is short-circuited by loaded_user_id being set the first time:

```
function load_from_local_storage_and_sync() {
    const user = store["config.user"];
   //...
    if (loaded_user_id === user.id) {
        return;
    }
```

This results in the first `remote_sync` not happening till rather later when the socket connects.

src/lib/data.ts
```diff
@@ -645,6 +649,10 @@ function load_from_local_storage_and_sync() {
                 const key = full_key.substr(store_prefix.length);
                 try {
                     remote_store[key] = JSON.parse(localStorage.getItem(full_key)) as RemoteKV;
+                    if (remote_store[key].replication !== Replication.LOCAL_OVERWRITES_REMOTE) {
```
anoek, line 652: I think this needs to be `remote_store[key].replication === Replication.REMOTE_OVERWRITES_LOCAL` specifically doesn't it? We wouldn't want this to trigger with `REMOTE_ONLY`

GreenAsJade: Hmm - you are right again!

I did think about that, and I thought "oh, remote only is about when we are setting the value", just writing it out there, but of course it would come bouncing back through here.

What is the rationale for REMOTE_ONLY?

GreenAsJade: ... if we don't load the remote value here, when do we ever load it for this account?

anoek: We do load it, it's set in `remote_store`, which is what is used if `store` isn't set. When `get` is called, it first checks `store`, if nothing is in there it checks `remote_store`, if nothing is in there it uses the default value.

The intent with `REMOTE_ONLY` was to be able to effectively set defaults, so like default theme is black, but don't force other devices to change if they were set to white locally. Come to thing of it, I'm not sure this is any different than `LOCAL_OVERWRITES_REMOTE`.. I'd have to ponder on that a bit more, I've either forgotten the distinction, or erroneously thought there was a distinction before. But regardless, seems like we should be exact with this case.

GreenAsJade: Oh, cunning, right!

It's not the same as LOCAL_OVERRIDES_REMOTE, because it means you can say "if I ever add more devices, I want them to have this value that's different to any value currently set on my devices", wheras with LOCAL_OVERRIDES_REMOTE you'd get that value on your local device as a result of this action.

It's hard to imagine a direct case for this (a user thinking along those lines) but maybe it has an application where we want to do something tricky on their behalf...

anoek:
Hmm, so now when I toggle 

image

and refresh it doesn't remember me toggling it

GreenAsJade:
Man... I think I've been looking at this too long and am not seeing it anymore.    Yep, darn.

GreenAsJade:
*** as noted in the edited description, this needs https://github.com/online-go/ogs-node/pull/17

Waaay better with that simple event :)

anoek:
Thanks for working through all of that :)

PR #1967, opened 2022-08-15
Get rid of the  conditional moves button in the dock in rengo games

GreenAsJade:
Fixes [players messing things up by seeing what happens when you click the greyed out conditional moves button](https://forums.online-go.com/t/rengo-status/40415/1120?u=greenasjade)

## Proposed Changes

  Completely remove conditional moves button from Game Dock in rengo games.

(Note: previous implementation (probably by me) assumed that className `disabled` disables the link, but of course it doesn't.   In the nearby cases in the code, the greyed link gets clicked and something down stream checks and decides not to do it.  This is not practical in the case of conditional moves)


anoek:
Thanks for the fix!

PR #1976, opened 2022-08-23
Update legal.tsx 

Sofia-mal:
Another silly thing, I found it strange that some sentences ended with a period and some didn't. 
Maybe ";" could be better but I didn't feel like twisting things too much. Maybe it's not even wanted. 

(Again two empty commits, I did something wrong and can't get rid of them.  Sorry)


anoek:
Thanks :)

PR #1978, opened 2022-08-26
Put the game name in quotes so it is clear that it is a name

GreenAsJade:
Fixes confusion about what the text is in the top left of an invite card

## Proposed Changes

Put the game name in quotes:

Screen Shot 2022-08-26 at 10 37 43 am (3)

### DIFF
--- a/src/components/ChallengeLinkButton/ChallengeLinkButton.tsx
+++ b/src/components/ChallengeLinkButton/ChallengeLinkButton.tsx
@@ -39,7 +39,7 @@ export function ChallengeLinkButton(props: ChallengeLinkButtonProps): JSX.Elemen
             onClick={(event) => copyChallengeLinkURL(event.target as HTMLElement, props.uuid)}
             className={full_class}
         >
-            <i className="fa fa-share" />
+            <i className="fa fa-share-square" />
         </button>
     );
 }
@@ -88,7 +88,7 @@ export function copyChallengeLinkURL(ack_target: HTMLElement, uuid: string): voi
 export function showChallengeLink(challenge_link: string, target: HTMLElement): void {
     popover({
         elt: (
-            <div className="challenge- link - copy">
+            <div className="challenge-link-copy">
                 {pgettext(
                     "This is the label for a link (URL) that they created",
                     "Invitation link:",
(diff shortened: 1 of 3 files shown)


benjaminpjones:
Will it be weird if the InviteList has quotes on the title and the ChallengeList doesn't? Currently they look relatively similar stylistically:

Screenshot_20220825-220109_Firefox

GreenAsJade:
Yeah I guess that's wierd.

The problem is actually the "copy the link for sharing" button.  With the game name next to it, it makes it look like it's a label for that button, I think.

Darned if a better way is occuring to me right now.

anoek:
I think the quotes are good, but yeah lets make it consistent and do quotes for the challenges too?

benjaminpjones:
>Darned if a better way is occuring to me right now.

Not sure if this addresses the problem (tbh I didn't notice the issue myself), but just to throw a couple of options out:

Screen Shot 2022-08-26 at 1 00 12 PM

GreenAsJade:
I like the Copy Link, except ... does that work on the Play page?  

The FAB is less than ideal, because that part of the interface is for "doing things to the challenge itself".

benjaminpjones:
Ah makes sense on the "FAB". And are the invite cards on the play page? I don't think that copy link button would fit on the rows as is (although something with a label might be more discoverable)

Either way, the quotes work for me too, just wanted to throw some stuff out there.

GreenAsJade:
The invite card isn't on the Play page, but the button to copy the challenge link is.  So ... whatever we use here, we should use it in that (right most column of challenge table, for own-games) as well!

anoek:
In the above example I like the aesthetics of the fab version better, it's a clear actionable thing so I think it's fine there. I think the link icon is fine to use in the table too, though I wouldn't want a full fab there I don't think, although I also think the full copy link text and icon might be appropriate for the table version too. So long as the icon is consistent between the two I think that would be fine from a ux perspective? Might have to see it in action to see if that's a bad idea or not, but in my minds eye it seems like it'd be fine?

GreenAsJade:
I tried various options, and ended up thinking "make the challenge list match with quotes" was the simplest and tidiest.

anoek:
Thanks!

PR #1979, opened 2022-08-26
Converted RengoTeamManagementPane to a function component.

GreenAsJade:
Fixes future ugliness using context in class based components: it's easier to add rdh to function components.

## Proposed Changes

Convert to function component.


anoek:
Awesome, thanks!

PR #1980, opened 2022-08-26
Converted RengoManagementPane to a function component

GreenAsJade:
Fixes future ugliness when adding rdh context

## Proposed Changes

Convert RengoManagementPane to be a function component.

### DIFF
--- a/src/components/RengoManagementPane/RengoManagementPane.tsx
+++ b/src/components/RengoManagementPane/RengoManagementPane.tsx
@@ -1,18 +1,18 @@
 /*
  * Copyright (C) 2012-2017  Online-Go.com
  *
- * program is free software: you can redistribute it and/or modify
+ * THis program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as
  * published by the Free Software Foundation, either version 3 of the
  * License, or (at your option) any later version.
  *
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Note: adding rdh to a class based component is not _that_ bad, I just prefer to have one way of doing it if possible!

src/components/RengoManagementPane/RengoManagementPane.tsx
```diff
@@ -1,18 +1,18 @@
 /*
  * Copyright (C) 2012-2017  Online-Go.com
  *
- * This program is free software: you can redistribute it and/or modify
```
anoek, line 4: I reckon we should retain the This's in the legal preamble :)

GreenAsJade: Whoa, how on earth did I delete that line!!?

GreenAsJade: Oh LOL ... I deleted all the "this" :D

BHydden: Good old replace all haha wHaT cOuLd Go WrOnG 🤪🫠😂

anoek:
Thanks, even though you now spell it as `THis` for one case :rofl:

PR #1982, opened 2022-08-28
Help for rengo guest

GreenAsJade:
Fixes guests on challenge links to rengo games not knowing what's going on

## Proposed Changes

Add a help flow item that tells rengo challenge link guests that their game will commence in due course.

And also trigger the password help flow for these guests.

(And pull in rdh 2.3.0 that has better input validation)

** requires #1981

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "2.3.0",
+        "react-dynamic-help": "2.3.1",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "2.4.1",
+        "react-dynamic-help": "2.4.2",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
         "markdown-it": "^12.3.2",
         "moment": "^2.29.4",
         "moment-duration-format": "^2.3.2",
-        "moment-timezone": "^0.5.31",
+        "moment-timezone": "^0.5.35",
         "performance-now": "^2.1.0",
         "prop-types": "^15.5.0",
         "query-string": "^5.0.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "2.4.0",
+        "react-dynamic-help": "2.4.1",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "^2.1.7",
+        "react-dynamic-help": "2.2.0",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "react-datetime": "^3.0.4",
         "react-dom": "^18.0.0",
         "react-dropzone": "^12.0.5",
-        "react-dynamic-help": "2.3.1",
+        "react-dynamic-help": "2.4.0",
         "react-linkify": "^1.0.0-alpha",
         "react-number-format": "^3.0.2",
         "react-resize-detector": "^7.1.1",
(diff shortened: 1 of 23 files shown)


anoek:
As my normal user, when I go to `/settings/help` I get a red screen and this on the console:

image

GreenAsJade:
😮 

Hmm - that falls in the category of "this can't happen" 🤣 

It seems to mean that a flow is being registered with an undefined id.

Could you turn on rdh debug and send me the console log that leads up to this?

data.set("debug-dynamic-help", true)

GreenAsJade:
I've been unable to reproduce this.  I've also been unable to set up a new user that does not have any prior `data`.   Somehow the combination of browser memory and local storage is defeating me - despite whatever I've tried so far (including dumping localk storage in guest profiles, different browsers (which all have had prior access to OGS) etc) I always end up with dark theme!

I'm sure there's something simple that I'm overlooking ... I wonder what it is.

anoek:
Here's the console export: [console-export-2022-8-30_21-6-1.txt](https://github.com/online-go/online-go.com/files/9457994/console-export-2022-8-30_21-6-1.txt)

anoek:
image
image
image

And console pics if you find them easier to read

GreenAsJade:
I'd hoped that leaving this for a bit and coming back would make it clearer, but the above sequence of events is still in the "that can't actually happen" list ... either that or I don't properly understand what

`TypeError: _a[flowID] undefined` means   

(or both).

I could just code against flowID being undefined and move on, but it's unnerving how it could get that way.

Could you try again with the added debug?   Could you also ask the debug tools to show us what exact source code line it is - what precisely is the _a that it is complaining about...

Thanks!

benjaminpjones:
> I could just code against flowID being undefined and move on, but it's unnerving how it could get that way.

I don't believe `flowID` being undefined is the problem here.  The problem is that _a[flowID] is undefined.  This could be the case if you call `_a["i-never-set-this-key"]`.  Furthermore, you won't get this error just by indexing `_a`, you will only see the error once you try to do something with the undefined output.  Consider the following code:

```
const m = {}

// okay, just prints undefined
console.log(m[undefined])
// okay, just prints undefined
console.log(m["i-never-set-this-key"])
// error, tries to access a property of an undefined value
console.log(m["i-never-set-this-key"].seen)
```

> Could you try again with the added debug? Could you also ask the debug tools to show us what exact source code line it is - what precisely is the _a that it is complaining about...

Not a line number, but the error message does say what functions it's in: GetFlowInfo > info, which does have a lot of `arr[flowID].some_member`.  Maybe check if the keys of `registeredFlows` and `systemState.flows` matches?

https://github.com/GreenAsJade/react-dynamic-help/blob/5297c41325032ff80d0b50ebdaa878992eae6089/src/components/HelpController.tsx#L305-L313

GreenAsJade:
Thanks for taking a look!     I did trace the message back to `getFlowInfo`  ... what I was hoping @anoek might be able to see is something more specific in the transpiled js.

The thing that has thrown me is that in your third example, "error", the message is 

"Cannot read properties of undefined (reading 'seen')."

That is a very easy message to debug: it says what you said in the comment.

But ... what is `TypeError: _a[flowID] undefined` trying to say?  And why is it different?

I guess it is conceivable that, as you say it is simply complaining that   `this.systemState.flows[flowId]` is undefined.   

I didn't go down that path because:

1) I thought it would say "Cannot read properties of undefined (reading 'description')" in that case
2) It appears that all the flows are registered (in the trace) 

And more fundamentally, in order for there to be an entry in the array that is being mapped, addHelpFlow has to have run, which means those things are set.

🤔    Though now that I type that... let me go see about something 😊

GreenAsJade:
... ha: I love the "explain why this bug is impossible" debugging technique  😊  .   Darn storage initialisation _still_.   😭

benjaminpjones:
>"Cannot read properties of undefined (reading 'seen')."

Just tell @anoek to get back on Chrome 😋

jk, but seriously the reason you got a different error message is because you're not in Firefox.

GreenAsJade:
rocky-and-bullwinkle-moose-and-squirrel

GreenAsJade:
> > "Cannot read properties of undefined (reading 'seen')."
> 
> Just tell @anoek to get back on Chrome 😋
> 
> jk, but seriously the reason you got a different error message is because you're not in Firefox.

What is this fiery rodent you speak of, hmm ??

anoek:
Yay, crash is fixed!

I'm seeing some funky alignment / jumping issues though:

image

BHydden:
> fiery rodent

Umm... I'm fairly sure foxes aren't rodents 🤔

GreenAsJade:
> I'm seeing some funky alignment / jumping issues though:

I may need to rename myself to Bullwinkle!   😭    😁

GreenAsJade:
> I'm seeing some funky alignment / jumping issues though

Was there some scrolling involved?   I can't tell from the screenshot whether you're at the top of the page, and the jumpiness was "immediate", or whether this happened after you scrolled the blue box into view?

GreenAsJade:
... oh, I'm starting to see things though: at least parts of this page are rerendering on socket events, aren't they?   So that's going to be the trigger for a jumpy thing to happen for sure 🤔

GreenAsJade:
I reproduced it.   The refresh that causes it seems related to the table freeze and unfreeze associated with mouse movement, rather than the socked, FWIW.

GreenAsJade:
I don't even dare say "this time for sure", this time...

3295580_orig

anoek:
Yay, looks great! Thanks for persisting!

GreenAsJade:
http://www.tzr.io/yarn-clip/e4f18b33-5131-4ab1-8bd5-6897a5579bab

So that only leaves the question of the rodent.    This is the bad comedian's dilemma isn't it?   Whether to explain the joke, because it was so bad it needs explaining, so that at least it's understood ... or to leave it forever a question mark.

Or maybe just provide a [hint](https://tropedia.fandom.com/wiki/What_Is_This_X_You_Speak_Of?#:~:text=Also%20seen%20as%20%22What%20Is,the%20context%20of%20the%20discussion.) of what was behind it, and leave it at that ...

Issue #1971, opened 2022-08-17
Chat notification settings unintuitive


benjaminpjones:
Context: https://forums.online-go.com/t/disabling-the-new-chat-notification/28104/18

A couple issues with the chat settings widget:

<img width="153" alt="Screen Shot 2022-08-17 at 10 38 48 AM" src="https://user-images.githubusercontent.com/25233703/185206110-ad1226bc-1266-4f33-a924-ab073d0d25c6.png">

- [x] The layout is such that "unfollow mentioned" is not fully visible, so it gets truncated to "unfollow..." (Fixed in #1986)
- [ ] Negative verbs (**un**follow vs follow) obscure the meaning of actions.

Proposal:

Borrow UI from chat apps Discord and Slack by specifying which notifications we want to **follow** (not unfollow) and changing this to a radio button (to communicate that you can only select one):

<img alt="Slack" src="https://ogs-forums.s3.dualstack.us-east-1.amazonaws.com/original/3X/a/2/a2822c045accca7f08849ef62b7cba9e2aea2abc.jpeg" width="250"/>
<img alt="Discord" src="https://ogs-forums.s3.dualstack.us-east-1.amazonaws.com/original/3X/8/7/87a3148f31eea5f6afb4090b59576c6fc066b5c0.jpeg" width="250"/>

hakanyilmazz:
Hi Benjamin,
I fixed that problem (The layout is such that "unfollow mentioned" is not fully visible, so it gets truncated to "unfollow..."). You should see it.
[Pull Request Link](https://github.com/online-go/online-go.com/pull/1986)

PR #1986, opened 2022-09-04
fix: chat notification settings unintuitive

hakanyilmazz:
Problem: The layout is such that "unfollow mentioned" is not fully visible, so it gets truncated to "unfollow..." (https://github.com/online-go/online-go.com/issues/1971)

Result (Fixed): "Unfollow Mentioned" is visible now.

onlinego

### DIFF
--- a/src/components/Chat/ChatDetails.styl
+++ b/src/components/Chat/ChatDetails.styl
@@ -32,7 +32,7 @@
         justify-content: space-between;
 
         > button {
-            min-width: 98%;
+            min-width: 100%;
             text-align: left;
             .fa, .ogs-goban {
                 padding-right: 0.1rem;
--- a/src/components/Chat/ChatDetails.tsx
+++ b/src/components/Chat/ChatDetails.tsx
@@ -171,8 +171,8 @@ export class ChatDetails extends React.PureComponent<ChatDetailsProperties, Chat
                             <i className="fa fa-comment" />
                             {" " +
                                 (this.state.notify_mentioned
-                                    ? _("unfollow mentioned")
-                                    : _("follow mentioned"))}
+                                    ? _("Unfollow Mentioned")
+                                    : _("Follow Mentioned"))}
                         </button>
                     )}
                     {this.state.subscribable && (
@@ -183,8 +183,8 @@ export class ChatDetails extends React.PureComponent<ChatDetailsProperties, Chat
                             <i className="fa fa-comment" />
                             {" " +
                                 (this.state.notify_unread
-                                    ? _("unfollow unread")
-                                    : _("follow unread"))}
+                                    ? _("Unfollow Unread")
+                                    : _("Follow Unread"))}
                         </button>
                     )}
                     {this.props.partFunc ? (
--- a/src/global_styl/global.styl
+++ b/src/global_styl/global.styl
@@ -443,7 +443,7 @@ button, a.btn, .btn {
     padding-top: 0.1em;
     padding-bottom: 0.1em;
     border-radius: 0.3rem;
-    height: 2em;
+    height: fit-content;
     vertical-align: middle;
     box-sizing: border-box;
     line-height: font-size-normal;


GreenAsJade:
What do those buttons even mean?   I was thinking "these should really be pgettext, so the translators have a reasonable hint at what this is", then I realised that I don't even know what it is talking about.

benjaminpjones:
@GreenAsJade see the referenced issue (#1971 ) or the forums thread where flovo's explanation in the forums (https://forums.online-go.com/t/disabling-the-new-chat-notification/28104/20)

GreenAsJade:
@benjaminpjones  Right... so this PR fixes the obscured text ... yay - nice job, thanks, which is bullet one of the PR 👍 

I guess if a followup fix for the verbs is coming, then you could argue that `pgettext` for these words is a bit academic, _but_  those isolated words suggested for the full fix will even-more need `pgettext` context, so why not add it now ;)

(One answer to this question is that if you add it now, you will force a retranslation, which is Not To Be Done Lightly (TM) )

src/components/Chat/ChatDetails.styl
```diff
@@ -32,7 +32,7 @@
         justify-content: space-between;
 
         > button {
-            min-width: 98%;
+            min-width: 100%;
```
GreenAsJade, line 35: @BHydden   It'd be fascinating to know why you set this to 98%.    When I see a number that specific, I find myself wondering "what was special about the missing 2%, is it OK to change this?"

BHydden: I have no conscious recollection of that commit or indeed ever working on that component... 🤔

At a guess I set it to 98 because the previous setting was 48 and so I just changed the leading character? 🤷‍♂️

GreenAsJade: Hah I can imagine that too :)

anoek:
Good stuff, though I think we need to make the overall box a little wider, I'm getting word wrapping now:

image

anoek:
Also, I concur, if we're going to change the text, we should switch to using pgettext and give the translators some context

src/components/Chat/ChatDetails.styl
hakanyilmazz: Hey folks, i hope the issue was solved, i think online-go website should be recreated, there are too many bugs in the website. Or maybe we should consider about recreating this online-go with flutter? Flutter is really cool technology.

hakanyilmazz:
> What do those buttons even mean? I was thinking "these should really be pgettext, so the translators have a reasonable hint at what this is", then I realised that I don't even know what it is talking about.

I will work on it. But i don't promise. It can be really hard issue. Because, if i create a new button text, then i may need to create new logic for them.

hakanyilmazz:
> 

Is this my committed code? omg, it was

hakanyilmazz:
> Good stuff, though I think we need to make the overall box a little wider, I'm getting word wrapping now:
> 
> image

Is this my committed code? omg it was working. :(

benjaminpjones:
> I'm getting word wrapping now

> Is this my committed code? omg it was working. :(

Note: this is an improvement to previous behavior. Before it would just cut off the second word 😂

hakanyilmazz:
> > I'm getting word wrapping now
> 
> > Is this my committed code? omg it was working. :(
> 
> Note: this is an improvement to previous behavior. Before it would just cut off the second word 😂

hahaha my man 😎😂

hakanyilmazz:
Also when we load the chat page, chat messages aren't loading. If we press f5 key then chat messages are loading. Why? Should we open a new issue for this problem?

hakanyilmazz:
> Also when we load the chat page, chat messages aren't loading. If we press f5 key then chat messages are loading. Why? Should we open a new issue for this problem?

GreenAsJade:
> Is this my committed code? omg it was working. :(

Story of my life!!   ([firefox maybe](https://github.com/online-go/online-go.com/pull/1982#issuecomment-1238065678)? ;) )

GreenAsJade:
> I will work on it. But i don't promise. It can be really hard issue. Because, if i create a new button text, then i may need to create new logic for them.

Yeah - I think that would be a separate PR.  At the time I wrote my comment, I hadn't realised that "fix the layout" and "fix the meaning of the buttons" were already identified as two separate issues (thanks to benjito for pointing that out)

anoek:
Had a look into this issue, though it struck me that the button system we were using here didn't really follow the same patterns we were using elsewhere, I'm considering obviating this PR with an alternative PR / look, see #1993 which makes things look like:

hakanyilmazz:
My solution was temporary solution. Yeah, you're right. Nice job anoek 🎉

PR #1987, opened 2022-09-04
Keep page-controls visible at all times

turtoise0:
## Proposed Changes

Previously, the page controls would move with the PaginatedTable when you scrolled.
Now, only the table entries move so the page controls are always visible.

I added a div around the table and applied the css there, because  `overflow-x: auto` didn't seem to have an effect on tables.

My workaround is based on this: https://stackoverflow.com/q/8706503

### DIFF
--- a/src/components/PaginatedTable/PaginatedTable.styl
+++ b/src/components/PaginatedTable/PaginatedTable.styl
@@ -18,19 +18,23 @@
 .PaginatedTable {
     position: relative;
     max-width: 100vw;
-    overflow-x: auto;
-    table {
-        width: 100%;
-        max-width: 100vw;
-        border-collapse: collapse;
-        tr {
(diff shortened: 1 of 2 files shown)


anoek:
On the surface this looks fine and I have no objections to merging, but I'd like to see the original problem and the fix - where were you seeing the problematic behavior initially?

turtoise0:
Before:

https://user-images.githubusercontent.com/99894371/188838796-92c1c67c-ddbc-4511-b8f9-56f25b25a963.mp4

After:

https://user-images.githubusercontent.com/99894371/188838783-ef4bc90c-3522-49c0-aa2c-aa46583ff645.mp4

anoek:
Awesome thanks for that, and thanks for the patch!

PR #1990, opened 2022-09-06
Re-try fetching pages after search/filter fails

### LINKED ISSUE #1717: Paginated Table won't update if on a greater page # than filter allows
**Describe the bug**
I was looking at the Game History and tried to filter on a player, the table was not updating.  Realized it's because I was on a page that was higher than the number of pages returned for that player.

**To Reproduce**
1. Go to https://beta.online-go.com/player/1420/
2. View Game histor
3. Go to page 3/3
4. Try to filter by opponent = GnuGo
5. Notice the table doesn't change

**Expected behavior**
Games shown should only be the player specified.

**Screenshots**
Screen Shot 2022-02-18 at 12 23 30 AM

turtoise0:
Fixes #1717

## Proposed Changes

  - Resend request, but with `page = 1` after 404 error.

### DIFF
--- a/src/components/PaginatedTable/PaginatedTable.tsx
+++ b/src/components/PaginatedTable/PaginatedTable.tsx
@@ -168,6 +168,9 @@ function _PaginatedTable<RawEntryT = any, GroomedEntryT = RawEntryT>(
             })
             .catch((err) => {
                 console.error(err);
+                if (err.status === 404) {
+                    setPage(1);
+                }
                 setLoading(false);
                 if (load_again.current) {
                     load_again.current = false;


anoek:
Thanks!

GreenAsJade:
Out of curiosity, why isn't this an infinite loop if page = 1 is hard-failing with 404?

Is this simply - "yeah, that's not going to happen?"

I can dream up a situation where the URL supplied to PaginatedTable is just wrong for some error of context, so the whole table becomes 404 maybe?

BHydden:
Yeah let's deliberately add an infinite loop, what could go wrong 🤣

GreenAsJade:
I don't know for sure it is, nor if it matters ... it just caught my eye as maybe worth a double check.

benjaminpjones:
I don't believe this will infinite loop because calling `setPage(1)` won't trigger an update if `page` is already 1.

GreenAsJade:
Cool that makes sense 😅

PR #1991, opened 2022-09-06
Bugfixes needed to make the move tree variation saver take the correc…

GreenAsJade:
Fixes it not being clear that the move-tree variation-saver will honour the "chat type" selection.

## Proposed Changes

 - Move related css 'up' out of `GameChat` and into `Game` scope.
 - Fix interpolation of chat-type into variation-tree-saver class name

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -24,6 +24,18 @@
     bottom: env(safe-area-inset-bottom);
     right: 0;
     left: 0;
+
+    .hidden, .hidden.active, .hidden:focus, .hidden:hover {
+        themed background-color hidden-bg
+    }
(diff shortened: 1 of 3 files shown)


src/views/Game/Game.styl
```diff
@@ -35,6 +47,8 @@
     top: 0;
 }
 
+// Is this really intended as a global scope css directive?
+// I don't dare scope it for now, but it looks odd: if it's global, it should probably be in ogs.styl?
```
anoek, line 51: Looks like pretty old code, I think it's only used in this view and should be properly scoped

anoek:
Hah whoops, thanks for the fix!

PR #1993, opened 2022-09-08
Chat notification / highlight settings update to us checkmarks instead of buttons

anoek:
This PR adjusts the look of the chat notification settings. The original issue was brought to light in #1986, this PR revisits the look of the component entirely,

This PR makes the component look something like this:

image

### DIFF
--- a/src/components/Chat/ChatDetails.tsx
+++ b/src/components/Chat/ChatDetails.tsx
@@ -17,7 +17,7 @@
 
 import * as React from "react";
 import { browserHistory } from "ogsHistory";
-import { _, pgettext } from "translate";
+import { pgettext } from "translate";
 import { shouldOpenNewTab } from "misc";
 import { close_all_popovers } from "popover";
 import { close_friend_list } from "FriendList/close_friend_list";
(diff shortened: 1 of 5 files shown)


benjaminpjones:
Can probably mark #1971 complete.

I do have a couple questions about this though - 

- Why is the term "Highlight" instead of "Notify"?  It seems like "notify"/"notification" is the language used in both the PR description and translation hint, why would the UI be different?
- Is there any visible difference between these two states?
    - ✅ Highlight when mentioned ✅ Highlight on unread messages
    - 🔲 Highlight when mentioned ✅ Highlight on unread messages

src/components/Chat/ChatDetails.tsx
```diff
@@ -163,30 +162,40 @@ export class ChatDetails extends React.PureComponent<ChatDetailsProperties, Chat
...
+                        <div>
+                            <input
+                                type="checkbox"
+                                id="notify_mentioned"
+                                checked={this.state.notify_mentioned}
+                                onChange={this.toggleMentionNotification}
+                            />
+                            <label htmlFor="notify_mentioned">
+                                {pgettext(
+                                    "Don't notify on unread @user mentions in a chat channel",
```
benjaminpjones, line 176: Nit: I think having the hint be the opposite (don't notify) of the actual text (do notify) could lead to confusion for translators.

anoek:
I'm not wild about the verbiage "highlight" either, I feel like there's probably a better way to phrase that - however I did want to stay away from "notification" too as that's something that actively pops up somewhere, usually on the desktop, email, or phone, and since we use that word for those things elsewhere, I wanted to avoid confusion about what we're doing here which is simply lighting up the indicator and changing the color of the channel names, hence the word choice "highlight". 

That said I should update the pgettext, I'll do that rq.

anoek:
Also, looking at #1986 - I like the screenshots you posted better, I think we should change it to be a radio selector instead of this checkbox thing

benjaminpjones:
Makes sense about the highlight and notification thing difference, and I can't think of a better solution there so never mind haha

Thanks for taking a look at the issue!

PR #1994, opened 2022-09-09
Chat notification settings take 3

anoek:
This updates the chat notification settings once again to look more like how discord/slack present their settings. Also reverted the change of terminology from "highlight" to "notification", because we probably do want those to be actual notifications some day, so maybe it's best to just stick with that terminology.

New look:

image

@benjaminpjones

### DIFF
--- a/src/components/Chat/ChatDetails.tsx
+++ b/src/components/Chat/ChatDetails.tsx
@@ -35,66 +35,55 @@ interface ChatDetailsProperties {
     subscribable?: boolean;
 }
 
-interface ChatDetailsState {
-    channelId: string;
-    subscribable: boolean;
-    notify_unread: boolean;
-    notify_mentioned: boolean;
-}
+export function ChatDetails(props: ChatDetailsProperties): JSX.Element {
+    const channel = props.chatChannelId;
 
-export class ChatDetails extends React.PureComponent<ChatDetailsProperties, ChatDetailsState> {
-    constructor(props) {
-        super(props);
-        const channel = this.props.chatChannelId;
-        if (channel) {
-            this.state = {
-                channelId: channel,
-                subscribable: props.subscribable,
-                notify_unread: false,
-                notify_mentioned: false,
-            };
-        }
-    }
+    const [channelId, setChannelId] = React.useState(channel);
+    const [subscribable, setSubscribable] = React.useState(props.subscribable);
+    const [notify_unread, setNotifyUnread] = React.useState(false);
+    const [notify_mentioned, setNotifyMentioned] = React.useState(false);
 
-    componentDidMount() {
-        watchChatSubscriptionChanged(this.onChatSubscriptionChanged);
-    }
+    React.useEffect(() => {
+        setChannelId(channel);
+        setSubscribable(props.subscribable);
+        setNotifyUnread(getUnreadChatPreference(channel));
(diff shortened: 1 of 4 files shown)


src/components/Chat/ChatDetails.tsx
```diff
@@ -35,174 +33,147 @@ interface ChatDetailsProperties {
     subscribable?: boolean;
 }
 
-interface ChatDetailsState {
-    channelId: string;
-    subscribable: boolean;
-    notify_unread: boolean;
-    notify_mentioned: boolean;
-}
+export function ChatDetails(props: ChatDetailsProperties): JSX.Element {
```
benjaminpjones, line 36: ooh FC nice :)

src/components/Chat/ChatDetails.tsx
```diff
@@ -35,174 +33,147 @@ interface ChatDetailsProperties {
...
-            this.state = {
-                channelId: channel,
-                subscribable: props.subscribable,
-                notify_unread: false,
-                notify_mentioned: false,
-            };
-        }
-    }
+    const [channelId, setChannelId] = React.useState(channel);
+    const [subscribable, setSubscribable] = React.useState(props.subscribable);
```
benjaminpjones, line 40: Perhaps not an issue for this PR, but it's odd to me that subscribable is both a prop and a state variable...

It looks like there's a `useEffect` below that mirrors `props.subscribable` into `subscribable`, but it would be simpler and safer to just use `props.subscribable` directly.

benjaminpjones (commented):
Looks good to me!

Not on the fa-external-link: I actually preferred the "View Group"/"View Tournament" from the previous PR.  From the screenshot it looks like that might take me to a dedicated notifications settings page?

anoek:
> Note on the fa-external-link: I actually preferred the "View Group"/"View Tournament" from the previous PR because it's clear where it will take me. From this screenshot it looks like that might take me to a dedicated notifications settings page?

Yeah I agree, I couldn't fit the link on there in a way that didn't look bad though.. I had thought about simply having a link icon on the chat bar next to the settings icon or something, maybe that would be a better solution?

src/components/Chat/ChatDetails.tsx
anoek: Yeah that's totally true.. it was a very mechanical port, would be good to further clean that up

BHydden:
> a link icon on the chat bar next to the settings icon

Actually that sounds pretty good 🤔👍

benjaminpjones:
> Yeah I agree, I couldn't fit the link on there in a way that didn't look bad though.. I had thought about simply having a link icon on the chat bar next to the settings icon or something, maybe that would be a better solution?

Ah I see what you're saying - I played with it a bit and the Notifications header and the Group link don't fit very well together. Anyway - here's what I managed to come up with if you're interested ([source](https://github.com/online-go/online-go.com/compare/chat-notification-settings-take-3...benjaminpjones:online-go.com:chat-notification-settings)):

Screen Shot 2022-09-09 at 6 12 03 PM

But in the end, it may make more sense to move that link out of settings.

anoek:
Yeah that looks great, lets go with that. The icon on the bar would look good too I reckon, though I'm less sure about where the link should go in the notification chat list where this component is used too, so because of that, and because your modifications ltgm, lets just go with that :)

Thanks @benjaminpjones !

PR #1997, opened 2022-09-11
Variation links

benjaminpjones:
There was some discussion in the forums requesting the ability to specify a game variation via URL: https://forums.online-go.com/t/include-move-number-and-variation-in-url-to-game/44730

I think there are many ways to do this, but in this PR, I chose to use the chat_id (are those stable?) as the variation identifier.

output

## Proposed Changes

  - Clicking a variation in game chat adds a query parameter to the URL bar
  - Opening this URL highlights the variation text, and displays the variation on the board


mmartin:
This is awesome!

Maybe it would be better to remove `var_id` from URL bar when user closes analysis mode or leaves the variation?

Would it be possible to scroll to the variation chat message when opening a URL with a variation?

When I open a game from move link (`/game/<id>/<move>`) and then click on a variation, I get `/game/<id>/<move>?var_id=<variation>`. Opening such link makes it jump to the move and then to the variation. Maybe clicking on variation should remove `<move>` parameter from URL bar?

benjaminpjones:
Thanks for testing it out @mmartin!

So yeah, I think all of those ideas would be great enhancements, but unfortunately implementation might be a little hairy (at least for me 😅)

> Maybe it would be better to remove var_id from URL bar when user closes analysis mode or leaves the variation?

I went ahead and tried setting:

```
game_control.onPushAnalysisLeft = () => {
    leaveAnalysis();
    setSearchParams([]);
}
```

But the variation is still in the URL bar when I leave analysis?

> Would it be possible to scroll to the variation chat message when opening a URL with a variation?

In React this can be hard to get right because everything loads dynamically and you need to wait for it all.  But could be as simple as `chat_line_ref.current.scrollIntoView()` in the right place.

> When I open a game from move link (/game/<id>/<move>) and then click on a variation, I get /game/<id>/<move>?var_id=<variation>. Opening such link makes it jump to the move and then to the variation. Maybe clicking on variation should remove <move> parameter from URL bar?

Nice catch! I didn't even think to test thiat path.  One issue for this one is that we're currently on an old react-router version, so I don't think `useHistory()` was an option.  I'm not familiar with the old APIs, and it might be better not to rely on them if we're planning to upgrade any time soon.

anoek:
Very cool

I'm thinking we should make it so when the `Share` button is pressed, we replace the URL at that point too?

As for knowing when we have a move url, I *think* you can use `useParams` in any of those components on the Game view like this line: https://github.com/online-go/online-go.com/blob/devel/src/views/Game/Game.tsx#L75 and then just check `params.move_number` to see if it is defined.


I'm thinking, at least in some cases, we'll want to do a history replacement instead of pushing history on the stack. (ie use `setSearchParams(..., { replace: true });` so when the user hits the back button they go back to their previous page instead of back through the stack of variations they've clicked on. I'm open to counter arguments here, on the one hand if they click on a "link" thing, which the variations are in a sense, then one might expect to hit the back button to go back and be in the state they were in before they clicked the link. On the other hand, when a user clicks around a bunch of links in the same game, then hits "back to game", then hit's the back button, I feel like they shouldn't have to click the back button through all of the variations they had clicked through.  

I *think* having one variation in the history stack while a variation is being viewed might be the correct thing to do and might feel right, so maybe something along the following lines:
   * if we are not viewing a variation, and we click a variation link, we push the variation onto the history stack.
   * If we are in a variation already and click another variation link, we replace the history entry, so clicking the back button will take us to the original non variation state.
   * If we are in a variation and click the "Back to Game" button we use `useNavigate(-1)` to go back in the history stack to get our URL on a non var_id URL, but also makes us able to hit the forward button just as though we had clicked the back button.
   * When the browser back button is clicked while we are on a variation and our previous entry is the same game (IE we got there from clicking a variation link, not by following it some other way), we should exit analysis mode just as though the "Back to Game" button was pressed. 

Again, definitely open to some other thoughts on that navigation matter, it's kind of a weird case for the intermixing the browser url history stack and the state of the game view, but I think there's a pretty decent answer somewhere there that feels more or less natural.

benjaminpjones:
Hi anoek, great feedback on the nav - I actually just didn't realize this affected the back button 😅 I went ahead and added `{ replace: true }` in both places according to your first suggestion.  I think this is a totally reasonable option, and is my personal favorite because it is simple and maintains existing behavior.

> I'm thinking we should make it so when the Share button is pressed, we replace the URL at that point too?

I think this is another case of, "I think this is a good idea, but implementation may be hairy?".  When the user clicks the Share button, the chat id doesn't exist until the server updates the chat.  I dug into the `onShareAnalysis()` function, and it looks like it uses goban's `sendChat()` function to handle the network call, so I don't think we can immediately process the response (assuming that even includes the chat id).  Any ideas on how this should/could be done?

anoek:
Ah yeah that's problematic. 

Food for thought, since we're not doing anything else with the var_id other than looking for the variation, the other thing we could do if we wanted would be to simply encode the variation itself in the url. There's a helper function available in the engine, `getMoveDiff()` which will return `{from: 3, moves: "ldlglh"}` to mean from move 3, follow the variation by the move string in moves. There's also then a engine method `followPath` which takes a from move number and a move string and will go to that variation, so simply shoving the move number/move string in the URL instead of te var_id might ultimately be an easier solution and would allow you to set that url right when Share was clicked.

That said, what you have is working good too, so if that's not something you want to do then meh, we can merge and call it good, it's already a great improvement :)

benjaminpjones:
I looked into this last night, seems not too hard since that format is already used to store the variation.  And I like it because:

- it doesn't rely on stability of chat ids
- possible to not rely on sending the chat
- shorter in many cases

I think the big question is what URL pattern?

- path/to/game/12345?from=3&moves=ldlglh
- path/to/game/12345?v=3ldlglh (<- I am leaning towards this one)
- path/to/game/12345/3/ldlglh

Those are really just a few options, of which I'm currently leaning towards the second (compact, query parameters easier)

anoek:
Yeah the second is probably good, I like the first in terms of understandability, but since they don't make sense separated, the second option is probably the best choice.

BHydden:
I like the second, but the third is how move numbers already work, so it depends whether you want consistency or deliberate differentiation between the two features.

benjaminpjones:
Going to close this for now-  I plan on coming back to it when I get time, but I definitely think it's worth doing with the serialized variation (`getMoveDiff`) rather than the chat id.

PR #2003, opened 2022-09-14
Add "4x4" option in puzzle

tkoyama010:
Fixes #

## Proposed Changes
.
  - 張 栩 made puzzles of 4x4 board in https://www.amazon.co.jp/dp/B07CKY1VK5 . I want to solve it in the OGS puzzle.


anoek:
Thanks!

PR #2006, opened 2022-09-21
Fix ChatList links

mmartin:
Links in the ChatList are slightly broken:
* clicking on icon doesn't work
* for tournaments clicking inside `div` padding takes me to wrong URL

## Proposed Changes
* use `event.currentTarget` instead of `event.target`
* fix tournamets' `data-channel` attribute


anoek:
Hah good catches, thanks for the fixes!

PR #2007, opened 2022-09-22
Convert remaining main views from class views to functional components

anoek:
This is going to finally eliminate the need for the rr6ClassShim hack that we put in when we updated to react router 6. This shim is causing some problems that are affecting a number of people, so seems best to just try and eliminate it.


anoek:
Some of the remanining views are a real beast to port, I no longer thing it's worth it. Included in this patch will be the two I finished porting and a fix to the rr6ClassShim that should take care of the sentry error we were seeing.

PR #2009, opened 2022-09-23
State the type of the context, matching what RDH says it is.

GreenAsJade:
TBD I'm not totally sure this is better, but it _seems_ nicer to use the expicitly stated type of the context, rather than show that we know how surely it must have been created, and rely on that?


anoek:
Thanks!

PR #2010, opened 2022-09-25
potential fix for sort issue on game library page

andrewallred:
## Fixes

Game libraries are intermittently sorted in random order, and sometimes sorted correctly. See the below thread for some context and that other folks have noticed this.

https://forums.online-go.com/t/sgf-library-sort/42511/2

## Proposed Changes

I'm not a typescript expert, but looking through the code this is the only area where localeCompare is used without specifying the compare string, here we're passing just the b object to be sorted. It seems possible that this is the source of the issue, and if not it will at least make the code consistent on this point.

I'm a first time contributor, so please let me know if I should be doing anything differently in this process!

### DIFF
--- a/src/views/LibraryPlayer/LibraryPlayer.tsx
+++ b/src/views/LibraryPlayer/LibraryPlayer.tsx
@@ -177,8 +177,8 @@ class _LibraryPlayer extends React.PureComponent<LibraryPlayerProperties, Librar
 
                 for (const collection_id in collections) {
                     const collection = collections[collection_id];
-                    collection.collections.sort((a, b) => a.name.localeCompare(b));
-                    collection.games.sort((a, b) => a.name.localeCompare(b));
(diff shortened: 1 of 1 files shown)


anoek:
Hah, whoops - thanks for the fix! What's odd is I'm not sure why typescript wasn't yelling at us for the collections.. games was typed as `any[]` so that makes sense, but it seems like typescript should have known that was an error for collections.

benjaminpjones:
> What's odd is I'm not sure why typescript wasn't yelling at us for the collections.. games was typed as any[] so that makes sense, but it seems like typescript should have known that was an error for collections.

At line 110:

https://github.com/online-go/online-go.com/blob/ae79d61901da19d8d09a4f5b671de5ebd5d2e98b/src/views/LibraryPlayer/LibraryPlayer.tsx#L110

Adding a type annotation will cause TS to raise the expected error on the localeCompare:

```
const collections: { [idx: number]: Collection } = {};
```

PR #2011, opened 2022-09-25
Make language-picker stay on one line

GreenAsJade:
Fixes # language picker getting split into two lines

## Proposed Changes

use `white-space: nowrap` to stop it being split.


anoek:
Thanks!

PR #2019, opened 2022-10-04
Game.tsx: clear all marks in the tree when toggling AI mode.

jmoudrik:
This depends on PR in `goban` - https://github.com/online-go/goban/pull/85

Before this:
- marks in the root & in the branches were not cleared on the toggle.

Steps to replicate / test:
- open a game from SGF library
- toggle AI Mode On in the root node
- once some variations / best moves are showed, toggle the Ai Mode Off

Behavior before the PR:
- the letter marks stayed on the goban, unlike the blue circle & heatmap, which disappeared
- also, analysis in branches also retained some marks & AI variations after toggling off

Behavior now: 
- letter marks disappear as well as the blue circle & heatmap (as one would expect)
- also, marks and variations in branches disappear on toggle-off as well


anoek:
Thanks for the fix! 

Tip if you run into goban update problems again in the future, if you run `yarn` to update the `yarn.lock`, and commit that change as well as the `package.json` change, the build for the PR will succeed like you'd expect.

PR #2020, opened 2022-10-04
Preliminary paste SGF option

bhostetler18:
This pull request attempts to address [this feature request](https://forums.online-go.com/t/feature-request-paste-sgf/43876) for the ability to paste raw SGF data.

As I was testing this on the beta server, I noticed that uploaded files are not persisted and do not show up in the SGF library list, even when using the original upload button. The `POST` request from my new feature appears to work properly in that it receives the expected `{"success":"Files uploaded"}` response, but I can't fully test it for this reason. This behavior might be intentional (maybe to avoid storing a bunch of test .sgf files on the beta server?), but any guidance is appreciated.

Let me know if I should change anything else to make this acceptable for production! I think it will need some translation (for the button and placeholder text), but I'm not entirely sure how to start with that.

### DIFF
--- a/src/components/SGFPasteModal/SGFPasteModal.styl
+++ b/src/components/SGFPasteModal/SGFPasteModal.styl
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2012-2022  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.SGFPasteModal {
+    width: 30rem;
+    max-width: 100%;
+    height: 55vh
+    max-height: 100%;
+
+    .body {
+        display: flex;
+
+        align-items: stretch;
+        justify-content: stretch;
+        align-content: stretch;
+
+        textarea {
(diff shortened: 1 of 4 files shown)


anoek:
What's the use case for this? Namely, what application are you using that makes it easier to paste rather than upload an SGF?

anoek:
Never mind I read the original feature request. I guess I don't have any objections, seems like a pretty niche feature but that's not necessarily a problem.

bhostetler18:
Yeah, it may be a bit niche. There seemed to be some interest on that original post, but if you feel it adds unnecessary clutter I understand.

anoek:
As long as more than a couple of people want it I think it's a fine addition, and it sounds like that. Nice job with the implementation so far, caught all of the translations too which a lot of people miss, so cudos :)

I'm thinking the name "pasted.sgf" is going to get a bit old fast for those that are using this feature. Any chance we could either generate a name looking at the game name field of the pasted sgf (if it exists), and/or have an input field for the name in that modal?

bhostetler18:
Thank you :)

That sounds like a great idea and I’d be happy to! I’ll update this PR when I have a chance.

bhostetler18:
Just added a filename input area which automatically pulls the game name from any pasted SGF data. A few questions:

- Should there be any further validation on the filename? Right now it just ensures that the extension ".sgf" is used, but someone could conceivably cause trouble by using `.`, `/`, or other unwanted characters. I could just strip these before uploading.
- Would you prefer something other than "pasted.sgf" when there is no SGF game name and the user hasn't typed a filename?
- For the title `textarea`, I made a new CSS class with left, right, and top margins matching that of `body` to achieve the layout shown in the attached screenshot. This seems prone to errors if the global style ever changes, so I'd rather inherit these values if possible, but I couldn't figure out a way to do this without introducing a bottom margin (and without making the title box expand vertically like the SGF data area does). I am not very familiar with web development, so any guidance is appreciated.

title textarea

bhostetler18:
On a related note, I just realized that SGF files downloaded from OGS can include special characters in the `GN` field, and other servers might do the same. Right now these names won't be automatically detected or used as the filename at all, but maybe it would be better to recognize them and strip the characters?

anoek:
Awesome, thanks!

Issue #2013, opened 2022-09-29
SGF Library not sorted by last uploaded anymore


gezabohus:
The SGF library used to place the most recently uploaded file on top, it doesn't any more. It seems to do it alphabetically, but the earlier system was much more logical.

1. Go to SGF Library
2. Upload a game
3. It won't be on top unless its name happens to be at the beginning of the list.

**Expected behavior**
Newly uploaded file should be on top or one should have a way of sorting by upload date.

**Screenshots**
<img width="465" alt="image" src="https://user-images.githubusercontent.com/10329508/193073839-3a6b552c-967b-4eb5-a740-a45bac710595.png">

**Desktop (please complete the following information):**
 - OS: MacOS Catalina
 - Browser: Brave, Safari

**Smartphone (please complete the following information):**
 - Device: Android
 - Browser: Brave

Also, there seems to be limit on the number of games one can upload before starting a new directory, but it doesn't say so, just fails quietly.

benjaminpjones:
This would be because of #2010, although that PR actually fixed a bug.  The fact that it was sorted by most recent was possibly a happy accident.

Maybe good to discuss on the forums what should be the default sort?

gezabohus:
Thanks for the info @benjaminpjones. My vote would be for last uploaded on top, but perhaps others think differently. One could also have a clickable header to sort whichever way, but that requires a bit more development.

anoek:
SGF sorting was implemented so takes care of this issue

PR #2024, opened 2022-10-05
SGF sorting

bhostetler18:
This PR starts to address #2013 by adding various sort options and an ascending/descending toggle. It's not the prettiest, but the underlying functionality is mostly there. Currently, the three sort modes are `name`, `date added`*, and `game date`. Other suggestions welcome, especially if anyone has ideas for a better and more consistent UI.

*For this sort mode I currently use the `created` field on each game, but I'm not sure if that's appropriate. If that field doesn't represent the date a game was uploaded, would `entry_id` or `game_id` work instead (I'm assuming one of those is a sequential, increasing value)? That might be better even if `created` behaves as expected, since it would avoid parsing a new `Date` during each sort. @anoek 

~~In order to test this functionality, I've just been manually adding fake game entries since the beta server doesn't appear to save anything (as mentioned in #2020). Still curious if there's a better way.~~
Fixed with #2027 

sort


bhostetler18:
sorting

I've made a layout similar to the game list page. It currently defaults to sorting the most recently added files on top since that seemed popular. Let me know if anything seems amiss or you want something else with the layout/functionality!

src/views/LibraryPlayer/LibraryPlayer.tsx
```diff
@@ -69,9 +74,27 @@ class _LibraryPlayer extends React.PureComponent<LibraryPlayerProperties, Librar
             games_checked: {},
             new_collection_name: "",
             new_collection_private: false,
+            sort_order: "date_added",
+            sort_descending: true,
```
anoek, line 78: I reckon it'd be nice if we remembered these, that's usually done with the preferences system, which goes something like this

1) Add your preference keys and default values to `lib/preferences.ts", so something like `"sgf.sort-order": "date_added"`
2) `import * as  preferences from "preferences";`
3) 
```
sort_order: preferences.get("sgf.sort-order"),
sort_descending: preferences.get("sgf.sort-descending"),
```
4) When you change your sort order, call `preferences.set("sgf.sort-order", whatever)`

bhostetler18: Will do!

anoek:
Awesome, thanks for the patch and going the extra mile!

PR #2027, opened 2022-10-10
Fix file uploads when using the dev server

bhostetler18:
This just adds a small wrapper which stops the proxy from parsing the request body when uploading files.

See [this issue](https://github.com/villadora/express-http-proxy/issues/127) for more details.


anoek:
Awesome, thanks for digging into that and getting a fix put together!

PR #2028, opened 2022-10-11
Provide an ObserveGamesComponent filter for Malkovich games.

GreenAsJade:
Fixes limited discoverability of Malkovich function

## Proposed Changes

 Provide a "malkovich only" filter on the Watch page

** note: requires https://github.com/online-go/ogs-node/pull/20 in order to actually work.


GreenAsJade:
Note: prior to "Private logs" this would have been a dubious thing to do, because folk were using Malk for private logs, and this feature would have been tempting to use to find those.

Nowadays since we have private logs, the only time Malk would be expected is in a context where spectators are truly expected, and the comments are intended for spectators.   

It is a slight cheat-opportunity-enhancer for those games, but since you can find out anyhow if you are interested, by loggin out,  it seems like an OK thing to do this to promote Malk spectatoring.

anoek:
Cool, fun stuff, thanks!

Issue #1985, opened 2022-09-03
Unable to remove rengo games from active games list after timeout/resign game


hackeysack01:
**Describe the bug**
Unable to remove rengo games from active games list after leaving game

**To Reproduce**
1. join rengo
2. timeout or resign
3. game stays in active list

**Expected behavior**
Game is removed from active list.

An opt out option such that the game is removable but the default action is to continue watching would satisfy my need.

Am willing to concede this may be a feature request.

Might also expand to a 'follow games' opt in feature with similar flags to be able to watch other correspondence games that are public.

edit:
My forum post to ogs.
https://forums.online-go.com/t/remove-rengo-from-active-list-after-leaving-game/44660

benjaminpjones:
Thanks for opening the GitHub issue, I'm linking several other reports from the forums:

- https://forums.online-go.com/t/rengo-status/40415/17
- https://forums.online-go.com/t/rengo-issue/44262
- https://forums.online-go.com/t/removing-game-from-start-page/41388
- https://forums.online-go.com/t/how-to-close-finish-this-game/44057

PR #2032, opened 2022-10-11
Rengo display fix when timed out/resigned

### LINKED ISSUE #1998: [rengo] Rengo columns misaligned in active games table
**Describe the bug**

In casual rengo games that you are no longer an active participant in, a player appears in the "Opponent's clock" column, shifting remaining columns to the right.

**To Reproduce**
1. Have casual rengo games going.
2. **Resign or timeout from them**
3. Go to the home page
4. Observe the table entry for rengo game looks like:

```
Move    | Game         | Opponent    | Clock  | Opponent's Clock | Size
10      | Rengo Match! | some_player | 5 Days | another_player   | 18h46m | 19x19
```

**Expected behavior**

```
Move    | Game         | Opponent    | Clock  | Opponent's Clock | Size
10      | Rengo Match! | some_player | 5 Days |      18h46m      | 19x19
```

**Screenshots**
Screen Shot 2022-09-11 at 10 41 57 PM

**Additional context**
I believe the same component is used on the ObserveGames page, but the columns do include both White and Black's name since it makes sense in that context.  Possibly related?

bhostetler18:
This PR starts to fix #1998, which as @benjaminpjones noted is caused when a player resigns or times out from a rengo game. In these cases, the `rengo_teams` list does not contain the current player and this causes `GobanLineSummary` to think it is displaying in the style used on the https://online-go.com/observe-games page (the unwanted behavior where another player column is added).

All I have done is detect this scenario and display an empty cell across the three columns (opponent, clock, opponent clock) which cannot be determined. We could also stick some text in that empty space explaining that the game has been left, but it felt a bit cluttered when I tried this. I can think of a couple of ways to proceed:

1. Display the proper data in those columns. I believe this would need backend support, since there doesn't seem to be a way to obtain which team a player was originally part of after they leave a rengo match. I could be wrong, however.
2. Push these entries to the bottom of the games list, and potentially even have the option to hide them. I think I saw this requested somewhere on the OGS forums.

rengo


benjaminpjones:
> 1. Display the proper data in those columns. I believe this would need backend support, since there doesn't seem to be a way to obtain which team a player was originally part of after they leave a rengo match. I could be wrong, however.

@GreenAsJade can probably speak to the feasibility of getting the right fields into the API response (if they aren't there already).

> 2. Push these entries to the bottom of the games list, and potentially even have the option to hide them. I think I saw this requested somewhere on the OGS forums.

See https://github.com/online-go/online-go.com/issues/1985.  I commented a list of forum posts requesting removal of old games from this list.  Personally, I'm in favor of this option.

bhostetler18:
@GreenAsJade How does something like a “dropped rengo games” section sound for a compromise? This section could potentially be collapsible so that games aren’t removed entirely, but people who don’t want to see them can avoid the clutter.

benjaminpjones:
> It's actually painful to get that info where it needs to go, to take care of it :)

Isn't this PR evidence that it's possible to detect stale rengo without changes to the API?

>I just didn't like the idea of people dropping out from casual rengo games, and didn't want to encourage it.

I have thoughts one but won't get into them here 🙂

> Some people actually like following their ex-game

I agree this is a concern.  A few of the forum posts I linked to have a comment stating that there should at least be an option to view these games.

> How does something like a “dropped rengo games” section sound for a compromise?

This is something I think a lot of folks would appreciate, and would fix the display issue since you can just make it 3 columns shorter!

GreenAsJade:
I had thought that when people complain about this, they were complaining that the games still show in their game indicator circle.

Or at least, that's what I'd always imagined.  I never really checked TTTT, because I'd just assumed that's what it was!   😊 

If they do show in the circle indicator, then this change won't fix that complaint.

That said, the idea of "dropped games" section totally matches my "feeling" that dropping out of casual rengo games should be called what it is 😄, and would be easy to make optional, and apparently doesn't need a back end change, so it's got a lot going for it!

bhostetler18:
> I had thought that when people complain about this, they were complaining that the games still show in their game indicator circle.

I just checked and they don't show on the game indicator, even when tapping it to cycle between games.

> That said, the idea of "dropped games" section totally matches my "feeling" that dropping out of casual rengo games should be called what it is 😄, and would be easy to make optional, and apparently doesn't need a back end change, so it's got a lot going for it!

I'll start woking on that when I have time then! To be fair, there is a difference between resigning and timing out of a rengo game in terms of the negative connotation.

GreenAsJade:
Thanks for checking!   

It sounds like your proposal is a winner then :)

 > To be fair, there is a difference between resigning and timing out of a rengo game in terms of the negative connotation.

Actually, resignation from casual rengo games probably needs some work.   There isn't a mechanism for a team to resign, and we can't tell the difference between a resignation that says "this is a losing position" and one that says "I'm sick of this".

As you may or may not know, casual rengo was a mode that was added "as an afterthought", in response to the problem of people tiring of their rengo game before it was finished, and dropping out.  In normal rengo rules, this terminates the game, losing it for their team.

Casual rengo is therefore a "defence against the act of someone getting sick of the game".

It's not intended (by me) to say "it's OK to tire of the game you signed up for", but rather to say "at least everyone else doesn't have to suffer because that person got sick of it".

bhostetler18:
Got it, thanks for the context! Team resignation does seem like a good feature for the future.

GreenAsJade:
Yes - team resignation got put aside when we realised we need a proper concensus mechanism, which would likely find other uses.    And which in itself is not truly difficult, but warranted enough due care and attention that it wasn't the right time, back then.

So if you happen to know the right way to build a concensus mechanism for this type of thing, we'd certainly love it :)

bhostetler18:
Just finished a prototype of the changes we discussed. It involved a lot of refactoring of the `GameList` component, which hopefully makes it easier to reason about and harder to introduce bugs like the one that started this PR. Since this component is used heavily (and in different configurations) across the Overview, User, and Observe pages, we should probably do some more rigorous testing of these changes. Happy to defer to anyone else about UI preferences – I don't know if there's a standardized way to make collapsing sections throughout the site, and the dropped game table layout could be improved. 

open
collapsed

GreenAsJade:
Probably a good one to put up on Beta and point Sofia at it eh @anoek

bhostetler18:
Hey @Sofia-mal are you available and willing to help test this?

Sofia-mal:
Yes, of course, in a few hours :) 
But how? Is it enough to create new games?

bhostetler18:
@Sofia-mal 

Testing will involve creating rengo games and dropping out of them, then checking that they display correctly in your home/overview page (under the new "dropped games" section). Similarly, we should also check that dropped games are displayed correctly when navigating to the profile page of a user other than yourself. 

The new game list is also used on the "observe" page so that can be tested as well, though I'm pretty sure I've covered and tested all the possibilities there.

Throughout all this, I just want to double check that everything behaves as expected across browsers, screen sizes, different amounts of active/dropped games, and values of the "game thumbnail list threshold" setting under general preferences. If anything looks weird or behaves strangely (even just a UI element that could be adjusted slightly) let me know!

I'm happy to help facilitate this by creating and joining games with other accounts. Thank you!

Sofia-mal:
I have found other "problems", but nothing that seems related to this :)

anoek:
Awesome, thanks for all your work there @bhostetler18 ! And thanks for testing and reviewing @benjaminpjones , @GreenAsJade , and @Sofia-mal !

PR #2036, opened 2022-10-14
Fix #1738: Ladder position error wording

### LINKED ISSUE #1738: Ladder pages use the term "rank" incorrectly
**Describe the bug**
When you click on a player in a ladder that you are ineligible to challenge, it says either "your rank is too low" or "your rank is too hitgh".

It is trying to say that you are placed to low or too high on the ladder: nothing to do with rank.

It should say "this player is beneath you on the ladder" or "this player is too far above you on the ladder", or similar.

cassidoo:
Fixes #1738

## Proposed Changes

Changes error wording to say, "Your ladder position is too high" instead of using the word "rank"


anoek:
Indeed, thanks!

PR #2039, opened 2022-10-15
Add types to TimeControlPicker.

benjaminpjones:
I was digging around in `ChallengeModal`/`TimeControlPicker` and thought it might be easier to code in here if we have stronger types.  This PR *should* only be types, no changes to the JavaScript output.

## Proposed Changes

  - Add (most) types that are necessary when `noImplicitAny` and `strictNullChecks` are on
  - Add some comments about what still needs to be fixed.


anoek:
Awesome thanks!

PR #2042, opened 2022-10-17
Double the resolution of nation flags in games.

### LINKED ISSUE #1824: Blurry flags on player card (mobile)
**Describe the bug**
The flags shown in games on the player card are more blurry than necessary.  Here's an example screenshot taken without zooming on a phone with a 2,400x1,080 resolution:

Screenshot_20220504-222012

We have higher-resolution flag images on the profile pages; those should just get used here.

Using the `f32` class instead of `f16` on the outer span and adding `transform: scale(0.5);` to the `span.flag` seems to mostly fix it, but someone would need to adjust the offset.

benjaminpjones:
This is done by using the 32px asset, scaled down by 50%.

Fixes #1824 

Note: There's probably room for improvement further down the stack.  In other words, I have only "fixed" this for games (where these flags are seen most often).  However, the flag is still blurry in, for example, the popover for the `Player` component.  But if we want to fix blurry flags everywhere, it might mean updating world-flags to contain 64px assets (or even vector graphics).  Just a thought.

Tested: This is difficult, but I tried all layouts that have their own css - desktop, portrait, and squashed views, as well as odd shaped flags (Swiss, Nepal, LGBTQ+).  The one major base I haven't covered is testing on an actual low res screen.

<details><summary>Screenshots (Before then After)</summary>

Desktop:
Screen Shot 2022-10-16 at 9 27 46 PM
Screen Shot 2022-10-16 at 9 27 28 PM

Squashed:
Screen Shot 2022-10-16 at 9 26 50 PM
Screen Shot 2022-10-16 at 9 26 44 PM

Portrait:
Screen Shot 2022-10-16 at 9 25 56 PM
Screen Shot 2022-10-16 at 9 25 49 PM

</details>

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -346,7 +346,7 @@ function PlayerFlag({ player_id }: { player_id: number }): JSX.Element {
     }, [player_id]);
 
     if (country) {
-        return <Flag country={country} />;
+        return <Flag country={country} big />;
     }
     return null;
 }
--- a/src/views/Game/Players.styl
+++ b/src/views/Game/Players.styl
@@ -152,6 +152,7 @@
             .player-flag {
                 top: 1rem;
                 left: 1rem;
+                transform-origin: 6px 10px;
             }
         }
     }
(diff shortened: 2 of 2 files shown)


anoek:
Awesome, thanks for digging into that and getting it fixed!

PR #2045, opened 2022-10-18
Online league support

GreenAsJade:
Fixes having nothing to show yet :)

## Proposed Changes

Routes and view to support the URLs that we provide to online leagues.   They don't _do_ anything, but they do at least display a page.

Requires backend update: https://github.com/online-go/ogs/pull/1754


GreenAsJade:
Note *** when running the 0099_nagf migration, take note of the key that it prints out: the NAFG client needs this key.

anoek:
Fun stuff ahead :)

BHydden:
Always ;)

PR #2052, opened 2022-10-21
Show folk's OOLs on their profile

GreenAsJade:
Fixes not being able to tell what leagues someone is in, and whether their membership id is correct

## Proposed Changes

Display people's OOL in their Profile, with the membership id visible if the backend sends it (which it only does if you're looking at your own profile, though we should add moderators  as well).

Needs https://github.com/online-go/ogs/pull/1757 to be able to show anything.

### DIFF
--- a/src/models/user.d.ts
+++ b/src/models/user.d.ts
@@ -154,6 +154,10 @@ declare namespace rest_api {
             name: string;
             icon: string; // URL
         }>;
+        online_leagues: Array<{
+            name: string;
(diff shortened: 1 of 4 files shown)


GreenAsJade:
I can't tell why the build failed - it shouldn't need the backend PR just to run.   The build failure message is non-informative :)

WFM  🤷‍♀️

anoek:
Thanks!

Issue #2050, opened 2022-10-19
Play graph improvements


bhostetler18:
When I started using OGS, I was completely mystified by the colors in the graph on the play page. I eventually found [this forum post](https://forums.online-go.com/t/play-graph/34662/2), but it seems worthwhile to create some kind of legend explaining what the colors mean. To that end, I started messing with a layout like this:

<img width="557" alt="legend" src="https://user-images.githubusercontent.com/35665417/196793561-4a9a0c4e-2917-4862-9432-dae7287a08ae.png">

I was thinking that mousing over/tapping on each box in the legend would display the type of game challenge  (either in a popup or in canvas text directly below the legend). Before I continue, any alternative ideas or suggestions? Also, while I'm messing around with this code anyway, does anyone have improvements/changes they would like to see made to the play graph?

BHydden:
For some reason I can't initiate reactions on mobile so

`+1` 😅🔥

anoek:
I would merge that, that'd definitely be an improvement.

As for alternative idea's, improvements, and all of that - I'm very open here. This code is pretty darn old by now and it looks antiquated, I'd welcome an overhaul so sky is the limit here.

PR #2053, opened 2022-10-22
Seek graph improvements

bhostetler18:
As proposed in #2050, this PR contains a number of improvements to the seek graph.

Notable changes:
- Improved resolution of seek graph
- Added shapes representing various challenge types (ranked, unranked, rengo)
- Moved challenge filters away from the bottom of the page, combining them with a legend explaining the symbols in the seek graph
- Made challenge filters affect the seek graph in addition to the challenge lists
- Added a challenge filter to hide/show rengo games
- Adjusted seek graph colors to match the current site theme
- Made the seek graph immediately respond to site theme changes
- Fixed a bug where the popover list would occasionally not show when mousing over a challenge in the bottom right of the seek graph

| Before  | After |
| ---------- | ---------- |
| old_seekgraph | new_seekgraph  |

### DIFF
--- a/src/components/SeekGraph/SeekGraphLegend.tsx
+++ b/src/components/SeekGraph/SeekGraphLegend.tsx
@@ -141,12 +141,6 @@ export function SeekGraphLegend(props: SeekGraphLegendProps): JSX.Element {
         <div className="seek-graph-legend">
             <Card>
                 <table>
-                    <thead>
-                        <tr>
-                            <th>{_("Board Sizes")}</th>
-                            <th>{_("Game Types")}</th>
-                        </tr>
-                    </thead>
                     <tbody>
                         {[0, 1, 2, 3, 4].map((i) => (
                             <tr>
--- a/src/components/SeekGraph/SeekGraphLegend.tsx
+++ b/src/components/SeekGraph/SeekGraphLegend.tsx
@@ -26,7 +26,7 @@ import {
 } from "./SeekGraphSymbols";
 import { ChallengePointStyle, SeekGraphColorPalette, SeekGraphPalettes } from "./SeekGraphPalettes";
 import * as data from "data";
-import { ChallengeFilter, ChallengeFilterKey } from "src/views/Play";
+import { ChallengeFilter, ChallengeFilterKey } from "challenge_utils";
 // import { SeekGraph } from "./SeekGraph";
 // import { SeekGraphColorPalette } from "./SeekGraphPalettes";
 
--- a/src/components/SeekGraph/SeekGraphLegend.tsx
+++ b/src/components/SeekGraph/SeekGraphLegend.tsx
@@ -32,6 +32,7 @@ import { ChallengeFilter, ChallengeFilterKey } from "challenge_utils";
 
 interface SeekGraphLegendProps {
     filter: ChallengeFilter;
+    showIcons: boolean;
     toggleHandler: (key: ChallengeFilterKey) => void;
 }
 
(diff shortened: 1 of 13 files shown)


bhostetler18:
@anoek @benjaminpjones This is probably another good candidate for testing on the beta site first. There are a lot of code and visual changes, and some of the new behavior is opinionated enough to warrant community feedback.

BHydden:
Can the key/legend/filters be minimized? That might be a good option to include if not.

benjaminpjones:
I tried running this, but not seeing anything in the SeekGraph now.  I got an error about using keys in a map, but fixing it didn't help.

System: Firefox/Mac

Screen Shot 2022-10-21 at 8 44 16 PM

src/components/SeekGraph/SeekGraphLegend.tsx
```diff
@@ -0,0 +1,192 @@
...
+        ),
+    ];
+
+    return (
+        <div className="seek-graph-legend">
+            <Card>
+                <table>
+                    <tbody>
+                        {[0, 1, 2, 3, 4].map((i) => (
+                            <tr>
```
benjaminpjones, line 146: Here's that key think I was talking about. ([React Docs: Lists and Keys](https://reactjs.org/docs/lists-and-keys.html))

```suggestion
                            <tr key={i}>
```

bhostetler18: Thanks for the reminder! I’d actually meant to fix that but it completely slipped my mind.

bhostetler18: The SeekGraph issue is probably entirely unrelated – it’s a different component than the legend. What happens when you inspect the area where it should show up?

bhostetler18: Never mind, fixed. Firefox doesn’t support a certain font measure I was using, which caused a padding value to be NaN, which cascaded and made all rendering coordinates NaN…

anoek:
What about putting the legend underneath the graph, and just squishing the graph vertically a little bit? I appologize for my GIMP skills but something kinda like this:
image

bhostetler18:
I definitely considered it, but thought it might be unclear since the checkboxes also affect the challenge list. Happy to consider it further though, and I definitely agree that layout changes could be made.

Considering that people here and on the forums have asked for the option to hide the graph entirely – if the legend did go directly beneath the graph, should hiding the graph cause the legend to fill the full right card (on desktop at least)?

src/components/SeekGraph/SeekGraphLegend.tsx
benjaminpjones: yep the keys thing was def unrelated haha, just the first thing i saw in the console when I hit inspect.

Thanks for figuring out the Seek Graph issue - that was a strange root cause!

anoek:
Hmm, I see what you're saying. 

On the game observation page we sorta solved this by having a filter icon that expands the filter options, though I'm not sure where such an icon would live on this page. 

Maybe for now, what about having those filters span the width of the automatch and seek graph, maybe 5 columns of two rows , and just have the groups wrap if there's not enough room (ie on mobile)?

bhostetler18:
Sounds like a good idea! I’ll try it out.

I do wonder if it’s worth keeping the current layout on mobile, since it seems about as compact and nice-looking as possible:

74D9DADE-E06F-4FD6-9B4E-72EF29625B21

anoek:
Yeah that sounds good

PR #2054, opened 2022-10-22
useData()

benjaminpjones:
This is a React Hook that does all the watch/unwatch stuff that one would ideally want to call when interacting with the data store.

## Proposed Changes

  - add useData()
  - replace the `get`/`watch`/`unwatch` in `<Experiment />` as an demonstration
  - use the new function in `useUser()` as well
  
## Open questions

- undefined values
    - Current behavior: may be undefined if the value is not set in the store, however the state variable will not be set to undefined if the value is unset during the lifetime of the component.
    - Do we want to have a `default_value` parameter like `data.get()`?  A `call_on_undefined` parameter like `watch`?

### DIFF
--- a/src/components/Experiment/Experiment.tsx
+++ b/src/components/Experiment/Experiment.tsx
@@ -16,7 +16,7 @@
  */
 
 import * as React from "react";
-import * as data from "data";
+import { useData } from "hooks";
 import { Default } from "./Default";
 
 interface ExperimentProps {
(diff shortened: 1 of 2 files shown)


anoek:
Slick.

Would it be canonical to return `[value, setter]` instead of just the value?

I feel like providing a default value would be a good thing. It's unclear to me if call_on_undefined would be applicable in this case, my gut says it's not, but I haven't thought through it.

benjaminpjones:
> Would it be canonical to return [value, setter] instead of just the value?

Probably a plus to have this align more with `useState()`.  The main downside I see is that it can be kind of easy to forget the array destructuring when you're only getting the value:

```
// silently wrong
const selected = useData(...);

// right
const [selected] = useData(...);
```

I'm honestly not sure which is the right way to go.  With `usePreference()`, I did include the setter 🤷‍♂️.  Anyway, I changed it in the most recent commit and added a `default_value` parameter which guarantees a not-undefined return!

PR #2057, opened 2022-10-26
Migrate to use new computeAverageMoveTime function

bhostetler18:
Average time-per-move calculations now consider the board size. The most important side-effect is that `isLiveGame` also considers the board size, which makes sense.

This PR also aligns the seek graph more closely with the site definitions of blitz, live, and correspondence games.


anoek:
Thanks @bhostetler18 !

PR #2061, opened 2022-11-04
Bug Fix: Resend validation email button missing.

benjaminpjones:
Fixes issue discussed at [OGS Forums: Account Verification](https://forums.online-go.com/t/account-verification/45360/6). The regression likely happened in 1722cd6fb33de4f46eb13754b6ef6215747227a9, although I didn't actually go back and run it.

The issue is that this message/button is missing:

Screen Shot 2022-11-04 at 12 41 45 AM

## Proposed Changes

  - Fix Bug
  - Write a test that would have caught said bug


anoek:
Thanks!

PR #1865, opened 2022-05-20
Fix Original Game link missing from reviews.

benjaminpjones:
Fixes #1864

Screen Shot 2022-05-20 at 1 53 50 PM

PR #2062, opened 2022-11-05
Bug fix: Link to review SGFs if both game_id and review_id are present.

benjaminpjones:
Fixes issue from [OGS forums: Some reviews do not have "SGF with comments"](https://forums.online-go.com/t/some-reviews-do-not-have-sgf-with-comments/45362/3).  Regression likely occurred in https://github.com/online-go/online-go.com/pull/1865.

- Bug: "SGF with comments" button was missing from reviews.
- Root cause: `game_id` is present on reviews, because `game_id` is used to link to the original game.

## Proposed Changes

  - Remove `game_id` and `review_id` from `GameDockProps`, and get these values from the goban.
  - Write a test that fails on the bug.
  - Fix the bug.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -1462,8 +1462,6 @@ export function Game(): JSX.Element {
                         {((view_mode === "portrait" && !zen_mode) || null) && (
                             <GameDock
                                 annulled={annulled}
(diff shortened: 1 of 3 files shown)


anoek:
Thanks!

PR #2063, opened 2022-11-05
chore: add npx husky result with git-lfs

PabloLION:
After running commands in https://github.com/online-go/online-go.com/blob/devel/CONTRIBUTING.md
I got 4 new files.

Not sure what to do:

1. remove them and change husky settings
2. add them to .gitignore
3. merge them to `devel` branch


benjaminpjones:
Hmm, I noticed those showed up in the other repo as well, I didn't realize they were generated files.  Apparently git-lfs does this when it detects git-lfs files and husky? https://github.com/typicode/husky/issues/926

Anyway seems okay to merge to `devel` (at least that's what was done in https://github.com/online-go/goban/commit/6cba6b0aafe037e161abf358731f1b8619f1310e)

anoek:
I'm not sure what the correct thing to do is either, but having them checked in doesn't seem to have broken anything on goban, so sure lets do it :)

PR #2064, opened 2022-11-05
fix: add visual feedback for 500 error

### LINKED ISSUE #50: Loading games with missing or corrupted data should display a useful error message
### Summary

[This (old) game](https://online-go.com/game/437044), appears to be eternally loading

### Environment

 * Operating system: Ubuntu 16.04.1
 * Browser: Chrome
 * Browser Version: Version 54.0.2840.100 (64-bit)

### Steps To Reproduce:

1. Go to game linked above
2. Do nothing
3. The board appears superimposed with the message "Loading..."

### Additional Information:

The game chat doesn't work either, Zen mode is buggy and so are most of the right panel options.

PabloLION:
Fixes #50

This commit only adds a visual feedback and doesn't change anything else.

## Proposed Changes

- Add an `alert` only on game file not found.

## Possible Improvements
- refactor the `.then`/`.catch` with `try..catch`+`async`/`await`
    The `async`/`await` is a syntactic sugar doesn't change any functionality (so we don't lose anything). Yet they offer SLIGHTLY more readability and improve debugging experience. With the current implementation, when debugging, we have to split the `.then()` to many `.then().catch().then().catch()`... and in comparison with `try..catch`, we just remove the outer `try{}` and we can see clearly where the error is thrown. But the change can be too so subtle for the work needed.
- improve error handling when web socket returns a 404 / 500: they should have consistent behaviour
    Currently, the https://github.com/online-go/online-go.com/blob/a0cb2908be96f79d1f0b30f8934cb7623f3c90e2/src/main.tsx#L281-L282 (from https://github.com/online-go/online-go.com/issues/50#issuecomment-1304527096) handles 404 with a 10 second delay correctly, but the 415/500 was not dealt with the same procedure. And I reckon this as the real reason that causes #50. We can either extract the error handling function from the `errorAlerter` or handle all WS errors in in `sockets`. (I'm not very sure about this, just feeling the inconsistency.)

Does any of these two worths creating an issue?

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -1302,7 +1302,21 @@ export function Game(): JSX.Element {
                         goban_div.current.removeAttribute("data-tournament-id");
                     }
                 })
-                .catch(ignore);
+                .catch((e) => {
+                    if (e.statusText === "abort") {
+                        console.error("Error: abort", e);
+                        return;
+                    }
+                    if (e.statusText === "Not Found") {
+                        console.error("Error: not found, handled 10s later by socket.ts", e);
+                        return;
+                    }
+                    console.error(e);
+                    void alert.fire({
+                        title: "Failed to load game data: " + e.statusText,
+                        icon: "error",
+                    });
+                });
         }
 
         if (review_id) {


benjaminpjones:
Looks good!

> - refactor the try..catch with async/await

I didn't see any `try..catch` or `async`/`await` in the diff.  Is there a missing commit?

PabloLION:
@benjaminpjones my understanding of "Proposed Changes" was like future todo..  But when I see I've misunderstood reading it again. I've edited the PR description.

benjaminpjones:
@PabloLION oh I see!  That makes sense, and seems like a valid interpretation of "Proposed Changes".  Anyway, thanks for clarifying the description :)

anoek:
Thanks @PabloLION ! And if you feel like doing those other two things, those are welcome improvements too :)

PabloLION:
@anoek I think the refactor is kinda needed and TBH I like refactoring myself for the fact that it makes it clearer for me how the code works. 
But the refactor requieres more awareness of the structure so I'm afraid that I have to delay it after 2 or 3 PRs.

For the second idea of handling web socket error, I want to know how the backends works but I didn't see any repo in this organization.

But I can convert these two ideas to issue if needed. Also, can you help with some recommendation on my next issue? @benjaminpjones @anoek 

And something off topic: https://github.com/online-go/gtp2ogs/issues/383

PR #2065, opened 2022-11-06
Display ool match name

GreenAsJade:
Fixes OOL players not being clear if they are at the correct match

## Proposed Changes

Display the match name provided by the OOL.


GreenAsJade:
Ooops, hadn't intended to append this to #2061

PR #2066, opened 2022-11-08
Hide moderator tools

Sofia-mal:
I hope this works!

It's not a necessary thing but it wouldn't hurt, I guess. 

Edit: I don't understand this error, close it if it is not trivial!


anoek:
There's a lot of other places that would need to be updated for that option to work as expected, so closing for now until someone wants to tackle it.

Issue #1378, opened 2021-02-01
Automatch blitz with Fischer time counted as a live game


TeXitoi:
**Describe the bug**

I've selected Fischer time in the preferences of the blitz game. I've done such a game, and it appears as a live game in the history and rank calculation. See for example https://online-go.com/game/30655665

**To Reproduce**

1. Go to 'play'
2. Click on 'settings' of 'Quick match finder'
3. Blitz Time control "prefer Fischer"
4. close preferences
5. blitz automatch
6. The time limit will be "Fischer: Clock starts with 30 seconds and increments by 10 seconds per move up to a maximum of 1 minute."
7. Play the game
8. Look at the game in your game history, it is "live" and not "blitz".

**Expected behavior**
When playing a blitz game with automatch, the game is counted as blitz.

shinuito:
It would be nice if these issues were fixed.

When the rating system changed in January it reclassified all games, so many games that were blitz got bundled into the live ratings instead, which I think was probably a bad idea for anyone who cares about these individual ratings instead of the overall.

Anyway aside from that I believe that 

https://github.com/online-go/online-go.com/blob/8a8d763db1ecc71465d31ac69eac2836932c4a00/src/components/TimeControl/util.ts#L168

and 

https://github.com/online-go/goban/blob/3897a698b130c9c6c21f1c01eb5ae0238728ffe2/src/GoUtil.ts#L217

are the lines where these thresholds are defined.

I've made some test unranked games and I think the calculations are pretty poor the way they are now for classifying Blitz games and I've logged some examples in the forums thread.

https://forums.online-go.com/t/blitz-game-problem/38453/16?u=shinuito

Basically from what I can see, there's an arbitrary `90` in the formulae, for calculating the time per move. Maybe this was some simplification to make time settings consistent across board sizes? At least it was mentioned that this number used to be different for different board sizes.

Anyway the `90` is arbitrary, but the more important issue I believe is how one can get very silly games classified as "Live" and "Blitz".

For example the current automatch settings for "Blitz" use 10s byo-yomi and since there's a strict <10s in the code, none of these games, the automatch or the sitewide "Blitz" tournaments count as Blitz.

Furthermore one can have silly things like 5m+1x6s byo-yomi and 0s+100x9s counted as Blitz but 0+1x10s counted as live, since the code doesn't take into account the number of byo-yomi periods.

I haven't thought about or tested the Canadian byo-yomi one, but if I have more time I'll think about it.

[...]

PR #2080, opened 2022-12-13
TimeControlPicker/ChallengeModal improvements

bhostetler18:
There were a number of issues with the old `TimeControlPicker`, including but not limited to:
- When making a lot of changes to drop-downs in the `ChallengeModal`, it was possible to get into malformed but hard-to-reproduce states with nonsensical time control settings including:
  -  Pause on weekends on a live/blitz game
  -  As discovered by @honganhkhoa, challenges that had conflicting "system" and "time_control" values (e.g. "simple" and "byoyomi" on the same game)
- The "force system" behavior that occurs for casual rengo games was buggy: when checking "casual" after a system other than "simple" was initially selected, the system drop-down would be disabled but wouldn't show that the time control had in fact been changed to "simple" (until you unchecked "casual" or tried to make a change to the time control properties).
- A similar issue occurred when changing the system to "live" after previously selecting "correspondence" and "none" as the system

In order to fix these issues, `TimeControlPicker` has been rewritten as follows:
- `TimeControlPicker` is now "dumb" (stateless and pure functional component) and simply displays the time control that is fed to it. This time control object is immutable from the perspective of the `TimeControlPicker`.
- Any changes to the time control are fed directly to the enclosing component after passing through a series of well-defined "update" functions. These update functions perform validation on the time control and can easily be modified as new validation requirements or issues are discovered.
- Repeated JSX code has been condensed considerably and TypeScript types have been fully implemented where possible

This is still a work in progress, and similar design patterns should be applied to `ChallengeModal` (I have started to do so, but only in code directly related to time control settings).

These changes (paired with the previous PR #2057) should make future UX improvements easier, especially regarding inconsistencies between the selected speed bracket and how the server actually classifies a game (similar to issue #1378). They should also make testing easier, and I have added a few example tests to be expanded later as we continue to figure out the best time control UX.


bhostetler18:
The failing test is intentional: currently, the default time settings for a given speed bracket are not always treated as that speed bracket by the server. For example, the default for a blitz game is 
```
{
  "initial_time": 30,
  "time_increment": 10,
  "max_time": 60,
  "pause_on_weekends": false,
  "speed": "blitz",
  "system": "fischer"
}
```
but the server treats this as a live game on 9x9 boards. The failing test reflects this.

src/components/TimeControl/TimeControlPicker.tsx
```diff
@@ -16,805 +16,309 @@
...
+                    key={"tc-stones-per-period"}
+                    id={"tc-stones-per-period"}
+                    name={_("Stones per Period")}
+                    property={"stones_per_period"}
+                    min={default_time_settings[tc.speed].canadian.stones_per_period_min}
+                    max={default_time_settings[tc.speed].canadian.stones_per_period_max}
+                    default={default_time_settings[tc.speed].canadian.stones_per_period}
+                    valueGetter={(ev) => parseInt(ev.target.value)}
+                    onChangeProperty={onChangeProperty}
+                ></TimeControlPropertyInput>,
```
anoek, line 114: Minor nit: `/>` instead of `></TimeControlPropertyInput>`

anoek (commented):
This is awesome, thanks for taking this on! Very glad to have this refactored and cleaned up.

src/components/TimeControl/TimeControlPicker.tsx
bhostetler18: Will change it! Sloppy autocomplete on my part.

We should probably continue to discuss potential time control changes on Slack when you have time, since this rewrite currently only highlights rather than fixes mismatches between user selections and server treatment.

anoek:
Is this still a work in progress or should we bump it out of draft mode and get it merged?

bhostetler18:
> Is this still a work in progress or should we bump it out of draft mode and get it merged?

I have left it in progress since it highlights but does not fix buggy behavior (the mismatch between server and client speed classifications). There is also an intentional failing test. Both can be resolved, but probably warrant more discussion since any fixes will involve larger functionality changes that I didn’t want to make unilaterally.

benjaminpjones:
Would it make sense to split the PR into refactor and fixes/functional changes?  Most of this PR doesn't functionally change the challenge modal right?  I see the failing test, but it might be easier to review and discuss the proposed "larger functionality" changes if it's 100 lines of code instead of 1000.

bhostetler18:
> Would it make sense to split the PR into refactor and fixes/functional changes? Most of this PR doesn't functionally change the challenge modal right? I see the failing test, but it might be easier to review and discuss the proposed "larger functionality" changes if it's 100 lines of code instead of 1000.

Definitely! Wouldn’t require too much splitting either – as you noted there aren’t really any functionality changes (other than the bug fixes and warning text). I can remove the failing test and warning text so we can get the bug fixes and refactored code into production.

anoek:
That sounds great, and I agree I think it's a good idea to get the refactor merged and then move on to improvements and other fixes.

bhostetler18:
Done! Let me know if there's anything else you'd like me to do before this gets merged.

anoek:
Thanks a ton for the hard work on this refactor! It looks great, and I'm very glad to have this switched over to be a functional component.

PR #2084, opened 2022-12-20
Add display of "See also" info in OJE

GreenAsJade:
## Proposed Changes

Use the `see_also` field from the server to tell users that there are other identical positions in OJE

Needs https://github.com/online-go/ogs/pull/1768


anoek:
Awesome, thanks!

PR #2089, opened 2022-12-26
Fix player not being able to pass if first move and white.

### LINKED ISSUE #2088: Pass button is not shown if white to play on first move
**Describe the bug**
Pass button is not shown if white to play on first move.

This is not an issue for most games, but could happen in two scenarios: Japanese handicap games and forked games.  Usually people don't want to pass on the first move, but it is still usually an option.  In my case, this bug interferes with debugging #2086, when I tried to fork and go to scoring.

**To Reproduce**
1. Create a Japanese game, with handicap.  Set your color to White.
2. Note that the pass button is missing, even though it says "your move".

**Expected behavior**
Pass button should be shown.

**Screenshots**
Screen Shot 2022-12-26 at 12 27 16 AM

benjaminpjones:
Fixes #2088 

## Proposed Changes

  - Add a test for this
  - Attempt at a fix by using playerToMove() instead of last_official_move
      - Note: I'm not confident this won't break something else. Commit 6fe7e8f88cf8184bf8f01200e1621db00dc26bdb introduced `real_player_to_move`, and I assume there was a reason for it.  Hopefully this at least helps pinpoint the error even if we go with a different fix.

### DIFF
--- a/src/views/Game/PlayButtons.test.tsx
+++ b/src/views/Game/PlayButtons.test.tsx
@@ -488,4 +488,29 @@ describe("PlayButtons", () => {
 
         expect(screen.queryByText("Accept Undo")).toBeNull();
     });
+
+    test("forked game where it's the first move, white to play, my move.", () => {
+        const goban = new Goban({
+            initial_player: "white",
+            players: {
+                black: { id: 456, username: "test_user2" },
+                white: { id: LOGGED_IN_USER.id, username: LOGGED_IN_USER.username },
+            },
+        });
+
+        render(
+            <GobanContext.Provider value={goban}>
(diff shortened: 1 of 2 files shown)


benjaminpjones:
The most recent change doesn't change behavior, but it is good practice to use the goban hooks rather than getting properties from Goban directly.

anoek:
I sure wish I left myself a better commit message for that change I did, I can't recall what I was trying to fix with 6fe7e8f88cf8184bf8f01200e1621db00dc26bdb

Anyways, thanks for hunting this down and fixing my bug! I'll put it up on beta and maybe whatever issue I was trying to fix will make itself apparent.

PR #2096, opened 2022-12-31
PlayControls: unsubscribe from goban events after unmount.

benjaminpjones:
Fixes memory leaks caused by subscribing to goban without cleaning up. In actual practice, this might happen when resizing the window.  Unlikely to have significant memory impact, but a leak nonetheless.

## Proposed Changes

  - Write a test to catch memory leak caused by `PlayControls` subscription.
  - Add unsubscribes to the cleanup function.

## Dependency

Unit test relies on https://github.com/online-go/goban/pull/107

### DIFF
--- a/src/views/Game/GameHooks.ts
+++ b/src/views/Game/GameHooks.ts
@@ -45,17 +45,31 @@ export function generateGobanHook<T, G extends GobanCore | null>(
                 return;
             }
 
-            const events_with_load: Array<keyof GobanEvents> = ["load", ...events];
-            for (const e of events_with_load) {
-                goban.on(e, syncProp);
-            }
-            return () => {
-                for (const e of events_with_load) {
-                    goban.off(e, syncProp);
-                }
-            };
+            return subscribeAllEvents(goban, events, syncProp);
         }, [goban]);
         return prop;
+    };
+}
+
+/**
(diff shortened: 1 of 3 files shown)


benjaminpjones:
While I do expect that type error in the test, I'm a little surprised it's showing up in the webpack step.  Do I need to update the webpack config to ignore `.test.tsx` files?
 
Screen Shot 2022-12-31 at 2 31 20 PM

anoek:
I think it might be the `ts-loader` module, so I reckon that's just kind of part of the process of getting things ready for webpack in terms of compiling the TS files into JS. It does look like there's ways to excluding those files though if we want to. I'm not sure how necessary it is though, we often have compilation errors until we bump the goban library, so running into that same issue with test files doesn't seem to be anything new, and type checking our test files seems like a good thing too so I reckon it's fine as is?

I did double check to make sure those files weren't ending up in the final build, and as far as I can tell (grepping through the generated map files), it doesn't look like it so I think we're in good shape.

And thanks for finding and fixing the leak, and the cleanup!

benjaminpjones:
Okay cool, as long as they aren't ending up in the bundle, I don't think it's a problem to get this check on the web pack step.

Thanks for taking a look and merging!

PR #2099, opened 2023-01-02
Promote V6 navbar to production

anoek:
This makes the new "V6" navbar the default (and only) navbar.

@GreenAsJade - This also updates the help flows, removing the legacy flow and renaming the v6 flow. It seemed to work with my testing, but might be worth a once over before it goes to production.


kevincoleman:
Thanks for seeing this change through! I know it’s a small thing, but I do enjoy trying to contribute to a platform I love. Much appreciated. :)

benjaminpjones:
Woohoo, excited for the 100% rollout! Thanks @anoek and @kevincoleman for working on this!

PR #2108, opened 2023-01-10
Load last-used time control setting when creating a custom challenge

### LINKED ISSUE #2103: Time control settings are not saved after making a challenge
**Describe the bug**
When creating a challenge, the previously used settings should be the default values. This is no longer the case .

Likely related to #2080 

**To Reproduce**
Create a challenge, changing the time settings. Create another challenge, notice the time settings have not been saved.

bhostetler18:
Fixes #2103


anoek:
Thanks for the fast turnaround!

bhostetler18:
No problem! Oversight on my part.

PR #2112, opened 2023-01-11
Get rid of `profile-edit-link` help item reference from User.tsx...

GreenAsJade:
...  it is no longer a thing.

Fixes - being forced to wrap that in a try to make things work.

## Proposed Changes

Don't refer to `profile-edit-link`, because it was a old nav system thing.


GreenAsJade:
Note: this will be followed up with a helper system fix to make it give a proper error message when this happens.

anoek:
Thanks!

PR #2118, opened 2023-01-15
Keyboard shortcut to focus the search input in the navigation bar.

mmartin:
The old navigation bar had a keyboard shortcut to open the left pane and automatically focus the search input field.

Even though the new navigation bar has the search input always visible, I'd still like to have a keyboard shortcut for focusing it.

## Proposed Changes
  - Add keyboard shortcut for focusing search input.


anoek:
I too have missed that :) I suppose in hindsight that's a bit of a nonstandard key, but I don't think it gets in the way of normal usage.

PR #2119, opened 2023-01-18
Further prevent users from selecting time control values outside the allowable min/max range

### LINKED ISSUE #1628: Players can create games with null x 10s byo-yomi

bhostetler18:
Fixes #1628 (see https://forums.online-go.com/t/0-periods-byo-yomi/40409/ as well) and cleans up some unused code.

The two properties affected were "periods" on Byoyomi games and "stones_per_period" on Canadian games.

There were already some checks to ensure proper values, but it seems some users were (accidentally or otherwise) pressing "create challenge" before the value could be clamped to the allowable range. As with previous time control picker bugs, these malformed settings had the potential to be persisted in local storage and come back again and again.

Note that even after this fix, users can still create malformed challenge requests (not using the UI), so there should still be more validation on the backend. There may be validation already for "stones_per_period", but it appears there isn't any on "periods".

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -39,8 +39,6 @@ import { PlayerIcon } from "PlayerIcon";
 import {
     shortShortTimeControl,
     isLiveGame,
-    getDefaultTimeControl,
-    TimeControlTypes,
     TimeControlPicker,
     timeControlSystemText,
     TimeControl,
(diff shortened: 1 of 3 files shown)


anoek:
Interesting, we have this line in our backend sanitization `ret["periods"] = min(300, max(1, int(time_control_parameters["periods"] or "0")))` so we were certainly *trying* to ensure that value wasn't  something too crazy, but obviously something isn't right if we were seeing `0` get all the way through. I'll look into that more.

Regardless, thanks for the PR!

PR #2121, opened 2023-01-20
Update UserByName tests to work with React Router 6

benjaminpjones:
Fixes my OCD about commented code ;P

## Proposed Changes

  - Use a [MemoryRouter](https://reactrouter.com/en/6.7.0/router-components/memory-router) to set router params per this StackOverflow thread: https://stackoverflow.com/questions/58117890/how-to-test-components-using-new-react-router-hooks/58206121#58206121
  - use react testing library's render method instead of React's deprecated render function.


anoek:
Yay, thanks!

PR #2123, opened 2023-01-22
Update Report.tsx

Sofia-mal:
Small changes.

### DIFF
--- a/src/components/Report/Report.tsx
+++ b/src/components/Report/Report.tsx
@@ -1,4 +1,3 @@
-
 /*
  * Copyright (C)  Online-Go.com
  *
--- a/src/components/Report/Report.tsx
+++ b/src/components/Report/Report.tsx
@@ -28,6 +28,7 @@ type ReportType =
     | "stalling"
     | "inappropriate_content"
     | "score_cheating"
+    | "harassing"
     | "ai_use"
     | "sandbagging"
     | "escaping"
@@ -71,8 +72,8 @@ export const report_categories: Array<ReportDescription> = [
         type: "inappropriate_content",
         title: pgettext("Report user for inappropriate content", "Inappropriate Content"),
         description: pgettext(
-            "Report user for inappropriate content or harassing another user",
-            "User is posting inappropriate content or harassing another user.",
+            "Report user for inappropriate content",
+            "User is posting inappropriate content.",
         ),
         min_description_length: 20,
     },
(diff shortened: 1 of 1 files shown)


BHydden:
I think harassment is usefully distinct from inappropriate content and doesn't need to be removed.

Rest of the changes look pretty good though 👍

Sofia-mal:
I think I'm giving up. If it is easy to fix and makes sense do it, otherwise close it. 

Regarding "inappropriate content" and "harassment" I think that making only one category makes sense. I wouldn't know which one to choose for a user that swears in chat. But 2 is fine too I guess.

Sofia-mal:
I thought it was an easy thing to do from the web. It's a space/header problem again though. Sorry.

BHydden:
Before the first error, one line was fine but you needed a space after the comma. I think that is the only problem code-wise? 🤔

BHydden:
Re: types as I see it

Inappropriate content: one bad isolated incident (such as swearing or other abuse)
Harassment: prolonged behaviour where each incident might not be bad in isolation but represents a pattern of negative directed behaviour

Sofia-mal:
> Before the first error, one line was fine but you needed a space after the comma. I think that is the only problem code-wise? 🤔

I don't know, but I am not going to try again today.



> Re: types as I see it
> 
> Inappropriate content: one bad isolated incident (such as swearing or other abuse) Harassment: prolonged behaviour where each incident might not be bad in isolation but represents a pattern of negative directed behaviour

It makes sense, but then the cases of 'harassement' are very rare indeed and perhaps can be included in 'other'. 
I have seen several confused users, now that I think about it. And some  still use "stalling" for escaping cases.

Sofia-mal:
Finally seems ok. Sorry for my stupidity.

benjaminpjones:
Argh it shouldn't be this hard. Prettier/formatting is great for adding consistency to the codebase, but I see that it's painful for folks who are submitting through the web interface. Sorry for the trouble @Sofia-mal !

One idea: I recently submitted a [PR to typeshed](https://github.com/python/typeshed/pull/9324), and one of the cool things they had was a bot that automatically fixed lints and appended a commit. I can dig into that and see if it's easy enough to set up for this repo.

anoek:
That sounds pretty slick

anoek:
Thanks for the wording improvements @Sofia-mal !

PR #2128, opened 2023-01-26
Made right navbar drop down a hover drop down like the rest of the menu items

anoek:
This changes the right drop down menu to work on hover instead of clicking it

image

This only applies to desktop, in mobile mode it presents an icon that is clickable as before

image

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -378,40 +378,7 @@ export function NavBar(): JSX.Element {
                             to={`/user/view/${user.id}`}
                             className="profile desktop-only"
                         >
-                            <Link to={`/user/view/${user.id}`}>
-                                <PlayerIcon user={user} size={16} />
-                                {_("Profile")}
-                            </Link>
-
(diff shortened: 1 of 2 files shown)


src/components/NavBar/NavBar.tsx
```diff
@@ -362,17 +363,56 @@ export function NavBar(): JSX.Element {
...
+                            title={
+                                <span className="icon-container">
+                                    <PlayerIcon user={user} size={64} />
+                                    <span className="username">{user.username}</span>
+                                </span>
+                            }
+                            to={`/user/view/${user.id}`}
+                            className="profile desktop-only"
+                        >
+                            <Link to={`/user/view/${user.id}`}>
```
benjaminpjones, line 381: Given that these are the same contents as desktop (and we don't want them to go out of sync) it might be a good idea to break this into its own component and using it in both places.

benjaminpjones (commented):
Just tried it out, I like it!

src/components/NavBar/NavBar.tsx
anoek: Yep totally

PR #2132, opened 2023-01-30
create preview environments for every PR using Uffizzi

waveywaves:
This PR adds github workflows which will trigger uffizzi preview environments for PRs to this repo. A PoC has been created at https://github.com/waveywaves/online-go.com/pull/1. Once this PR is merged, you should see comments like https://github.com/waveywaves/online-go.com/pull/1#issuecomment-1406198574 pinned to all subsequent PRs to this repo. These comments are created after a preview env has been deployed for the PR.

To understand this PR better do check this article below on two stage workflows
https://www.uffizzi.com/preview-environments-guide/github-actions-two-stage-workflow

fixes https://github.com/online-go/online-go.com/issues/2124

### DIFF
--- a/.github/uffizzi/docker-compose.uffizzi.yml
+++ b/.github/uffizzi/docker-compose.uffizzi.yml
@@ -1,0 +1,38 @@
+version: "3"
+
+x-uffizzi:
+  ingress:
+    service: online-go
+    port: 8080
+
+services:
+  online-go:
+    image: '${ONLINE_GO_IMAGE}'
+    container_name: online-go
+    ports:
+      - "8080:8080"
+    entrypoint:
+    - "sh"
+    command:
+    - "-c"
+    - "npm run gulp"
+    deploy:
+      resources:
+        limits:
+          memory: 2000M
+
+  online-go-server:
+    image: '${ONLINE_GO_IMAGE}'
+    container_name: online-go
+    ports:
+      - "8080:8080"
+    entrypoint:
+    - "sh"
+    command:
+    - "-c"
+    - "npm run gulp dev-server"
+    deploy:
+      resources:
+        limits:
+          memory: 4000M
+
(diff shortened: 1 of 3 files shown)


anoek:
tl;dr - I think you might be looking for `npx gulp dev-server` which would start just the server portion and serve the externally built bundles and other files.

As you have undoubtedly figured out, the way this repo works is by building the client .js bundle with Gulp and then ,currently embedded within the Gulp runtime, there is also a dev server which basically just serves that bundle and proxies any API requests out to the beta server, this is how we get around browsers complaining about different domains and in general just keeping things consistent between dev, beta, and production. That dev server code is all here: https://github.com/online-go/online-go.com/blob/devel/Gulpfile.js#L167-L364 with a couple of supporting functions underneath it, but all in all, it's pretty simple and there's nothing "gulp" about it, we just bolt it on to the gulp process to streamline the whole process for our contributors.

I'm assuming if you're bumping into resource constraints I'd guess Gulp and all of it's build subprocesses are what's eating up resources. If you wanted to build those assets and then have a separate process come along and do the server / proxy stuff I think there are two options: 
  1) run `gulp dev-server` which will just start that server part and should be pretty light weight as it shouldn't load up any of the build stuff into memory
  2) if that's still too heavy for whatever reason, split out that dev server code into another file and just run it as a node server without Gulp.

I don't see any reason why you couldn't do either of those within a docker image if that's easier for your workflow.

anoek:
I'm going to mark this as draft since you labeled it as `WIP`, but if merging it now would help in your development efforts I'm happy to press the button. Otherwise just mark as ready for review whenever you're ready!

anoek:
Very cool, thanks for all of your work @waveywaves!

anoek:
@waveywaves  I'm getting an error "No compose file passed and no deployment found. No operation." : https://github.com/online-go/online-go.com/actions/runs/4064987419  - any ideas?

PR #2133, opened 2023-01-31
Update Go Commentary video link

erictleung:
Original link is broken as the domain name lapsed. The new link points directly to YouTube Channel instead.

## Proposed Changes

- Updates Go Commentary video link


anoek:
Thanks for the fix!

PR #2134, opened 2023-01-31
Adds "next game" arrows on PlayerCards

GreenAsJade:
For moderators to step through a person's games

## Proposed Changes

arrow-buttons on PlayerCards, for moderators, which go to the player's next and prev games
moderator-option to hide these arrows, which default  on

**Needs https://github.com/online-go/ogs/pull/1780

**Improved by:  https://github.com/online-go/ogs/pull/1781  (sort by ended, started instead of id)

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -130,6 +130,7 @@ export const defaults = {
             visible: boolean;
         };
     },
+    "moderator.hide-next-game-arrows": false,
 
     "table-color-default-on": false,
 
--- a/src/views/Game/PlayerCards.styl
+++ b/src/views/Game/PlayerCards.styl
@@ -3,3 +3,10 @@
         color: #FF8B00;
     }
 }
+.next-game-arrows {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    margin: 0 0.5rem 0 0.5rem;
+    align-items: center;
+}
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -16,12 +16,15 @@
  */
 
 import * as React from "react";
+import { get } from "requests";
+
 import { Goban, GobanCore, PlayerScore, JGOFPlayerSummary } from "goban";
 import { icon_size_url } from "PlayerIcon";
 import { CountDown } from "./CountDown";
 import { Flag } from "Flag";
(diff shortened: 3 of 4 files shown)


BHydden:
What did we ever do to deserve you? ❤️❤️❤️

Side note... how did none of us think of this until now? 😅

BHydden:
This should probably work fine for live players, but IINM this steps through by `$game_id` not time completed which might be confusing for corr since games don't necessarily finish in the order they started in? 🤔

GreenAsJade:
It's a good point (and a good insight: how did you know that it steps through game id?!)

One could argue "that's the design, they step through in the order they were created"  😆

Does that lead to problems?  I suppose one thing is that it certainly results in not doing it in the order of GameHistory.

Or one could argue that "actually, I don't have to use `game_id` even though at one point I thought I was forced to`.

I guess that for the front end ( this PR ) it makes no difference - it's a back end change that can be made separately.   I'll have a look at that.

BHydden:
> how did you know that it steps through game id?!

Umm... i read your code changes and two lines say get game/game id 🤷‍♂️



> not doing it in the order of GameHistory

Yeah, hard to say how "bad" that will be, but that's what I thought of

GreenAsJade:
> > how did you know that it steps through game id?!
> 
> Umm... i read your code changes and two lines say get game/game id 🤷‍♂️

Hah!  A fascinating case of wrong reason leading to right answer 😆

> > not doing it in the order of GameHistory
> 
> Yeah, hard to say how "bad" that will be, but that's what I thought of

Now we have two options in backend PRs, one with gameid-stepping and the other with ended/started stepping (the latter seems better but may be more expensive on the server).

BHydden:
Fascinating indeed! Care to elaborate? 😅

GreenAsJade:
`void get(`games/${game_id}/prev/${player.id}`).`   

This is asking the back-end "please give me the previous game played by `player.id` starting from `game_id`"

But it doesn't stipulate how "previous game" should be calculated - that is up to the back end.

(And this request remains unchanged which ever solution anoek choses for the back end)

BHydden:
Huh. There you go! I thought previous game could only ever be previous game id... neat! Thanks for explaining

anoek:
Pretty slick, thanks!

GreenAsJade:
The trick is that the previous game id will be the game that was created immediately previous to this one.

But it's not the previous game _for this player_ ... :)

PR #2135, opened 2023-01-31
Show report centre link to moderators only

GreenAsJade:
Fixes fascinating link that does nothing appearing for users.


anoek:
Thanks for the fix!

PR #2136, opened 2023-02-01
Testing out uffizzi previews

anoek:
Add a little color to the site

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -8,6 +8,7 @@ let BACKEND = process.env.OGS_BACKEND || "BETA";
 const spawn = require("child_process").spawn;
 const fs = require("fs");
 const gulp = require("gulp");
+const path = require("path");
 const execSync = require("child_process").execSync;
 const livereload = require("gulp-livereload");
 const stylus = require("gulp-stylus");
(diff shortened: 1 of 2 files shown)


anoek:
@waveywaves is a fast failure on "[Use Remote Workflow to Preview on Uffizzi / Create, Update, or Delete Preview on Uffizzi](https://github.com/online-go/online-go.com/actions/runs/4067471571/jobs/7004858834#logs)" with an error message "No compose file passed and no deployment found. No operation." an expected thing when first pushing? I don't see a red x here but I'm getting that job failure message still

https://github.com/online-go/online-go.com/actions/runs/4067471571/jobs/7004858834

anoek:
Looks like it's getting close to working, exciting stuff! https://pr-2136-deployment-13116-online-go-com.app.uffizzi.com/ still shows that several of the built resources aren't being served though, FYI @waveywaves

anoek:
It looks like the resources are being requested on port 8080 explicitly instead of just the same domain, however even using the same url without the port specification (which *should* work I think) is not resolving and is just returning the fall through case where we just return the index html file, so I'm thinking it can't find those static assets either. I think both of those will need to be fixed. 

I can take a stab at trying to get those resolved.

anoek:
Success! Thanks for getting all of that put together @waveywaves!

anoek:
> > is a fast failure on "[Use Remote Workflow to Preview on Uffizzi / Create, Update, or Delete Preview on Uffizzi](https://github.com/online-go/online-go.com/actions/runs/4067471571/jobs/7004858834#logs)" with an error message "No compose file passed and no deployment found. No operation." an expected thing when first pushing? I don't see a red x here but I'm getting that job failure message still
> 
> https://github.com/online-go/online-go.com/actions/runs/4067471571/jobs/7004858834
> 
> I believe you are right. I am escalating this to the team, there has got to be a better way to handle this.

FTR this doesn't seem to be happening now, maybe something got fixed along the way during a merge or something? Not sure, but regardless it seems to be working without error for these past couple of test PRs, so hooray!

Thanks again for putting together the feature, and thanks uffizzi for making the service, it will be very useful :)

PR #2152, opened 2023-02-14
Copy analysis branch to conditional move planner

### LINKED ISSUE #239: Allow branches to be instantly copied from 'Analyze Mode' to 'Conditional Move Planner', or merge the Analyzer with the Planner
It is not uncommon for me to make plans in 'Analyze Mode', and then manually copy them to the 'Conditional Move Planner', move by move. After all, if I tested branches while in the 'Conditional Move Planner', I would have to delete often.
It would be nice to be able to click a button to make a branch in 'Analyze Mode' to make the branch a conditional move plan.
If we add this feature, we may as well also show in 'Analyze Mode' which branches are conditional move plans by highlighting them and moving them closer to the main branch than everything which is not a conditional move plan.
Then may as well have a button in 'Analyze Mode' to remove a move and its branches from the 'Conditional Move Planner'.
At this point, we can remove the 'Conditional Move Planner' because all of its features have been absorbed by 'Analyze Mode'. Another advantage to merging the Analyzer and the Planner is that it would allow conditional move plans that can deal with passes.

turtoise0:
Fixes #239 and https://forums.online-go.com/t/could-we-go-straight-from-analyse-to-plan-conditional/12605?u=turtoise

## Proposed Changes

  - Adds a button that copies the selected analysis branch to the conditional move planner

### DIFF
--- a/src/views/Game/PlayControls.tsx
+++ b/src/views/Game/PlayControls.tsx
@@ -1199,10 +1199,22 @@ function automateBranch(goban: Goban): void {
     // TODO: check player to move?
 
     const before = goban.conditional_tree.duplicate();
-    const tree = diffToConditionalMove(diff.moves);
+    const tree = mergeConditionalTrees(before, diffToConditionalMove(diff.moves));
+
     goban.setConditionalTree(tree);
     goban.saveConditionalMoves();
     goban.setConditionalTree(before);
     toast(<div>{_("Copied branch to the conditional move planner")}</div>, 2000);
-    // TODO: don't delete existing conditional moves
+}
+
+// Merges two conditional trees into one. If there are conflicts, the branch in
+// `b` overwrites the one in `a`.
+function mergeConditionalTrees(a: GoConditionalMove, b: GoConditionalMove): GoConditionalMove {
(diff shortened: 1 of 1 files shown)


src/views/Game/PlayControls.tsx
```diff
@@ -749,6 +750,17 @@ export function AnalyzeButtonBar({
                     <i className="ogs-label-x"></i>
                 </button>
             </div>
+            {(!is_review || null) && (
+                <div className="btn-group">
+                    <button
+                        onClick={() => automateBranch(goban)}
+                        title={_("Copy branch to conditional move planner")}
+                        disabled={goban.engine.rengo}
```
anoek, line 758: Should we add `data.get("user").id === currentPlayer(goban)` in this disabled statement too so the button is disabled if it's your move to make, or does that cause problems as we progress through the moves?

turtoise0: There shouldn't be any problems. Also, maybe we should hide the button from rengo players since the button doesn't ever get enabled in rengo games?

anoek:
Awesome, thanks @turtoise0 !

PR #2158, opened 2023-02-17
Don't talk about email validation in chat placeholder text

GreenAsJade:
Fixes misunderstandings about what really is required to chat.

## Proposed Changes

Just talk vaguely about "account validation"  😄

### DIFF
--- a/src/components/Chat/ChatLog.tsx
+++ b/src/components/Chat/ChatLog.tsx
@@ -559,7 +559,7 @@ function ChatInput({
             autoFocus={autoFocus}
             placeholder={
                 !user.email_validated
-                    ? _("Chat will be enabled once your email address has been validated")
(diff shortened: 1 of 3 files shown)


anoek:
Hmm, on the one hand the change is more accurate, but otoh talking about the email address gives the user a clear indication about what they need to do in order to chat (other than wait a day.)

GreenAsJade:
How about that!   I nearly put this change in with the other one I was doing, but then thought "hang on this is independent"!
 

😎

Yeah I agree,.

GreenAsJade:
I've remembered why this caught my attention in the first place: it's what spectators see when they arrive at the lobby of an OOL game and they aren't logged in.

And the reason for this is that `ChatLog` in particular doesn't check if we're logged in, in the way that GameChat does.

So different fix needed.

PR #2159, opened 2023-02-17
Ool spectators landing update

GreenAsJade:
Fixes confusion from players who arrive here not realising it is the spectator page.

And makes it match the player page improved layout.

## Proposed Changes

Make spectators see basically what players see, except the I'm ready button.

Needs: https://github.com/online-go/ogs/pull/1788


anoek:
Nice, thanks!

PR #2161, opened 2023-02-17
Say "reported by" instead of "reporting user"

GreenAsJade:
Fixes potential confusion about who did what

## Proposed Changes

Say "reported by" instead of "reporting user"


anoek:
:+1:

PR #2169, opened 2023-02-21
Don't show analysis-conditional button in finished games

turtoise0:
## Proposed Changes

 The "copy branch to the conditional move planner" button was being shown in finished games. This PR fixes this.


anoek:
Thanks for the fix!

PR #2174, opened 2023-02-27
Rating calc

shinuito:
## Proposed Changes

  - Adds a new page containing a Glicko 2 rating calculator linked in the nav bar under "Tools".
  - Can input rating, deviation and volatility manually to see new ratings or autofill using player autocomplete, recycled from Game History table component.
  - Code is based on python code in https://github.com/online-go/goratings/tree/master/goratings/math, but converted to javascript/typescript with some liberties taken. Global constants and configure functions were added in as parts of the class and methods - unsure if that's the best idea or not.
  - Can choose handicap between 0 and 9 inclusive, exports get_handicap_adjustment from rank_utils.
  - Unit tests aded again based on https://github.com/online-go/goratings/tree/master/unit_tests and thanks to @benjaminpjones for the CI steps.
  - gor functions currently unused by the calculator, but one could set up a gor calculator if one wanted, similarly I imagine (could be nice for comparison for instance at some point).
  - The layout and design could probably be improved by someone with better css and imagination :)

Images:

Nav bar icon:

</img>

Example form filling and calculation from beta:
</img></img> 

Random amybot-ddk game calculation comparison:

</img></img> 

Black player:
image

amybot (wins):
image


shinuito:
Based on request here https://forums.online-go.com/t/a-small-experiment-on-ogs-new-player-ranking/46810/6?u=shinuito

Generally there's a lot of questions about ratings, whether one can reach a certain rank by beating people of another rank, how much rating points you might lose to a new account and so on.

shinuito:
Good questions :)

> I noticed if either of the players are blank, clicking the "Use Players" button doesn't fill in the player that is not blank, it just doesn't do anything, would it be troublesome to make that so it just fills in the player that's provided?

I can have a look at doing that now - feels like it should be easy enough.
 
> Do we need the "Use Players" button, or would it make sense to simply fill in the current parameters if the player changes?

I guess it depends on whether one imagines mistyping and getting the wrong players, or accidentally overriding numbers on the wrong side and whether it's a hassle to refill in some of these numbers, whether they were manually typed in or not.

So more of a perspective of having to confirm a mistake vs an immediate update wipe away something you didn't want it to.

It's probably not really an issue.

> Instead of a calculate button, should we just automatically calculate if any of the parameter fields changes?

I suppose we could, whenever it's a number - ignoring partially typing in a decimal point. It might be distracting in one sense, if it keeps updating as you're typing, but in another sense it might be more fun for experimenting quickly.

> Should we default either the white or black player to the current player (if the user is logged in) ?

That's an interesting idea - one could default both even maybe, if one presumes that people likely want to compare themselves to something/someone and then they just have to delete one field.

I'll experiment with it a bit and push an update. 

Maybe another idea might be a toggle to swap the ratings to ranks like in the ratings tabels, but I'm not sure how that might work.

shinuito:
> I noticed if either of the players are blank, clicking the "Use Players" button doesn't fill in the player that is not blank, it just doesn't do anything, would it be troublesome to make that so it just fills in the player that's provided?

This should be implemented now, changing logic of the checks, and setting state conditional on what's defined.

> Do we need the "Use Players" button, or would it make sense to simply fill in the current parameters if the player changes?

I've added the `fill_player_info` function to the setState as a callback, so this updates as soon as the autocomplete completes and after the state is set I believe.

> Instead of a calculate button, should we just automatically calculate if any of the parameter fields changes?

I've added the `compute_new_ratings` function to the setStates as a callback, so this updates once the input boxes change from a manual change. I've also added the `compute_new_ratings` function as a callback within the setStates of the `fill_player_info` functions. The idea would be if the autocomplete finishes it then calls `fill_player_info` which updates the input boxes, and when that's done it computes a ratings update.

> Should we default either the white or black player to the current player (if the user is logged in) ?

I've defaulted both players to the user if they're logged in (checking if id is positive, which I hope is ok?).

Let me know how it feels testing it out anyway, and happy to make any more changes. The issue I was running into with the set states, was that I couldn't use componentDidUpdate with my functions because they get into a setState loop, so I settled on the callbacks idea.

anoek:
Very cool :) I'll get this merged soon. Thanks for making this!

anoek:
You know, the thing you're going to get asked for right away is to see the conversion to ranks too. Is that something you want to do as well, or should we deploy it and see if it's an ask first?

For those we usually just do `rankString(rating_to_rank(rating, deviation), true)` 
https://github.com/online-go/online-go.com/blob/devel/src/lib/rank_utils.ts#L211 
https://github.com/online-go/online-go.com/blob/devel/src/lib/rank_utils.ts#L77

shinuito:
> You know, the thing you're going to get asked for right away is to see the conversion to ranks too. Is that something you want to do as well, or should we deploy it and see if it's an ask first?

I didn't have much time at the weekend, mostly away from my laptop. Yeah I can have a look at it.

> For those we usually just do `rankString(rating_to_rank(rating, deviation), true)` https://github.com/online-go/online-go.com/blob/devel/src/lib/rank_utils.ts#L211 https://github.com/online-go/online-go.com/blob/devel/src/lib/rank_utils.ts#L77

I'm thinking that if this produces the r ± d type string, then I can just include it below each collection or something, rather than duplicating everything, or complicating it by making it that you can compute starting from ranks etc.

I'll have a look at it this evening probably :)

shinuito:
I've added in the conversion to ranks in the format of 

https://github.com/online-go/online-go.com/blob/2ccee72a3ce88bd0d1321687306b4983dacf00b1/src/components/ProfileCard/ProfileCard.tsx#L55-L58

and similar by using the output of `getUserRating` since that's what's used above

https://github.com/online-go/online-go.com/blob/c9b1feb0eb544c31c2a530ce3e7da6f522fdaab8/src/lib/rank_utils.ts#L137

I've tried to limit handicaps to integers. It was working fine with non-integer values, and large handicaps but I suppose it makes sense to keep it relevant :)

Let me know again what you think

BHydden:
The rating calc currently shows players losing rating points if they lose against a [?] player.

I thought I remembered @anoek saying this isn't the case?

anoek:
Well two things, one calculator is not running all of the possible scenarios when ratings aren't applied to one or another.

The second though is that might have been true some time ago but I think as of v5 you still gain/lose some points, the deviation system pretty much makes that a moot point though, which is probably why it's not a thing with v5.

BHydden:
Good to know, thanks for clarifying 👍

kobakitty:
Just noticed that the calculator isnt listed anywhere on mobile version. 
IMAGE 2023-04-01 17:17:29

BHydden:
I see it between SGF Library and Announcement Centre 🤔
Still shows for me even when not signed in.

Try refreshing? Maybe clear your cache?

kobakitty:
ah yeah.... now i see it ^_^

i kinda feel like an idiot for not trying that first xD

PR #2180, opened 2023-03-04
Say in the console why we are getting 400 from the server

GreenAsJade:
I thought declaring the `get` as void would basically quiet everything, but apparently not, so at least now it says in the console what's going on.

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -229,17 +229,25 @@ function PlayerCard({
     const show_next_game_arrows = user.is_moderator && game_id && !hide_next_game_arrows;
 
     const jumpToPrevGame = () => {
-        void get(`games/${game_id}/prev/${player.id}`).then((body) => {
-            const prev_game = body.id;
-            browserHistory.push(`/game/${prev_game}`);
-        });
+        get(`games/${game_id}/prev/${player.id}`)
+            .then((body) => {
+                const prev_game = body.id;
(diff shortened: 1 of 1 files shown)


GreenAsJade:
I think I'm not a fan of Uffuzzi comments in github issues ... it's a bit of unnecessary spam...

BHydden:
> I think I'm not a fan of Uffuzzi comments in github issues ... it's a bit of unnecessary spam...

They're not in issues, only pull requests, and the link is to a live implementation of the committed code, so you don't have to clone everyone's repo to test if a new change breaks anything.

GreenAsJade:
Yeah, I know what they are, but it's a shame that they appear as comments, which send email notification...

BHydden:
Only one per PR 🤷‍♂️ I think the benefit highly outweighs the (very slight) inconvenience personally

GreenAsJade:
No each email in this discussion is unique.

I wonder if we could throw Uffuzi into the build pipeline and have the URL always available in the build check details...

anoek:
Thanks for the clarification!

GreenAsJade:
I realised just now that the reason Uffuzi emails are a pain in the neck is because it used to be the case that when you get an email from git immediately after a PR, it's because the build failed.

Now I get an email from git, and it's "whoa - what! .... Oh, that again".

BHydden:
You don't sit on the git page waiting for the test to finish? Wow what confidence!  😅

GreenAsJade:
It takes quite a while these days...

PR #2184, opened 2023-03-08
use useUser in ChatLog (for tidier mocking)

GreenAsJade:
Fixes needing to `data.set(user` as well as mock `useUser`, at least for `ChatLog`

## Proposed Changes

use `useUser` instead of `data.get(user)`


anoek:
Thanks!

PR #2186, opened 2023-03-09
Add clear command ahead of jest in npm run test

GreenAsJade:
Fixes difficulty seeing which output is new and which is from before.

## Proposed Changes

Clear the terminal screen before running jest.


benjaminpjones:
Hmm I can see how this is useful, but I feel like devs should get to decide when to `clear` the terminal..

Some tangentially relevant numbers, ngl I'm surprised `clear &&` has as many hits as it does!

- ["test": "jest",](https://github.com/search?q=%2F%22test%22%3A+%22jest%22%2F+path%3Apackage.json&type=code&ref=advsearch) 198k hits
- ["test": "clear && jest",](https://github.com/search?q=%2F%22test%22%3A+%22clear+%26%26+jest%22%2C%2F+path%3Apackage.json&type=code&ref=advsearch) 54 hits

GreenAsJade:
Hmm - I guess it depends who "devs" is ;)

When it's "just me" writing tests, then I feel like I get to decide this by doing it the easiest way possible ;)

If it is "you and me" then let's discuss.

When a third party starts doing UI tests, and objects to the screen being cleared, let's parameterize it ... after the party 🥳

benjaminpjones:
> If it is "you and me"

Obviously it's this case 😄 I'll just come out and say it - I don't want my terminal clearing every time I run the tests.  The only reason I shared the GitHub numbers was to show I'm not a wacko for preferring that...

GreenAsJade:
Hah - now we know.

Since I never type the command, it doesn't matter what it is (I just type uparrow) ... so I gave myself a different one 😇

PR #2191, opened 2023-03-10
Suppress wearerequired/lint-action

benjaminpjones:
It's not working on forks (that was the whole point!) will look into it later


anoek:
Sounds good

PR #2193, opened 2023-03-12
Actual cool tests

GreenAsJade:
Fixes there being no actual tests of COOL UI

## Proposed Changes
* Test that COOL player landing page looks functionally correct when a person arrives.

 ** Adds jest-chain package


tsconfig.json
```diff
@@ -41,7 +41,7 @@
         "jsx": "react"
     },
     "files": [
-        "./src/main.tsx"
-    ],
+        "./src/main.tsx",
+        "node_modules/jest-chain/types/index.d.ts"],
```
anoek, line 45: I didn't know you could do this, cool.

Pedantic and slightly tangential, my only though there is if we go with this style, I think we should move the `/// <reference path="../typings_manual/index.d.ts" />` to be included in this files? Or move this and put it as a reference path in the tests (or in main.tx if you can't put reference path in tests - I'm not sure?)

I don't think I have a preference as to which is better, do you? How about you @benjaminpjones ?

benjaminpjones: Oh cool! I don't have a strong preference, but according to [this stackoverflow answer](https://stackoverflow.com/a/40861142/5001502), the tsconfig approach is better practice than `/// reference`

> What I've learned so far is that /// < reference >-ing a module with reference comments is not a good method.
>
> For example: in case you have a file Foo and a file Bar. Both files use jquery, but only file Foo has a reference comment to jquery. If file Foo is deleted for some reason, your file Bar is broken, because the reference is missing.

package.json
```diff
@@ -15,7 +15,7 @@
         "webpack-watch": "webpack --watch --progress --color",
         "lint": "eslint src/ --ext=.ts,.tsx",
         "lint:fix": "eslint --fix src/ --ext=.ts,.tsx",
-        "test": "jest",
+        "test": "`jest`",
```
benjaminpjones, line 18: Why?

GreenAsJade: Whoa, what is that?   No idea - it looks like something a prettier might have done???

tsconfig.json
GreenAsJade: followed up in #2194

PR #2194, opened 2023-03-12
Took individual file /// references out ...

GreenAsJade:
(but left collector in typings_manual)

Fixes using unrecommended /// reference approach to individual typings

## Proposed Changes

Use the tsconfig method instead.

Except for typings_manual/index , where it seems to make sense to have explicit collection of references?


anoek:
Awesome thanks, and thanks for the COOL tests too!

anoek:
Thanks!

PR #2196, opened 2023-03-13
Switch from socket.io to GobanSocket

anoek:
This PR is a work in progress as we migrate from using socket.io to communicate to the server to GobanSocket, which will give us type checked direct websocket based communications to the server. 

- [X] Type check existing socket.io implementation
- [X] Update existing messaging to reflect updates to the protocol
- [x] Updates for AI server direct connections with types

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,9 +29,12 @@ minjs:
 analyze:
 	ANALYZE=true npm run analyze
 
+test:
+	npm run test
+
 #NODE_PATH=$(NODE_PATH) PATH=$(PATH) PRODUCTION=true webpack --optimization-minimize --devtool=source-map --profile --json > analyze.json
 #npm run webpack-bundle-analyzer dist/ analyze.json
 
-.PHONY: dev min minjs mincss
+.PHONY: dev min minjs mincss test analyze pretty prettier lint-fix .husky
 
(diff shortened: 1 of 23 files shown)


anoek:
@benjaminpjones any idea's why the failing tests would be throwing an error that `GobanSocket` isn't a constructor?

benjaminpjones:
I think all these tests would be mocking socket, but I'm not sure why they would fail at the GobanSocket constructor.

When I get to a computer, I'll take a closer look!

benjaminpjones:
Still haven't figured it out, but this has the same smell as an issue I was seeing before: https://github.com/online-go/goban/pull/69

`import {...} from "goban"` wasn't working (all imports were undefined).  I fixed it by enabling UMD in the webpack config.  Did something about that change recently?

anoek:
Thanks for the pointer @benjaminpjones , looks like it didn't like this change: https://github.com/online-go/goban/commit/4d9f008cae27311a265abd749cfb75117e7a2542#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L5-L6 

Working better now though :)

benjaminpjones:
Woohoo, glad you got it worked out!

PR #2201, opened 2023-03-17
Play page rengo layout

GreenAsJade:
Fixes nothing probably :)

## Proposed Changes

Put live Rengo games before correspondence normal games on Play

### DIFF
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -102,6 +102,15 @@
         cursor: help;
     }
 
+    .challenge-row-heading {
+        // make sure these don't mess with column widths in the following rows
+        max-width: 5rem;
+    }
+
+    .challenge-row-heading.sub {
+        padding-left: 1rem !important; // cells already hav a padding-left specified
+    }
+
     .challenge-row {
         padding-top: 0.2em;
         padding-bottom: 0.2em;
--- a/src/views/Play/Play.styl
+++ b/src/views/Play/Play.styl
@@ -83,8 +83,124 @@
 
     }
 
+    // Flex to escape from the containing table formats
+    .custom-games-list-header-row {
+        display: flex;
+        font-weight: bold;
+        white-space: nowrap;
+        padding-bottom: 0.5 rem;
+        justify-content: flex-start;
+    }
+
+    .challenge-list {
+    display: table;
+    width: 100%;
+    .ineligible {
+        themed color shade3
+        font-size: 0.9em;
+        vertical-align: middle;
+        cursor: help;
+    }
+
+    .challenge-row {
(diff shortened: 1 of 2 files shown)


GreenAsJade:
This is a PR so we can let folk play with it, not proposed for Prod yet

GreenAsJade:
Looks like Uffizzi is broken:

[Use Remote Workflow to Preview on Uffizzi / Create, Update, or Delete Preview on Uffizzi](https://github.com/online-go/online-go.com/actions/runs/4451893505/jobs/7819048929#step:10:1)
Unexpected input(s) 'compose-file-cache-key', valid inputs are ['script', 'github-token', 'debug', 'user-agent', 'previews', 'result-encoding', 'retries', 'retry-exempt-status-codes']

GreenAsJade:
Curiously, it has come good 🤷‍♀️

What back-end does Uffizzi point at?

BHydden:
I swear it used to be live 🤔

Doesn't uffizzi lose a lot of its usefulness if it's not pointing at the main server?

benjaminpjones:
> I swear it used to be live 🤔

https://github.com/online-go/online-go.com/commit/f0c2cfa2db28834a3765e3dc3022fcbc2b191daf

GreenAsJade:
I can imagine that the thinking is "if you are trying something out, rather than using it in earnest, then you should not be affecting Prod data"

BHydden:
That's why it makes sense to have both. 
Upload the breaking tests to beta,
Have uffizzi point ar prod for the tests that need real world data that is hard to fake.
Having uffizzi point at the beta server makes it almost redundant. All it does is stop anoek needing to manually push to beta.

GreenAsJade:
No, it means we can independently test on a real server multiple different branches without having to merge them.

anoek:
It was at live, but I switched it to beta a few (days?) ago. I figured that was safer in general.. I thought about making it something you can switch though, both are definitely useful use cases.

BHydden:
> https://makeameme.org/meme/i-always-test
> 
> ;)

OGS thuglife 😎

anoek:
FYI: f0c2cfa2db28834a3765e3dc3022fcbc2b191daf is the commit I changed it to beta, so I think you just need to swap the `OGS_BACKEND` comment in `.github/uffizzi/docker-compose.uffizzi.yml` and you're all set. Feel free to swap it back, I don't have a strong preference.

GreenAsJade:
The real attraction of Prod is definitely-seeing-what-exists-there-in-the-way-it-will-look.

I wonder whether we could just as well have two uffizzis per PR as one?

GreenAsJade:
This is a challenge for another time.

PR #2204, opened 2023-03-22
Make handicap initial default

BHydden:
Make handicap initial default value for first custom game for live and corr games to match behaviour in automatch.

Set values to `-1` as per 
https://github.com/online-go/online-go.com/blob/73c13b5899043c5d5963df9f2c1d764c5a8b2181/src/components/GameAcceptModal/GameAcceptModal.tsx#L102-L104

I'm hoping this is the correct way to implement this, a stint in beta to make sure I didn't break anything is probably warranted.


BHydden:
Whelp. Looking at uffizzi it seems my changes had no noticeable effect on either play page create custom game or player direct challenge 😅 more digging is in order...

benjaminpjones:
Congratulations! You've found an OGS easter egg!  The code you modified is only used on this page: https://online-go.com/crash-course-learn-to-play-go/11, Cool stuff, but I wonder if that can all be deleted given that it's not really discoverable and kind of broken.

As far as where the actual change would go, I think some defaults are set in the `ChallengeModal` constructor, but I'm not sure how one would switch on Corr/Live/Blitz.  Maybe @bhostetler18 can help.

BHydden:
> constructor

That wouldn't be line 150 that I changed in my initial commit and then changed after second guessing myself, would it? 😅

BHydden:
Ok so that worked, create custom game now defaults to automatic instead of none. However as ben suspected it applied to all 3 formats. I guess that's fine right? Nothing wrong with blitz handicap.

BHydden:
> That wouldn't be line 150 ... would it?

Fun fact. It was 😅🤦‍♂️

anoek:
My only hesitation is that I thought this was a decision made by the community a long long time ago.. But it seems good and I can't find a forum thread about it, so :man_shrugging:

PR #2205, opened 2023-03-22
Delete unused Tutorial and NewGameModal.

benjaminpjones:
Noted in #2204, many of the exported variables/functions in ChallengeModal are only used in NewGameModal, which is only used in the old tutorial.  It seems to live at https://online-go.com/crash-course-learn-to-play-go, but it's broken in the sense that it redirects you to the new tutorial (LearningHub) on the first page.

## Proposed Changes

  - Delete old Tutorial
  - Delete NewGameModal
  - Delete unused exports in ChallengeModal
  - Remove export keyword where in places where it is only used inside the file.


anoek:
Awesome thanks for the cleanup :)

PR #2206, opened 2023-03-24
Windows dev detailed steps

GreenAsJade:
Fixes Windows developers struggling to get started.

## Proposed Changes

Add a tried / tested detailed process for getting development environment up on Windows.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,7 +78,9 @@ This project is largely built with TypeScript and React. If you are unfamiliar w
 4. In a command window, run
 
 `nvm install lts`
+
 `npm install -g yarn`
+
 `npm install -g husky`
 
 5. Open VSCode
@@ -100,6 +102,7 @@ You could poke around in src/ if you are already curious
 6. in a command window, cd to the folder that was created when you cloned the repo and do
 
 `yarn`
+
 `npm run dev`
 
 This should result in a bunch of packages being installed, then a server starting up and telling you it's running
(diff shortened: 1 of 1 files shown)


anoek:
Nice, thanks!

BHydden:
1. Install linux on virtual machine
2. ...
3. Profit.

benjaminpjones:
VM is no longer necessary - I hear WSL is pretty decent!

BHydden:
I think I would still prefer a clean dev environment 🤔

GreenAsJade:
.... _NOW_ everyone comes out with alternatives for helping Windows people ;) 😝

I was more expecting this review to bail me up about the choice of crlf handling.    I'm not sure if that is wise advice or not, it just WFM.  😆

And, in all seriousness, I think that if there's an option that doesn't invoke "Bah, you unix people are just trying to force me to use unix", then that's probably a good thing.

Also, I'm serious when I typed that t"hese instructions seem helpful for a unix person", especially a person new to most of the toolchain.

nvm for installing node was painless - I think windows and unix - and VSCode has surprisingly good git support (which I didn't know before because I use SmartGit).

I don't recall my own getting-started on linux working these things out along the way being as smooth as following that procedure for windows.

BHydden:
I didn't notice people had reported it being an issue. I remember I tried at first, and it was a huge headache, so before the end of the first day I picked the low hanging fruit of setting up a linux VM dev environment haha

PR #2215, opened 2023-04-07
Fix call moderator logico

GreenAsJade:
Fixes users not being able to use the call moderator button, due to only anonymous users being allowed.

## Proposed Changes
Allow non-anonymous users to use call moderator 😝


BHydden:
Nice find! Good fixing the logico ❤️

BHydden:
Fixes https://github.com/online-go/online-go.com/issues/2214

anoek:
Thanks for finding and fixing that :)

PR #2222, opened 2023-04-20
Format warnings with white-space: pre-line to honour paragraphs

GreenAsJade:
Fixes warning template with paragraphs getting smooshed up

## Proposed Changes

format warnings with white-space: pre-line

PR #2224, opened 2023-04-20
Added in Mark5000 moderator templates (with tweaks)

GreenAsJade:
Fixes missing templates

## Proposed Changes

Put in Mark's suggestions, with added newlines and slight edits

Needs #2222


anoek:
Awesome, thanks for compiling these

Issue #1449, opened 2021-04-11
Game control buttons are not accessible


romgrk:
The game control buttons are using `span` elements with onclick handlers, however span elements are not typically detected as activatable elements by accessibility tools, which renders them inaccessible for people with disabilities. The solution would be to either use `button` elements, or the ARIA attribute [role=button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role).

![Screenshot from 2021-04-11 00-24-21](https://user-images.githubusercontent.com/1423607/114292403-58a7e900-9a5c-11eb-8e10-4b7665b49c6e.png)

anoek:
I feel like these should be changed to a slider anyways, not sure why we have 7 "buttons" there.

romgrk:
TBH the buttons are practical, I use them often. I feel like a slider would be harder to use when going 1 move forward/backward.

anoek:
Sounds good, I feel like I've had that conversation before and got that same feedback now that you mention it, so buttons it is :) ideally made to be accessible.

BHydden:
Ideally we'd have both, but that brings us back to the screen real-estate dillema

saizai:
Would adding the Aria role even make any noticeable GUI change? It seems like an easy improvement to make with near zero cost.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

I believe that the set of buttons should also be enclosed in an Aria nav role.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

romgrk:
No noticeable changes to the GUI except that a11y tools would be able to present things correctly to disabled (eg blind, limited arm/hand mobility, etc) users. Most of the time I'd recommend just using `button`, because you don't need to remember to add `role=button` and `tabIndex=0` (required for keyboard navigation; crucial if the user cannot use a mouse), those things are done by default for native elements. In general, any custom made element (modals, dropdowns, etc) should have ARIA attributes to make them accessible.

Also note that if the buttons are labeled only by image icons, it's important to add [invisible labels](https://stackoverflow.com/a/19758620/6303229) for screen-readers only. In this case as @saizai mentions, playing go while blind is probably rare due to the skill level required, but still nice to have just in case, and it's not too long to add. Building an accessible website takes a bit more effort but not doing it is like having a building without wheelchair access; not cool for disabled people.

saizai:
[...]

PR #2227, opened 2023-04-24
Issue #1449: Span elements to button

krystofffff:
Span elements to button

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -806,35 +806,35 @@ export function Game(): JSX.Element {
             <div className="action-bar">
                 <span className="icons" />
                 <span className="controls">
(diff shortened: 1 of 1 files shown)


GreenAsJade:
I think something is up with the build system.  Last time I had one looking like this it went away when I pushed an update!

anoek:
Yeah it looks like the node build containers switched to v20 and one of our libraries had a dependency issue with that. It's fixed on devel now.

anoek:
Thanks for the patch!

benjaminpjones:
It sounds like this changed the layout - may need to update the css/styl files as well.

See: https://forums.online-go.com/t/ui-broke-scroll-bar-time-control-display/47787?u=benjito

PR #2230, opened 2023-04-26
Annul with blame

GreenAsJade:
Give moderators a button on the player card which fills in the player number markup in the moderator note for them while annulling.

Fixes  ongoing oopsies in the mod logs

## Proposed Changes
 - Factor out "doAnnul" work from the couple of places that have almost identical implementations
 - Replace the "hide arrows" icon with a "hammer" that annuls the game and provides a player-markup into the mod log.
 
 (Arguably, `doAnnul` might have a more appropriate home, but one did not leap out at me)


anoek:
https://github.com/online-go/online-go.com/pull/2230/files#diff-e05e0aa698471ed3ccabe5593b43424e53842ae4c49672f5828c50f6f01a2849R254-R255

I think the build part is failing here, `hideNextGameArrows` is no longer needed?

GreenAsJade:
Hah - a good opportunity to make the user experience nicer, the previous was a bit clunky

GreenAsJade:
Soooo  ... it seems I also didn't notice the circular dependency error message ... and my hunch that doAnnul does not belong in misc was correct 😅

anoek:
Awesome thanks :)

PR #2238, opened 2023-04-29
Add a class of report as a way for a moderator to warn someone directly

GreenAsJade:
Fixes having to raise a report then warn from there.

## Proposed Changes

 - Add moderator-only category of reports
 - Add "warning" category
 - Send the warning instead of filing a report, for "warning" category

### DIFF
--- a/src/components/Report/Report.tsx
+++ b/src/components/Report/Report.tsx
@@ -24,6 +24,7 @@ import { PlayerIcon } from "PlayerIcon";
 import { post } from "requests";
 import { alert } from "swal_config";
 import { setIgnore } from "BlockPlayer";
+import { useUser } from "hooks";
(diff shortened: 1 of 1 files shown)


anoek:
Awesome thanks :) I'm not 100% sure about removing the report button all together, I've seen mods file a report on users that they feel they shouldn't handle from time to time, like too close to the subject or they want a second opinion or something like that. But the warning part is great :)

GreenAsJade:
For clarity, moderators still have a report button.  This change gives them an extra "kind" of report, on the resulting form, which has the effect of warning.

I think the downside of this is that it doesn't add another button called "Warn", so if you don't know that it exists, it's a teeny bit unobvious to find :)

PR #2241, opened 2023-04-30
Do single game annulling using the /moderation endpoint

GreenAsJade:
Fixes two ways to do the same thing being used.

## Proposed Changes

Do single-game annul using the /moderation endpoint.

*** Needs https://github.com/online-go/ogs/pull/1803

This also adds the API definition file for that new endpoint, hopefully coincides with @RubyMineshaft  definition


anoek:
Yay, thanks!

PR #2243, opened 2023-04-30
Mass annul

RubyMineshaft:
## Proposed Changes

  - adds selection mode to game history table, which adds games to an annulment queue
  - adds modal for inspecting games within the queue
  - passes array of ranked, non-annulled games to back end for annulment. 

~I've also written a draft of usage documentation for the forums that I can send if the screenshots and info there would be useful.~ That was before I saw the Uffizzi thing. How neat! Certainly a lot better than screenshots.  :)


anoek:
That's a pretty fancy interface, thanks a ton!

PR #2250, opened 2023-05-08
Annul queue modal bugfix 

RubyMineshaft:
Currently, the annul queue modal is being rendered outside of the Router which causes the modal to crash if there is a review linked in the game chat. 

## Proposed Changes

  - Render annul queue modal from within the game history table instead of outside the Router


anoek:
Thanks for the fix!

PR #2252, opened 2023-05-12
Restored display of komi in player cards

GreenAsJade:
Fixes people complaining about komi not appearing anymore 😝

## Proposed Changes

Restore the code that displays komi in PlayerCards

Note that to make the test work, I had to change the import of React in PlayerCards.tsx.

Not sure what that's about, but it seems to work.

I also exported PlayerCard so the test could import it.  Not sure if that's "normal" 🤨  I suspect that in theory if something needs to be tested, it should be in it's own file, but meh 😝


anoek:
Nice, thanks for the fix and tests!

PR #2256, opened 2023-05-13
Add BID Alias column to aliases table

GreenAsJade:
Fixes some uncertainty about account aliases

## Proposed Changes

Display the new BID match info to moderators.


anoek:
This is up on beta FTR, I'll merge when we deploy the backend to production (probably tomorrow)

Issue #2229, opened 2023-04-26
UI Enhancement request -- add tooltips to the right-hand-side slide-out menu buttons.


bradwood:
**Is your feature request related to a problem? Please describe.**

It's not really a problem, but more of an annoyance. I'd like to keep the slide-out menu on the right permanently minimised so I can see the stuff behind it. But, I keep forgetting what the icons do when I do this.

**Describe the solution you'd like**

My request is simple: add a descriptive tool-tip which pops up on hover over those buttons, describing what they do. This way I can keep the slide-out menu permanently minimised, but still remember what each button does.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

benjaminpjones:
Seems to be fixed by #2259 - @bradwood @loganbruce let me know if I'm missing anything!

PR #2259, opened 2023-05-19
Adding new tooltip component to game dock

loganbruce:
Fixes #
Proposal for[ Issue #2229 ](https://github.com/online-go/online-go.com/issues/2229)

## Proposed Changes

  - Created a new tooltip component
  - Added to the game dock to show option name when not pulled out
  - Only appears when user has preferences set to disable game dock pop out
  - Future consideration: Included option for 'position' property to customize where the tooltip appears **but** stylings still need to be coded and applied


GreenAsJade:
This is cool.

I'd use it :)

Aside from the lint errors, I think the css needs theming, and some of the buttons pop up tooltips (pause and sound at least)

Screenshot 2023-05-19 at 4 22 23 pm (2)

loganbruce:
I have added the extra tooltips for the other dock items and I added theming to the css. 

It's my first time using the theme so please let me know if there's improvements to be made 🥳😄

GreenAsJade:
Nice, works for me 👍

src/views/Game/GameDock.tsx
```diff
@@ -325,123 +329,169 @@ export function GameDock({
...
+                        value={volume}
+                        min={0}
+                        max={1.0}
+                        step={0.01}
+                    />
+                </a>
+            </Tooltip>
+            <Tooltip
+                tooltipRequired={tooltipRequired}
+                title={"Zen Mode"}
```
anoek, line 358: Thanks for the PR, looks good! We do need to get those translated though, but that's fairly easy. To handle translations we wrap it with `_("text")` or `pgettext("some translation context", "text")`, these translate the strings into whatever language the user is using.

loganbruce: Not a problem! Is there a difference/preference between the two options?

anoek: The difference is that `pgettext` provides some context for translators. For instance "rank" might be translated differently depending on whether we're talking about a ladder rank, or your playing ability, so you would provide some additional string context to both differentiate the string for translation and provide that string to translators to let them know how it will be used.

If that extra context isn't needed though, just use `_`.

In this case you can probably usually use whatever was used for the string that's visible when the dock slides out. If you use the same string and same call, it will use existing translations which is preferable.

loganbruce: image

I have noticed that some of the strings in the dock currently do not have anything for translation (Fast AI Review for example). Have these been left out intentionally? If not, happy to add them now as well

anoek: Yeah, those are things restricted for super users, so I don't include them there so the translators aren't asked to translate stuff that's not visible to the general population

src/views/Game/GameDock.tsx
```diff
@@ -325,123 +329,157 @@ export function GameDock({
...
+                </Tooltip>
             )}
-            <a onClick={showLinkModal}>
-                <i className="fa fa-share-alt"></i>{" "}
-                {review ? _("Link to review") : _("Link to game")}
-            </a>
-            {sgf_download_enabled ? (
-                <a href={sgf_url} target="_blank">
-                    <i className="fa fa-download"></i> {_("Download SGF")}
+            <Tooltip tooltipRequired={tooltipRequired} title={_("Link to game")}>
```
anoek, line 471: I reckon this title should be `review ? _("Link to review") : _("Link to game")` to be consistent?

src/views/Game/GameDock.tsx
```diff
@@ -325,123 +329,157 @@ export function GameDock({
...
-            </a>
+            </Tooltip>
+
             {game && (
-                <a
-                    onClick={onAnalyzeClicked}
-                    className={phase !== "finished" && goban.isAnalysisDisabled() ? "disabled" : ""}
-                >
-                    <i className="fa fa-sitemap"></i> {_("Analyze game")}
+                <Tooltip tooltipRequired={tooltipRequired} title={_("Toggle AI review")}>
```
anoek, line 369: I might be making this up, but I seem to recall once having "Toggle AI review" as the text here, and it being confusing enough to people that I ended up doing what we do down below which is to be more explicit with `i_review_enabled ? _("Disable AI review") : _("Enable AI review")`

src/views/Game/GameDock.tsx
```diff
@@ -325,123 +329,157 @@ export function GameDock({
...
-                <a href={sgf_url} target="_blank">
-                    <i className="fa fa-download"></i> {_("Download SGF")}
+            <Tooltip tooltipRequired={tooltipRequired} title={_("Link to game")}>
+                <a onClick={showLinkModal}>
+                    <i className="fa fa-share-alt"></i>{" "}
+                    {review ? _("Link to review") : _("Link to game")}
                 </a>
+            </Tooltip>
+            {sgf_download_enabled ? (
+                <Tooltip tooltipRequired={tooltipRequired} title={_("SGF Download")}>
```
anoek, line 478: Let's use the same string here, `"Download SGF"` to avoid extra translation work (also to maintain consistency)

anoek (changes requested):
Thanks for all the work on this, looking good!

src/views/Game/GameDock.tsx
loganbruce: Rewording to "Download SGF" causes an issue with one of the tests as two elements have the same string.  I am unsure of how to approach this?

image

anoek (approved):

BHydden:
Maybe it's a coincidece but after this was merged this section lost its alignment on mobile.

https://forums.online-go.com/t/what-happened-to-mobile-game-options/48036?u=bhydden

Are the two events connected?

GreenAsJade:
I just discovered (after I turned off dock pop out 🎉 ) that we don't have a tooltip for the "private game" lock icon at the top of the dock in private games.

GreenAsJade:
@loganbruce  : https://join.slack.com/t/online-go/shared_invite/zt-1vf15ez6u-ztsWJj_~n~2LNICOXtfleA

GreenAsJade:
There a fix of sorts here: https://github.com/online-go/online-go.com/pull/2266

A more thorough fix might go through and find the now-unneeded css for <a> in the dock, but ... meh?

GreenAsJade:
So it turns out that it _is_ broken and it does need to be fixed.

=> Dock.style needs a go-over to make sure we understand what each part is doing, when, and how to get it to apply to the TooltipContainer.

I did some experiments with just setting all the `<a>` properties onto the TooltipContainer, but it doesn't suffice ... I suspect that the children of the `<a>` also need treatment.

PR #2265, opened 2023-05-21
Selecting a template claims a report

GreenAsJade:
Fixes doing all the stuff without remembering to claim

## Proposed Changes

- Add selection callback to MessageTemplate, use it to claim the report
- Tidy up report claiming, so it happens the same way in all three cases


anoek:
Nice, thanks!

PR #2266, opened 2023-05-23
Fix Dock layout for mobile

GreenAsJade:
Fixes scrambled dock layout in mobile

## Proposed Changes

Apply the needed flex properties to Dock children.

### DIFF
--- a/src/components/Dock/Dock.styl
+++ b/src/components/Dock/Dock.styl
@@ -85,13 +85,8 @@ dock-inline-height=500px
     }
 }
 
-// make the elements in the Dock line up properly in mobile mode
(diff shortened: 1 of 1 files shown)


BHydden:
Yeah looks better. I think they're bigger than they used to be, but they're at least now aligned properly 👍

GreenAsJade:
> Yeah looks better. I think they're bigger than they used to be, but they're at least now aligned properly 👍

Ah - I kind of noticed that, but didn't twig that it needed attention.

I'll take a look.

anoek:
Thanks for the quick fix!

PR #2274, opened 2023-05-28
Issue 2273 - Default chat mode

### LINKED ISSUE #2273: Preference for defaulting chat mode on the game page
[From the forum](https://forums.online-go.com/t/default-chat-preference/48072?u=greenasjade):

Is it possible for users to set the chat preference to “personal or Malkovich chat” as the default option? This would be greatly appreciated, especially for users with multiple games, as it would eliminate the need to constantly toggle the chat setting whenever they click on the chat button.

I fear of forgetting to toggle chat, so my evil plan gets accidentally leaked to my opponent :person_facepalming:

andymarden:
Fixes #2273 

## Proposed Changes
Changes in GamePreference.tsx and preferences.tsx to select the default mode (defaults to "main")
Changes in Game.tsx and GameChat.tsx - anywhere where it was hardcoded to default to "main" for the chat mode it now uses the preference
Snuck in an unrelate change to th3 description of the Variation Move count to make the setting clearer.
  -
  -
  -

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -34,6 +34,7 @@ export const defaults = {
     "chat.show-all-group-channels": true,
     "chat.show-all-tournament-channels": true,
     "chat.user-sort-order": "rank",
+    "chat-mode": "main",
     "desktop-notifications": true,
     "desktop-notifications-require-interaction": false,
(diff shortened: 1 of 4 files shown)


anoek:
Thanks for the feature!

PR #2276, opened 2023-05-28
issue-2275: reduce clickable area for invite-only games

### LINKED ISSUE #2275: Move the Invite-only checkbox (or clickable area) to stop it being checked by mistake on small screen devices
See https://forums.online-go.com/t/invite-only-by-mistake/48114

Will experiment with a few options.

andymarden:
Fixes #2275 

## Proposed Changes
Simple change in ChallengeModel.tsx to remove the htmlFor tag from the label alongside the checkbox, making it none-clickable 
  -
  -
  -

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -966,7 +966,29 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                         )}
                     </div>
                 </div>
+                {!(this.props.playerId || null) && (mode === "open" || null) && (
+                    <div className="form-group">
+                        <label className="control-label" htmlFor="challenge-invite-only">
+                            {pgettext(
+                                "A checkbox to make a challenge open only to invited people who have the link to it",
+                                "Invite-only",
+                            )}
+                        </label>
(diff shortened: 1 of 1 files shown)


anoek:
For consistency with everywhere else, I think the htmlFor should remain. I think we should probably fix this issue by moving that checkbox somewhere else so it's not easy to click by accident, or give it some more margin.

andymarden:
If that is what's required, I think I will move it to to the section above the Create Game button - personally I don't feel it is useful for most people to leave the htmlfor in since it is likely to be relatively rare when you want to switch between these but happy to do so.

Another option has just occurred to me - there is easily enough space have a "Create Open Game" and a "Create Private Game" next to each other horizontally where the existing button is, rather than have to remember to click on and off a check box.

What do we think?

anoek:
Yeah moving it above is the right thing to do I think, under `Private` maybe?

andymarden:
OK - will do that

andymarden:
OK - done - now under Private - I agree that's much better

anoek:
Cool looks good, though I'm seeing a lot of other commits here too, could you try rebasing your changes to see if that clears all of those up?

PR #2279, opened 2023-05-31
Remove max container width

BHydden:
Fixes pages not filling screen with content

https://forums.online-go.com/t/wish-layout-of-watch-page-similar-to-home-page/48139/7?u=bhydden

### DIFF
--- a/src/global_styl/bootstrap-esq.styl
+++ b/src/global_styl/bootstrap-esq.styl
@@ -19,8 +19,6 @@
 
 .container {
     width: 100%;
-    max-width: 1200px;
-    max-width: 1200px;
     margin-right: auto;
     margin-left: auto;
     padding-left: 0.5rem;


BHydden:
Looks ok on my mobile but should be checked on a bigger monitor.

GreenAsJade:
I think this needs some work for wider screens - the items at the top are all kinda far-flung and "lost looking":

Screenshot 2023-06-02 at 4 35 00 pm (2)

Screenshot 2023-06-02 at 4 39 55 pm (2)

BHydden:
Hmm 🤔 thanks for the feedback. It does look a bit haphazard. 
Maybe if I pinned refresh to the left edge of show boards box on the top right rather than having it in centre, would that help? It's the best I can think of since I don't think we want to make those buttons big enough to fill the screen.

BHydden:
As for the No games being played advisory... I think that can stay centred.

PR #2291, opened 2023-06-12
Added confirm dialog for undo button

GreenAsJade:
Fixes [accidentally asking for undo](https://forums.online-go.com/t/feature-request-cancel-undo-request/48099?u=greenasjade)

## Proposed Changes

Have a confirm dialog for undo.

### DIFF
--- a/src/views/Game/PlayButtons.tsx
+++ b/src/views/Game/PlayButtons.tsx
@@ -84,7 +84,19 @@ export function PlayButtons({ show_cancel = true }: PlayButtonsProps): JSX.Eleme
             data.get("user").id === goban.engine.playerNotToMove() &&
             goban.engine.undo_requested !== goban.engine.getMoveNumber()
         ) {
-            goban.requestUndo();
+            void alert
+                .fire({
+                    text: _("Request undo?"),
+                    confirmButtonText: _("Yes"),
+                    cancelButtonText: _("No"),
+                    showCancelButton: true,
+                    focusCancel: true,
+                })
+                .then(({ value: accept }) => {
+                    if (accept) {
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Note: there's some discussion in the forum about whether this is a Good Thing (TM) and whether it needs to be tweaked.

Draft for now.

anoek:
Hmm, maybe we put this behind an setting? I haven't heard this being a particularly notable problem in the past, so I think we might get some more push-back from having it defaulting a prompt whenever undo is clicked.

GreenAsJade:
The suggestion in the forum is that if you have submit for click then you get this dialog as well.

I dunno, I suspect that the real answer is going to be to provide "cancel undo request", which is kind much more hassle :)

anoek:
Yeah agreed on both accounts.

anoek:
FWIW I do think cancel undo request is feasible

anoek:
Misclick on the close, sorry - I do think the cancel undo request is feasible fwiw, I don't think there would be any major blockers, just takes tracing that through all the different layers. It'd definitely be a good thing to have though, I should have implemented that right away when I did undo.

I think this is probably going to be useful even when we finally get around to undo-undo though so :+1:  once we get it behind a check for submit-to-move

BHydden:
> Misclick on the close, sorry

The irony is just so delicious ❤️

GreenAsJade:
Yeah cancel-undo should be readily doable, just not in the "oh, I'll wack that out right now" bucket :)

BHydden:
Can we implement the confirm dialog until GaJ is ready to code up the cancel button?

PR #2293, opened 2023-06-12
Issue 2292 - fix first character deletion in input field and change some to ranges

### LINKED ISSUE #2292: Number input fields on settings don't allow backspace of first character
Makes it difficult to change esp on mobile UI.

In Game Preferences, the following are affected:
- AutoPlay Delay
- Variation Stone Transparency
- AI Variations Shown 

Will change those to sliders and any others that are affected and it makes sense to.

andymarden:
Fixes #2292 

## Proposed Changes
Game Preferences:
- AI Variation Moves: converted to range with 0 displayed as Off
- Stone Transparency: converted to range
- Auto Delay: left as input field but borrowed fix for first character deletion from the Threshold in General Preferences.

  -
  -
  -


anoek:
Definitely an improvement!

* Was there a technical issue why Autoplay was left as a numeric input instead of slider? Now I'm wondering if it shouldn't just be a slider too.

* I get why how we ended up where we are for the "AI variations shown" slider, but looking at it now, `"off"` really meaning `"all"` might be a bit confusing. How hard would it be to start at `1` and then do some magic so if you slide it to the end it turns to `"all"`?

andymarden:
One the first point, the reason I left this as a numeric input (with the ability to backspace now) was that there is no limit on the upper bound of this value and the step is 0.1, so having it start at 0.1 go up in increments of 0.1 and an upper limt of say 60 (1 minute) would make it impossible to accurately select the time you want. If we change the step (or had a distinct list of say 0.1, 0.2, 0.5, 1.0 etc) and had an upper limit so that there were only max 20 values selectable, that would work as a slider. Thoughts?

As for All vs Off, I guess I did this as it the option read "Limit AI Variations Move Shown" but can easily change it. All is a bit misleading since it ever shows All moves but just whatever it thinks is useful. So which?

anoek:
Re the slider: I think distinct steps would probably work well there, I don't think we need folks to be able to set 56.3 seconds or whatever.

Re the all vs off, how about "Max" instead of "All"?

andymarden:
Ok - redone:
1. Autoplay delay is now slider 1-20 (secs)
2. Default is now Max (10) vs Off (0) and AIReview.tsx modified accordingly

anoek:
Thanks @andymarden !

PR #2294, opened 2023-06-14
Type use of Reports in ViewReport

GreenAsJade:
Fixes untyped state variable

## Proposed Changes

 Declare the type of `report` and extend the type to add the extra fields that come from the server.


anoek:
Yay, thanks for the typing!

PR #2297, opened 2023-06-15
Terminology around "ban".

GreenAsJade:
Fixes a sense of unease about banning people.

Well, probably doesn't "fix" it per-se, but moderators agree it reflects the intent better.

## Proposed Changes

Say "your account has been suspended" rather than "you have been banned", where possible.


GreenAsJade:
Under review by moderator panel

PR #2305, opened 2023-06-22
Use the same selected_chat_log for GameChat and PlayControls.

benjaminpjones:
Fixes this issue from the forums: [Bug: Variations Going to Chat when Malkovich is On](https://forums.online-go.com/t/bug-variations-going-to-chat-when-malkovich-is-on/48157)

## Proposed Changes

  - use the selected_chat_log from the top level Game component instead of a separate variable in the GameChat component.
 - Note: if you navigate to a different game (say by clicking the TurnIndicator), this will change the chat back to the default mode.  I think this was the intent in #2274, tagging @andymarden for confirmation.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -44,7 +44,7 @@ import { setExtraActionCallback, PlayerDetails } from "Player";
 import * as player_cache from "player_cache";
 import { notification_manager } from "Notifications";
 import { Resizable } from "Resizable";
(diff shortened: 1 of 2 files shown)


anoek:
Oh nice, thanks for figuring out that one!

Spoekenkieker:
Awesome, THANK YOU!
You’re saving not only me but probably others also from embarrassments 😆

PR #2306, opened 2023-06-22
Reduce time that the Accept Removed Stones button is disabled

### LINKED ISSUE #2303: accept score button disabled too long
**Describe the bug**
By design, the **Accept Score** button gets disabled for about a second after the territory markings change.  This also happens if your opponent accepts the stones.

Occasionally (and this is the bug), the button is disabled much longer than one second.  I'm not sure what triggers this, but I think it's some combination of the useEffect's dependency array (TODO: add a code pointer) and score events passed through the socket.

**To Reproduce**
1. Play a game, get to a scorable position
2. After passing, let your opponent accept first.
3. Note how long your own button is disabled after your opponent accepts. This is usually 1 second, but could be as much as 5 seconds

**Expected behavior**
Button should only be paused up to 1 second in the absence of changes.  In addition, perhaps it should stay enabled if your opponent accepts but does not change the score/territory.

**Screenshots**

This doesn't happen like this every time, but here is an example of where it happens for about 5 seconds.

https://github.com/online-go/online-go.com/assets/25233703/8c5d8f3b-fed2-430a-8697-786e69aaabc8

 - Device: Samsung Galaxy S22
 - OS: Android
 - Browser: Firefox

benjaminpjones:
Fixes #2303 

## Proposed Changes

  - Lower timeout on mobile to 1500ms (same as desktop)
  - Only disable the button if the removed stones changed

### DIFF
--- a/src/views/Game/PlayControls.tsx
+++ b/src/views/Game/PlayControls.tsx
@@ -30,7 +30,6 @@ import {
     MoveTree,
     PlayerColor,
 } from "goban";
-import device from "device";
 import { alert } from "swal_config";
 import { challengeRematch } from "ChallengeModal";
 import { Clock } from "Clock";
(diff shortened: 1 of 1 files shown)


anoek:
Slick, thanks!

PR #2307, opened 2023-06-23
Make it so that a report is claimed for a moderator if they simply send a message

GreenAsJade:
Fixes moderators having to explicitly claim in the case where they don't select a template.

## Proposed Changes

- Give MessageTemplate an onMessage callback, called when the template view is used to send a message
- Use MessageTemplate.onMessage to claim the report


anoek:
Thanks!

PR #2308, opened 2023-06-23
Update US rank search

BHydden:
Temporary fix while main site is under upgrades to fix rank links.

https://forums.online-go.com/t/updating-aga-rating-linkage/48338?u=bhydden


anoek:
Thanks!

PR #2310, opened 2023-06-26
Add a proxy for the firewall config backend routes

GreenAsJade:
Fixes 404s when trying to test firewall config features

## Proposed Changes

 - Add a backend proxy for `/firewall` to the dev server


anoek:
Ah yep, thanks for the fix :)

PR #2311, opened 2023-06-26
Firewall action: report

GreenAsJade:
Fixes moderators wanting to be able to generate reports from the firewall.

## Proposed Changes

Have a firewall action type "REPORT" which the backend can use to generate reports for firewall configs.

Needs https://github.com/online-go/ogs/pull/1817 for backend support.

This is built on top of https://github.com/online-go/online-go.com/pull/2310 since I couldn't test it without that!

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -239,6 +239,7 @@ function dev_server(done) {
     devserver.use("/disconnect", backend_proxy("/disconnect"));
     devserver.use("/OGSScoreEstimator", backend_proxy("/OGSScoreEstimator"));
     devserver.use("/oje", backend_proxy("/oje"));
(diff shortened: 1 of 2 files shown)


GreenAsJade:
I couldn't cheat with pretending to be the system user reporting this, in the way that I did with rengo timeout comments - the "reporter" has to be a real user or stuff breaks.

So I made the reporter be "anoek".   :D    Open to other ideas of course.

GreenAsJade:
(In fact, what I'd like to do is return the matching rule from firewall_match instead of the action, then the consumer can look at things like what the rule name is and who created it - the reporter in this case could then be the creator of the rule... if the rule has a creator ... which it probably should eh!    Can of worms :) )

anoek:
Haha, alrighty fine for now

PR #2317, opened 2023-06-30
GameTiming function component refactor

GreenAsJade:
Fixes out of date implementation.

## Proposed Changes
 
Convert `GameTiming` to be a function component.

### DIFF
--- a/src/views/Game/GameTimings.tsx
+++ b/src/views/Game/GameTimings.tsx
@@ -1,18 +1,18 @@
 /*
  * Copyright (C)  Online-Go.com
  *
- * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


src/views/Game/GameTimings.tsx
```diff
@@ -1,27 +1,29 @@
 /*
  * Copyright (C)  Online-Go.com
  *
- * This program is free software: you can redistribute it and/or modify
```
benjaminpjones, line 4: Gotta turn on the case-sensitivity 😄

GreenAsJade: That's one way of looking at it, though in reality the problem was I have regex match turned on but I said `this.` 🤦‍♀️

PR #2322, opened 2023-07-05
Try to make bid_match column clearer in as few words as possible

GreenAsJade:
Fixes ... remains to be seen.

## Proposed Changes

  - Call that narrow column Shared BIDs instead of BID Alias


BHydden:
I think the greater confusion is exactly *what* a BID is and how two may be correlated, rather than the semantic differences between "shared" and "alias"

PR #2323, opened 2023-07-05
Don't show the ranks of players in unranked games

GreenAsJade:
Attempts to address folk getting wound up about the rank of their opponent in unranked games.

## Proposed Changes

 - Don't show rank of players in the game in the PlayerCards or game chat if the game is unranked and not rengo.

### DIFF
--- a/src/views/Game/GameChat.tsx
+++ b/src/views/Game/GameChat.tsx
@@ -513,6 +513,11 @@ export function GameChatLine(props: GameChatLineProperties): JSX.Element {
     let chat_id = props.review_id ? "r." + props.review_id : "g." + props.game_id;
     chat_id += "." + line.channel + "." + line.chat_id;
 
+    const show_rank =
+        !(line.player_id in goban.engine.player_pool) ||
+        goban.engine.config.ranked ||
+        goban.engine.rengo;
+
     return (
         <div className={`chat-line-container`} data-chat-id={chat_id}>
             {move_number}
@@ -526,7 +531,9 @@ export function GameChatLine(props: GameChatLineProperties): JSX.Element {
                         ]{" "}
                     </span>
                 )}
-                {(line.player_id || null) && <Player user={line} flare disableCacheUpdate />}
+                {(line.player_id || null) && (
+                    <Player user={line} rank={show_rank} flare disableCacheUpdate />
+                )}
                 <span className="body">
                     {third_person ? " " : ": "}
                     <MarkupChatLine line={line} />
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -308,7 +308,11 @@ export function PlayerCard({
 
             {((player && player.rank !== -1) || null) && (
                 <div className={`${color} player-name-container`}>
-                    <Player user={player.id} historical={(!engine.rengo && historical) || player} />
+                    <Player
+                        user={player.id}
+                        historical={(!engine.rengo && historical) || player}
+                        rank={engine.config.ranked || engine.rengo}
+                    />
                 </div>
             )}
 


anoek:
Was there a forum discussion or something this is request is coming from? This seems like maybe we'd get some backlash for doing this.

benjaminpjones:
Personally, I prefer to see rank even in unranked games.

If rank visibility is an issue, shouldn't the user just toggle the "Hide ranks" feature?

GreenAsJade:
There's discussion in the moderator chat.

It's a suggestion from Bhydden: trying to give people a pause for thought before reporting sandbagging in unranked games, which is an ongoing "sore".

Many opponents we meet in unranked games only play unranked, so their rank is meaningless (and often lower than their skill).   This means that folk in fact shouldn't really be "preferring to see rank", because it doesn't necessarily mean anything.

I just submitted this PR on the back of the suggestion to see what it would look like, it was easy enough - but the moderators also are not united on it being a good idea, so I'll throw this into draft for now.

benjaminpjones:
> Many opponents we meet in unranked games only play unranked, so their rank is meaningless (and often lower than their skill). 

Shouldn't these specific players be marked with a `[?]` then?

I just don't see why a subset of unranked games should lose a feature that exists across the site.

> It's worth noting that if this is the solution, it probably should be extended into match making

Totally agree, and I bet a lot of folks who complain about these "pseudo-sandbaggers" are looking at the rank on the `/play` page as well.

GreenAsJade:
> Shouldn't these specific players be marked with a `[?]` then?

Yes, you would think so.

But the ranking system doesn't know that these people have played a lot more go than the ranked games that it sees.

We'd have to tweak that somehow to achieve the effect you mention.

kobakitty:
How about showing their rank as ? if their rank is uncertain, as in their ± -value is larger than 1 full rank? Brings back the provisional ranks we had in the Elo days ^^

Doesnt help with cases where player has solid but inaccurate rank, but might reduce the sandbagging reports that are caused by players who only have handful of games under their belt.

GreenAsJade:
The rank display kinda does work like that already.    At a certain threshold [?] kicks in.

Maybe it'd be as simple as tightening that up, but we might find that players who play only ranked also start showing up as [?] due to their personal variation.

benjaminpjones:
That would address some, but not all of the accounts that draw complaints.  Especially if they are doing it on purpose.  I think it would be a good idea to factor in time to the deviation calculation.  A crude, but potentially workable solution might be:

```
if user played a ranked game in the last X months:
    return actual deviation
else:
    return 350
```

A more complete implementation would, of course, have less of a cliff.

shinuito:
> if user played a ranked game in the last X months:

> A more complete implementation would, of course, have less of a cliff.

Glicko is supposed to have a feature where the deviation increases from inactivity right? 

https://github.com/online-go/goratings/blob/4c677e360440273429bef196fd09269c71067f35/goratings/math/glicko2.py#L49

It was just that I don't think this gets called until they play a game again after the inactivity if at all. I guess it probably becomes a lot of server work to just bump every users deviation on various windows of time.

There is actually a chance though that some players ranks won't change much with a certain amount of inactivity, though for others it will drastically as they improve, playing elsewhere.

I'm not sure exactly what the solution should be.

BHydden:
> if user played a ranked game in the last X months:
>     return actual deviation
> else:
>     return 350

This seems pretty elegant actually.

GreenAsJade:
> It was just that I don't think this gets called until they play a game again after the inactivity if at all. I guess it probably becomes a lot of server work to just bump every users deviation on various windows of time.

Maybe we should update the glicko of each player every _unranked_ game as well.... then their rank would degrade over time as intended...

GreenAsJade:
@anoek   Can we call ranking at the end of unranked games?       Would that have the effect?

anoek:
I don't think that'd cover the case when a player is gone for a few months and either got notably worse or better during that time, in that case I reckon we'd want to just do something when they start a session maybe?

Before making a change like that though I'd want to run it through some full rating tests to see how things play out.

GreenAsJade:
Am I remembering wrong that glicko has some actual passage-of-time factor in it?  I thought that the uncertainty increases with time.   That's what I was hoping this suggestion would take advantage of - right now people who only play unranked don't get re-assessed at all by the ranking engine, so a passage-of-time factor doesn't get applied to them.

BHydden:
IIRC anoek turned that off 🤔 the justification escapes me though, or even where to look to certif verify this

anoek:
Yeah, Glicko was designed with in person tournaments in mind, so a bunch of games played in one session, then presumably a few months or years or whatever inbetween. On OGS though a lot of our games are correspondence games so it gets kind of weird, you can start and finish many games while some other game is going on, and you could start and be actively playing games for a year before one even finishes, things like that - so measuring the time between games as a measurement of end times isn't really what you'd want to do in a lot of cases.

All that said, I did play with enabling that and fiddle with it and whatnot, and as I recall it either didn't really make a notable difference overall for us, with whatever parameters I was using (I did try many as I recall).

shinuito:
> > if user played a ranked game in the last X months:
> > return actual deviation
> > else:
> > return 350
> 
> This seems pretty elegant actually.

It's also worth (re)considering what issue this is trying to address. 

Is it trying to deal with the idea that people are playing against players with the wrong rank? That still happens in unranked games no matter what you show the player, a ? or a not fully calibrated rank.

Is it trying to fix the idea that people are being misled by someone with a rank that's not accurate? Probably this won't fix it in the short term, as someone can play one ranked game every X months to have maintain an incorrect rank if that's what they want then continue to play unranked.

So it's not clear what issue it's addressing and if it fixes the issue and for what part of the userbase.

benjaminpjones:
> Is it trying to fix the idea that people are being misled by someone with a rank that's not accurate?

This is the issue it is trying to address

> someone can play one ranked game every X months to have maintain an incorrect rank

It's true that someone could play an occasional game to maintain a rank, but if they regularly play ranked games and still maintain a wildly inaccurate rank, IMO they lose "plausible deniability" for sandbagging reports.

As I mentioned earlier, this was a crude solution to the issue, and more gradual degredation of the `+/-` would probably be more effective to this end.

GreenAsJade:
The specific problem (to underline it) is the steady flow of reports "Hey this person is sandbagging" ... no, they just play unranked games.

BHydden:
> Can we call ranking at the end of unranked games?

> if user played a ranked game in the last X months...

Maybe a combination of these two proposals wouldn't totally fix the problem but they both sound useful and positive. Can anyone think of a downside to implementing them? I don't think we need to avoid good solutions in the hopes we'll find a perfect one.

BHydden:
Stale after a month only? Is the bot broken???

benjaminpjones:
PRs go stale faster than issues (fwiw I thought this PR *had* stalled, with some alternative solutions)

GreenAsJade:
Yes, I think so too.

PR #2335, opened 2023-07-28
Make the reporter-templates include the reported player name

GreenAsJade:
Fixes reporters wondering which report a moderator is responding to

## Proposed Changes

Make `MessageTemplate` also take the reported person as a parameter, and provide template markup to substitute the reported person's name.


anoek:
Nice :)

PR #2338, opened 2023-07-28
Confirm dialog for rengo challenge delete

GreenAsJade:
Fixes people accidentally deleting rengo challenges

## Proposed Changes

Add an alert to confirm deletion of rengo challenge


GreenAsJade:
Ooops this was on top of the variation glitch fix, by accident!

Issue #2271, opened 2023-05-25
Standardise internal variable names in Settings code


andymarden:
Some start with _ and some don't. Standardise on _ since they are internal? Spotted in GamePreferences - but probably rife elsewhere.

GreenAsJade:
In general, I think this ticket is aimed at code like this

```
    const [variation_stone_transparency, _setVariationStoneTransparency] = usePreference(
        "variation-stone-transparency",
    );

    function setVariationStoneTransparency(ev) {
        const value = parseFloat(ev.target.value);

        if (value >= 0.0 && value <= 1.0) {
            _setVariationStoneTransparency(value);
        }
    }
```

... where the correct function to call to set the value is `setVariationStoneTransparency`, then the underlying preference setter should be preceded by `_`

As Andy commented - "a common pattern in Prefs, but possibly elsewhere"

@anoek   Good First Issue ?

anoek:
Well, what's here is the expectation that for  a given state/preference `xyz` there exists a corresponding function `setXyz` that changes that value. In the case where we need some extra logic, we provide `setXyz` as the function we use, but we have `_setXyz` that that function calls after doing whatever special logic it needs. In this case, having the `_setXyz` is "internal" in the sense that it is not meant to be directly called by ourselves, but it remains very clear that that is the actual setter, but that we should be using `setXyz`.

GreenAsJade:
I think the gist of this ticket was "there's meaning behind the scheme, but it's inconsistently applied, so a first-time coder could get to know the code and make it applied consistently"

anoek:
In that case, open for business!

sofhasissuesx:
hi! is this issue still open to work on?

wyattmccurdy12:
Hi everyone, I use this site all the time and I'm also trying to build my programming CV: I'd like to contribute if I can. 
I noticed that this was a good first issue, so I want to ask a clarifying question about how the "hidden functions" work. 

I found a couple areas where `_setXyz` is used outside of the function definition of `setXyz`. For example, in `src/components/PaginatedTable/PaginatedTable.tsx`, I found that `setPage` and `_setPage` exist. However, `_setPage` is called outside of `setPage`, on line 171. 
[...]

PR #2346, opened 2023-08-19
Standardization #2271

JoshuaNascimento:
# Summary of Issue - [#2271](https://github.com/online-go/online-go.com/issues/2271)

The Settings folder was noted to contain a preferred scheme for the setters of state and preferences.
Where the set function prefixed with an underscore ("_setterExample"), acts as the "real" internal setter.
While another setter without the prefix ("setterExample"),
offers to perform any extra logic required prior to then calling the internal setter.
However, this format was not applied universally to the files within the Settings Folder.

## Description of Changes
Within the files of the Settings folder all state and preferences which did not use a prefixed setter were altered.
The prefix setter was added and a function to the external setter was implemented with type parameters when able.

For Instance: **_HelpSettings.tsx_**
`const [reload, setReload] = React.useState(false);`
`React.useEffect(() => {
        if (reload) {
            setReload(false);
        }
    });`

Was changed to:
`const [reload, _setReload] = React.useState(false);`
`function setReload(tf: boolean) {
        _setReload(tf);
    }`
`React.useEffect(() => {
        if (reload) {
            setReload(false);
        }
    });`

As noted, this change was implemented to create commonality within the Settings folder.
Furthermore, while the above example does not currently contain additional logic prior to changing the setReload state.
We are now provided a clear function to handle any logic we may wish to implement in the future, for any state/preference.

# Testing

To ensure these changes didn't introduce any issues, the tests provided within the repo were run with results below.

image

### Aside

👋 Hey! 

First pull request for an open-source project. Looking to get my feet wet and contribute to real codebases outside of a couple small personal projects. I would greatly appreciate any constructive criticism regarding these changes 😊

### DIFF
--- a/src/views/Settings/AnnouncementPreferences.tsx
+++ b/src/views/Settings/AnnouncementPreferences.tsx
@@ -33,7 +33,7 @@ import { Toggle } from "Toggle";
 import { PreferenceLine } from "SettingsCommon";
 
 export function AnnouncementPreferences(): JSX.Element {
-    const [blocked_players, setBlockedPlayers]: [
+    const [blocked_players, _setBlockedPlayers]: [
         Array<any> | null,
         (x: Array<any> | null) => void,
     ] = React.useState(null);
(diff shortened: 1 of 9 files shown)


anoek:
Welcome @JoshuaNascimento, thanks for your effort, however I think I like it the way we have it. Basically how to approach and think of it is if you see a setter with an underscore, you know there's a function defined down below that does some extra stuff, otherwise you can expect the setter to be called directly. This change, while adding consistency, makes it a bit harder to see at a glance what has special logic associated with it, it also makes it more verbose, so overall I like it how it is better. Thanks all the same though!

JoshuaNascimento:
I see the intent of the variable scheme now thanks for the clarification. Attempting to tackle the issue allowed for a good introduction into this project for myself. So I’ll poke around and make myself helpful, though maybe with some clarifying questions prior to jumping head first into an issue 😅.

PR #2347, opened 2023-08-23
More ViewReport compaction tweaks

GreenAsJade:
Ensure that the close buttons end up sensibly-near to the messaging buttons always, stop the related reports from wrapping and taking up more vertical space (and also tweak a message template title so it fits in the space)

Fixes stuff going off the bottom of the page

## Proposed Changes

Fix the layout after previous squishing up.


anoek:
Thanks!

PR #2348, opened 2023-08-23
Improve handling reviews/demos without historical rating

jsonn:
Freshly created demo boards don't necessarily provide the historical ratings object. This can easily be observed by loading an appropriate demo board with Firebug open and "Pause on exceptions" active.


anoek:
Thanks for the fix!

Issue #2302, opened 2023-06-18
UI does not display user after SSO sign in


benjaminpjones:
**Describe the bug**
When I log in using Google SSO, I am redirected to the Watch page and it looks like I am not logged in.  Refreshing the page fixes this issue.

**To Reproduce**
1. If logged in, log out.
2. Go to the Sign In page (`https://online-go.com/sign-in`)
3. Click **Sign In with Google**
4. Observe: you are redirected to the homepage, but it looks like you are logged out.
    - In the top right, a link to the sign in page
    - The central content is the **Observe Games** page, not the user's games

**Expected behavior**
The nav bar should show the user that is logged in, and the central content should be the user's games and other info.

**Screenshots**

https://github.com/online-go/online-go.com/assets/25233703/29ef97be-c580-44fb-8272-980db076eb5d

**Desktop (please complete the following information):**
 - OS: macOS Ventura 13.4
 - Browser: Firefox 114.0.1 [64-bit]

JoshuaNascimento:
This seems like a relatively simple fix as when you login with a direct account the login method also forces a refresh of the window which then allows it to update the user information. 

But with logging in through third-party this functionality doesn't exist. I could definitely take a crack at this but I am unsure how I would get these social logins to work on a local server to do testing.

@anoek or perhaps @GreenAsJade would you mind pointing me in the right direction to get social logins to work locally? 

Looking at the Network Activity in browser it's clear that when running locally a call is made to "http://dev.beta.online-go.com:8080/login/google-oauth2/" but the auth flow stops there and i'm simply redirected back to the site.

anoek:
Yeah unfortunately I don't think it's going to work locally, SSO providers lock down the urls that you can redirect to pretty tightly.

JoshuaNascimento:
Would there be anyway for me to test this prior to making an official PR?

GreenAsJade:
I think not, which is perhaps one reason why it's a laid there for a while.   Does SSO work on beta, when logging in direct?   If that's the case, then you can make a PR and get it deployed there for testing.    

It's painful enough that personally I think we'd love you to try no matter what ... let's get it fixed eh 😅

PR #2349, opened 2023-08-29
Properly update profile UI with SSO login.

JoshuaNascimento:
# Summary of Issue - #2302
It was noted that when logging in using the Google SSO a user can successfully login. However, when redirected back to the site, the UI does not update with the user's profile information until after a manually page reload.

This is because the behaviour of direct and third-party login is different. When a user logs in with a site-specific account the page is also refreshed. But when logging in via a third-party they are taken through the OAuth process and returned back to the site, after which the user config (_the data stored in useUser_) is updated . Due to how the user data is stored in the _NavBar_ component a re-render is never triggered.

## Description of Changes
Within _NavBar_ state was added using the previous _user_ variable name to allow the component to determine if a re-render is necessary. This state updates via a useEffect call of the _userData_ const which calls _useUser()_. Therefore, when a change occurs to _userData_ the useEffect hook is also called setting _user_ state to the new value. Which in turn causes a component re-render.

In addition, reformatted _ProfileAndQuickSettingsBits_ function to use optional chaining when referencing _user_ state to prevent errors in the case that user has not yet been set (user is not logged in).

## Testing
Provided the current test it appears these changes do not introduce any issues*
image

**\*** Due to the inability to test third-party login on my local system, I was unable to check if the issue is actually resolved with these changes. The changes will need to be deployed somewhere for proper testing and feedback. Using my best judgement as to the issue and potential solution, this should theoretically work 🤞.

## Additional Comments
I'm unsure of the best practice to use but noticed both _main.tsx_ and _routes.tsx_ make redundant calls to config.user, assigning it to the same variable across multiple functions. From what I'm aware of it would be best to introduce Redux to store this and other important state for the sake of a structured data flow. Not sure if this is something that would want to even be addressed but noticed it and wished to comment on it here.

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -69,9 +69,9 @@ const setThemeDark = setTheme.bind(null, "dark");
 const setThemeAccessible = setTheme.bind(null, "accessible");
 
 export function NavBar(): JSX.Element {
-    const user = useUser();
     const location = useLocation();
-
+    const userData = useUser();
+    const [user, setUser] = React.useState(userData);
     const [search, setSearch] = React.useState<string>("");
     const [search_focus, setSearchFocus] = React.useState<boolean>(false);
     const [omniMouseOver, setOmniMouseOver] = React.useState<boolean>(false);
@@ -81,7 +81,6 @@ export function NavBar(): JSX.Element {
     const search_input = React.useRef<HTMLInputElement>(null);
     const [force_nav_close, setForceNavClose] = React.useState(false);
     const [banned_user_id] = useData("appeals.banned_user_id");
-
     const { registerTargetItem } = React.useContext(DynamicHelp.Api);
 
     const { /* ref: toggleRightNavButton, */ used: rightNavToggled } =
@@ -137,6 +136,12 @@ export function NavBar(): JSX.Element {
         // remote-storage update mechanism, which doesn't call setTheme()
         data.watch("theme", _update_theme);
     }, []);
+
+    // Whenever the data held in user config changes we set user state to that
+    // Changing user state causes the component to re-render with the correct user information
+    React.useEffect(() => {
+        setUser(userData);
+    }, [userData]);
 
     //const valid_user = user.anonymous ? null : user;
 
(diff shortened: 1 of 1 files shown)


anoek:
I've deployed this branch to beta for testing.

JoshuaNascimento:
Still seems to be an issue logging in with the third-party methods

anoek:
Well, the video shows the normal username/password login which was working just fine. The issue is with SSO, which apparently seems to still be the same :( (although not surprising, the `useUser` *should* cause a refresh, so in theory the updated code wouldn't change the behavior.. but good to have confirmation of that :)

JoshuaNascimento:
> 

Realized I checked the wrong login behaviour with the video demo shortly after making it my bad.

Perhaps passing in the data from useUser as a prop for NavBar would result in the correct behaviour?
Currently my only other solution would be utilizing the Google specific library to handle the login. Creating an async function where we could then call a page refresh once the promise has been fulfilled. With the caveat being we would have to do something similar for the other SSOs, provided they offer a similar library.

I'll reset and do some more thinking about this.

JoshuaNascimento:
> I'm wondering if this is at a lower level, a backend issue perhaps

Was assuming this was a potential cause as well. Though I am unaware of how the backend processes the auth. With the fact that the issue self-corrects on a manual page refresh I'm assuming it might just be a "simple" issue regarding flow of the app.

login auth -> redirect uri -> backend processing

Means the frontend will always be acting ahead of the data the backend receives then updates. and without a current way for the frontend to passively observe or be notified of changes to the user config, were left with the active method of either manually refreshing the page or some bandaid solution like pinging the useUser method for changes every "x" amount of time.

anoek:
Ok so I looked into what was going on, I think the crux of the issue is our user data wasn't getting updated when we came back from our `ui/config` fetch. https://github.com/online-go/online-go.com/pull/2350 seems to make things work, however I think before merging that I'd like to go over it with some fresh eyes in the morning / next few days. It might be a better fix exists.

anoek:
Closing in favor of https://github.com/online-go/online-go.com/pull/2350

PR #2350, opened 2023-08-29
Set user and config.user when we get ui/config results

### LINKED ISSUE #2302: UI does not display user after SSO sign in

anoek:
Fixes #2302

It's late and this is not vetted, I feel like this shouldn't be necessary so I think before merging I'd like to understand why this seems to be necessary. A better fix might exist.

### DIFF
--- a/src/lib/cached.ts
+++ b/src/lib/cached.ts
@@ -69,6 +69,8 @@ export const cached = {
                     disable_refresh_callback_on_user_change = true;
                     try {
                         data.set(cached.config, config);
+                        data.set("user", config.user);
+                        data.set("config.user", config.user);
                         data.set("config", config);
                         if (cb) {
                             cb();


JoshuaNascimento:
Got back in and took a look through the _data.tsx_ file out of curiosity for this issue.
Would a potential solution be to simply call _watch()_ defined in _data.tsx_ on user.config?

Since we know the user information is updated eventually from the backend. Then by creating a watcher on that data we can provide it with some functionality to either update the user state in a parent component and pass it down into _NavBar_. Or we could have the watcher simply in _NavBar_ checking the value of user.config against what we have as the current user in that component, then updating/refreshing the page if these values are different.

Assuming this is a potential direction we would like to go down I can put up a PR when I get back in. Otherwise, what you already have may be the better of the two solutions as it seems slightly simpler.

anoek:
Once upon a time we were doing something similar to what you're suggesting, watch on "config" and then setting "user" and "config.user" from that.. however I can't recall if we had troubles with that or removed it for some other reason.

It's probably worth exploring, but definitely needs some testing to make sure nothing is wonky

benjaminpjones:
I don't think I ever figured out the difference between "config.user" and "user" 😄

anoek:
There's no difference these days I don't think (and I'm not sure if there ever was)

JoshuaNascimento:
Hmm, so if I'm understanding right and the useData hook functions identical to calling a watcher to monitor live changes, and we were already using that hook within NavBar. Then how that component is receiving user info should be fine and not require change. Instead, how the user is being set within the frontend to begin with being the root issue. Again making this change the best option?

anoek:
Yep I think so. I went through the code again this morning and I believe this is safe and seems to fix it up.

PR #2351, opened 2023-08-29
Community Moderator support: Moderator power: annul

GreenAsJade:
Fixes too few people attending to score cheat reports.

## Proposed Changes

Users can be given `MOD_POWER_ANNUL` as a `moderator_power`, via the edit user modal.

Users with `moderator_powers` can see score cheat reports, and claim them.   They can send warnings, system PMs.

Users with the `MOD_POWER_ANNUL` bit set can  annul games.

All community moderator messages, warnings are logged (forced, no choice)

Some sections of reports remain hidden (the anticipation is that these would be shown to users with `MOD_POWER_SEE_AI_TOOLS`).

Needs https://github.com/online-go/ogs/pull/1827 and https://github.com/online-go/ogs-node/pull/40

### DIFF
--- a/src/components/ModerateUser/ModerateUser.tsx
+++ b/src/components/ModerateUser/ModerateUser.tsx
@@ -24,6 +24,8 @@ import { proRankList } from "rank_utils";
 import { Modal, openModal } from "Modal";
 import { lookup } from "player_cache";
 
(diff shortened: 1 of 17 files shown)


anoek:
Very cool, thanks!

PR #2353, opened 2023-08-30
Create Watcher to monitor and change user state when config.user is altered

JoshuaNascimento:
Alternative to #2350 

Unsure if this is the best way to go about it but interested as to if this solution also works. 
Setter isn't named correctly as `setUser` already shares a namespace with a `Sentry` function and not certain if that would be useful here. Though now that I think about it `_setUser` is probably the proper name as we implement some additional logic prior to setting the state.

There may also be some merit in having this logic outside of `NavBar` and passing it in as a prop. But for now just interested in if this works and is a solution we wish to proceed with.

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -69,7 +69,7 @@ const setThemeDark = setTheme.bind(null, "dark");
 const setThemeAccessible = setTheme.bind(null, "accessible");
 
 export function NavBar(): JSX.Element {
-    const user = useUser();
+    const [user, setUserData] = React.useState(useUser());
     const location = useLocation();
 
     const [search, setSearch] = React.useState<string>("");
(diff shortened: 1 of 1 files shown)


benjaminpjones:
> There may also be some merit in having this logic outside of NavBar and passing it in as a prop.

Hooks are also pretty easy to compose into new hooks (which keeps related logic in the same place).  Could do something like:

```
function useConfigUser() {
    const [user, setUserData] = React.useState(data.get("config.user"));

    // Watch for changes in config.user checking if these changes differ from current user state
    data.watch("config.user", (userConfig) => {
        // Check if user state current id differs from that in config.user
        if (userConfig.id !== user.id) {
            setUserData(userConfig);
        }
    });
    return user;
}
```

But actually, we have a [convenience hook for watching data](https://github.com/online-go/online-go.com/blob/42d9957567628aad9471fff81f654404e7fa9671/src/lib/hooks.ts#L22-L51) changes, which probably makes the need for a new function unnecessary:

```
const user = useData("config.user")
```

anoek:
Closing in favor of https://github.com/online-go/online-go.com/pull/2350

Issue #2164, opened 2023-02-19
Dancing screen: another chat variation-related bug


turtoise0:
**Describe the bug**
The size of the player cards sometimes changes when you hover over a variation link, which causes the chat area to move and the the cursor to move away from the link.

**Expected behavior**
The size of the player cards shouldn't change, or the chat area shouldn't move.

**Screenshots**

https://user-images.githubusercontent.com/99894371/219916772-a65462d1-42ac-4732-a104-3852648f3ba3.mp4

**Desktop (please complete the following information):**
- firefox on linux

turtoise0:
similar bugs: #1923, #1941, #2145

benjaminpjones:
Marking this as "good first issue".  I think this can be addressed with some more robust CSS.

A little hard to repro, but I think these things need to be in place:
 - The text width of the points on the player cards need to change when hovering the variation (helps if the variation is early game and points are single-digit)
 - The dimensions need to be just right.  I had luck in the 1600 pixel width range
 - In order to get the dancing variation, the chat needs to be scrolled up a bit (I wasn't able to get variation links to move if the chat was scrolled all the way to the bottom)

## Code points

- [PlayerCards.ts](https://github.com/online-go/online-go.com/blob/42d9957567628aad9471fff81f654404e7fa9671/src/views/Game/PlayerCards.tsx) - most likely would not need to be updated, but just for reference
- [PlayerCards.styl](https://github.com/online-go/online-go.com/blob/42d9957567628aad9471fff81f654404e7fa9671/src/views/Game/PlayerCards.styl) - this is likely where a change would need to be

If you're not familiar with Stylus, see https://stylus-lang.com.  It's very similar to CSS, but with some powerful additional features.

PR #2355, opened 2023-08-31
added margin to center-col to resolve a dancing screen issue related …

JoshuaNascimento:
# Summary of Issue
Originally was attempting to recreate the issue described in #2164. Instead, managed to find a similar "dancing screen" bug related to the horizontal resizing of the site in a browser window. It seems at certain horizontal sizes the overlap between the center-col and adjacent containers would cause the flex-basis value to infinitely cycle between a range.

https://github.com/online-go/online-go.com/assets/59673146/e505101a-5aab-42fd-8a7d-6c63876dcc73

## Description of Changes
Within the styling of `center-col` in the `MainGobanView` container, a margin of 0.5 rem was added. This was done to create a consistent spacing between `center-col` and `right-col.` which in combination with the scroll bar appearing seemed to prevent `center-col` from choosing a consistent width/height value.

## Testing
This issue could potentially be browser specific so testing was done firstly in Opera GX then in Firefox. Both seeming to no longer contain this bug.

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -288,8 +288,6 @@ goban-view-bar-width=400px
         }
     }
 
-
-
     .goban-container {
         flex-grow: 1;
         display: flex;
@@ -395,6 +393,7 @@ goban-view-bar-width=400px
         flex-direction: column;
         align-items: stretch;
         width: 100%;
+        margin: 0.5rem;
     }
     &.portrait .center-col {
         .cancel-button {


JoshuaNascimento:
> If this is a horizontal scroll thing, does changing margin: 0.5rem to something like margin-right: 0.5rem and/or margin-left: 0.5rem also resolve the issue?

Altered it to margin-right so the left-side of the page has a little extra space for the game board to take up.

Also from my interpretation of this issue it seems to be a weird interaction between `center-col` expanding via flex-grow/shrink without taking into account the `right-col`. After it expands it is within the bounds of the right-col element so it then shrinks itself to readjust. But this interaction keeps flipping back and forth between grow and shrink. So to address it we can add the margin to prevent center-col from expanding into right-col. 

This might be the issue of #2164 and if i can find a way to recreate that bug I'll look into addressing that as well.

anoek:
Thanks for finding and fixing this!

Issue #1941, opened 2022-07-15
Disappearing goban title (above chat area) causes goban and chat area to flicker [seizure warning]


turtoise0:
**To Reproduce**

0. start a game
1. on your turn, create variation with even number of moves
2. place a stone on the board, but don't submit the move
3. hover on the variation

**Screenshots**
[flicker.webm](https://user-images.githubusercontent.com/99894371/179184658-41a9b3df-0d5e-4238-ba4c-c75d688a4749.webm)

**Additional context**
It produces a similar effect to #1923.

anoek:
This seems to have been fixed at some point

PR #2356, opened 2023-09-01
Created .styl for PlayControls and added a min width to avoid compone…

JoshuaNascimento:
# Summery of Issue
It was noted in issue #1941 that when a user creates a variation then hovers over said variation. The chat area shakes and creates a visual bug.

The perceived cause of this issue being when a user hovers a variation, it removes the span `game-state`. This causes the `play-controls`  div to shrink in size with the chat area filling in the missing space. Since this moves elements on the screen the mouse will no longer be hovering over the variation and so `game-state` will now have a span element rendered again. increasing its size. Which in turn pushes the chat area back down placing the user's cursor on the variation once again, cycling infinitely.

## Description of Changes 

Added a .styl file for `PlayControls` and gave it a min-height. Allowing it to maintain a consistent size whether or not the `game-state` span is rendered.

5 rems was chosen as play-controls takes up a consistent 77.59 pixels when testing in my browser. Given the default size of a rem is 16px, we get 16*5 = 80 pixels which while technically larger is visually identical to the previous dimension.

## Testing
https://github.com/online-go/online-go.com/assets/59673146/e52d4cfc-6585-4b2d-bc16-23c01fa23a68

### DIFF
--- a/src/views/Game/PlayControls.styl
+++ b/src/views/Game/PlayControls.styl
@@ -1,0 +1,3 @@
+.play-controls {
+  min-height: 5rem;
+}


anoek:
Thanks for fixing that!

PR #2357, opened 2023-09-05
Don't show the score during stone removal.

GreenAsJade:
Fixes tempting players to change the score by seeing the actual score there.

## Proposed Changes

Don't show score during stone removal.

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -277,7 +277,7 @@ export function PlayerCard({
     const highlight_their_turn = their_turn ? `their-turn` : "";
 
     const show_points =
-        (engine.phase === "finished" || engine.phase === "stone removal") &&
+        engine.phase === "finished" &&
         goban.mode !== "analyze" &&
         engine.outcome !== "Timeout" &&
         engine.outcome !== "Resignation" &&


benjaminpjones:
This also makes it harder to detect errors in scoring (and the majority of players aren't cheating!)

GreenAsJade:
Any thoughts about how we can tackle this?    Maybe we need to go "back to the forum" to re-explore.

One thing I've been thinking is that score-griefers don't need to know the actual score to decide to grief ... that's why I only referred to "temptation" rather than "actual fix" in the description.

benjaminpjones:
I think it would be hard to remove the temptation from score griefers as long as score exists.  I don't really have constructive ideas on how to make them stop wanting to do it.  Maybe the forumites have some ideas, but personally I am pessimistic about changing griefers' motivations.

As far as detecting griefers so that they can be warned or removed.. I think it's possible to use signals like:

- KataGo estimate vs outcome
- Number of back-and-forths on the score

kevincoleman:
It seems to me like if you’re adept at counting during the game then showing the score during stone removal would be one additional tool for making sure you actually scored correctly. We probably shouldn’t damage the experience for the honest just because it’s also usable by griefers.

PR #2362, opened 2023-09-11
improve conditional display of provisional rank in mini-goban

kevincoleman:
Fixes #
Adds a check to display [?] on the minigoban

## Proposed Changes
Just a bug fix.
  -
  -
  -


anoek:
Thanks for fixing that!

Issue #2210, opened 2023-04-02
in-game "Add to library" option to add SGF file to your SGF library


csessh:
**Is your feature request related to a problem? Please describe.**
At the moment, I have to download the game SGF with `Download SGF` button

![Screenshot 2023-04-02 at 9 51 25 am](https://user-images.githubusercontent.com/5171823/229323933-63c8bf76-661b-46b7-bd1c-a253946244eb.png)

and then go to `Tools > SGF library` to upload said downloaded SGF file. 

**Describe the solution you'd like**
Perhaps a button `Save SGF to library` along side with `Download SGF` to prompt user with a selection of existing libraries

PR #2364, opened 2023-09-12
Add to SGF Library Directly From Game

JoshuaNascimento:
## Summery of Issue and Implementation
There is a feature request to add an in game option to save the given game to a user's SGF Library #2210.

I wish to attempt this issue and have begun implementing a very rough Modal and Tooltip component to allow for this request.
However, it seems I've run into 2 issues while working with this.

Firstly, It seems from looking through other post requests made in the code. The only way to add a file to library is to make a post to `me/games/sgf/%%`. With the additional parameters of `collection_id` and a `SGF file`. But the only way to obtain the SGF file without the user having to manually download then upload it themselves is to access the `api1('games/${game_id}/sgf')` endpoint directly. This doesn't seem like a major issue, we would simply use api1 to receive the file at the endpoint and use that to then make the post call to have the same file added to the user's library.

Unfortunately I run into my second issue being that the api1 endpoint does not work in the development environment (or at least not for me). So I'm unsure if simply calling a fetch (or get?) on it would actually result in obtaining the file, and without being able to test that myself I can't move forward with seeing if I can post the fetched file into the user's library.

I'm hoping for a bit of guidance from those more comfortable with this codebase to either confirm my current thinking and setup "should" work or if there is a simpler/better way to access the SGF file using some request I'm unaware of. From there I would make a more official PR to test the functionality and move on with my implementation from there.

### DIFF
--- a/src/views/Game/GameDock.tsx
+++ b/src/views/Game/GameDock.tsx
@@ -313,6 +313,7 @@ export function GameDock({
             : engine.playerNotToMove();
 
     const showLibraryModal = () => {
+        // Pull user data prior to opening modal preventing component from rendering prior to obtaining needed data
         const user = data.get("user");
         const promise = get("library/%%", user.id);
         promise
--- a/src/views/Game/GameDock.tsx
+++ b/src/views/Game/GameDock.tsx
@@ -20,7 +20,7 @@ import * as data from "data";
 import * as preferences from "preferences";
 import { MAX_DOCK_DELAY } from "SettingsCommon";
 import { useUser } from "hooks";
-import { api1, post, del } from "requests";
+import { api1, post, del, get } from "requests";
 import { Dock } from "Dock";
 import { Link } from "react-router-dom";
 import { toast } from "toast";
@@ -28,6 +28,7 @@ import { _, pgettext } from "translate";
 import { openACLModal } from "ACLModal";
 import { openGameLinkModal } from "./GameLinkModal";
 import { openGameLogModal } from "./GameLogModal";
+import { openGameLibraryModal } from "./GameLibraryModal";
 import { sfx } from "sfx";
 import { alert } from "swal_config";
 import { challengeFromBoardPosition } from "ChallengeModal/ForkModal";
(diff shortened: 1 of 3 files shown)


GreenAsJade:
@JoshuaNascimento  Can you PM me on OGS (site or forum) and we'll set about helping.

JoshuaNascimento:
@GreenAsJade did so via site PM!

JoshuaNascimento:
I think with my latest commit I have a majority of the feature complete. Ill implement some sort of feedback mechanic so users know if their submission to add to library and/or create a library was successful. But asides from that and final styling it should be good enough to play around with and get feedback from yall.

src/views/Game/GameDock.tsx
```diff
@@ -311,6 +312,18 @@ export function GameDock({
             ? engine.playerToMove()
             : engine.playerNotToMove();
 
+    const showLibraryModal = () => {
+        // Pull user data prior to opening modal preventing component from rendering prior to obtaining needed data
```
anoek, line 316: We need to open the modal and show a Loading indicator, you can probably just drop in `<Loading />` while you wait for the data, that way people know their action has been registered and they just need to be patient while we wait for the internet.

anoek (changes requested):
Nice start! It seems functional, we need to work on the aesthetics a bit though. Here's what I'm seeing:

When there are no collections:
image

When there is one collection:
image

When there are a few collections:
image

I think we need to make this experience more consistent and scalable, as some folks have a lot of collections, and except for the case when you have no collections, there's no way to create a new collection here. There's also no indication whether a collection is private or not.

Two possibilities come to my mind, one would be to have a drop down to select your collection along with a way to add a new collection/sub-collection.  The other would be to mirror what we have in the SGF library so it's a more consistent experience, something like:

image

For that I suspect you could rip out a lot of the code in the SGF library and make it a component that you then load in this modal kind of thing, and have a callback to save the SGF.

csessh:
@anoek @JoshuaNascimento you guys mind if I picked this up where it's currently left off?

Issue #2360, opened 2023-09-06
Shorter disconnection timer


raylillywhite:
**Is your feature request related to a problem? Please describe.**
I've only been playing for two weeks and I've already had to report 5 users who just close the tab or something when they're losing, without resigning. This is nowhere near as bad of a problem on chess.com and lichess because they forfeit you after a percentage of the total game time if you haven't reconnected, even if you still have time on your clock. One game someone stalled for multiple minutes so I started another game, then they started playing again, so I lost both games because I tried to play both and ran out of time.

**Describe the solution you'd like**
Shorten the time that a user has after disconnecting before the game is forfeit. Base the disconnection timeout on a percentage of estimated total game time, so that faster games have shorter timeouts.

**Describe alternatives you've considered**
- More aggressively ban users who disconnect without resigning (also good)
- Detect if they're clearly losing when they disconnect (this would be great but I assume it's harder)

anoek:
There's a claim victory on potentially abandoned games now after a short time, which I think addresses this issue

PR #2372, opened 2023-09-20
Anti griefing features

anoek:
This patch includes two features to help thwart people being rude and unsportsmanlike.

The Anti-stalling code will appear if your opponent is not passing, the game is toward the end, and KataGo thinks there is a >99% chance that one of the players will be victorious. In that case, the players will be given a special control box under their play controls to end the game:

image

Similarly, the anti-escaping code will appear if a player has seemingly left a live game without finishing it. This box will show up after 30 seconds has passed and gives the opportunity to the remaining player to end the game immediately or continue waiting to see if their opponent returns.

image

The person who left the game will also be shown a `Toast` in the upper left informing them that this will happen, clicking the toast will bring them back to the game.

image

### DIFF
--- a/src/views/Game/AntiGrief.tsx
+++ b/src/views/Game/AntiGrief.tsx
@@ -21,8 +21,63 @@ import { pgettext, _ } from "translate";
 import { useGoban } from "./goban_context";
 import { useUser } from "hooks";
 import { JGOFClockWithTransmitting, JGOFTimeControl } from "goban";
+import { browserHistory } from "../../ogsHistory";
+import { toast } from "toast";
+
+let on_game_page = false;
+let live_game = false;
+let live_game_id = 0;
+let live_game_phase = null;
+let last_toast = null;
+
+function checkForLeavingLiveGame(pathname: string) {
+    try {
+        const goban = window["global_goban"];
+        const was_on_page = on_game_page;
+        const was_live_game = live_game;
+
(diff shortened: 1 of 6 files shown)


bhostetler18:
Looking forward to this one!

src/views/Game/AntiGrief.tsx
```diff
@@ -0,0 +1,283 @@
...
+                )}
+            </div>
+            <div>
+                <button
+                    className="danger"
+                    onClick={() => goban?.sendPreventEscaping(my_color, false)}
+                >
+                    {pgettext(
+                        "This button is shown when one player has left the game, it allows the other player to end the game and claim victory",
+                        "End game",
```
raylillywhite, line 188:
```suggestion
                        "Claim victory",
```

I think lichess uses the button title Claim Victory, which I think makes it more clear how the result will be counted if you press it, but "End game" is fine too if it's clarified in the message above by saying "or end the game and claim victory."

raylillywhite (commented):
Both these features look great! 🙇🏻 

And can consider https://github.com/online-go/online-go.com/issues/2360 closed after this is merged.

src/views/Game/AntiGrief.tsx
anoek: Yeah that's better, thanks!

PR #2379, opened 2023-09-29
Improve styles in chat view (Issue #2375)

### LINKED ISSUE #2375: Community Tab UI Improvements
### Discussed in https://github.com/online-go/online-go.com/discussions/2373

Result of discussion: go ahead and try the improvements, possibly including some niceties like resizing the sidebars manually (maybe iteration number one will just use media queries wisely).

- [x] Add padding throughout interface
- [x] Improve sidebar colors for visual heirarchy
- [x] Test in both light and dark mode
- [x] Test in variety of sizes (media queries)
- [ ] Consider/implement resizable sidebars

kevincoleman:
Fixes #2375 

## Proposed Changes

Improve colors and styling on chat page
Does not yet implement splits for the sidebars, but instead makes them size-responsive (with limits)

Note: this “re-activates” the light theme’s `card-background-color`. It used to just lighten `#ffffff`, which was useless. This darkens it, but only very slightly, so it shouldn’t cause a big upset, but it is now useful.


anoek:
Thanks for the update!

kobakitty:
After using this new chat for a while, i gotta admit that i did like the older one more, it had nice old-school IRC feel to it ://

Would it be possible to have some option in the settings to choose between the new chat UI and the older one?

kevincoleman:
I could see the case for an old skool mode. I also dig IRC. However, if we do that I think it'd be better implemented as a site-wide style, not just this one view.

PR #2382, opened 2023-10-02
Get rid of pinging from UI code, take latency update from GobanSocket

GreenAsJade:
Fixes pinging twice as often as we imagined.

## Proposed Changes

  - Get rid of pings from UI code, let GobanSocket do that
  - Get latency updates from GobanSocket

### DIFF
--- a/src/lib/sockets.ts
+++ b/src/lib/sockets.ts
@@ -16,7 +16,7 @@
  */
 
 import Debug from "debug";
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Now needs https://github.com/online-go/goban/pull/129 , to support  configurable ping interval

anoek:
Thanks for cleaning that up, this was my intent I'm sure, I must have forgotten to cycle back to it after the goban code was finished or something :)

PR #2383, opened 2023-10-02
Network out notifier

GreenAsJade:
Fixes not noticing that there's a problem with the socket to OGS

## Proposed Changes
  - Display a flashing icon if we've lost contact (ping timeout or disconnect)

** needs https://github.com/online-go/goban/pull/130
  
Screenshot 2023-10-02 at 5 05 25 pm (3)

### DIFF
--- a/src/components/NetworkStatus/NetworkStatus.tsx
+++ b/src/components/NetworkStatus/NetworkStatus.tsx
@@ -24,15 +24,18 @@ export function NetworkStatus(): JSX.Element {
     const [hidden, setHidden] = React.useState(true);
 
     React.useEffect(() => {
-        socket.on("latency", () => {
-            setHidden(true);
-        });
-        socket.on("timeout", () => {
-            setHidden(false);
-        });
-        socket.on("disconnect", () => {
-            setHidden(false);
-        });
+        const show = () => setHidden(false);
+        const hide = () => setHidden(true);
+
+        socket.on("latency", hide);
+        socket.on("timeout", show);
+        socket.on("disconnect", show);
+
+        return () => {
+            socket.off("latency", hide);
+            socket.off("timeout", show);
+            socket.off("disconnect", show);
+        };
     }, []);
 
     return (
--- a/src/components/NetworkStatus/NetworkStatus.tsx
+++ b/src/components/NetworkStatus/NetworkStatus.tsx
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


GreenAsJade:
(Build fails 'cause it needs the new goban)

src/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,44 @@
...
+    React.useEffect(() => {
+        socket.on("latency", () => {
+            setHidden(true);
+        });
+        socket.on("timeout", () => {
+            setHidden(false);
+        });
+        socket.on("disconnect", () => {
+            setHidden(false);
+        });
```
benjaminpjones, line 35: Probably doesn't have consequence in this case since this component is only mounted once, but it's good practice to unsubscribe from events in the [cleanup function](https://legacy.reactjs.org/docs/hooks-effect.html#effects-with-cleanup) of `useEffect`:

```
useEffect(() => {
    const show = () => setHidden(false);
    const hide = () => setHidden(true);

    socket.on("latency", hide);
    socket.on("timeout", show);
    socket.on("disconnect", show);

    return () => {
        socket.off("latency", hide);
        socket.off("timeout", show);
        socket.off("disconnect", show);
    }
}, [])
```

src/lib/sockets.ts
```diff
@@ -16,12 +16,15 @@
...
 import Debug from "debug";
-import { GobanSocket, niceInterval, protocol } from "goban";
+import { GobanSocket, protocol } from "goban";
 
 const debug = new Debug("sockets");
 
 export const socket = new GobanSocket(window["websocket_host"] ?? window.location.origin);
 
+socket.options.ping_interval = 10000;
+socket.options.timeout_delay = 2000;
```
anoek, line 26: Soo, depending on the day and the network weather, we can actually have folks who are using the site that have some fairly extreme latencies:

image

How about instead of `2000ms` we make that more dynamic? Something like, start it off at say 10s, and then scale it when we get a new latency to say either 2x or 3x their latency? That way we trigger sooner for those that have been on a very fast connection, and a lot slower for those that are across an ocean on a saturated link?

src/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,47 @@
...
+            socket.off("latency", hide);
+            socket.off("timeout", show);
+            socket.off("disconnect", show);
+        };
+    }, []);
+
+    return (
+        <div className={"NetworkStatus" + (hidden ? " hidden" : "")}>
+            <i className="fa fa-2x fa-ban" />
+            <i className="fa fa-wifi" />
```
anoek, line 44: I'm thinking we should probably have some text here describing what's going on for those not so tech-literate

anoek (changes requested):
Great feature to have, thanks for making it happen :)

src/lib/sockets.ts
GreenAsJade: Yeah - I had no data about what latencies we experience, but I did wonder about basing the timeout on latency like that.

Okie dokie

src/components/NetworkStatus/NetworkStatus.tsx
GreenAsJade: Okie dokie

src/lib/sockets.ts
GreenAsJade: Okie dokie

GreenAsJade:
Note: dynamic ping interval needs https://github.com/online-go/goban/pull/131 , but this PR still works without that (the ping interval just doesn't change)

anoek (approved):

anoek:
Nice, thanks @GreenAsJade !

PR #2385, opened 2023-10-03
rounded corners, enum for state, comment update

GreenAsJade:
Fixes petty nit picking :D

(tweak while pondering tracking slow internet)


anoek:
Good stuff, thanks :)

PR #2386, opened 2023-10-04
Handle went away

GreenAsJade:
Fixes "disconnect" showing when the socket goes down for legitimate reasons, like browsing away

## Proposed Changes
 * Notice if it went down for a legitimate reason, don't show an error.

Needs: https://github.com/online-go/goban/pull/133

Is on top of #2385


anoek:
Cool, thanks for getting that figured out

PR #2387, opened 2023-10-04
Go slooow

GreenAsJade:
Fixes too much alerting people that their internet is slow

## Proposed Changes

  - Wind up the ping interval and timeout each time we time out.
  
(on top of #2386)


GreenAsJade:
Note: With a 6s delay in `pong` there's still no red box displayed.

The only time we should get them now is if they start with a latency >8s OR if their connection deteriorates at any point.   It's deterioration, not absolute value, that causes the "Slow Internet" timeout.

PR #2394, opened 2023-10-14
Show the networks status warning only to live game participants on the page

GreenAsJade:
(also added engine validity check in sockets just in case)

Fixes seeing the distracting banned unless absolutely necessary

## Proposed Changes

  - Show banner only to participants on a live game page.


GreenAsJade:
This'd be a good one for beta-first to test in real not-localhost conditions.

PR #2399, opened 2023-10-18
Fix bugged check for live game

GreenAsJade:
and  add console debug for users with trouble cases.

Fixes slow internet banner appearing on correspondence games

## Proposed Changes

Check for "live game" properly
Add console log showing user's network status parameters more clearly


GreenAsJade:
Hmm I wonder why I didn't experience that 🧐

PR #2416, opened 2023-11-08
Reports center bug fix 

RubyMineshaft:
Fixes bug caused by checking `detected_ai_games.length` on older reports

### DIFF
--- a/src/views/ReportsCenter/ViewReport.tsx
+++ b/src/views/ReportsCenter/ViewReport.tsx
@@ -490,9 +490,7 @@ export function ViewReport({ report_id, reports, onChange }: ViewReportProps): J
                         </div>
 
                         <div className="actions-right">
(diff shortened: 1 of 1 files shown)


src/views/ReportsCenter/ViewReport.tsx
```diff
@@ -490,7 +490,9 @@ export function ViewReport({ report_id, reports, onChange }: ViewReportProps): J
                         </div>
 
                         <div className="actions-right">
-                            {reportState !== "resolved" && report.detected_ai_games.length > 0 ? (
+                            {reportState !== "resolved" &&
+                            report.detected_ai_games &&
+                            report.detected_ai_games.length > 0 ? (
```
anoek, line 495: `report.detected_ai_games?.length > 0` would do this too FYI

PR #2420, opened 2023-11-12
Tighten up game list entries that aren't paused

### LINKED ISSUE #2418: Game list entries use twice the necessary vertical space
**Describe the bug**
Every game in the game list that has a clock and is NOT paused is followed by a blank line. This results in using twice the necessary vertical space.

**To Reproduce**
1. Start a few correspondence games that use some (any) time control.
2. Pause one of them.
3. Start a few games that don't use time control (no clock).
4. Start enough other games that the game list is a textual table, rather than pictures of boards (I think 12 or more does it?).
5. Observe the vertical space for game entries.

Note that every game with a clock takes two vertical rows, whether paused or not. The games that are not paused have a blank line. The paused games have a line that's mostly blank, but says "paused" (or "vacation", etc.).

**Expected behaviour**
Most games should take one row of vertical space. Only the paused game(s) should take two rows.

 - OS: macOS
 - Browser: Safari
 - Version: 17.1

**Additional context**
This used to do the expected thing, until cae93c52519786af0ebeaab9a31f54a1b5d2373e. That commit added a `<span class=transmitting>` to the relevant `<div>` for something to do with blitz games, and changed the `<div>` to always be emitted (whether or not the game was paused).

I have a patch that seems to fix the visuals for the case I'm bothered by (the display of a long list of correspondence games) by again suppressing the emission of the `<div>` when it's empty. But I'm not sure whether there's a reason it won't work correctly for the new `<span>` (I'm not sure how to trigger or test its behaviour).

The patch is commit f2ebda37ecbe88076c2116977ef834fb15e0c160, currently on a branch in my fork at https://github.com/dexonsmith/online-go.com/tree/hide-pause-and-transmit-when-empty. I can submit a pull request if it's the right general direction.

(Also, let me know if I should have started with a PR, rather than creating this issue... this is my first patch for the project.)

dexonsmith:
cae93c525 added the display of transmission problems to the game list. Perhaps inadvertently, this doubled the vertical space used for all games in the list.

Only use the vertical space when there is something to show.

Fixes #2418

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -28,11 +28,13 @@ export function Clock({
     color,
     className,
     compact,
+    lineSummary,
 }: {
     goban: Goban;
     color: clock_color;
     className?: string;
     compact?: boolean;
+    lineSummary?: boolean;
 }): JSX.Element {
     const [clock, setClock] = useState<JGOFClockWithTransmitting>(null);
     const [submitting_move, _setSubmittingMove] = useState<boolean>(false);
@@ -101,8 +103,6 @@ export function Clock({
         const need_small_main_time_font = prettyTime(player_clock.main_time).length > 8;
 
         const show_pause = !compact && clock.pause_state;
-        const show_transmitting =
-            (submitting_move && player_id !== data.get("user").id) || transmitting > 0;
 
         return (
             <span className={clock_className}>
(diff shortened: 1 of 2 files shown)


dexonsmith:
I've tested this manually by spinning up a local instance, creating a few games at beta.online-go.com, and seeing that the list looks correct. Please let me know if there's other testing I should be doing.

Also, based on @anoek's comment in #2418, maybe the `transmitting` span could/should be skipped entirely when the clock is displayed in the game list. Not sure whether that's worth figuring out as part of this PR... let me know if you want me to dig into it.

dexonsmith:
> Also, based on @anoek's comment in #2418, maybe the `transmitting` span could/should be skipped entirely when the clock is displayed in the game list. Not sure whether that's worth figuring out as part of this PR... let me know if you want me to dig into it.

I had a quick look and I think b2f2565 makes this cleaner. It adds a `lineSummary` flag to `Clock`, and uses that to decide whether to emit the transmitting span. This has the side effect of documenting that the clock is used in summaries.

Issue #2392, opened 2023-10-08
Branches disappearing during review


benjaminpjones:
**Describe the bug**
This pops in the forums quite a bit, so I am creating a tracking issue:

[When reviewing a game and using the left arrow key to navigate, MoveTree branches dissapear](https://forums.online-go.com/t/when-reviewing-a-game-and-using-the-left-arrow-key-to-navigate-movetree-branches-dissapear/49362) (Oct 2023)
[Bug in reviews - disappearing branches](https://forums.online-go.com/t/bug-in-reviews-disappearing-branches/47821) (May 2023)
[Variations and comments in reviews disappearing](https://forums.online-go.com/t/variations-and-comments-in-reviews-disappearing/30709) (Sep 2020)

**Steps to repro**

More detail in the linked forum posts, but if you use the left arrow key to navigate backwards, any branches attached to the node you are moving *to* get erased.

**Screenshots**

![](https://ogs-forums.s3.dualstack.us-east-1.amazonaws.com/original/3X/1/c/1c7f6ff335ef553b0e4e404bcf1f654c091ecfc6.gif)

GreenAsJade:
Somehow the phrase "during review" threw me off the scent here, I thought this was like the "sometimes the branches all disappear" somewhat ephemeral bug.

But this is "any time that you are looking at a review, the review information can disappear".

Fortunately, it it still there on a reload, but still it's pretty yuk.

anoek:
I'm having trouble reproducing this, are you two seeing it?

bpjatfb:
Hmm.. I'm not able to reproduce either (Firefox + Mac)

benjaminpjones:
Hmm.. I'm not able to reproduce either (Firefox + Mac)

GreenAsJade:
I upgraded the priority because... yep, I see it.   Chrome + Mac.   I'll try to find a case...

anoek:
Oh excellent, I've still been unable to reproduce it, glad you can

GreenAsJade:
Hah - dunno about "reproduce".  I've definitely experienced it, will go hunting.

GreenAsJade:
Here's a public review where it happens for me - backarrow across any branch removes the branch:

https://online-go.com/review/738788

I can't find a pattern in why some reviews exhibit this behaviour and some don't.

I haven't been able to reproduce it locally.

GreenAsJade:
[...]

PR #2422, opened 2023-11-15
Analysis is always enabled for finished games

dexonsmith:
Update the `isAnalysisDisabled` hook for the Goban engine to always return `false` when the game is finished.

This is a speculative fix for #2392. I can't test right now because beta.online-go.com isn't available. Posting anyway in case someone else gets a chance to test before I do.

### DIFF
--- a/src/lib/configure-goban.ts
+++ b/src/lib/configure-goban.ts
@@ -64,6 +64,10 @@ export function configure_goban() {
         },
 
         isAnalysisDisabled: (goban: Goban, perGameSettingAppliesToNonPlayers = false): boolean => {
+            if (goban.engine.phase === "finished") {
+                return false;
+            }
+
             // The player's preference setting to always disable analysis overrides the per-game setting for
             // their own games.
             if (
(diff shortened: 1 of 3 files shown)


dexonsmith:
Good news; this patch seems to fix the problem.
- Played a game here: https://beta.online-go.com/game/16046
- Made a review here: https://beta.online-go.com/review/403 (just one branch, not very exciting)
- Triggered the bug by clicking "left arrow" using beta.online-go.com
- Does not happen when I click "left arrow" using dev.beta.online-go.com:8080 (with my fix)

anoek:
I think it makes sense, when the game is finished analysis should always be enabled so I think it's correct to have that function return false when the game is over

dexonsmith:
Okay, give me a moment, and I'll push a cleanup commit that cleans up other call sites.

anoek:
Awesome, thanks for hunting and fixing this.

I'm about to hop on a plane so it might be a bit before I merge and deploy, but in the mean time, @benjaminpjones @GreenAsJade  - does this fix all of the cases you two hit?

benjaminpjones:
I hadn't run into this bug myself, but this seems to fit the description of the Forum reports.

Looks much cleaner this way! Nice fix @dexonsmith!

PR #2425, opened 2023-11-17
Shorten names for tournament/ladder games in game lists

### LINKED ISSUE #2424: Relevant parts of game names not visible for tournaments and ladders in game lists
**Describe the bug**
Game names for tournaments and ladders are long always start with the same prefix. This is the only text visible in game lists, whether active games or lists of previously played games, because the relevant text overflows.

This is borderline between an enhancement and a bugfix. Let me know if you'd prefer a discussion on the forums before proceeding. In any case, I have a patch, and I'll post a pull request in a moment.

dexonsmith:
Show more of the relevant part of the game name in game lists before the text overflows, by replacing:

- the "Tournament Game:" prefix with the trophy icon and
- the "Ladder Challenge:" prefix with the ladder icon.

Without this change, the now-shortened prefix was the only visible text in such games.

This fixes both `GobanLineSummary` and `GameHistoryTable`.

Fixes #2424

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -28,11 +28,13 @@ export function Clock({
     color,
     className,
     compact,
+    lineSummary,
 }: {
     goban: Goban;
     color: clock_color;
     className?: string;
     compact?: boolean;
+    lineSummary?: boolean;
 }): JSX.Element {
     const [clock, setClock] = useState<JGOFClockWithTransmitting>(null);
     const [submitting_move, _setSubmittingMove] = useState<boolean>(false);
@@ -101,8 +103,6 @@ export function Clock({
         const need_small_main_time_font = prettyTime(player_clock.main_time).length > 8;
 
         const show_pause = !compact && clock.pause_state;
-        const show_transmitting =
-            (submitting_move && player_id !== data.get("user").id) || transmitting > 0;
 
         return (
             <span className={clock_className}>
@@ -168,9 +168,10 @@ export function Clock({
                         </React.Fragment>
                     )}
 
-                {(show_pause || show_transmitting) && (
+                {(show_pause || !lineSummary) && (
                     <div className="pause-and-transmit">
-                        {show_transmitting ? (
+                        {(submitting_move && player_id !== data.get("user").id) ||
+                        transmitting > 0 ? (
                             <span
                                 className="transmitting fa fa-wifi"
                                 title={transmitting.toFixed(0)}
--- a/src/components/GobanLineSummary/GobanLineSummary.tsx
+++ b/src/components/GobanLineSummary/GobanLineSummary.tsx
@@ -35,6 +35,13 @@ interface UserType {
     professional?: boolean;
 }
 
+export type GameNameForList = {
+    original: string;
+    text: string;
+    span?: string;
+    prefix?: string;
+};
+
 interface GobanLineSummaryProps {
     id: number;
     black: UserType;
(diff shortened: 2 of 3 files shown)


dexonsmith:
BTW, I'm new to React and TypeScript... the initial patch shares the core logic for this change between the two places, but duplicates the rendering code. Not sure if this is okay, or too spaghetti-like.

I had a look at sharing the rendering code but what I came up with (changing the new `GameNameForList` type to a `class` and adding a `render()` function) seemed too heavyweight for such a small use case. Let me know if you want me to look more in that direction.

dexonsmith:
Just pushed d4f047b5f318ed84582845035dc4ffbd9d3db62b to respond to CI failure (I hadn't seen failures locally when testing with beta.online-go.com... looks like the CI checks more thoroughly!).

benjaminpjones:
> I had a look at sharing the rendering code but what I came up with (changing the new `GameNameForList` type to a `class` and adding a `render()` function) seemed too heavyweight for such a small use case.

React has Functional Components that are a lot lighter weight than class components.  In modern React these are preferred anyway ([the React docs](https://react.dev/reference/react/Component) only cover class components as a "Legacy API").

This PR looks good to me as-is, but I've opened https://github.com/dexonsmith/online-go.com/pull/1 as an example of how you can dedupe the render code too.  IMO pretty light-weight or maybe lighter since there is no longer a need for an intermediate object.

dexonsmith:
> This PR looks good to me as-is, but I've opened PR [dexonsmith#1](https://github.com/dexonsmith/online-go.com/pull/1) as an example of how you can dedupe the render code too.

Awesome; your version is what I might have hoped for, so I merged it in.

dexonsmith:
Some screenshots.

Ladders before / after / after with title:
LadderBefore
LadderAfter
LadderTitle

Tournaments before / after / after with title:
TournamentBefore
TournamentAfter
TournamentTitle

dexonsmith:
> This PR looks good to me as-is, but I've opened PR [dexonsmith#1](https://github.com/dexonsmith/online-go.com/pull/1) as an example of how you can dedupe the render code too.

(BTW, if you preferred my patch before your change, I can force push back to that version. But it seems useful to me for the archived games and current games to share the rendering code, in case there are other tweaks to make, and this seems pretty clean.)

BHydden:
Long overdue. Thank you ❤️

BHydden:
I don't know if it's in scope, but could the ladder games say which ladder they belong to? Many ladder players are in multiple ladders and it's hard to tell them apart.

dexonsmith:
> I don't know if it's in scope, but could the ladder games say which ladder they belong to? Many ladder players are in multiple ladders and it's hard to tell them apart.

I think better to do separately (IMO, out of scope). It might also be better server-side... depending on where the logic for creating/naming ladder games (I haven't looked yet).

IMO, it'd be better to actually change the names of the games to include the ladder name, rather than just prettifying it in game lists (which is what this patch does). I.e., the game-creation logic could be changed such that the name is something like:

    Ladder Challenge: Name of Ladder: challenger(#3) vs challenged(#2)

And then this patch would (continue to) prettify the display of the name in game lists to use the ladder symbol instead of the words "Ladder Challenge:".

Unlike this patch, it wouldn't retroactively change names of old ladder games, but new ones would include the ladder name.

dexonsmith:
> I think better to do separately (IMO, out of scope). It might also be better server-side... depending on where the logic for creating/naming ladder games (I haven't looked yet).

Yeah, confirmed the names are generated server-side. This is what online-go.com sends to the server:
```
post(
    `ladders/${this.props.ladder.props.match.params.ladder_id}/players/challenge`,
    {
        player_id: ladder_player.player.id,
    },
)
```
I'd be happy to write a patch for that, but I don't think I have access to the server code right now.

PR #2429, opened 2023-11-23
Update GoResources.tsx

shinuito:
Jeth's book is available in some new languages and has requested flags be added for these on the resource page.

### DIFF
--- a/src/views/docs/GoResources.tsx
+++ b/src/views/docs/GoResources.tsx
@@ -86,6 +86,8 @@ export const GoResources = () => {
     const gr = "gr";
     const br = "br";
     const _in = "in";
+    const ro = "ro";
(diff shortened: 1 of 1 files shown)


shinuito:
That does make sense - the main site linked has links to all the versions, but there's a lot of them

image

shinuito:
In this case, for Jeth's book

image

the link we currently have points to https://www.gobook.eu/ and if you scroll down the page you see the large list of options mentioned above. 

So it's just one link to the webpage that has all those file links in different languages, hence I think why multiple language/flags were just beside the resource link itself (I think - just guessing).

anoek:
I'm sorry, I misread things a bit and thought this was being applied to the 81 little lions, reading fail on my part. This is great, thanks!

PR #2434, opened 2023-11-28
Use last_emitted_clock in AntiEscaping

dexonsmith:
Since the type being used JGOFClockWithTransmitting, the intent must have been `last_emitted_clock`, not `last_clock`.

N.B., from src/Goban.ts in the goban repo:

    public last_clock?: AdHocClock;
    public last_emitted_clock?: JGOFClockWithTransmitting;

Also drop the `as any` which was probably added to suppress the type error.

### DIFF
--- a/src/views/Game/AntiGrief.tsx
+++ b/src/views/Game/AntiGrief.tsx
@@ -103,13 +103,13 @@ function AntiEscaping(): JSX.Element {
     const user = useUser();
     const goban = useGoban();
     const [phase, setPhase] = React.useState(goban?.engine?.phase);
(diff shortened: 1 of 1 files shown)


dexonsmith:
@anoek, I haven't really tested this yet, just that making moves still works in games against bots. What kind of testing would you like to see? (Or maybe you'd rather test yourself?)

Note that I suspect the core logic in `AntiEscaping` has been disabled because of this bug since I figure this early return was triggering:
```
    if (!clock || clock.pause_state) {
        return null;
    }
```
... although I'm not sure.

anoek:
Seems to work over here, thanks for finding and fixing that! (They both worked.. but this feels better, typing and all)

PR #2435, opened 2023-11-28
Fix strict-weak-ordering in clock sort

dexonsmith:
Fixes the comparator when sorting by clock to follow strict-weak-ordering. Previously, the comparator had a fallback result of `0` (compares equal with everything). Now, there's fallback data for each game.

The effect is:

- Old: might find 99 <= THROWS <= 1.
- New: always finds 1 <= 99 <= THROWS.

If exceptions are ever thrown here, this will fix resulting sorting bugs. (Does NOT fix anything related to pause ordering though...)

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -80,67 +80,49 @@ export class GameList extends React.PureComponent<GameListProps, GameListState> 
         }
     };
 
+    extractClockSortFields(game: GameType) {
+        try {
+            const clock =
+                game.goban && game.goban.last_clock ? game.goban.last_clock : game.json.clock;
+            return [
+                1,
+                clock.current_player === this.props.player.id ? 1 : 0,
+                clock.expiration,
+                game.id,
+            ];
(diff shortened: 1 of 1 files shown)


dexonsmith:
Note: the tuple I'm using (with no named fields!) is ugly. Happy to clean it up in some way, but thought I'd get feedback on the direction first.

(Wanted to fix this latent bug before fixing pause-sorting.)

dexonsmith:
(I'm also curious whether the `try` blocks are actually necessary, or whether a previous refactoring made them redundant... not sure if anyone knows...)

anoek:
Thanks!

As for the try catch blocks, it's possible there's was or still is some racing that happens that might result in some access errors to some of those fields, but I agree they shouldn't generally be needed, especially if we armor the code a bit with `?.` or more explicit `null`/`undefined` checking. Or maybe none of that is needed at all, it might have been something thrown in there in the past to quickly get past an ongoing issue that is no longer needed, I'm not sure.

PR #2440, opened 2023-12-03
Document why pause state is not considered when sorting by Clock

dexonsmith:
GameList doesn't consider the pause state when sorting by Clock. Document why, and explain how it could be fixed in the future. (Also document other limitations of the current sort.)

The short answer: during GameList's initial sort (before rendering), it doesn't have access to the pause state. The Goban (which has a pause state) won't be initialized until each line is rendered. Re-sorts could look at Goban.last_emitted_clock.pause_state. I tried this and it made the UI glitchy.

This commit also replaces the tuple returned by extractClockSortFields() with a dictionary to make it more obvious what's going on. No functionality change intended.


anoek:
Awesome, thanks!

PR #2442, opened 2023-12-09
Sink paused games to the bottom of the game list

### LINKED ISSUE #424: Date sort puts games with opponent on vacation at the very top
### Summary

When you have the game list sorted by date, games where the opponent is on vacation appear to be stuck to the very top no matter how much time is actually remaining. For example, this screenshot:

At least, I'm _assuming_ it is my opponent's vacation flag that is causing this unexpected sort, obviously the game with 6 hours left should be at the top. If anything, games where the opponent is on vacation should be at the very _bottom_ of the list, since they require the least immediate attention.

I've lost games on time because in thumbnail view, a quick glance at the top row or two didn't show any games where I was in danger of timing out, because of lots of opponents on vacation, but there was one lurking right beneath the fold :(

### Environment

 * Operating system: Arch Linux
 * Browser: Chrome
 * Browser Version: 60.0

### Steps To Reproduce:

1. Sort by Clock
2. Have a game where the opponent is on vacation
3. See it at the very top regardless of how many other more urgent games there are.

dexonsmith:
Using `JSON.stringify()`, I discovered that AdHocPauseControl is, in fact, in `GameType[]` from the server. Expose it in the GameType.json interface and use it for sorting, preferring the one in Goban.config when available.

Fixes #424

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -21,7 +21,7 @@ import * as preferences from "preferences";
 import { MiniGoban } from "MiniGoban";
 import { GobanLineSummary } from "GobanLineSummary";
 import { Player } from "Player";
-import { AdHocClock, AdHocPackedMove, Goban } from "goban";
+import { AdHocPauseControl, AdHocClock, AdHocPackedMove, Goban } from "goban";
 
 interface UserType {
     id: number;
@@ -37,6 +37,7 @@ interface GameType {
     json?: {
         clock: AdHocClock;
         moves: AdHocPackedMove[];
+        pause_control: AdHocPauseControl;
         rengo_teams: {
             black: Array<UserType>;
             white: Array<UserType>;
@@ -80,6 +81,13 @@ export class GameList extends React.PureComponent<GameListProps, GameListState> 
         }
     };
 
+    isPaused(pause_control: AdHocPauseControl) {
+        for (const _key in pause_control) {
+            return true;
+        }
+        return false;
+    }
+
     extractClockSortFields(game: GameType) {
         try {
             // In the initial sort, game.goban is null. In later sorts, it's
@@ -99,24 +107,9 @@ export class GameList extends React.PureComponent<GameListProps, GameListState> 
             // directly with the game list.
             const clock =
                 game.goban && game.goban.last_clock ? game.goban.last_clock : game.json.clock;
-
-            // Unfortunately, game.json lacks an AdHocPauseControl. Thus, the
-            // initial sort cannot take into account the pause state.
-            //
-            // In later sorts, we have game.goban, which has the pause state in
-            // game.last_emitted_clock.last_clock.
-            //
-            // However, applying the pause state later makes the list
-            // "glitchy", since sometimes the first re-sort isn't for a while,
-            // and paused games will hang out in the wrong place until that
-            // happens. For now, continue to ignore the pause state.
-            const paused = false;
-            // const paused =
-            //     game.goban &&
-            //     game.goban.last_emitted_clock &&
-            //     game.goban.last_emitted_clock.pause_state
-            //         ? true
-            //         : false;
+            const paused = game.goban
+                ? this.isPaused(game.goban.config?.pause_control)
+                : this.isPaused(game.json?.pause_control);
 
             // AdHocClock.expiration is useful, but insufficient, for sorting.
             //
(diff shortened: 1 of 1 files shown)


dexonsmith:
7ac13061552abd57827802f19bdde54c098a8308 fixed the declaration of `GameType.pause_control` (recognizing that it's optional). The code using it was already correct, since I hit the situation where it was missing during testing, but I forgot to update the declaration.

src/components/GameList/GameList.tsx
```diff
@@ -80,6 +81,13 @@ export class GameList extends React.PureComponent<GameListProps, GameListState>
         }
     };
 
+    isPaused(pause_control: AdHocPauseControl) {
```
anoek, line 84: I might be missing something, and given the compiler isn't complaining there's a good chance I am - but above we have `pause_control?: AdHocPauseControl` indicating it can be `undefined`, which I think is correct. Down below we just pass that field right along to this function, so it seems to me like this needs to also be `pause_control?: AdHocPauseControl` along with a corresponding `undefined` check before we do that `key in pause_control`?

benjaminpjones: > given the compiler isn't complaining

This could just be because strict null checks are off in this repo ([config](https://github.com/online-go/online-go.com/blob/b2448c1fa98213ee2f3e5823e4f80205dd2a1cba/tsconfig.json#L34))

anoek: Ahh, geeze yeah about time to fix that I reckon, I think this is the last repo that doesn't have that enabled. (Obviously not in this patch, just mental note)

dexonsmith: Nice catch. Should be fixed in 6567958b04238d4346323d9ff187da96a55468c8.

dexonsmith: Specifically, here's the error I got (on the other branch):
```
ERROR in ./src/components/GameList/GameList.tsx:99:58
TS1016: A required parameter cannot follow an optional parameter.
     97 |     }
     98 |
  >  99 |     computeRemainingTime(time_control?: JGOFTimeControl, clock: AdHocPlayerClock | number) {
        |                                                          ^^^^^
    100 |         if (!time_control) {
    101 |             return Number.MAX_SAFE_INTEGER;
    102 |         }
```

dexonsmith: I think the previous logic was correct. I suspect any parameter can be `null` any time, and optional chaining with `?.` just accounts for it. I've force pushed back to the previous state (dropping 6567958), but happy to be overridden (and add it back) if you're confident.

dexonsmith: (Regardless, I've tested this in cases where `pause_control` is null, so the code works... if there's a type error that'll be exposed when we turn on strict null checks, we can fix it then...)

anoek: Yep so marks the parameter as optional, and when not provided when called, the default value is `undefined`. All optional parameters need to be at the end of the function, no required parameters after optional so that's what it's complaining about. So the two ways to fix this are to either specify it as `time_control: JGOFTimeControl | undefined` (which differs from `?` in which would essentially expand to `time_control: JGOFTimeControl | undefined = undefined`), or move `clock` to be before the `time_control` parameter.

dexonsmith: That helps, thanks. I hadn't internalized that `undefined` is an independent type from `null` (although both "null-ish"). I've also now read the typescript docs on `?:` and `?.` (previously I inferring from other languages, context, and compiler errors). I'll send a new version of the fix-up.

> it will ever actually be undefined though so might be fine

`game.json?.pause_control` can be undefined; based on dumps, it's missing from `game.json` for games that have never been paused (after unpausing it's defined but equal to `{}`). A bit confused I didn't see this causing a problem... if `for (key in undefined)` raises a TypeError exception, then never-paused games should have been sorted last (that's where exceptions go), but that's not what I saw... based on observation, I assumed the `for (key in pause_control)` was throwing `undefined` to the `default:` case.

Anyway, I'll get this cleaned up and test again with some fresh data.

anoek: You know what, `for (key in undefined)` doesn't throw an error, I was testing the wrong thing (`key in undefined`, the key existence check as opposed to an iterator, does throw an error). So, actually it's totally fine as is, though still good to mark the parameter as being able to accept undefined.

dexonsmith: Aha, that explains it, and good to know. I pushed faafa791e1e17ce2f921d25abb299e531a86aee0 to resolve this.

PR #2444, opened 2023-12-13
Cleanup to enable various stricter typescript features

anoek:
This PR adds necessary extra typing, checking, assertions, and bug fixes to enable various stricter flags in `tsconfig.json`. I'm trying to keep it to just benign refactors but some things need actual fixing before we can pass the tests, so there will be some logic testing and before merge some full regression testing will need to be done.

[X] strictNullChecks
[X] strictBindCallApply
[X] strictFunctionTypes
[X] strictPropertyInitialization
[X] noImplicitAny
[X] noImplicitReturns


anoek:
This branch is up on beta for anyone who wants to play spot the big-red-crash-screen :)

benjaminpjones:
Thanks for the tag @GreenAsJade :) As y'all know, I find incremental changes easier to review than doing everything at once.  That said, ripping the bandaid off has its own benefits - maybe we can tap in dedicated Forumites to help QA?

Anyway, here's one!  You can get it by visiting the homepage without logging in:

Screenshot 2023-12-15 at 4 11 46 PM

I submitted a crash report via the Sentry modal :)

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -1948,10 +1971,11 @@ export function challenge_text_description(challenge) {
         (g.disable_analysis ? ", " + _("analysis disabled") : "");
     if (c.challenger_color !== "automatic") {
         let yourcolor = "";
+
         if (
             data.get("user") &&
-            ((c.challenger && c.challenger.id !== data.get("user").id) ||
-                (c.user && c.user.id !== data.get("user").id))
+            ((c as any)?.challenger.id !== data.get("user").id ||
+                (c as any)?.user.id !== data.get("user").id)
```
benjaminpjones, line 1978: Should include `?` after `user` - this member is not set in in `ChallengesList` (me/challenges). Example challenge:

```
{
		"id": 7065,
		"challenger": {
			"id": 1300,
			...
		},
		"challenged": {
			"id": 1371,
			...
		},
		"game": { ... }
		},
		"group": null,
		"challenger_color": "black",
		"aga_rated": false
	}
```

benjaminpjones:
Clicking the "Create" button on the Play page also does not work for me (no red screen, just doesn't do anything, and the rest of the page becomes unresponsive)

benjaminpjones:
Seeing "Tournament starts Invalid date" on open tournaments

Screenshot 2023-12-15 at 10 46 46 PM

src/views/Game/GameChat.tsx
```diff
@@ -128,8 +128,10 @@ export function GameChat(props: GameChatProperties): JSX.Element {
             debouncedChatUpdate();
         };
 
-        for (const line of goban.chat_log) {
-            onChat(line);
+        for (const log of goban.chat_log) {
+            for (const line of log.lines) {
+                onChat(line);
+            }
```
benjaminpjones, line 134: Not sure this change is valid. I get a red screen when game has chat and I exit Zen mode

src/components/ChallengeModal/ChallengeModal.tsx
anoek: Good catch, thanks!

src/views/Game/GameChat.tsx
anoek: Most definitely wasn't, had some bad typing driving that. Thanks for the catch!

anoek:
Great catches, thanks for rattling the cage!

> I find incremental changes easier to review than doing everything at once. That said, ripping the bandaid off has its own benefits

Yeah sorry, this is a horrible PR. Both `strictNullChecks` and `noImplicitAny` both touched almost every file to get it happy, and I figured I'd run out of steam dicing that task up too much, so with the holidays here I figured this was probably a good time power through all those changes and drop something that would otherwise have a high chance at producing merge conflicts for folks.

benjaminpjones:
Makes sense! Even if there are a few breakages, there's going to be a lot of benefit of turning this all on in the long run :)

src/components/SeekGraph/SeekGraph.tsx
```diff
@@ -298,11 +323,13 @@ export class SeekGraph extends TypedEventEmitter<Events> {
                         ]);
                     }
                 }
-                this.challenges[entry.challenge_id] = entry;
+                entry.rank = entry.ranking;
```
benjaminpjones, line 326: I think this was right before, see: https://forums.online-go.com/t/broken-graph-on-play-page/50060

PR #2450, opened 2023-12-16
gitignore logs in .vscode

GreenAsJade:
Fixes untracked file

## Proposed Changes

  - gitignore logs in .vscode


GreenAsJade:
@anoek  I'm not sure if a PR onto your branch is the right way to go but ... here it is anyhow 😝

anoek:
Yeah that's great, thanks :)

PR #2453, opened 2023-12-19
Switch GameList to sort based on JGOFClock

### LINKED ISSUE #2451: Sorting of paused clocks stymied by corrupt `last_clock` field
**Describe the bug**
#2443 added sorting of not-running clocks (including paused clocks). It works by computing the remaining time on the `AdHocPlayerClock`.

- If `!game.goban` (first sort), it reads the player clock from `game.json.clock`.
- If `game.goban` (later sorts), it reads the player clock from `game.goban.last_clock` (which should get updated).

Since it's the weekend, I had a look at some other players' sorted games and noticed they were out of order. Using the developer console, I found:

- They are sorted correctly according to the `AdHocPlayerClock`s in `game.goban.last_clock` (either `black_time` or `white_time`, depending)
- But, `game.goban.last_clock` was wrong and disagreed with the times in `game.goban.last_emitted_clock` (JGOFClock)
- (Confirmed that `game.goban.last_clock`'s incorrect time is also in `game.json.clock`)

It appears that we can't currently depend on the `AdHocPlayerClock` to be correct. We need to either:
- Fix the `AdHocPlayerClock`s to be correct.
- Change to use `last_emitted_clock`.

The main problem with sorting based on `last_emitted_clock` is that the initial sort doesn't have one available. If we go that route, need to either:
- Ensure the `game.json.last_clock` has the correct value (unlikely, if we can't trust `game.goban.last_clock`)
- Make a `JGOFClock` available for the initial sort (delay the initial sort, or send extra data)
- Live with a glitchy sort (initial sort uses bad ad-hoc data, but subsequent sorts use the good JGOF data)

dexonsmith:
Change GameList over to sorting based on the JGOFClock instead of AdHocClock. Computing remaining time is simpler and easier to get right, especially for paused games.

The downside is that we don't initially have a JGOFClock available.  Here's how that's handled:

- New callback to `GobanLineSummary`, called the first time a `JGOFClock` is available.
- New field `force_render` in `GameList`'s state, which gets bumped by the callback.  When `render()` gets called again, the sort will be reevaluated.

Note: `force_render` is named generically because there might be other (non-clock) reasons to force `GameList.render()` to be called in the future.

Fixes #2451

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -23,9 +23,9 @@ import { GobanLineSummary } from "GobanLineSummary";
 import { Player } from "Player";
 import {
     JGOFTimeControl,
-    AdHocPauseControl,
-    AdHocClock,
-    AdHocPlayerClock,
+    JGOFPauseState,
+    JGOFClock,
+    JGOFPlayerClock,
     AdHocPackedMove,
     Goban,
 } from "goban";
@@ -42,10 +42,7 @@ interface GameType {
     white: UserType;
     goban?: Goban;
     json?: {
-        clock: AdHocClock;
         moves: AdHocPackedMove[];
-        pause_control?: AdHocPauseControl;
-        time_control?: JGOFTimeControl;
         rengo_teams: {
             black: Array<UserType>;
             white: Array<UserType>;
@@ -71,6 +68,7 @@ type DescendingSortOrder = `-${SortOrder}`;
 
 interface GameListState {
     sort_order: SortOrder | DescendingSortOrder;
+    force_render: number;
 }
 
 export class GameList extends React.PureComponent<GameListProps, GameListState> {
@@ -78,6 +76,7 @@ export class GameList extends React.PureComponent<GameListProps, GameListState> 
         super(props);
         this.state = {
             sort_order: "clock",
+            force_render: 0,
         };
     }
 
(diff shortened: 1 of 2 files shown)


src/components/GameList/GameList.tsx
```diff
@@ -98,107 +97,63 @@ export class GameList extends React.PureComponent<GameListProps, GameListState>
...
     computeRemainingTime(
         time_control: JGOFTimeControl | undefined,
-        clock: AdHocClock,
-        player_clock: AdHocPlayerClock | number,
+        clock: JGOFClock,
+        player_clock: JGOFPlayerClock,
     ) {
+        if (clock?.start_mode) {
+            return clock.start_time_left || 0;
+        }
```
dexonsmith, line 105: Currently `start_time_left` is returned here regardless of whose turn it is. Is that reasonable? Or should it only be returned if it applies to the selected `player_clock`?

src/components/GameList/GameList.tsx
```diff
@@ -320,12 +275,21 @@ export class GameList extends React.PureComponent<GameListProps, GameListState>
...
         }
     }
+
+    onFirstClock() {
+        // Now that (more) gobans have their first clock, force render() to be
+        // called again so that the default sort (by clock) works.
+        this.setState({
+            force_render: this.state.force_render + 1,
+        });
+    }
```
dexonsmith, line 292: Or maybe this function should just be called `forceRender()`. (The callbacks would still refer to it as `onFirstClock()`.)

anoek: Either way, it's easy to trace through this way, on this flip side `forceRender`  describes exactly what it does, both are good imho.

anoek (approved):

anoek:
Thanks for the continued effort and iteration on this one!

PR #2454, opened 2023-12-19
Fix byoyomi off-by-one in game-list sorting

dexonsmith:
#2451 had an off-by-one in game-list sorting of byoyomi time control. Note the "SD" (sudden death) entry in this screenshot:
Image

In the screenshot, we're sorting as if there's an extra 24h period. The reason is that `JGOF.periods_left` includes the current, partially used, period, not just the full ones (as the calculation incorrectly assumed).


anoek:
Thanks for the quick fix!

PR #2456, opened 2023-12-19
Unify sorting of MiniGobanList with LineSummaryTable

dexonsmith:
Unify the logic between LineSummaryTable and MiniGobanList to fix sorting of MiniGobanList.

- Set `game.goban` so that `applyCurrentSort` can find the clock.
- Per 85a08f237e905d240b5c42917b47dc81d27983b2, force rendering once we have clocks for the gobans.

Before 85a08f237e905d240b5c42917b47dc81d27983b2, the initial sort for MiniGobanList was correct but wouldn't have updated over time. But that commit started depending on the link to `game.goban`, which `MiniGobanList` never set, and so it broke sorting entirely.

A couple of notes:

- Renamed `onFirstClock()` to `forceRender()` to better descript what it does.
- Moved onFirstClock logic from `LineSummaryTable.initializer()` to new method, `GameList.onGobanCreated()`, which can be reused. - Also added the setting of `game.goban` to the same method.
- Changed MiniGobanList to send in `onGobanCreated`. - Hypothetically, this would override `MiniGobanProps.onGobanCreated`, but users of `GameList` never set that. The only existing users of `onGobanCreated` use are directly using `MiniGoban`.

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -279,17 +279,36 @@ export class GameList extends React.PureComponent<GameListProps, GameListState> 
                     currentSort={this.state.sort_order}
                     player={this.props.player}
                     lineSummaryMode={this.props.lineSummaryMode}
-                    onFirstClock={() => this.onFirstClock()}
+                    onGobanCreated={(game: GameType, goban: Goban) =>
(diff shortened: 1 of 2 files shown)


anoek:
Thanks for the quick fixes!

PR #2457, opened 2023-12-20
Mark games where it's the opponents move

dexonsmith:
When looking at another player's list of games, the current user's games are already highlighted as "green".  Also mark (as "grey") the other games where it's their opponent's move.

This only changes the appearance of OTHER players' game lists:

- No effect on the current user's active games.
- No effect on the list of games in the "Watch" tab.

EDIT: Updated screenshots as of b95112cd3324a795ee6419e6fe3cd90987ff4238.

Here's the line summary look for light:
light-list

dark:
dark-list

And for mini-goban:
light-goban

dark:
dark-goban


dexonsmith:
Updated screenshots in the main PR to match b95112cd3324a795ee6419e6fe3cd90987ff4238.

anoek:
Awesome thanks!

I'm going to flip this around a bit so the games highlighted are the ones where it is the player's move instead of the opponent move, I *think* that'll be more what folks are expecting.

PR #2458, opened 2023-12-20
Bring back AdHocClock to seed sorting of GameList

dexonsmith:
85a08f237e905d240b5c42917b47dc81d27983b2 switched to sorting using JGOFClock, which is more accurate, especially for paused games. This brings back sorting based on the AdHocClock until a JGOFClock is available.

[As reported on the forums](https://forums.online-go.com/t/new-better-sorting-for-active-games/50061/4), waiting until a JGOFClock is available made the mini-gobans (or line summaries) jump around a lot as things settled. If we seed the sort with AdHocClock, then only the parts it gets wrong will have to update.

Note that we can't use `AdHocClock.expiration` because we have nothing to compare it against on the `JGOFClock` side.

### DIFF
--- a/src/components/GameList/GameList.tsx
+++ b/src/components/GameList/GameList.tsx
@@ -26,6 +26,9 @@ import {
     JGOFPauseState,
     JGOFClock,
     JGOFPlayerClock,
+    AdHocClock,
+    AdHocPlayerClock,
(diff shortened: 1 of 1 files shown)


anoek:
Thanks again!

PR #2459, opened 2023-12-21
Sanitize input from contributor fork workflows.

axisofentropy:
This update sanitizes user input from contributor forks. This fixes a GHA workflow security vulnerability.


anoek:
Thanks for the fix!

PR #2460, opened 2023-12-22
Changes the default joseki filter to blank

aesalon12:
The default filter hides a lot of sequences, the intention with the change was to increase engagement with the filter feature but there are too many cases of users being confused about not seeing many variations.

There is probably more code that can be removed regarding forced filter but simply removing this tag does the job.

### DIFF
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -1443,7 +1443,7 @@ class PlayPane extends React.Component<PlayProps, any> {
             this.props.current_filter.source === null) {
             // Set up a Joseki filter by default
             this.props.set_variation_filter({
-                tags:[null,
+                tags: null,
                 contributor: null,
                 source: null
             });
--- a/src/views/Joseki/Joseki.tsx
+++ b/src/views/Joseki/Joseki.tsx
@@ -1443,7 +1443,7 @@ class PlayPane extends React.Component<PlayProps, any> {
             this.props.current_filter.source === null) {
             // Set up a Joseki filter by default
             this.props.set_variation_filter({
-                tags:[this.props.joseki_tag_id],
+                tags:[null,
                 contributor: null,
                 source: null
             });


GreenAsJade:
This needs discussion.

The intent of having the default filter was to 1) engage use of the filter  but also 2) remove from view all the extraneous sequences that are not Joseki.

OJE is primarily about high quality Joseki sequences.    The existence of non-joseki variations is a secondary "nice to have".

If we make the default filter "none" then I'd be inclined to _remove_ a lot of "questionable" variations, those that did not earn themselves a Joseki tag.

aesalon12:
Between the two options, having that tag as a default filter is the larger problem.

A few examples: [The basic slide joseki](https://online-go.com/joseki/16028) recently had that tag removed. This pattern shouldn't be grouped with "nice to have" or questionable patterns.

Non joseki moves like [this hane mistake]( https://online-go.com/joseki/22736), [this incredibly common trickplay/mistake hane]( https://online-go.com/joseki/18651), or [invading an enclosure](https://online-go.com/joseki/23380) should be visible by default.

I am in a lot of Go Discords and tune in for joseki conversations there or on Reddit, these are common questions and the users I interact with in Discord are almost never initially aware that there are a lot of hidden sequences.

Previous discussions and your recent comment of a dynamic help would be a good patch. The right patch would probably be a whole new tag that isn't as exclusive as the Joseki:Done tag but that would require a lot of revision to the tags.

GreenAsJade:
It's great to have the feedback about usage problems!

Curiously, OJE has been very quiet for quite some time, so I'm pretty oblivious to issues like this.  

From time to time I check in, and there are many 10's of thousands of hits on it per day so it is solidly use, but as far as contributions or discussion goes aesalon has been the sole contributor for ages, and basically no discussion takes place in OGS channels.   Even the "known bugs" don't generate anything in the way of comments or feedback ongoing.

If you could post links to comments about OJE from elsewhere that "we" (me?) need to be aware of in one of the OGS OJE chats would be great, to improve awareness.

I'd like to understand how tags that should be in place get removed: can we fix that?

It seems as simple as "if a patterns shouldn't be grouped in 'nice to have' then it should have a Joseki tag".

I could be persuaded I'm wrong, but right now I see the tag system as "the method that people can use to know that they are looking at quality sequences", while also allowing contributors to add less qualified material

Of course, this only works if contributors use the tag system, follow the rules (specifically, entering only qualified sequences) and don't "vandalise" (eg remove tags).

Adding dynamic help to inform people about the filtering that's in place should be an easy step in the right direction: I'll do that ASAP.

aesalon12:
> I'd like to understand how tags that should be in place get removed: can we fix that?
> 
> It seems as simple as "if a patterns shouldn't be grouped in 'nice to have' then it should have a Joseki tag".

There are just contrasting purposes of the tag. There is a reasonable argument that the basic slide pattern might be not seem like joseki but I do not think there is an argument that it's an unnecessary pattern to have in a joseki tool.

Not sure about that suggested use of the joseki tag but it is possible. Just seems a little loaded and misleading if that's how it were used.

GreenAsJade:
I've submitted #2461 for the immediate "make sure users know what's going on".

aesalon and I are discussing further improvements and "refinement of OJE policy" to better meet everyone's needs.

anoek:
Closing in favor of #2461 for the time being

PR #2464, opened 2023-12-25
Use REMOTE_OVERWRITES_LOCAL for 'ignored-reports'

benjaminpjones:
## Motivation

I am finding that I need to ignore the same report across multiple devices.  I feel like the "right thing" is to have the "Ignore" propagate to all devices.

## Proposed Changes

See title.


anoek:
Yep thanks!

PR #2469, opened 2023-12-27
Ordered rengo

GreenAsJade:
Fixes people not being able to control the order of players in rengo

## Proposed Changes

 - Provide a button in `RengoTeamManagementPane` that puts it into 'sort the players by drag and drop' mode
 - Send the whole new team list to the server when they finish doing that
 - Improve the overall UI feedback on server response to rengo team changes (the seek graph update)

** Looks very big because I pulled in dndkit's Multiple Sortable Containers example, so that people can drag players between teams - that brought in a lot of files.

** Looks big also because it is :)


anoek:
Not sure if this is an easy change or not, but while I was first testing it I thought that the order wasn't being updated after saving. Upon being slightly more patient it turns out it was taking somewhere between about 1 and 5 seconds to refresh the list, despite the fact that the `teams` endpoint completes essentially immediately (`< 20ms`). It feels to me like maybe some react state thing isn't getting update when the teams comes back or something, causing the ui to wait until something else on the page triggers a re-render? I'm not sure about that, but that's my hunch.

anoek:
How hard would it be to remove the `li` dots here, and maybe move the handle over to the right side so they align ? Alternatively, make the player name boxes fixed width so the handles align on the right side.

image

anoek:
Would it be a better UX to have the lists always be sortable? So managers can simply just drag and drop people around without clicking the "re-order" people?

GreenAsJade:
> Would it be a better UX to have the lists always be sortable? So managers can simply just drag and drop people around without clicking the "re-order" button?

Yeah, ideally ... but it feels harder than it's worth.

The reason is that if you do that, you have to send an update to the server every time they drag, and that means that you have to feed back into Sortable that an external change happened to its idea of the lists - because the server may have a different idea by then.

Right now you get to tell Sortable what is where at the beginning (basically, on mount), then it takes over and controls what is where after that.

This has a crude solution to the problem "what if someone joins while you're reordering" which is "bad luck, they get dumped, and have to try again if they really should be in your game".

I'm open to taking the trouble to mess with Sortable to make it support this, if you think its too ugly this way ... getting to this point I was just relieved to have it working at all 🤪

---
(I had an afternoon of debug due to this code:

 ```
       if (set_teams):
            with transaction.atomic():
                RengoBlackChallengeTeams.objects.filter(challenge=self).delete()
                RengoWhiteChallengeTeams.objects.filter(challenge=self).delete()
                self.rengo_nominees.all().delete()

                black_team_memberships = [
                    RengoBlackChallengeTeams(challenge=self, player_id=player_id, team_position=index)
                        for index, player_id in enumerate(set_teams["rengo_black_team"])
                ]
                RengoBlackChallengeTeams.objects.bulk_create(black_team_memberships)

                white_team_memberships = [
                RengoWhiteChallengeTeams(challenge=self, player_id=player_id, team_position=index)
                    for index, player_id in enumerate(set_teams["rengo_white_team"])
                ]
                RengoWhiteChallengeTeams.objects.bulk_create(white_team_memberships)

                for id in set_teams["rengo_nominees"]:
                    self.rengo_nominees.add(id)
```

Even ChatGPT couldn't spot the very obvious bug in there, despite it having the warning message the same as I did (which is "constraint violation while updating a player")

🫠

)

GreenAsJade:
> Not sure if this is an easy change or not, but while I was first testing it I thought that the order wasn't being updated after saving. Upon being slightly more patient it turns out it was taking somewhere between about 1 and 5 seconds to refresh the list, despite the fact that the `teams` endpoint completes essentially immediately (`< 20ms`). It feels to me like maybe some react state thing isn't getting update when the teams comes back or something, causing the ui to wait until something else on the page triggers a re-render? I'm not sure about that, but that's my hunch.

I've submitted a fix for this.

The trick is that the RengoTeamManagementPanes are getting their info from SeekGraph, which takes longer to update.

In theory we could "merge" the input from SeekGraph with the input from the return value of the teams endpoint, but that takes you into the territory of "when do I ignore props because I know better at the moment".

This problem has been there since the start: it's the same problem with the pre-existing team management buttons: they have that delay.

The "fix" I submitted is to more clearly show that we're waiting.

It's another one of those things that could be solved with effort, but I haven't yet learned the "pattern" for "here's how you sometimes ignore props and sometimes use them" ...

GreenAsJade:
> How hard would it be to remove the `li` dots here, and maybe move the handle over to the right side so they align ? Alternatively, make the player name boxes fixed width so the handles align on the right side.
> 
> image

Yeah definitely: fixing that layout is on my list.

Also, it's really supposed to be the case that when you drag over an element, it moves out the way to show you where you are going to drop.

That functionality works in the Sortable example that I imported, but got lost on import.   I suspect its due to some funky csss being lost due to difference in scss vs stylus scoping, but it also could be a dumb bug that I introduced while I added the "value payload" to the drag item.  😭

It's another one in the bucket of "man, at least it works, let's try it out, while I gather strength/motivation to do better" ...

GreenAsJade:
@anoek  Latest update is a big step forwards: the styling is picked up properly now, and as predicted, the animation of dragging over target spots works.

It's a teeny bit yucky that you have to tell stylus about every layer of directory that might have stylus files, but it seems to be true.

anoek:
Something weird is still going on with updates: The video below shows two windows, the white one is a player account in chrome, the dark is the creator in firefox. Updates seem to be pretty instant with Chrome, but firefox there are some wicked delays sometimes. 

weird-rengo-update-delays


Side note, the graying out seems to work visually but I can still click everything so we should probably fix that.

GreenAsJade:
> Something weird is still going on with updates: The video below shows two windows, the white one is a player account in chrome, the dark is the creator in firefox. Updates seem to be pretty instant with Chrome, but firefox there are some wicked delays sometimes.
> 

Ah - I have an idea about this.   I bet it's the "freeze the Play pane when the mouse is over it" !!  🤦‍♀️

Egads!   Thanks for the demo!

GreenAsJade:
Yeah la la la I'm an idiot, sorry about that.  The mysterious "seek graph slower than expected update" is solved. And there was much rejoicing.

PR #2472, opened 2023-12-29
[Report Center] Use a thumbnail instead of coordinate list

benjaminpjones:
## Summary

The coordinate lists are hard to grep.  Mods have way to mark the coordinates in the Game view (https://github.com/online-go/online-go.com/pull/1759), but [that hasn't made its way into the reports center](https://github.com/online-go/online-go.com/blob/790c053581845556f04ce88cdae299b4f3d98e62/src/views/ReportsCenter/ReportedGame.tsx#L273).  By adding these thumbnails, (community) mods will be able to view the game as the players see it.

## Why is this a draft?

This PR depends on some updates to what gets logged.  Namely, I would like two additional fields in the logs:

- `move_number`
    - Reasoning: if the game enters scoring phase, and later resumes, we need to jump back in the move tree before displaying the board.
    - Alternative considered: theoretically, one could hackily match timestamps to move numbers, but this is complicated and prone to synchronization issues.
    - How to get it: `engine.cur_move.move_number`
- `current_removed_stones`
    - Reasoning: so that we can show the full board state, not just the group that was just toggled.
    - Alternative considered: We could try to "replay" the score events to calculate the removed stones, but I don't think it is guaranteed that the logs are complete.  If we try to replay the data and any information is missing, the errors will propagate.
    - How to get it: `engine.getStoneRemovalString()`

## Testing

I included a route [/dev/fake-game-log](https://pr-2472-deployment-43641-online-go-com.app.uffizzi.com/dev/fake-game-log) to demonstrate the UI with the additional log fields.

Additionally, I reviewed [R157](https://beta.online-go.com/reports-center/all/157) with these changes to verify the page doesn't crash with the existing log structure.

## Screenshots

before and after screenshots

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -74,6 +74,7 @@ import { AccountWarning } from "AccountWarning";
 import { NetworkStatus } from "NetworkStatus";
 
 import * as docs from "docs";
+import { FakeGameLog } from "./views/FakeGameLog";
 
 /*** Layout our main view and routes ***/
 function Main(props: { children: any }): JSX.Element {
@@ -314,6 +315,7 @@ export const routes = (
                 {/* these aren't meant to be linked anywhere, just entered by hand
                 for developers looking to test and play with things */}
                 <Route path="/dev/styling" element={<Styling />} />
+                <Route path="/dev/fake-game-log" element={<FakeGameLog />} />
                 <Route path="/docs/about" element={<docs.About />} />
                 <Route path="/docs/privacy-policy" element={<docs.PrivacyPolicy />} />
                 <Route path="/docs/terms-of-service" element={<docs.TermsOfService />} />
--- a/src/views/FakeGameLog.tsx
+++ b/src/views/FakeGameLog.tsx
@@ -1,0 +1,533 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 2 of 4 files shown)


GreenAsJade:
> Namely, I would like two additional fields in the logs

I'm on it...

GreenAsJade:
@benjaminpjones  I could conceivably add `move_number` to every event log.   One would think surely there must be a move number for every event in a game :)

However, current_removed_stones is only going to be meaningful in some events.

That being the case, could you nominate which events you expect to see these in.  I wonder if the `force_*` ones are on your radar?

Screenshot 2023-12-29 at 9 30 33 pm (3)

benjaminpjones:
> I could conceivably add move_number to every event log. One would think surely there must be a move number for every event in a game :)

Seems reasonable!

> could you nominate which events you expect to see these in

- Definitely: `stone_removal_stones_set`
- Nice-to-have: `stone_removal_stones_accepted`
- Don't need it: `stone_removal_stones_rejected`, `stone_removal_phase_entered`, `stone_removal_phase_resumed_game`

> I wonder if the force_* ones are on your radar?

Definitely not on my radar, do these trigger on the new "three pass" rule?  Hard to say if we need it, but maybe a hint is whether there is already a "stones" field on these logs.  Currently the code is something like

```
if (data.stones) {
     showBoard(data.stones, data.current_removed_stones)
}
```

GreenAsJade:
> Definitely not on my radar, do these trigger on the new "three pass" rule? Hard to say if we need it, but maybe a hint is whether there is already a "stones" field on these logs.

Actually, they just do:

                if (black_offline && white_offline) {
                    this.logEvent("force_stone_removal_acceptance_abandoned", {
                        notes: "both players are offline, autoscoring",
                    });
                    this.enterStoneRemovalPhase(gamedata, true); // true for force autoscore


... so you'll see them as "enter stone removal" anyhow.

GreenAsJade:
@anoek  

https://github.com/online-go/ogs-node/pull/52
https://github.com/online-go/ogs-node/pull/53

benjaminpjones:
Thanks so much GaJ!  Makes sense to only instantiate when necessary (I was hoping one would be floating around in all the game code.. oh well 😅) I'll go ahead and mark this PR "ready to review"!

anoek:
Awesome, thanks for this! It'll go live Jan 3rd.

PR #2475, opened 2023-12-31
Let community moderators see and edit the mod notes.

GreenAsJade:
Fixes Community Moderator desire to explain why they are escalating

## Proposed Changes

  - Let Community Moderators see and edit Mod Notes

 
Needs https://github.com/online-go/ogs/pull/1866

### DIFF
--- a/src/views/ReportsCenter/ViewReport.tsx
+++ b/src/views/ReportsCenter/ViewReport.tsx
@@ -457,7 +457,7 @@ export function ViewReport({ report_id, reports, onChange }: ViewReportProps): J
                     </div>
                 )}
 
-                {(user.is_moderator || null) && (
+                {(user.is_moderator || user.moderator_powers || null) && (
                     <div className="notes">
                         <h4>Moderator Notes</h4>
                         <textarea value={moderatorNote} onChange={setAndSaveModeratorNote} />


GreenAsJade:
Draft: while this might be nice, I've realised it opens up a communication channel between Community Moderators.   I'm not convinced this will be a Good Thing.

anoek:
What if we add a notes field to the `ModerationActionVote` entry, so everyone would have their own dedicated space per issue kind of thing?

GreenAsJade:
That's a better idea.   I originally thought this approach was "low hanging fruit" for an outcome, as opposed to a new feature we'd need to put in place.

Since it's turned into "questionable anyhow, and not almost free" I reckon we should park it for a while, and see if the purported need is even really there.

benjaminpjones:
> while this might be nice, I've realised it opens up a communication channel between Community Moderators.

Don't they already have a communication channel in the game chat? Sure it's public, but I'm pretty sure I've come across other community mod messages to explain why something was scored in a certain way (and I've done the same in one instance)

GreenAsJade:
They kind of do.    I suspect your experience (and others) is telling us that the urge to communicate is high.

FWIW the doc that I wrote for Community Moderators recommends that they resist this urge, because communication leads to argumentation, and argumentation leads to tears.

I also have this fuzzy feeling that the voting system works better if the first person to vote doesn't tell the others what they're voting.

Totally open to discussion and persuasion on these points though...

PR #2478, opened 2024-01-03
Display any "moderation power offer" on a person's home page

GreenAsJade:
Fixes folk losing their offer if the notification about it disappears

This looks like a big change-set, but it's not actually :)

A bunch of tests changed because of the added field to the config user object.

## Proposed Changes

  - Factor the "offer" component out of Notifications, so we can display it on the Home page as well
  - Display it on the Home page when appropriate
  
** needs https://github.com/online-go/ogs/pull/1872 to actually work (need to receive that field)


anoek:
Thanks!

PR #2482, opened 2024-01-04
Limit the game name length to manage Play pane layout.

GreenAsJade:
Fixes clowns entering silly clownish game names that make the layout go all scrolly (and specifically, put the "rengo management pane close button off the immediately visible screen)

## Proposed Changes

  - Limit game name entry to 50 chars

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -66,6 +66,8 @@ import {
     updateSystem,
 } from "TimeControl/TimeControlUpdates";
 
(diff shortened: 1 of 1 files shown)


anoek:
I don't oppose limiting game names in general, but what about (also?) handling those overflows in css so it's not a problem regardless of device?

GreenAsJade:
Yeah maybe that's the right answer I guess.   When I did this I was exasperated by:

"Help! I'm stranded on a desert island and this is my only means of communication with the outside world. The coordinates of the desert island are 48°52.6′S 123°23.6′W"

... but I suppose a more mature attitude would be to let the person have their fun, and fix the layout responsiveness.

PR #2486, opened 2024-01-06
Community mod note support

GreenAsJade:
Fixes CMs struggling to find a way to explain to moderators why they escalated, and moderators wondering why they did 😝

## Proposed Changes

  - Send a mod-note along with a vote for escalation
  - Show that mod-note to moderators, when it exists

(I also consolidated some css of the ViewReport div)


GreenAsJade:
Pff, committed on top of wrong branch

PR #2488, opened 2024-01-07
Fix whitespace between menu item icon and text for Play

lykahb:
Fixes #

For the Play there is a bigger gap than for the other menu items.

## Proposed Changes
Before
image

After
image


anoek:
Thanks for the catch and fix!

PR #2491, opened 2024-01-10
Fix marks getting set before themes load?

benjaminpjones:
Partially fixes slow board loads.

## Proposed Changes

Set marks *after* the Goban has fully loaded.  Apparently the themes are not initialized when `setMarks` is first called in the (post)constructor.  This raises an error that is logged to the console, and that is CPU intensive.  It's probably a bug in goban that marks cause an error when passed in the config, but this PR is meant to demonstrate the improvement.

### Before

The console.error inside setMarks accounts for most of the compute. On my machine, this accounted for about 16 seconds on 50 boards.

Screenshot 2024-01-09 at 11 39 07 PM

https://profiler.firefox.com/from-browser/flame-graph/?globalTrackOrder=b0wa&hiddenGlobalTracks=1w9&hiddenLocalTracksByPid=88828-1w4~88831-0&implementation=js&thread=g&v=10

### After

setMarks is no longer a major contributor to CPU usage.  On my machine, `setMarks` uses less than 80ms for 50 boards.  However, `load` and `jumpToOfficialMove` are still reasonably heavy (~4 seconds).

Screenshot 2024-01-09 at 11 39 48 PM

https://profiler.firefox.com/from-browser/flame-graph/?globalTrackOrder=c0wb&hiddenGlobalTracks=1wa&hiddenLocalTracksByPid=88828-1w5~88831-0&implementation=js&thread=j&v=10

### DIFF
--- a/src/views/Game/GameLogModal.tsx
+++ b/src/views/Game/GameLogModal.tsx
@@ -157,9 +157,9 @@ export function LogData({
                             key={k}
                             config={{
                                 ...config,
-                                marks,
                                 removed: "",
                             }}
+                            marks={marks}
                             move_number={data.move_number}
                             removal_string={data.current_removal_string}
                         />,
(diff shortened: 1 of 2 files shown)


anoek:
I think I fixed this in the goban repo here: https://github.com/online-go/goban/commit/b9ee6a00b899b91824af3b117607dc8a065abe8c 

Now themes should always be fully usable right away, so closing this in favor of fixing things over there.

PR #2492, opened 2024-01-12
Make AI Review toggle button more visible to users (CSS Idea)

shinuito:
## Add's an example color to the AIReview toggle button when hovering over the AI review

See https://forums.online-go.com/t/ai-enabled-by-default-hurts-teaching/50302 for discussion about AI Review

in particular 

>It clutters up the board when it’s not specifically wanted, and a surprising number of OGS users (not just new users) don’t even know about the right-side menu that pops out on hover, which is where many options are, including the option to turn off AI.

This might make the side menu button more visible


shinuito:
Visually, as an example, it would be not hovering over the Ai review

image

while hovering over the AI review

image

anoek:
I think my concern with this is that most people like the AI stuff, so we're adding a rather prominent indicator to appease a minority view, and in such a way that it calls out an action for everyone to do something they don't wan to do, turn off the AI.

shinuito:
I also would've imagined that the majority actually like AI reviews, and learning to turn off something you don't like seems sensible, over say hiding a feature it's good to know exists.

Still anyway it was fun to practice some css as a demo :)

anoek:
Indeed, looks like it works well! just don't think it's something we want

Issue #2445, opened 2023-12-14
Split new users into Beginner/Intermediate/Advanced streams


dexonsmith:
This feature is to split new users into Beginner/Intermediate/Advanced streams. Background in the [forum proposal thread](https://forums.online-go.com/t/proposal-new-users-choose-beginner-intermediate-advanced-and-drop-ranked-game-restrictions/49983).

- New users choose between Beginner, Intermediate, or Advanced.
    - Beginner corresponds to 25k+–12k. The range won't be shown.
    - Intermediate corresponds to 16k–1k. The range is not immediately visible, but is shown if the user hovers or clicks on a `?` or something.
    - Advanced corresponds to 4k–9d+. The range is displayed on hover, like Intermediate.
- Depending on the choice, the player is assigned an initial rating of 743 (beginner), 1096 (intermediate), or 1764 (advanced).
    - This is instead of 1500, the old starting rating.
    - Deviation still 350.
    - The rating-to-rank conversion puts that at 22k, 12k, and 2k. (Not advertised to users up front.)
- Stop depressing the effective rank when the deviation is high. (I.e., remove code that made 1500+/-350 act as-if 12k.)

GreenAsJade:
Done

anoek:
Not quite? We still have the a/b experiment, which I think we can probably decide upon, but we should keep this open until that's settled I'd reckon?

anoek:
I don't think we ever took out the A/B stuff for this, probably about time to wrap this up :)

PR #2493, opened 2024-01-13
Make specific ratings fall back to the overall one

dexonsmith:
Make specific ratings (like blitz-19x19) fall back to the overall rating in a cascade (bottomming out at "overall").  If the requested rating uses a fallback from the cascade, keep the default (provisional) deviation and volatility.

This change makes it so that once we set an `"overall"` rating for a user during setup, it automatically cascades down to all the sub-ratings. It also has the side effect that if a user plays games on only one board size, their other board size ratings will follow (with high deviation) until they play games there.

Note that even though the rating cascades from `"overall"`, the *displayed* rating/rank will be different (for now). That's because currently OGS decreases the effective rating/rank when the deviation is high. Eventually we want to stop doing that, but not in this PR.

Here are some screenshots:

Old:
Image
Image

New:
Image
Image 2

Relates to #2445

### DIFF
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -182,6 +182,46 @@ test("getUserRating", () => {
     // With a size/speed that does not exist on the user's rating
     expect(getUserRating(user, "blitz", 9)).toEqual(
         expect.objectContaining({
+            rating: 1465,
+            volatility: 0.06,
+            provisional: true,
+            deviation: 350,
+            unset: true,
+        }),
+    );
+
+    // With a size that does not exist on the user's rating
+    expect(getUserRating(user, "correspondence", 9)).toEqual(
+        expect.objectContaining({
+            rating: 1480,
+            volatility: 0.06,
+            provisional: true,
+            deviation: 350,
+            unset: true,
+        }),
+    );
+
+    // With a speed that does not exist on the user's rating
+    expect(getUserRating(user, "live", 0)).toEqual(
+        expect.objectContaining({
+            bounded_rank: 24,
+            bounded_rank_label: "6k",
+            deviation: 350,
+            professional: false,
+            provisional: true,
+            rating: 1465,
+            unset: true,
+            volatility: 0.06,
+        }),
+    );
+
+    // When the rating is completely unknown
+    const unknown = {
+        pro: false,
+        ratings: {},
+    };
+    expect(getUserRating(unknown, "overall", 0)).toEqual(
+        expect.objectContaining({
             bounded_rank: 24,
             bounded_rank_label: "6k",
             deviation: 350,
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -133,6 +133,11 @@ test("getUserRating", () => {
                 deviation: 62,
                 volatility: 0.0625,
             },
+            correspondence: {
+                rating: 1470,
+                deviation: 63,
+                volatility: 0.0625,
+            },
             "correspondence-19x19": {
                 rating: 1480,
                 deviation: 64,
@@ -193,7 +198,7 @@ test("getUserRating", () => {
     // With a size that does not exist on the user's rating
     expect(getUserRating(user, "correspondence", 9)).toEqual(
         expect.objectContaining({
-            rating: 1480,
+            rating: 1470,
             volatility: 0.06,
             provisional: true,
             deviation: 350,
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -209,14 +209,11 @@ test("getUserRating", () => {
     // With a speed that does not exist on the user's rating
     expect(getUserRating(user, "live", 0)).toEqual(
         expect.objectContaining({
-            bounded_rank: 24,
-            bounded_rank_label: "6k",
-            deviation: 350,
-            professional: false,
-            provisional: true,
             rating: 1465,
-            unset: true,
             volatility: 0.06,
+            provisional: true,
+            deviation: 350,
+            unset: true,
         }),
     );
 
--- a/src/lib/rank_utils.ts
+++ b/src/lib/rank_utils.ts
@@ -145,13 +145,23 @@ export function getUserRating(
     const ratings: RatingsType = (user.ratings as any) || {};
     ret.professional = (user as UserType).pro || user.professional || false;
 
-    let key: string = speed;
+    // Set up keys for the rating in a cascade, from most specific to least.
+    //
+    // E.g., for:
+    //
+    // - blitz   and size=19:  ["blitz-19x19", "blitz", "overall"]
+    // - overall and size=9:   ["9x9", "overall"]
+    // - live    and size=0:   ["live", "overall"]
+    const keys: string[] = [speed];
     if (size > 0) {
         if (speed !== "overall") {
-            key += `-${size}x${size}`;
+            keys.unshift(speed + `-${size}x${size}`);
         } else {
-            key = `${size}x${size}`;
+            keys.unshift(`${size}x${size}`);
         }
+    }
+    if (speed !== "overall") {
+        keys.push("overall");
     }
 
     let rating = {
@@ -160,9 +170,21 @@ export function getUserRating(
         volatility: 0.06,
     };
     ret.unset = true;
-    if (key in ratings) {
-        ret.unset = false;
-        rating = ratings[key] as CompactRatingType;
+    let is_fallback_rating = false;
+    for (const key of keys) {
+        if (key in ratings) {
+            const stored_rating = ratings[key] as CompactRatingType;
+            if (is_fallback_rating) {
+                // Propagate only the rating itself from the fallback.  Keep
+                // the default deviation and volatility.
+                rating.rating = stored_rating.rating;
+            } else {
+                ret.unset = false;
+                rating = stored_rating;
+            }
+            break;
+        }
+        is_fallback_rating = true;
     }
 
     ret.rating = rating.rating;


dexonsmith:
I modified the tests in src/util/rank_utils.test.ts. Is there a way to run those locally? (I couldn't figure out how...)

dexonsmith:
> I modified the tests in src/util/rank_utils.test.ts. Is there a way to run those locally? (I couldn't figure out how...)

Tests are passing now... if there's a way to run them locally let me know (less churn on the pull request that way).

anoek:
> if there's a way to run them locally let me know

Yep sure is: `npm run test`

dexonsmith:
> Yep sure is: `npm run test`

Thanks!

anoek:
Hmm, the main problem I see with this is it doesn't match how the rating system works. With this, if someone joins, plays a ton of 9x9 blitz games and gets to 9d in overall, that's going to cascade visually here to all other positions. The moment they go play a 19x19 live game though the rating system takes over and puts in 13k or whatever, now their 19x19 rank suddenly "drops" from 9d to 13k.

dexonsmith:
> Hmm, the main problem I see with this is it doesn't match how the rating system works. With this, if someone joins, plays a ton of 9x9 blitz games and gets to 9d in overall, that's going to cascade visually here to all other positions. The moment they go play a 19x19 live game though the rating system takes over and puts in 13k or whatever, now their 19x19 rank suddenly "drops" from 9d to 13k.

Aha, so I need to land this (or similar) in the backend before the frontend. I'll convert to a draft for now.

anoek:
In terms of fall through, that's probably not something to do in the actual ratings as it would create weird incentives to game the system I'd imagine

dexonsmith:
> In terms of fall through, that's probably not something to do in the actual ratings as it would create weird incentives to game the system I'd imagine

Oh, interesting. I assumed this would be uncontroversial... and since it happens to simplify the situation for #2445, seemed worth doing first.

From my perspective, this just another application of taking a "starting rating" from another source. The best guess for a specific rating you don't have yet is a more general one (albeit with high deviation).

What kind of incentives / gaming-of-the-system are you imagining?

dexonsmith:
> > In terms of fall through, that's probably not something to do in the actual ratings as it would create weird incentives to game the system I'd imagine
> 
> Oh, interesting. I assumed this would be uncontroversial... and since it happens to simplify the situation for #2445, seemed worth doing first.
> 
> From my perspective, this just another application of taking a "starting rating" from another source. The best guess for a specific rating you don't have yet is a more general one (albeit with high deviation).

Moreover, it seems pretty important to me when we're taking a starting rating beginner/intermediate/advanced from the user. Let's say they're wrong and say "beginner" and get a 22k rating. Then they play some 19x19 games and the system corrects them to 15k.  When they start playing 9x9 games, don't we want the same correction to apply?

(To me, it seems like falling back to a more general rating reduces the chances for abuse....)

anoek:
I hear ya, and in a number of ways it can be viewed as positive. Here are my concerns:

1) It's "not fair". Specifically, if you have two people that are about the same rank, one plays a bunch of games with only one setting then it lifts all of their ratings, where as if the other has played even just one game in some other category, now they're effectively hampered and "have to play more" to get to the same place as the first player. This is a weird incentive that seems like we should avoid, we want people to play whatever they want to play whenever they want to play.

2) The act of ranking up is part of the enjoyable and rewarding part of the experience, short cutting that process isn't necessarily desirable. 

3) People have legitimately different strengths for different settings, sometimes by multiple stones. It's probably true that starting them off in their fallback rank is closer to their true strength though, so maybe this isn't really a valid concern.

dexonsmith:
Thanks for explaining!

> I hear ya, and in a number of ways it can be viewed as positive. Here are my concerns:
> 
> 1. It's "not fair". Specifically, if you have two people that are about the same rank, one plays a bunch of games with only one setting then it lifts all of their ratings, where as if the other has played even just one game in some other category, now they're effectively hampered and "have to play more" to get to the same place as the first player. This is a weird incentive that seems like we should avoid, we want people to play whatever they want to play whenever they want to play.

Concretely, say there are three players, A, B, and C, which start at 22kyu, the correct starting rank, and they all play 1000 games and improve to 15kyu.
- A plays an even distribution of live/blitz/correspondance and 9x9/13x13/19x19
- B plays only live 19x19 games.
- C plays 1 game of each category to start, then all the rest are live 19x19 games

For correspondence 19x19 games:
- The current system will have accurate ratings for A and inaccurate ratings for B and C.
- My proposal will have accurate ratings for A and B and inaccurate ratings for C.

IIUC, you're concerned that it's unfair for B to have accurate ratings when C's ratings are inaccurate, and that could disincentivize players from experimenting with ratings categories.

Is that correct?

If so, I agree it's a downside.

IMO, it's better to fix B, which we know how to fix, and then potentially fix C as a follow-up. One option for this would be to have specific ratings age out. I'll follow with another comment that has an example of how we could do this.

> 2. The act of ranking up is part of the enjoyable and rewarding part of the experience, short cutting that process isn't necessarily desirable.

Note this goes in both directions, not just ranking up. The notional starting rank might be 6kyu (joined before beginner/intermediate/advanced split) or 2kyu (clicked "advanced" because they're better than their friends). Then they hit the slow descent to 25kyu because they're actually a beginner on OGS.

IMO, the primary purpose of ratings is to find fun/fair games. If we have a decent guess for a player's rating that helps them find fun/fair games, I think it's important to use it.

I agree that it's also fun to see yourself improve. But you always get to see that in your overall rating. It's not generally fun to retread old ground.

> 3. People have legitimately different strengths for different settings, sometimes by multiple stones. It's probably true that starting them off in their fallback rank is closer to their true strength though, so maybe this isn't really a valid concern.

Yeah, that's the main point. The deviation is high, too, so they'll adjust quickly anyway.

dexonsmith:
> IMO, it's better to fix B, which we know how to fix, and then potentially fix C as a follow-up. One option for this would be to have specific ratings age out. I'll follow with another comment that has an example of how we could do this.

NOT currently proposed as part of #2445, but here's a possible way to improve "player C". We could have stale ratings categories regress toward the mean, by adjusting the deviation and rating periodically.

I haven't thought this through thoroughly, but as a first crack, some pseudo-code:

- For the below, assume `r.parent` navigates up the rating tree (e.g., from blitz-9x9 to blitz to overall).
- A rating `r` has a potentially-stale computed property, true if `r` hasn't had a single game in the last 1000 games that `r.parent` has had, or if `r.parent` is potentially-stale:
```
function potentiallyStale(self) -> bool:
    if self.parent.potentiallyStale():
        return true
    return not (self.lastGame() in self.parent.last1000Games())
```
- A rating `r` has a stale computed property, true if the rating deviations don't cover the difference.
```
function stale(self) -> bool:
    if not self.potentiallyStale():
        return false
    if abs(self.rating - self.parent.rating) > self.deviation + self.parent.deviation:
        return true
    return false
```
- If a rating is stale, we do something to correct it, first increasing deviation, then if that doesn't cover it, either:
    - adjust the rating toward the parent to follow along with a deviation lag
    - record a "delete rating" event so the parent rating is inherited (with default deviation)

Obvious flaws with the above:
- "1000 games" might be too arbitrary; maybe there's some combination of "# of games" and "# of days" that should determine if a specific rating is potentially stale.
    - (Might also be interesting to increase deviation after a "# of days" metric alone? But that's a tangent.)
- Details not quite right for the tiered rating categories (i.e. if blitz-9x9 and blitz are both stale, what order do you update them?).

But there are also some nice properties in a system like this:
- Any category a player is actively using has accurate ratings that don't get modified.
- When something is stale, adjusting deviation first ensures that it'll become accurate again after a small number of games
- Player C gets fun/fair games, just like player B

anoek:
I'm warming up to the idea. It's probably going to need some good time stewing before committing to the final thing, and we'll want to model this in `goratings` too to make sure it's what we want to do to see if we need to do any other tuning to the system while we're at it. FYI, doing this is not as simple as switching it here, we'll need to retroactively apply this change so it's fair to everyone, but we'll need to do that anyways with the small board handicap problems so might as well take our time and figure out what's going to be the best solution and do that together (in the past re-crunching those stats has taken about two weeks to do on the production system, so we definitely want to get all the changes we want together and get them right)

Re the follow up idea, counting games is pretty expensive, but time since you last played makes a good bit of sense, glicko2 encourages it too. There are some experiments with similar things in the `goratings` repo, though I think I was more experimenting with grouping together blocks of games during a time period and expanding the deviation depending on how many time periods had passed kind of thing, which is similar but might not be directly copyable.

dexonsmith:
Per @anoek's suggestion, I've put the question to the forums: https://forums.online-go.com/t/should-never-used-ratings-categories-follow-the-overall-rating/50339

dexonsmith:
> I'm warming up to the idea. It's probably going to need some good time stewing before committing to the final thing, and we'll want to model this in `goratings` too to make sure it's what we want to do to see if we need to do any other tuning to the system while we're at it. FYI, doing this is not as simple as switching it here, we'll need to retroactively apply this change so it's fair to everyone, but we'll need to do that anyways with the small board handicap problems so might as well take our time and figure out what's going to be the best solution and do that together (in the past re-crunching those stats has taken about two weeks to do on the production system, so we definitely want to get all the changes we want together and get them right)

I hadn't considered that you'd want to apply it retroactively immediately.

WDYT of a compromise? As in, incrementally do a few things non-retroactively (unfairly), and then once we've got all the changes in that we think we'll make, go back and retroactively apply everything (make it fair again). To me, this staged approach feels less risky, in terms of getting to see how some of these changes shake out in practice. But maybe fairness trumps that?

> Re the follow up idea, counting games is pretty expensive, but time since you last played makes a good bit of sense, glicko2 encourages it too. There are some experiments with similar things in the `goratings` repo, though I think I was more experimenting with grouping together blocks of games during a time period and expanding the deviation depending on how many time periods had passed kind of thing, which is similar but might not be directly copyable.

Interesting that counting games is expensive; I wonder if there's a way to estimate quickly by keeping running totals / ids / something, but maybe it's not worth thinking about if you already explored it.

I do think we can do something reasonable just with time.

anoek:
> WDYT of a compromise? As in, incrementally do a few things non-retroactively (unfairly), and then once we've got all the changes in that we think we'll make, go back and retroactively apply everything (make it fair again). To me, this staged approach feels less risky, in terms of getting to see how some of these changes shake out in practice. But maybe fairness trumps that?

I think there can be a period of time where this is active prior to or while we are performing the retroactive update, however we should be performing the experiment in the goratings repo not kicking it into to production to see how it shakes out.

The ratings and ranking system is pretty important to a lot of players and they take it very seriously, whenever we do a change it needs to be for a demonstrably good reason and the change, along with numbers to back up why it's a good thing, communicated to the community in pretty good detail so they know what's coming and why.

So in that regard, yes we can phase it in, but no it's not going to be a fast merge, I'd like to get all of our ducks in a row, gets the numbers crunched in `goratings`, then we can work out the plan for what needs to happen when, I'll put together a detailed forum post about the upcoming ratings changes, and then we can roll it out accordingly.

dexonsmith:
I'm abandoning this PR... I think any fallthrough/fall-back we do is likely to take a slightly different shape...

PR #2494, opened 2024-01-13
Neutralize humble_rating

dexonsmith:
Update `humble_rating` to return the `rating` parameter directly, rather than modifying it using the deviation, effectively neutralizing it.

After this lands, we should follow up by deleting the function entirely, but this commit as-written is much less likely to develop conflicts while we're waiting to land it and/or if we have to revert it.

Relates to #2445.

(Starting with a draft pull request, since we're not ready to land this yet...)

### DIFF
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -337,7 +337,7 @@ test("allRanks", () => {
 });
 
 test("humble_rating", () => {
-    expect(humble_rating(1500, 350)).toBe(1150);
+    expect(humble_rating(1500, 350)).toBe(1500);
 });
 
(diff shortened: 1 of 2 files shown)


GreenAsJade (commented):
I think this PR is not sufficient.  In particular, humble rank is used by node in matchmaking...  if that isn't changed to match, then people will still be matched humble, but they will wonder why because the front end is showing their rank non-humble.

I guess we do the same in node and live happily ever after :D

dexonsmith:
Closing since @GreenAsJade has added an equivalent commit now in #2568.

> I guess we do the same in node and live happily ever after :D

Indeed :).

Issue #2446, opened 2023-12-14
Allow all players to play ranked games with any opponent


dexonsmith:
This feature is to allow all players to play ranked games with any opponent. Background in the [forum proposal thread](https://forums.online-go.com/t/proposal-new-users-choose-beginner-intermediate-advanced-and-drop-ranked-game-restrictions/49983).

- Remove restriction that prevents players from playing ranked games with opponents with a rank more than 9 away.
- Keep restriction that prevents playing ranked games with an effective handicap bigger than 9.
    - For small boards, this might require a new explicit check, since it's not enough to check the number of stones.
    - For ranked games with `handicap=auto`, the rank difference can be capped to 9.

Related to #2445.

PR #2495, opened 2024-01-13
Allow any player to have a ranked game with any opponent

dexonsmith:
This removes restrictions from the SeekGraph related to ranked games.

Relates to #2446, but doesn't complete it, as there are also backend changes necessary. Draft for now, since I suspect those should land first.


dexonsmith:
Backend PR is https://github.com/online-go/ogs/pull/1890.

dexonsmith:
Just realized I never sent this for review. We already landed the backend PR, which handled normal challenges. This one just makes a couple of changes to the seek graph.

anoek:
Thanks!

PR #2498, opened 2024-01-14
Allow users to supply a URL for custom stone images

### LINKED ISSUE #2481: Stone Images from URL
I want users to be able to set their stone image, separately for each color, using an img url. Basically the same as what currently exist for the Go board.

CSS inject lol

**Additional context**
On my project, I am letting users unlock stone images and goban images. I have a way for them to copy the url at the click of a button. I would like them to be able to use the graphics they unlock on OGS with a simple copy paste.

image

image

For discussions - https://forums.online-go.com/t/feature-request-stone-images/50210

GreenAsJade:
Fixes #2481

## Proposed Changes

 - Add the ability to supply a URL when the Goban stone theme "Custom" is selected.

Needs: https://github.com/online-go/goban/pull/145  , which adds this theme for stones.

Screenshot 2024-01-14 at 9 36 13 pm (3)

### DIFF
--- a/src/components/GobanThemePicker/GobanThemePicker.styl
+++ b/src/components/GobanThemePicker/GobanThemePicker.styl
@@ -23,7 +23,12 @@
     max-width: 100%;
     justify-content: center;
 
-    .theme-set {
+    .custom-theme-set {
+        justify-content: flex-start !important;
+        padding-left: 2rem;
+    }
+
+    .theme-set, .custom-theme-set {
         display: inline-flex;
         flex-direction: row;
         flex-wrap: wrap;
@@ -51,17 +56,48 @@
             border: 0;
         }
 
+        .board-color-selectors{
+            display: flex;
+            flex-direction: row;
+            justify-content: flex-start;
+        }
         .color-reset {
             vertical-align: top;
             margin-top: 0.4rem;
+            margin-left: 0;
+            margin-right: 0;
+            padding-right: 0;
+            padding-left: 0;
+            .fa-undo {
+                margin-left: 0 !important;
+                margin-right: 0 !important;
+                padding-right: 0;
+                padding-left: 0;
+            }
         }
 
-        .customUrlSelector {
-            width: 212px;
-            height: 24px;
-            border: 1px solid lightgrey;
-            margin-top: 0px;
-            margin-bottom: 10px;
+        .custom-url-selection {
+            display: flex;
+            flex-direction: column;
+            justify-content: center;
+            padding-left: 0.4rem;
+
+            .customUrlSelector {
+                width: 212px;
+                height: 24px;
+                border: 1px solid lightgrey;
+            }
+
+        }
+    }
+
+    .show-customize-selector {
+        flex-grow: 1;
+        display: flex;
+        justify-content: center;
+        margin: 0.5rem;
+        .show-customize-toggle {
+            margin-left: 0.3rem;
         }
     }
 }
--- a/src/components/GobanThemePicker/GobanThemePicker.tsx
+++ b/src/components/GobanThemePicker/GobanThemePicker.tsx
@@ -23,6 +23,7 @@ import * as preferences from "preferences";
 import { PersistentElement } from "PersistentElement";
 import * as data from "data";
 import { CustomGobanThemeSchema } from "data_schema";
+import { Toggle } from "Toggle";
 
 interface GobanThemePickerProperties {
     size?: number;
@@ -40,6 +41,7 @@ interface GobanThemePickerState {
     urlCustom: string;
     black_stone_url: string;
     white_stone_url: string;
+    show_customize: boolean;
 }
 export class GobanThemePicker extends React.PureComponent<
     GobanThemePickerProperties,
@@ -65,6 +67,7 @@ export class GobanThemePicker extends React.PureComponent<
             urlCustom: this.getCustom("url"),
             black_stone_url: this.getCustom("black_stone_url"),
             white_stone_url: this.getCustom("white_stone_url"),
+            show_customize: false,
         };
 
         for (const k in GoThemesSorted) {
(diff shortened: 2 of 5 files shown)


GreenAsJade:
It'd be nice if we had a URL to a suitable image as the default image for the "Custom stone" in the GobanThemePicker.

It kinda needs to be a  URL, not a direct link to an asset, so that it doesn't require special casing.    I see that we have CDN URLs for firefox reasons anyhow, so this could go in there.

It's not "awful" without it.

anoek (commented):
Neat, here's the anime one for testing, though I feel like if we're going to show this as an always present option we can't use this, we'll need to find another distinct image to use. 

https://cdn.online-go.com/goban/anime_black.svg
https://cdn.online-go.com/goban/anime_white.svg
https://cdn.online-go.com/goban/anime_shadow.svg
https://cdn.online-go.com/goban/anime_board.svg

That said, I think we need to work on some different idea's for how to present this to the user, right now I don't like introducing a second row with one option in it, it throws the look and feel off. I'm not really sure I'm a fan of making the whole thing wider either, but that is a possibility - needs testing on small devices.

I'm also not a fan of having two places to customize your stones, the first being Plain where you can select the color, and the second being this one where you can provide your URL. That said, it feels weird to put this under Plain, especially since we default users to Plain sometimes, so them opening up the menu and seeing all of this might be a bit much. (The color selector is already a bit much when that happens IMHO, but that's not really related to this PR.)

I find myself wondering if the thing to do wouldn't be to have a "customize" toggle at the bottom of this menu, clicking it would expand all the options and possibly add the second row for this if we wanted to keep it hidden until the customize toggle is clicked. Basically i they click a thing, I don't mind it being a little chaotic, but I want the default experience to be elegant.

Or maybe we just have a dedicated customize section in the actual settings page that gives us more room to work. 

Here are some mockups of what came to mind with the customize toggle, we don't have to go this route, but whatever we do needs to have an elegant first impression experience.

image

image

What's not covered in the mockups is reconciling the color vs url configuration for the stones, I think that could be notably improved, but I leave that for future discussions and ideas depending on what direction we end up wanting to take.

GreenAsJade:
I knew that this would be "a thing" I nearly marked it as draft myself, due to the ugly extra row :D

I like your mockup, I think it is the right way to go.

The only unsolved problem I think it has is that it is not clear that you can change the colour of "Plain".  I had no idea about that till I worked on this ticket :)

I wonder if the answer is as simple as putting a rainbow where you have grey _without_ the word "custom", on the Plain stone.

anoek:
> The only unsolved problem I think it has is that it is not clear that you can change the colour of "Plain". I had no idea about that till I worked on this ticket :) 

Yeah the placement of that has always been bothersome, I should have insisted on something better at the time. Would be nice to tidy that experience up in one way or another.

I'm kind of wondering if the thing to do (maybe this is going beyond what you wanted to get into) would be to leave the plain stones and board as their default values, when the custom toggle is active then show the custom row with the custom icons on the left like you have them. Make the color selector smaller and have it as two lines, one being the color selector, the other being the URL, so basically have the custom board/stone class do both the plain custom colors, or the url based stuff? 

Side detail, I'm thinking if they have selected a custom anything, the the custom toggle should be forced active until they select all non custom stuff?

GreenAsJade:
> Another option is to have a kind of expandable menu item, where you click the word "Board" and it just shows you the board options including the custom url, and another for "Stones" which can show you all the stones and the custom url spots etc.
> 
> Might free up some space for more descriptors or whatever it is you want to include.

Yeah - that's cool.  As I started playing with that, I realised that it's nice to just be able to mouse over our username and see our options - including the board and stone basic options - and click them right there to try them.    I'm loathe to put another click in the way of that experimentation.

GreenAsJade:
@anoek  How do you feel about getting rid of the second row of boards in the customise-off view?

296768102-9d9f1e1e-b093-4f9d-8c80-7a8ae3f5c160

anoek:
I think that looks really nice 🙂

GreenAsJade:
Screenshot 2024-01-20 at 9 57 21 pm (2)


Screenshot 2024-01-20 at 9 58 52 pm (2) copy

Screenshot 2024-01-20 at 9 58 52 pm (2)

anoek:
This is shaping up quite nicely!

Would it be possible to keep the Plain options visible at all time (without the color customization stuff)? Somewhat important because it's the fallback setting that gets set when there are canvas rendering issues. When the customize option is clicked we could either move the option down there or just have it duplicated down there.

That said I really like the square aesthetics, so I find myself wondering if we put those back in, maybe the thing to do would be to also add another board color scheme to square things out again so have 6x2 stone options and 12 board options so it makes a nice rectangle again.

GreenAsJade:
As discussed in the goban PR, the "custom board" controls are slightly quirky in this implementation - in that you can select either the rainbow or the board and you get the effect of both.     I think the effect is what we want, because you can set the background board colour and the image on top of it.

It is like this because that made it easiest to transition the layout to this shape.

If we like the design, I'm open to yet more customisation of the code in that little area to make it work better.

FWIW right now each section of the GobanThemePicker is "the same" as the others - a map iterating through the themes in that section creating theme-selection elements that instantiate one or more pickers... even when in fact some of those sections are "unique".

That was the easiest way to start (cut, paste, tweak :) ) and makes the input kind of regular.

GreenAsJade:
Yeah, that's better 😅

GreenAsJade:
> Would it be possible to keep the Plain options visible at all time (without the color customization stuff)? Somewhat important because it's the fallback setting that gets set when there are canvas rendering issues. When the customize option is clicked we could either move the option down there or just have it duplicated down there.

Is it too awful for the user to see the customize section opened in this case? (which is what would happen)

They might say "hey, why can't I close this?" (which would be true) and the answer would be "because your device is struggling" ...

GreenAsJade:
> It's not the end of the world, I think it'd be better if that weren't the case but if it's too much of a PITA then whatever.

Actually, let's explore a bit more what you meant.   Do you mean that simply the Plain selectors, black and white, with their colour selector, are there under the rows of other selectors?     And customise-on just adds the remaining custom selectors? 

That's straightforward.   

I was more balking at changing the rows of normal selectors to accomodate "Plain"

---

Yeah, oops, I need to recover the lines and A, and I definitely would like a default image for the stone-image selector.  The question is: what :)

GreenAsJade:
Riiiight ... soooo ... what are the two extra boards that you've added in that mockup? 😝

How do you feel about:

Plain

anoek:
> Riiiight ... soooo ... what are the two extra boards that you've added in that mockup? 😝

Well one would be "Plain", the redish one could pretty easily be created using the same pattern we do for several of the others: https://github.com/online-go/goban/blob/main/src/themes/board_plain.ts#L208-L239 

I'm not a fan of the customization options visible without the customization toggle selected.

I feel like the duplication of "Plain" is mainly an artifact of us hijacking "Plain" to do custom stone/board stuff, maybe we just create a "CustomizedPlain" stone or something?

GreenAsJade:
> I feel like the duplication of "Plain" is mainly an artifact of us hijacking "Plain" to do custom stone/board stuff, maybe we just create a "CustomizedPlain" stone or something?

🫡

GreenAsJade:
^^^ This is in now.  I created two new boards, which are different from the others in a fairly arbitrary way.

I created a new disc stone subclass, to be the "colour customisable one", so that the disc stone themes supplied with "Plain" truly are plain.

I didn't  find a solution for "default image for URL stone" yet.    Could come later :)    "Nothing to see here because you haven't supplied a URL" is vaguely tolerable?

anoek:
> I didn't find a solution for "default image for URL stone" yet. Could come later :) "Nothing to see here because you haven't supplied a URL" is vaguely tolerable? 

I went ahead and merged the color and url selection things, so now the default is the plain color, and then when the url is supplied the url is used, so it sidesteps that issue

Issue #2501, opened 2024-01-17
Allow custom komi to be specified on demo boards


dexonsmith:
See forum post: https://forums.online-go.com/t/request-custom-komi-for-demo-board/39065

> Komi cannot be chosen when making a demo board. The default komi of the chosen ruleset is used and cannot be edited later. But what if, for instance, I wanted to create a review of an old game that used 0 komi?

PR #2503, opened 2024-01-17
Save settings for demo boards

dexonsmith:
This PR:
- Makes the "Create Demo" modal remember/reuse whatever settings you had the last time. Before this patch, settings were not saved for demos.
- EDIT: Makes the demo and game challenge settings independent, so that they don't clobber each other when settings are saved.
    - Previously, some were independent, while some were not.
    - Most game state accessors/modifiers now use a helper to consistently select between `challenge.game` and `demo`.
    - Rendering code can still be shared (in fact, it's easier to do correctly now).
    - See the commit message of 5635a19 for more details.

~~I'm not sure if this should be merged as-is— because of asymmetric reuse of fields saved/restored from `data`, the demo board and custom challenge pane:~~
- ~~share the board-size setting~~
- ~~do NOT share the rules setting~~

~~My inclination is to make the demo board settings completely independent before saving settings. So, draft for now.~~

Relates to #2501

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -482,6 +482,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
         }
 
         console.log("Sending", demo);
+        this.saveSettings();
         this.close();
         post("demos", demo)
             .then((res) => {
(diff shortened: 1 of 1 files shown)


dexonsmith:
I updated the description to reflect 5635a19f52d4b3745bdc54f21f2567f764a5be95. This is now ready for review!

anoek:
Nice, thanks!

PR #2505, opened 2024-01-17
Add translation and fix "no info about `reported_at`" logic

GreenAsJade:
Fixes case where report was made on turn `0`, and lack of translation of game presentation.

## Proposed Changes
 - Fix logic with null coalesce
 - Translate stuff.


anoek:
Thanks!

Issue #1469, opened 2021-05-02
"score breakdown" incorrectly includes handicap points under Japanese rules


pdg137:
The "score breakdown" popups activated by clicking in the lower area of the play boxes seem to know most of the details of scoring.  However, they incorrectly assign compensation points to White for handicap stones under Japanese rules:

Example game: https://online-go.com/game/33079596

![image](https://user-images.githubusercontent.com/466760/116820962-7c76c000-ab2c-11eb-9188-182f44a7ba72.png)

Adding up the points for white, which usually works, you would get 89.5, but actually the score is 83.5.

Points for handicap stones should only be added when using area counting (except not with NZ rules?), [as described here](https://online-go.com/docs/go-rules-comparison-matrix).

pdg137:
Still an issue.

dexonsmith:
Note that the current state isn't even consistently showing the number of handicap stones:

- For Chinese/Ing, it shows both the handicap scoring bonus and the handicap, which happen to be equal.
- For AGA, it shows the handicap scoring bonus (which is `handicap-1`).
- For Japanese/Korean/New Zealand, it shows the handicap (handicap scoring bonus is 0).

PR https://github.com/online-go/goban/pull/146 should fix this (indirectly) by making the `goban.computeScoring()` API consistently return the handicap scoring bonus in the `handicap` field.

PR #2507, opened 2024-01-19
Show handicap separately from handicap bonus

dexonsmith:
Adapt to Goban change in https://github.com/online-go/goban/pull/146 and show the actual handicap stones for black separately from the handicap bonus.

To make it clear that this handicap field is not part of the score:

- Use a `<hr/>` to separate it
- Add `+` in front of scores that are supposed to sum together.

Also move white's handicap bonus to the end, after komi.
- This means that there's always a `+` in front of it, even at the beginning of the game, so it's implicitly clear that there's a bonus.
- I considered changing the text to `Handicap Bonus` but that wrapped.

Relates to #1469

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -479,36 +479,59 @@ function ScorePopup({ show, goban, color }: ScorePopupProps) {
     const scores = goban.engine.computeScore(false);
     const { stones, prisoners, handicap, komi, territory } = scores[color];
 
+    let first_points = 0;
     return (
         <div className="score_breakdown">
+            {color === "black" && !!goban.engine.handicap && (
+                <div>
+                    <span>{_("Handicap")}</span>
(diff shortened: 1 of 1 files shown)


dexonsmith:
Some examples of how this looks (in combination with https://github.com/online-go/goban/pull/146):

AGA:
aga
aga-after-moves

Japanese:
jp

NZ:
nz

anoek:
Nice improvement, thanks!

pdg137:
Yay, so nice to see things becoming consistent and easier to use across all rulesets.

In case it's easy to do, I'd like to also draw attention to a suggestion I made about displaying handicap along with the ruleset in the main game view:

[image
link](https://forums.online-go.com/t/the-ruleset-of-the-game-should-be-easier-to-see/39165/12?u=feijoa)

Since just like the captures, that's basic information about the game that you might need to use to analyze the position, so it should be available without having to click anything.

PR #2508, opened 2024-01-21
Show the handicap rank difference in Game Information

dexonsmith:
In the "Game Information" modal, show the handicap rank difference if it's different from the handicap.

Depends on https://github.com/online-go/goban/pull/147, https://github.com/online-go/ogs-node/pull/58, and https://github.com/online-go/ogs/pull/1903.

### DIFF
--- a/src/views/Game/GameDock.tsx
+++ b/src/views/Game/GameDock.tsx
@@ -147,7 +147,14 @@ export function GameDock({
     };
 
     const showGameInfo = () => {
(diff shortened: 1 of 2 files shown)


dexonsmith:
Here's are some screenshots from the "Game Information" panel to show what it looks like, when it shows up:

A 9x9 game:
Image 4

A 13x13 game:
Image 3

Note that on 19x19 boards it's always hidden, since the rank difference is always equal to the number of handicap stones:
Image 2

anoek:
Thanks!

PR #2510, opened 2024-01-22
[ReportCenter] truncate the game log

benjaminpjones:
Conrad mentioned there are more frozen reports in the reports center.  And indeed here is an example: [R814](https://online-go.com/reports-center/all/134814).  The event log is a whopping 5000 entries!  IIUC, rendering is still the bottleneck, so this is a quick and dirty fix to truncate the rendered list until the user opts in to showing the full thing.

## Proposed Changes

  - Truncate event log to 25 events before rendering
  - Add "Show all" button

## Screenshot

Screenshot 2024-01-21 at 4 20 34 PM

## Alternative considered

In a perfect world, we would do proper paging, but I think that is overkill for the mod log - returns diminish greatly after the first few entries.  That said, I'm guessing 5000 is also an artificial limit set on the backend.  It might be reasonable to lower that cap as well since we don't expect mods to look at thousands of entries.

### DIFF
--- a/src/views/ReportsCenter/ReportedGame.tsx
+++ b/src/views/ReportsCenter/ReportedGame.tsx
@@ -282,44 +282,58 @@ function GameLog({ goban }: { goban: Goban }): JSX.Element {
         [goban],
     );
 
+    const [shouldDisplayFullLog, setShouldDisplayFullLog] = React.useState(false);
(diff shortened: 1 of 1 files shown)


anoek:
Thanks!

PR #2514, opened 2024-01-25
Initialize custom komi to the default for the rules

dexonsmith:
If komi switches from "automatic" to "custom", initialize the custom komi to the default for the current ruleset.  This is a reasonable starting value for modification and still a predictable behaviour for users.


dexonsmith:
Draft for now.

@anoek, I think komi=0 is the most natural default komi for handicap games, even on small boards (where it won't match `komi=auto`). But I guess it's debatable.

I don't want to duplicate the small board handicap table logic here... I think either we set to even/no komi (like this patch) or we just pick up whatever happens to be saved from last time.

WDYT?

dexonsmith:
Oh, this came up when I was implementing #2501.

anoek:
That seems reasonable

dexonsmith:
Great. I'm going to wait for #2501 to land (since this has a merge conflict with it) before sending this for review.

dexonsmith:
Landed, and merged in c4df3a1. Re-did manual tests (checked all the rules for Demo, checked one set of rules with and without handicap in Direct challenge).

Ready for review now.

PR #2516, opened 2024-01-26
Fix display of 0 handicap rank difference in GameInfoModal

dexonsmith:
Without this, the handicap for new even games was being shown as `None0` instead of `None`.

Caused by 50d60910b6b32e145e7c01ccdfe983eb78809c08. Could have sworn I fixed this bug before push, but I guess it slipped through (or maybe it was a similar bug, another location...).


dexonsmith:
@anoek fixed this in 62ad6a714d01acf5e3f7f47970edc6938d1703b6.

PR #2519, opened 2024-01-26
Increase the automatch rankdiff maximum to 40

### LINKED ISSUE #2517: Allow free choice of rank restrictions in automatch settings
Follow up to #2511, same thing for the ± in automatch settings here:

Image 3

Should have free choice, but currently limited to ±9.

dexonsmith:
Limit automatch rank difference to 40, instead of 9.  This should be enough to play any opponent.

Fixes #2517

### DIFF
--- a/src/components/AutomatchSettings/AutomatchSettings.tsx
+++ b/src/components/AutomatchSettings/AutomatchSettings.tsx
@@ -165,13 +165,13 @@ export class AutomatchSettings extends Modal<
     }
     setLowerRankDiff = (ev: React.ChangeEvent<HTMLInputElement>) => {
         const settings = this.getSelectedSettings();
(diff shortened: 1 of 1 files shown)


dexonsmith:
Image 4

anoek:
I think this might need a corresponding change on the server side to be functional, I'm not sure if this is the only place but here's what caught my eye: https://github.com/online-go/ogs-node/blob/devel/src/comm-server/Automatch.ts#L265-L272

dexonsmith:
> I think this might need a corresponding change on the server side to be functional, I'm not sure if this is the only place but here's what caught my eye: https://github.com/online-go/ogs-node/blob/devel/src/comm-server/Automatch.ts#L265-L272

Looks like you're right. I'll look into it. Draft until I sort it out.

PR #2524, opened 2024-01-28
Make stone theme selectors show the active board styles

GreenAsJade:
So that the users get to see what the stone looks like on the background they've chosen.

It also is intended to help be clear what background is in fact selected - this can be a little unobvious especially when it's the custom background.

## Proposed Changes

  - Render the active background styles into the stone theme selectors.
  
Screenshot 2024-01-28 at 11 19 06 am (2)

Screenshot 2024-01-28 at 11 22 39 am (2)


anoek:
Very nice 🙂

PR #2528, opened 2024-01-28
Only show custom komi in challenge descriptions

### LINKED ISSUE #2527: Komi value of 5.5 displayed in preferred settings (and perhaps elsewhere) when komi is automatic
E.g.:

eaf8ccf3ace7fa6f1c594c215edbeb630fb0ede8

The komi should probably be skipped from the description. Or it could be "automatic". But it says "5.5".

This is just a display bug; when the game is created, it has the correct komi.

dexonsmith:
The `komi` field is inaccurate (and ignored) when `komi_auto !== "custom"`.  It might be nice to show the automatic komi, but right now that would be hard to do so consistently, since for small boards it depends on handicap tables that are server side.

For now, just stop showing komi at all unless `komi_auto === "custom"`.

Fixes #2527.


dexonsmith:
Below are some preferred settings screenshots. Only the final one specifies a custom komi.

Before:

Image

After:

Image 2

PR #2533, opened 2024-01-31
Use "opacity" instead of "transparency" to refer to the opacity setting for variation stones

GreenAsJade:
Fixes wrong terminology, confusing number in "variation stone opacity" setting

## Proposed Changes

  - say opacity instead of transparency

Full disclosure: this will cause people who changed this setting to have to change it again, they will lose their old value.

I wouldn't have worried about this, except I'm planning to add "Last Move Opacity"... seems like it should be consistent.


GreenAsJade:
Pff - this extends to Goban too I see.   I'm persisting 😝

GreenAsJade:
Needs https://github.com/online-go/goban/pull/149

PR #2534, opened 2024-01-31
Provide a preference to control the "last move opacity".

GreenAsJade:
Fixes folk having a big circle messing up their custom stone images.

We've also been requested to do this in the past for folk who think it's cheating or distracting to know which stone was played last.

## Proposed Changes

  - Add a "game" preference for last move opacity
  - Pass this into goban
    - In Minigoban, it goes in with the config of the  goban
    - In GobanContainer it gets set when we draw the goban by calling setLastMoveOpacity
       -  That's because it seemed to complicated and "all over the place" to try to set it via config for GobanContainer applications

Needs:  https://github.com/online-go/goban/pull/150


anoek:
Almost made it through them all! Got some conflicts here with the change from transparency to opacity is all.

GreenAsJade:
Heh heh that's what I get for loading up optional PRs in parallel :)

PR #2537, opened 2024-02-01
Fix tournament result colours and elimination graph for draws

### LINKED ISSUE #2536: Ties displayed visually as-if one of the players won in tournaments
I noticed while working on #2521 that, after a draw, the double-elimination graph makes it look like one of the players won the game. Here's an example tournament with six draws in a row:
https://beta.online-go.com/tournament/193

For example, these are all drawn games:
Image 3

Problems:
- dexonsmith-beta is coloured the "winner" colour (green). Maybe there should be a "draw" colour, but certainly both players should get the same colour.
- dexonsmith-beta3 doesn't have a line leading forward in the graph, even though this player was not eliminated.

There's a good chance these visual bugs affect tournaments other than elimination tournaments. This is just where I saw it.

dexonsmith:
Drawn games are recorded as "B+0.5,W+0.5", and were incorrectly being displayed as wins for black.

- Check the full string of match results instead of just the first character.
- Fix the elimination graph drawing to allow for both players to continue.
- Choose colours for "tie" so it looks different from a loss.
- For accessibility, use a previously-unused colour from the accessibility table that is dark (since light text will be put on top of it).

Fixes #2536

### DIFF
--- a/src/global_styl/global.styl
+++ b/src/global_styl/global.styl
@@ -591,6 +591,9 @@ display: inline-block;
 .result-won {
     themed-desaturate-and-lighten-darken background-color win 55% 30%
 }
+.result-tie {
+    themed-desaturate-and-lighten-darken background-color tie 55% 30%
+}
 .result-lost {
     themed-desaturate-and-lighten-darken background-color loss 55% 30%
 }
--- a/src/ogs.styl
+++ b/src/ogs.styl
@@ -192,6 +192,7 @@ light.chat-system                       = #1CA760
 light.private-chat-user                 = #111111
 
 light.win                               = #73D355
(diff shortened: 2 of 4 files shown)


anoek:
> Drawn games are recorded as "B+0.5,W+0.5"

Ah, an unforseen consequence of the `black_lost`, `white_lost` being `false`/`false` I reckon

dexonsmith:
Old:
old

New:
new

New dark:
new-dark

New accessible:
new-accessible

anoek:
>Not sure it's related— the code for "B+0.5,W+0.5" was in the tournament backend code already for a long time

Ooooh ok yeah I stand corrected, that's kinda funny.. should probably change that

PR #2538, opened 2024-02-02
Show the ruleset and handicap stones in the black player card

dexonsmith:
Now, the ruleset and handicap are directly shown in the black player card (except in Zen mode), in the same spot where the white player card has komi.

- The ruleset uses the two-letter country code for any rules named after a country, and "Ing" and "AGA" for the Ing and AGA rules.
- The rules matrix document now includes the the short code in parentheses.
- The handicap uses the unicode for a circled number up to 50. After that, it uses parentheses (such as "(51)"), which is the old way to circle numbers on a typewriter.  I think this won't come up in practice.

Things considered and not implemented (but could be tried again in the future).

- Link to the rules matrix.  I tried this out, but I was worried someone might accidentally abandon a game when trying to pop out the scoring panel.
- Using `pgettext()` to allow translation of the codes.  I wasn't sure if this was useful, but could implement if reviewer thinks it'd be better.
- Draw circle using CSS `border-radius` instead of the Unicode.  The existing choice is more compact and works nicely for all handicaps up to 50.
- Also show the long name of the rules in the pop-out scoring panel.  It seemed cluttered.

I'll post some examples of how this looks in a moment.

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -284,6 +284,10 @@ export function PlayerCard({
     const their_turn = player_to_move === player.id;
     const highlight_their_turn = their_turn ? `their-turn` : "";
 
+    // Only show the rules for black, since white has komi showing here.
+    const rules =
+        color === "black" ? rulesString(engine.config.rules, engine.config.handicap) : null;
+
     const show_points =
         (engine.phase === "finished" || engine.phase === "stone removal") &&
         goban.mode !== "analyze" &&
@@ -352,13 +356,18 @@ export function PlayerCard({
                         })}
                     </div>
                 )}
-                <NumCapturesText
-                    score={score}
-                    color={color}
-                    zen_mode={zen_mode}
-                    hidden={show_points && !estimating_score}
-                />
-                {!show_points && <div className="komi">{komiString(score.komi)}</div>}
+                {!show_points && (
+                    <NumCapturesText
+                        score={score}
+                        color={color}
+                        zen_mode={zen_mode}
+                        hidden={show_points && !estimating_score}
+                    />
+                )}
+                {rules && <div className="rules">{rules}</div>}
+                {!show_points && !!score.komi && (
+                    <div className="komi">{komiString(score.komi)}</div>
+                )}
                 <div id={`${color}-score-details`} className="score-details">
                     <ScorePopup goban={goban} color={color} show={show_score_breakdown} />
                 </div>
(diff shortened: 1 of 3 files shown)


dexonsmith:
During the game:
captures
captures-breakdown

After scoring:
points
points-breakdown

Huge handicap:
huge handicap

Dark mode:
dark

src/views/Game/PlayerCards.tsx
```diff
@@ -352,13 +356,18 @@ export function PlayerCard({
...
                     </div>
                 )}
-                <NumCapturesText
-                    score={score}
-                    color={color}
-                    zen_mode={zen_mode}
-                    hidden={show_points && !estimating_score}
-                />
-                {!show_points && <div className="komi">{komiString(score.komi)}</div>}
+                {(!show_points || estimating_score) && (
```
dexonsmith, line 359: This is a drive-by fix, where the captures text would sometimes take up extra space when it was hidden. Instead, don't call `NumCapturesText` at all.

src/views/Game/PlayerCards.tsx
```diff
@@ -352,13 +356,18 @@ export function PlayerCard({
...
+                    <NumCapturesText
+                        score={score}
+                        color={color}
+                        zen_mode={zen_mode}
+                        hidden={show_points && !estimating_score}
+                    />
+                )}
+                {rules && <div className="rules">{rules}</div>}
+                {!show_points && !!score.komi && (
```
dexonsmith, line 368: Checking for `score.komi` here, because otherwise the blank komi `<div/>` takes up space in the black player's panel.

dexonsmith:
A few fix-ups later (cleaning up cruft from my experiments and getting tests passing), I think this is actually ready for review now.

dexonsmith:
BTW, while I'm convinced this patch is better than the status quo, it'd be _really_ nice to have a link to <https://online-go.com/docs/go-rules-comparison-matrix>.  I just couldn't figure out a UI for that without adding risk that some users would click it when they were expecting a mini-modal or something.

anoek:
Nice idea, thanks!

>  I just couldn't figure out a UI for that without adding risk that some users would click it when they were expecting a mini-modal or something.

Yeah might have to play with that a bit. It should be pretty easy to move the contents of that page out into a component then use the component as content for a modal here and as the content for that page, if that's something we wanted to do. I'm not sure clicking the AGA or whatever should open it right on the game page directly, but it seems like having a link handy or maybe that modal accessible through some mechanism or another on the page might be nice.

PR #2540, opened 2024-02-02
Fix copy/paste bug in tournament grids

### LINKED ISSUE #2539: Round robins are showing loss/win for the wrong players
From the forums: <https://forums.online-go.com/t/tournament-bug-winning-player-showing-a-loss/50624>

For example, [this tournament](https://online-go.com/tournament/111792) has 3 points for [Daniel Daniel](https://online-go.com/player/1002659/Daniel%20Daniel), and indeed, that player won all three games, but the chart makes it look like a 1 win 2 loss record.
Image
Anecdotally, seems to show up as a win for the black players.

This is almost definitely fallout from the fix for #2536, which was fixing a display bug in elimination tournaments.

dexonsmith:
Fix a copy/paste bug introduced by 0d496f86840cb25b3411013438ba2d976c17296a, which caused half of the tournament grid to show the results backwards.

|    |   P1  |   P2  |   P3  |
|----|-------|-------|-------|
| P1 |       | wrong | wrong |
| P2 | right |       | wrong |
| P3 | right | right |       |

Fixes #2539

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -1087,9 +1087,9 @@ export function Tournament(): JSX.Element {
                             : "?"
                     : "?";
                 result_map[m.white + "x" + m.black] = m.result
-                    ? m.result === "B+1"
+                    ? m.result === "W+1"
                         ? "win"
-                        : m.result === "W+1"
+                        : m.result === "B+1"
                           ? "loss"
                           : m.result === "B+0.5,W+0.5"
                             ? "tie"
@@ -1105,9 +1105,9 @@ export function Tournament(): JSX.Element {
                             : "no-result"
                     : "?";
                 color_map[m.white + "x" + m.black] = m.result
-                    ? m.result === "B+1"
+                    ? m.result === "W+1"
                         ? "win"
-                        : m.result === "W+1"
+                        : m.result === "B+1"
                           ? "loss"
                           : m.result === "B+0.5,W+0.5"
                             ? "tie"


dexonsmith:
Alternatively, we could revert the changes I made to `Tournament.groupify` in Tournaments.tsx, and leave draws unhandled for now in the display of round robin (and similar) tables.

dexonsmith:
Honestly, I don't totally understand what's going on with https://online-go.com/tournament/111792.
- One problem will be fixed by this patch (the grid is a mirror right now, but shouldn't be).
- But there also seem to some wrong results in the bottom-left quartile. The "points" totals are all correct, but the results are not.
- I don't have time tonight to look any deeper.

Options:
1. Take this patch, which is definitely better than current state.
2. Revert changes to `Tournament.groupify` in 0d496f86840cb25b3411013438ba2d976c17296a.
3. Revert 0d496f86840cb25b3411013438ba2d976c17296a entirely.

dexonsmith:
> Honestly, I don't totally understand what's going on with https://online-go.com/tournament/111792.

I understand now. The bug made it so that:
- In the top-right half, "win" meant black won, "loss" meant white won.
- In the bottom-left half, it was the reverse.

> 1. Take this patch, which is definitely better than current state, and may fix the problem.

After doing this manual testing, I'm confident this patch fixes the bugs here. IMO, this is the correct option.

Here are screenshots:

> * https://beta.online-go.com/tournament/194

Before (what is currently live at beta.online-go.com; notice that the results are mirrored across the diagonal; black marked as "won" in the top-right, white marked as "won" in the bottom-left):
Image 3

After (fixed, winner marked as "won"):
Image 2

> * https://beta.online-go.com/tournament/195

Before:
Image 5

After:
Image 4

anoek:
Awesome thanks for following up and fixing that!

PR #2541, opened 2024-02-02
Allow groups to configure the rules for their ladders

dexonsmith:
Show the rules used in ladders on the Group page, and allow those rules to be changed when editing the group.

Fixes #1042 (at least, adds one kind of configuration)
Depends on https://github.com/online-go/ogs/pull/1928

PR #2542, opened 2024-02-02
Show the rules configured for a ladder

dexonsmith:
On the ladder page itself, show what the rules are for the ladder.

Relates to #2541 (but doesn't really depend on it)

Depends on https://github.com/online-go/ogs/pull/1929

### DIFF
--- a/src/views/Ladder/Ladder.styl
+++ b/src/views/Ladder/Ladder.styl
@@ -49,6 +49,18 @@
         }
     }
 
+    .ladder-configuration {
+        th {
+            padding-right: 0.5rem;
+            text-align: right;
+            width: 40%;
+        }
+        td {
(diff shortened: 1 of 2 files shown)


dexonsmith:
Examples:
Image 7
Image 6

dexonsmith:
Found a simpler way to implement that doesn't depend on an extra backend change.

dexonsmith:
> Found a simpler way to implement that doesn't depend on an extra backend change.

After thinking about it more, depending on the extra backend is the right implementation. My alternative was to grab the rules from the ladder's group, but then the frontend would need to duplicate the right answer for site-wide ladders (which don't have a group).

This approach keeps the frontend more "dumb", in a good way.

PR #2544, opened 2024-02-02
Make report indicator take everyone to the Report Centre direct

GreenAsJade:
Fixes people complaining about unnecessary clicks to get to report centre

## Proposed Changes

  - Make report indicator take everyone to the Report Centre direct
  - Adjust help flow accordingly

### DIFF
--- a/src/components/IncidentReportTracker/IncidentReportTracker.tsx
+++ b/src/components/IncidentReportTracker/IncidentReportTracker.tsx
@@ -16,19 +16,14 @@
  */
 
 import * as React from "react";
-import * as moment from "moment";
-import * as preferences from "preferences";
 import * as data from "data";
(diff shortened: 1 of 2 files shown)


anoek:
I think what you want is to direct the CM's, which seems like a good idea - but for reporters you can't direct them there since they can't see it. Instead, that drop down contains the stuff they've filed so they can cancel it or whatever:

image

GreenAsJade:
Hah dagnabit, I knew there must have been a reason why we left that there before 😭

Just as well someone with their brain switched on is reviewing these 😜

GreenAsJade:
Fun fact: I nearly rolled the nearby "fix help flow" into this one.   Yay that I didn't 💪

PR #2545, opened 2024-02-03
Remove redundant board size from ladder list

dexonsmith:
The board size is included in every ladder name.  Stop including it redundantly in the link from the ladder list.


dexonsmith:
Old:
Image

New:
Image 2

anoek:
Hah, thanks!

PR #2547, opened 2024-02-03
Don't show the "show all" button if the log is small

benjaminpjones:
Fixes minor annoyance where the "Show all" button showed up on reports with short game logs

https://github.com/online-go/online-go.com/assets/25233703/4ae94466-0e10-44de-a117-66485bfa8718


anoek:
Thanks!

PR #2558, opened 2024-02-04
Add titles for rules and handicap in player cards

dexonsmith:
Add titles for rules and handicap in the player cards.  For example:

- Expand "JP" out to "Rules: Japanese"
- Expand "③" out to "Handicap: 3"

### DIFF
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -463,7 +463,7 @@ function rulesAndHandicap(rules: string | null | undefined, handicap_stones: num
 
     return (
         <div className="rules">
-            {code && <span title={_("Rules") + ": " + rulesText(rules)}>{code}</span>}
+            {code && rules && <span title={_("Rules") + ": " + rulesText(rules)}>{code}</span>}
             {code && stones && " "}
             {stones && <span title={_("Handicap") + ": " + handicap_stones}>{stones}</span>}
         </div>
--- a/src/views/Game/PlayerCards.tsx
+++ b/src/views/Game/PlayerCards.tsx
@@ -26,6 +26,7 @@ import { ChatPresenceIndicator } from "ChatPresenceIndicator";
 import { Clock } from "Clock";
 import { useUser } from "hooks";
 import { Player } from "Player";
+import { rulesText } from "misc";
 import { lookup, fetch } from "player_cache";
 import { _, interpolate, ngettext } from "translate";
 import * as data from "data";
@@ -284,10 +285,6 @@ export function PlayerCard({
     const their_turn = player_to_move === player.id;
     const highlight_their_turn = their_turn ? `their-turn` : "";
 
-    // Only show the rules for black, since white has komi showing here.
-    const rules =
-        color === "black" ? rulesString(engine.config.rules, engine.config.handicap) : null;
-
     const show_points =
         (engine.phase === "finished" || engine.phase === "stone removal") &&
         goban.mode !== "analyze" &&
@@ -364,7 +361,7 @@ export function PlayerCard({
                         hidden={show_points && !estimating_score}
                     />
                 )}
-                {rules && <div className="rules">{rules}</div>}
+                {color === "black" && rulesAndHandicap(engine.config.rules, engine.config.handicap)}
                 {!show_points && !!score.komi && (
                     <div className="komi">{komiString(score.komi)}</div>
                 )}
@@ -437,11 +434,12 @@ function komiString(komi: number) {
     return komi > 0 ? `+ ${abs_komi}` : `- ${abs_komi}`;
 }
 
-function rulesString(rules: string | null | undefined, handicap_stones: number | undefined) {
+function rulesAndHandicap(rules: string | null | undefined, handicap_stones: number | undefined) {
     const stones = handicap_stones ? stonesString(handicap_stones) : "";
-    let code: string;
+    let code: string | undefined;
     switch (rules?.toLowerCase()) {
         default:
+            break;
         case "japanese":
             code = "JP";
             break;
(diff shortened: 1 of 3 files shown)


dexonsmith:
Here's how it looks when hovering:
Image
Image 2

dexonsmith:
Converted to draft because I'm looking at adding something to the drop-down, too.

dexonsmith:
And now the clickable pop-up looks like this:
Image 3

dexonsmith:
Adding parenthesized code to pop-up too:

Image 4

dexonsmith:
Now avoids showing code in parentheses if it's the same as the rules text (i.e., AGA):

Image 5

dexonsmith:
ef48420f560cdd456d144771c2c424313ab2fd8b has no visible change in English, but starts using `pgettext()` so that the codes can be translated (for languages that don't use roman characters), and moves `ruleCode` next to `rulesText` in the `misc` module.

src/views/Game/PlayerCards.tsx
```diff
@@ -541,12 +526,23 @@ function ScorePopup({ show, goban, color }: ScorePopupProps) {
...
-            {color === "black" && !!goban.engine.handicap && (
-                <div>
-                    <span>{_("Handicap")}</span>
-                    <div>{goban.engine.handicap}</div>
+            {color === "black" && goban.engine.config.rules && (
+                <>
+                    <div className="summary">
+                        <span>
+                            {_("Rules")}: {rulesText(goban.engine.config.rules)}
+                            {rulesParens(goban.engine.config.rules)}
```
dexonsmith, line 534: I'm wondering if we really need the `{rulesParens()}` or if we can just delete this line and the `rulesParens` function.
- If we delete this line, the user can still discover that "NZ" means "Rules: New Zealand" by hovering over the "NZ".
- If we leave it in, it makes the pop-up unnecessarily busy with "Rule: New Zealand (NZ)". I've noticed this wraps if your window is small, but it didn't wrap before I added the parens.

GreenAsJade:
I mentioned in the forum thread: in the past "hover" has been ruled out as a solution for any UI issue, due to the phone problem.

I fully support "Handicap: 3", and think it should be it's own PR that gets in ASAP.

dexonsmith:
> I mentioned in the forum thread: in the past "hover" has been ruled out as a solution for any UI issue, due to the phone problem.

Hold on, it sounds a bit like you're saying that `title=`-based helpers are disallowed on OGS, but I see plenty of evidence to the contrary.

Certainly, it's an insufficient solution—"doesn't work on mobile" means "not a complete solution"—but `title=""`-based helpers are used in lots of other places. They improve the interface for desktop users and for those that use screen readers. I first saw it in the game list (for long game names), and `git grep title=` brings up lots of other instances.

I understand that you feel strongly that having the rules-code on the black player card is bad, and we're discussing that in the forums.

In the meantime, this PR reduces user confusion about what's currently deployed, providing two ways to figure out what "JP" means (`title=`-based helpers, and clicking on it).

> I fully support "Handicap: 3"

The only real change for handicap in this PR is adding `title=`-based hover text, which is the concern you raised above. The reformatting in the clickable pop-up is only necessary/useful if we're adding "Rules: ..." to it.

GreenAsJade:
> Hold on, it sounds a bit like you're saying that title=-based helpers are **disallowed** on OGS, but I see plenty of evidence to the contrary.

Nah, I didn't say that.

What I was noting is that _solutions_ that are based on hover have typically been deemed "not acceptable".

That tends to mean "if you want a hover, fine, but don't argue that this solves any problem".

So for example. the hovers mentioned in this PR don't solve the problem of "what are these initials in the player card".

Also historically, workarounds involving hovers have not been approved - I assume because this tends to embed the problem.

I'm speaking in these terms because I can recall cases where I've been sent back to the drawing board after tendering hover based solutions  😝

So you could read all this as "hey, how come we'd consider this PR when my previous similar have been rejected" 😝

I misunderstood the proposal for "Handicap" in this PR - I see that it is also "hover".     I think we need to get rid of the cryptic symbols ...

dexonsmith:
> I see that the PR has popup as well - I like that.

Right, the hover is a further improvement for people that use screen readers and desktops. I'm not sure why you seem to be opposed to this incremental progress.

> In fact, I think that if we put an ⓘ instead of the initials, we'd be in business from my point of view. ⓘ clearly means "here is more information" (I suggested this in the forum as well)

It looks like a stone with a 1 in it to me; I'll respond in the forums later today (middle of the night for me right now); I'm confident we'll get a solution though.

> > Hold on, it sounds a bit like you're saying that title=-based helpers are **disallowed** on OGS, but I see plenty of evidence to the contrary.
> 
> Nah, I didn't say that.
> 
> What I was noting is that _solutions_ that are based on hover have typically been deemed "not acceptable".

Thanks for clarifying.

In that case, I don't understand why you're objecting to this PR. This PR makes things incrementally better, is low-risk, and is ready to go. Why not land it?

It's not as if landing this incremental improvement precludes finding/landing a solution.

anoek:
> I mentioned in the forum thread: in the past "hover" has been ruled out as a solution for any UI issue, due to the phone problem.

I think I'm "ok" with it in this case because it's using a mechanism that, for better and worse, we already use to expand out the scoring details. Adding it there seems like an OK place to discover what those, potentially cryptic, symbols mean, but only since we already do the expansion thing in that area.


Now I will say if some day in the future we become enlightened and figure out a better place to put that data that doesn't require discovering and learning then I'd be all for it, but real estate is at a premium on the game page, and that little corner of the black was being unused, so seems like a good fit. 

Of course ideally I would love to get rid of that score expansion thing all together, but that's a pretty deep rabbit hole that probably ends in a redesign of the game page.

GreenAsJade:
> In that case, I don't understand why you're objecting to this PR. This PR makes things incrementally better, is low-risk, and is ready to go. Why not land it?

To clarify this point: consistency.

In the past, PRs involving hovers have been rejected - despite the fact that we do have some hovers.

Approving another PR with hovers just adds to the inconsistency.

That's the reason why I was pushing back.

> It looks like a stone with a 1 in it to me;

Fair cop.   I hope we find a better solution soon.

PR #2560, opened 2024-02-04
Fix order of pgettext and interpolate for "Message {{who}}"

dexonsmith:
Invert pgettext and interpolate calls for the three instances of:

    "Message {{who}}"

... so that "message" can be translated in the suggested text in chats.

From the forums: https://forums.online-go.com/t/no-sgf-from-just-part-of-my-games-why/50453/17

FWIW, I manually tested that I still get the expected prompts in English in the three places in the interface.


anoek:
Thanks!

PR #2561, opened 2024-02-05
Fix double-elimination graph for if-games

dexonsmith:
When there's an if-game in a double-elimination tournament, both players have lost a game and are notionally in the "second bracket".  Update the drawing code to do something reasonable for that case.

- Always draw finals "in the middle" if either player came from the second bracket.
- Always connect lines back to the source.

This handles any backend change to support this forum proposal: https://forums.online-go.com/t/update-to-double-elimination-tournament/50521


dexonsmith:
Assuming we take the backend change...

Before:
Image 6

After:
Image 7

anoek:
Very cool, thanks for tracing that through and getting that working!

PR #2562, opened 2024-02-06
Allow groups to turn on handicaps for their ladders

### LINKED ISSUE #2553: Ladder configuration: handicap settings
Split off from #1042, enable groups to configure whether handicap is enabled on their ladders.

dexonsmith:
Add frontend support for handicaps in group ladders:

- Enable configuration on the group page.
- Show the setting on the ladder page.
- Turn on handicaps by default when creating new tournaments.

One slightly awkward thing is that the group page canonicalizes the type as a `number` (0 or -1) and tournament page uses a `string` ("0" or "-1"), depending on what the backend expects.  If you'd rather have consistency here, I can change the backend to expect a string for the group page and convert to a number in the backend.

Fixes #2553
Depends on https://github.com/online-go/ogs/pull/1933

### DIFF
--- a/src/views/Group/Group.tsx
+++ b/src/views/Group/Group.tsx
@@ -71,6 +71,7 @@ interface GroupInfo {
     has_active_tournaments?: boolean;
     has_finished_tournaments?: boolean;
     rules?: string;
+    handicap?: number;
 }
 
 // API: group/%id%/news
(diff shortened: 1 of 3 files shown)


dexonsmith:
Some screenshots from manual testing follow.

Confirmed games are still even when handicap is set to None:
Image 3

Here's how the setting looks, set to Automatic in these screenshots:
Image
on the ladder page:
Image 6

9x9 game created here has the appropriate 12-rank handicap:
Image 4
Image 5

New tournaments default to handicap on:
Image 2

anoek:
Thanks!

PR #2563, opened 2024-02-06
Move rules out of the black player card

dexonsmith:
In the black player card, the rules codes can be mistaken for initials of the black player.  Move them out, just below the player cards, into an area for condensed game information with small font.

- Rules, aligned right.
- Ranked games, "Ranked", aligned left.

This is an option for addressing [feedback from the forums](https://forums.online-go.com/t/the-ruleset-of-the-game-should-be-easier-to-see/39165).

Draft for discussion. I'll post screenshots in a moment.

### DIFF
--- a/src/views/Game/Game.tsx
+++ b/src/views/Game/Game.tsx
@@ -26,7 +26,7 @@ import { popover } from "popover";
 import { post, get, abort_requests_in_flight } from "requests";
 import { KBShortcut } from "KBShortcut";
 import { UIPush } from "UIPush";
-import { errorAlerter, ignore } from "misc";
+import { errorAlerter, ignore, rulesText } from "misc";
 import {
     Goban,
     GobanCanvas,
@@ -786,6 +786,25 @@ export function Game(): JSX.Element | null {
             shareAnalysis();
             return false;
         }
+    };
+    const frag_game_information = () => {
+        const config = goban.current?.engine?.config;
+        if (!config) {
+            return null;
+        }
+        const rules = config?.rules ? rulesText(config.rules) : null;
+        return (
+            <div className="condensed-game-information">
+                {(config?.ranked || null) && (
+                    <div className="condensed-game-ranked">{_("Ranked")}</div>
+                )}
+                {rules && (
+                    <div className="condensed-game-rules">
+                        {_("Rules")}: {rules}
+                    </div>
+                )}
+            </div>
+        );
     };
     const frag_estimate_score = () => (
         <EstimateScore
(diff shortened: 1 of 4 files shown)


dexonsmith:
Two examples:

An unranked handicap game:
Image 2

A ranked even game:
Image

GreenAsJade (approved):

GreenAsJade (changes requested):
I realised a couple of things:

 - When there is a player to move, it's getting in between the cards and this.   This should always be right below the cards.
 
  - This should be turned off in Zen mode

dexonsmith:
> * When there is a player to move, it's getting in between the cards and this.   This should always be right below the cards.

It looks correct to me:
Image 5
Image 6

Am I looking at the right thing?

dexonsmith:
Looking at the code, I just noticed there's a rengo header (somewhat confusingly) inside of `<PlayerCards/>`. I can move that text to this level.

dexonsmith:
> > * When there is a player to move, it's getting in between the cards and this.   This should always be right below the cards.
> 
> It looks correct to me: 
> [...]
> Looking at the code, I just noticed there's a rengo header (somewhat confusingly) inside of `<PlayerCards/>`. I can move that text to this level.

@GreenAsJade, are you able to test f21c792? I don't know how to set up rengo games.

> * This should be turned off in Zen mode

72c6c92

dexonsmith:
> @GreenAsJade, are you able to test [f21c792](https://github.com/online-go/online-go.com/commit/f21c79262c246fd0cafafd6fbd839421f79b8362)? I don't know how to set up rengo games.

Or, point me at a rengo game at beta.online-go.com where there's a title? I only see four active ones there (e.g., [this one](https://beta.online-go.com/game/13966)), and none of them sees to have "show title".

GreenAsJade:
Weird - this is what I saw this morning (sorry for rushing out without doing the screenshot then):

Screenshot 2024-02-07 at 12 58 15 pm

(browser tab still open :) )

However, when I click into the same game in the Uffuzi VM now, I see this:




🤷‍♀️

GreenAsJade:
> > @GreenAsJade, are you able to test [f21c792](https://github.com/online-go/online-go.com/commit/f21c79262c246fd0cafafd6fbd839421f79b8362)? I don't know how to set up rengo games.
> 
> Or, point me at a rengo game at beta.online-go.com where there's a title? I only see four active ones there (e.g., [this one](https://beta.online-go.com/game/13966)), and none of them sees to have "show title".

I'll take a look as soon as I can ... it might be "cruft", but some Rengo stuff does belong in PlayCards in general because all the players are listed in there (as you've probably seen in the code).  In Rengo, the PlayerCard is the "team card".

I have a vague idea that initially games in Rengo mode had a title that said something like "Rengo mode courtesy of <whoever it was that sponsored for it, I've forgotten ☺️>"   ... maybe we stopped that showing but didn't completely remove the element 🧐

dexonsmith:
> Weird - this is what I saw this morning (sorry for rushing out without doing the screenshot then):

That's... curious. I'm not really sure how to fix it if we can't reproduce.

> I have a vague idea that initially games in Rengo mode had a title

The only content in the thing I moved comes from the `useTitle()` hook, so I think your vague idea must be it. There's also a `useShowTitle()` hook, which appears to be false for all the games I've looked at.

src/views/Game/Game.tsx
```diff
@@ -1663,20 +1701,24 @@ export function Game(): JSX.Element | null {
...
+                                        black_ai_suspected={bot_detection_results?.ai_suspected.includes(
+                                            historical_black?.id,
+                                        )}
+                                        white_ai_suspected={bot_detection_results?.ai_suspected.includes(
+                                            historical_white?.id,
+                                        )}
+                                    />
+                                    {frag_game_information()}
+                                    {frag_rengo_header()}
+                                </>
```
dexonsmith, line 1721: Maybe, this (and the one above for portrait mode) should be a `<div>` (not a `<>`) to make the player cards and the condensed-game-info indivisible? I only have a vague idea about how layout works, but maybe the condensed-game-info is allowed to float down below the "White's move" stuff, and sometimes that's happening.

dexonsmith: 163956d defensively does this. I don't fully understand if/why it's necessary, but...

dexonsmith:
Changed to "ready for review" since the forum seems happy enough with this. (Some debate about whether to show only "Unranked" or only "Ranked" or whichever one is relevant, but I think showing it either way will satisfy everyone.)

anoek:
Awesome, thanks for working through that all!

PR #2565, opened 2024-02-07
Keep absolute time indication with main time

dexonsmith:
Encourage CSS layout to keep the `/*` absolute time indication with the main time by putting it in the same `<span>`.  When the main time has a line break, this prevents creating an extra one just for the `/*`.

I.e., instead of:

    6 Days 23
        Hours
           /*

We get:

    6 Days 23
      Hours/*

EDIT: after feedback in the forums, seems `+0` is more intuitive:

    6 Days 23
      Hours+0

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -115,8 +115,7 @@ export function Clock({
                         {prettyTime(player_clock.main_time)}
                         {time_control.system === "absolute" && (
                             <>
-                                <span className="periods-delimiter">/</span>
-                                <span className="period-moves boxed">*</span>
+                                <span className="absolute-time">+0</span>
                             </>
                         )}
                     </span>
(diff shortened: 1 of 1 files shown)


dexonsmith:
Old:
Image 8

New:
Image 9

dexonsmith:
Running a poll at https://forums.online-go.com/t/poll-should-absolute-time-say-time-time-0-or-time-abs/50699.

anoek:
Hah, 0 votes at `/*` .. made immediate sense to me, reckon that's the programmer symbol recognition.

Thanks for taking and listening to feedback and making it happen!

dexonsmith:
> Hah, 0 votes at `/*` .. made immediate sense to me, reckon that's the programmer symbol recognition.
> 
> Thanks for taking and listening to feedback and making it happen!

Hah, yeah, even I voted for `+0` once someone pointed out the similarity to byo-yomi. I prefer the look/feel of `/*` here, but since few people use Canadian time there's nothing to latch on to.

PR #2568, opened 2024-02-08
New User Rank Chooser

GreenAsJade:
Functionality as per 

https://docs.google.com/document/d/1LRjbgZKYfpxSO1mCCdaq-JwXRAOC8mbSh8KWGsk61R0/edit?usp=sharing

* Needs https://github.com/online-go/ogs/pull/1943


anoek:
Let's keep this as draft until the backend is ready for merging, just helps me keep track of things that are ready vs not ready to merge a little easier

GreenAsJade:
> Let's keep this as draft until the backend is ready for merging, just helps me keep track of things that are ready vs not ready to merge a little easier

Yep - just if anyone wants to try this (Uffuzi or for debug) at least it works :)

dexonsmith:
> Yep - just if anyone wants to try this (Uffuzi or for debug) at least it works :)

I just tried using Uffizzi (I've never used it before; seems neat):

1. clicked and waited for load
2. registered new user
3. clicked on profile
4. clicked on play

I never got a prompt with buttons.

Is this expected? (because the backend changes are missing?)

(Hopefully I'll finally get some time this weekend to get back into this so we can finish it off and ship it...)

GreenAsJade:
Oh yeah, you'll need https://github.com/online-go/ogs/pull/1935   to see anything.

In that PR it's hardwired "on".

GreenAsJade:
Now the back end is there, I've found  bunch of places in the front end expecting 1500 as the starting rank, so this PR needs those sorted, I'll get to that next.

BHydden:
IIRC,  Humble Rank was introduced initially with the intention of allowing brand new users to both play 25k accounts if they are beginners and also to rank up quickly if they are already strong. I don't really see a reason to keep Humble Rank if we are to allow new users to indicate their relative strength before any ranked games are played,  especially if the ability to play ranked games beyond a ±9 rank limit is also brought in,  as I believe was done recently.

GreenAsJade:
Yeah - the thing is that users are not _forced_ to choose an initial rank, so there will still be users at the default 1500 rating needing to be able to "play lower than that"

dexonsmith:
I already have a patch up (draft) that neutralizes humble. Part of the plan as discussed in the forums was to remove it as part of adding starting rank. I still think it makes sense to remove it. On Mar 3, 2024, at 20:10, GreenAsJade ***@***.***> wrote:﻿
Yeah - the thing is that users are not forced to choose an initial rank, so there will still be users at the default 1500 rating needing to be able to "play lower than that"

BHydden:
That's why I also mentioned removing the rank restriction, as the main reason for humble rank was new accounts not being able to play beginners... which if there are no rank restrictions won't be a problem. 

Another option might be to have the new account rank chooser available at any point before your first ranked game,  in case someone chooses wrong option at account creation.

GreenAsJade:
> I already have a patch up (draft) that neutralizes humble. Part of the plan as discussed in the forums was to remove it as part of adding starting rank. I still think it makes sense to remove it. On Mar 3, 2024, at 20:10, GreenAsJade ***@***.***> wrote:﻿ Yeah - the thing is that users are not forced to choose an initial rank, so there will still be users at the default 1500 rating needing to be able to "play lower than that" —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: ***@***.***>

I think that because we did not insist on people choosing a starting rank, we need to leave humble there: the reasons that it is there still apply to people who didn't choose.

I don't feel strongly about that, but I finished up the backend for this leaving humble in for this reason.

But it doesn't stop us taking it out, separately.

dexonsmith:
When we added “skip”, the plan was to put them in at 12kyu, not 1500.

anoek:
Looking pretty nice man, thanks for doing all of this.

On the profile page, if you've skipped, could we fix the spacing on the left here?

image

GreenAsJade:
Fussy fussy!   😅

(Yeah I totally agree... )

GreenAsJade:
^^ That's done.

GreenAsJade:
Good catch 🤦‍♀️ 

I'll do some fink.

anoek:
This is awesome @GreenAsJade , thanks!

PR #2570, opened 2024-02-08
Add link to selected AI Player profile page in Challenge dialog.

### LINKED ISSUE #2569: Add convenience link to AI Player profile when creating a computer challenge
Sometimes I would like to view the profile page for a bot before I challenge it, but there's no link provided on the challenge dialog.

Workaround is to close the dialog, and search for the bot's name, if you can remember it.

A simple link to the bot's profile page would suffice:
Screenshot 2024-02-08 at 6 42 10 AM

ecyrb:
Addresses #2569

## Proposed Changes

In the challenge dialog, adds a link to a selected AI Opponent's profile page, using the PlayerIcon (if present) and username of the bot. 

Screenshot 2024-02-08 at 7 06 06 AM

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -969,7 +969,7 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
                                     target="_blank"
                                     title={_("Selected AI profile")}
                                 >
-                                    <i className="fa fa-link" />
+                                    <i className="fa fa-external-link" />
                                 </a>
                             </div>
                         </div>
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -38,7 +38,7 @@ import {
 } from "TimeControl";
 import { sfx } from "sfx";
 import { notification_manager, NotificationManagerEvents } from "Notifications/NotificationManager";
-import { one_bot, bots, bot_count, bots_list } from "bots";
+import { one_bot, bot_count, bots_list } from "bots";
 import { goban_view_mode } from "Game/util";
 import {
     Goban,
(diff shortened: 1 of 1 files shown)


src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -949,19 +949,27 @@ export class ChallengeModal extends Modal<Events, ChallengeModalProperties, any>
...
+                                        <option key={idx} value={bot.id}>
+                                            {bot.username} ({rankString(getUserRating(bot).rank)})
+                                        </option>
+                                    ))}
+                                </select>
+                            </div>
+                            <div>
+                                <a href={`/user/view/${this.state.conf.bot_id}`}>
+                                    <PlayerIcon id={this.state.conf.bot_id} size={16} />{" "}
+                                    {bots()[this.state.conf.bot_id].username}'s profile
```
dexonsmith, line 970: Not sure if we want this spelling, but if we do, this should be translated.  Something like:

    interpolate(_("{{who}}'s profile"), {who: bots()[...]})

ecyrb: Makes sense. I just now found the [wiki page on Internationalization and Localization](https://github.com/online-go/online-go.com/wiki/Internationalization-and-Localization). I was sort of expecting a mention of translation in CONTRIBUTING.md — that's where I looked and didn't see anything. Not sure if pointing to the wiki from the source is what should occur, but some pointer on translation in that file would've been useful for me.

dexonsmith:
Looks correct to me now!

I like it as-is, but maybe there are other thoughts on the wording (@benjaminpjones / @anoek / @GreenAsJade ?). A few options:
- "{{who}}'s profile" (as-is)
- "{{who}}" (i.e., skip the translation; maybe not obvious enough that it's a profile link though)
- "See {{who}}'s profile for guidelines" (something actionable... but maybe overly verbose and clunky)

anoek:
Doubling up the name like that looks a little strange to me. What about doing something like a little link icon by the drop down or something instead? Something along the lines of this: 

image

I believe the link icon is just `<i className="fa fa-link" />`. Feel free to push back if y'all like the current design or something else pops into mind of course.

ecyrb:
I like the link icon. Maybe add some mouseover and alt-text?

We don't want it to pop-out, right? (Screenshot 2024-02-10 at 5 59 15 AM) — I ask because the back button won't get you back to the open computer challenge dialog.

anoek:
Oh pop-out might be better eh? If we open in a new window then there's no expectation that the back button will get you back

benjaminpjones:
Definitely agree we don't need the name when it's right above. I was thinking something like "bot profile" for clarity, but if it's a pop out, then it's relatively low-stakes to click the icon and see :)

ecyrb:
> I was thinking something like "bot profile" for clarity,

I had "profile" initially, but when the selection changes and the link text didn't, it felt weird, like it didn't seem like the link wasn't actually being updated.

ecyrb:
Latest looks like this (mouseover ("title") text included):
Screenshot 2024-02-12 at 12 10 16 AM

Was going to use a [different icon](https://fontawesome.com/icons/up-right-from-square?f=classic&s=solid), but that _might_ require a font-awesome upgrade?

anoek:
FA4 has it too: `fa-external-link`  https://fontawesome.com/v4/icon/external-link

anoek:
Awesome, thanks!

PR #2572, opened 2024-02-09
Delay reloading tournament when saving edits

### LINKED ISSUE #2571: "Save tournament" request aborted on success, leaving tournament in edit mode
**Describe the bug**
If you click "Edit Tournament" then "Save Tournament", you're left in edit mode.

**To Reproduce**
1. Create a tournament at https://beta.online-go.com
2. Click "Edit Tournament"
3. Click "Save Tournament"
4. Notice that you are still in "edit" mode.

**Expected behavior**
You're out of edit mode (unless there's some error!), and can see the results of saving the tournament.

**Additional context**
In fact, if you're looking carefully, the tournament leaves edit mode briefly, then goes back into edit mode on a successful save. The problem is that the PUT request **_aborts itself on success_**.

- The "Save" button sends out a `put()` request and exits edit mode.
- The [put request](https://github.com/online-go/online-go.com/blob/devel/src/views/Tournament/Tournament.tsx#L1288) calls `resolve()` on success and `onError()` on failure.
    - ~~Since it's successful, it calls `resolve()`.~~ EDIT: misdiagnosed. The server sends a push notification that aborts the request before sending `"success"`.
- [resolve](https://github.com/online-go/online-go.com/blob/764699c392813ae1deb8c262a1a3853bc75dcaef/src/views/Tournament/Tournament.tsx#L260)'s first move is to abort requests in flight.
    - ~~This aborts the request currently being resolved and calls its `onError()` function.~~ EDIT: misdiagnosed. `resolve()` is never called.
- [onError](https://github.com/online-go/online-go.com/blob/764699c392813ae1deb8c262a1a3853bc75dcaef/src/views/Tournament/Tournament.tsx#L1282) kicks us back into edit mode.

dexonsmith:
If the tournament director clicks "Save Tournament" after making edits, when they're accepted by the server, the server will send a PUSH notification on the tournament channel asking all clients to reload before sending back "success".

Delay this reloading action until we have a response from the "Save" request. Previously, the reload was aborting the about-to-succeed request, kicking the client back into edit mode.

Fixes #2571

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -148,7 +148,7 @@ export function Tournament(): JSX.Element {
         document.createElementNS("http://www.w3.org/2000/svg", "svg"),
     );
 
-    const [edit_save_state, setEditSaveState] = React.useState("none" as EditSaveState);
+    const [edit_save_state, setEditSaveState] = React.useState<EditSaveState>("none");
     const [, _refresh] = React.useState(0);
     const refresh = () => _refresh(Math.random());
     const [loading, setLoading] = React.useState(true);
@@ -1287,15 +1287,15 @@ export function Tournament(): JSX.Element {
         //tournament.round_start_times = round_start_times;
 
         if (tournament.id) {
-            setEditSaveState("saving" as EditSaveState);
+            setEditSaveState("saving");
             put(`tournaments/${tournament.id}`, tournament)
                 .then(() => {
-                    setEditSaveState("none" as EditSaveState);
+                    setEditSaveState("none");
                     resolve();
                 })
                 .catch((err: any) => {
                     const should_reload = edit_save_state === "reload";
-                    setEditSaveState("none" as EditSaveState);
+                    setEditSaveState("none");
                     setEditing(true);
                     errorAlerter(err);
                     if (should_reload) {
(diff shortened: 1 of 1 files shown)


src/views/Tournament/Tournament.tsx
```diff
@@ -129,6 +129,8 @@ interface TournamentInterface {
     opengotha_standings?: boolean;
 }
 
+type EditSaveState = "none" | "saving" | "reload";
```
dexonsmith, line 132: Not sure this is really providing any type safety benefit when using React's useState. Is there some way to declare a type in `useState` that I missed?

anoek: Yep like this: `React.useState<EditSaveState>("none")`

anoek: (Once that's in you shouldn't nee dto do any of the "as EditSaveState" stuf fin either the useState call or the setEditSaveState)

dexonsmith: Yeah, and there are examples of that nearby too... done in 3cb14c68159b83fb4c7c4bf20b5483011e5e54f8.

anoek:
Thanks!

PR #2573, opened 2024-02-09
WIP: Allow tournaments to be cloned

dexonsmith:
Tournaments are super error-prone to create, because there are a lot of settings to get right. It should be easy to clone a previous tournament as a basis for a new one.

This enhancement allows tournaments to be cloned by adding routes:

- `/tournament/clone/7` clones tournament 7 without a group
- `/tournament/clone/7/37` clones tournament 7 into group 37

This works if you manually navigate to the URLs.  It behaves like a new tournament, loads tournament 7, and copies everything in.

But, the shiny new "Clone Tournament" button doesn't work.  We need a fresh instance of the `<Tournament/>` component, but it's reusing the existing one.

I'm not sure what right way to fix this is. Draft for now.

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -262,8 +262,8 @@ export const routes = (
                     element={<Tournament />}
                 />
                 <Route path="/tournament/clone/:src_tournament_id" element={<Tournament />} />
-                <Route key="current" path="/tournament/:tournament_id" element={<Tournament />} />
-                <Route key="current" path="/tournaments/:tournament_id" element={<Tournament />} />
+                <Route path="/tournament/:tournament_id" element={<Tournament />} />
+                <Route path="/tournaments/:tournament_id" element={<Tournament />} />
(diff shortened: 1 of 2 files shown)


dexonsmith:
Seems like I need to add `src_tournament_id` to the useEffect dependency chain.

dexonsmith:
> Seems like I need to add `src_tournament_id` to the useEffect dependency chain.

It's not that simple. I think to fix this properly I need to convert `tournament_ref` into a `useState` (instead of `useRef`) and change most other state functions into `useMemo` (memoization functions derived from the tournament state). I started pushing that through, but it's pretty hard to reason about what's actual vs. redundant state without a major refactoring.

I've just pushed a PR for one piece of that in #2574, which clarifies the logic for the elimination graph. In particular, because that patch moves the guts into top-level functions, it's more clear which pieces of state are actually being used and/or modified by `updateEliminationTrees`.

dexonsmith:
Abandoned in favour of #2579.

PR #2579, opened 2024-02-12
Allow tournament administrators to clone tournaments

dexonsmith:
Add a "Clone Tournament" button to the tournament page that allows an existing tournament to be cloned.

- Button shows up for anyone that can administer the tournament.
- Takes the user to the "new tournament" page in the same group.
- All tournament settings are pre-loaded to match the source tournament.

This makes it easy to create new tournaments in a series, without the laborious and error-prone process of re-entering the settings.

Replaces #2573.

In case it's easier to look at them individually, there are three commits here, the first two of which are prep commits which would be nice to land even if this PR doesn't:
- d6ec4d5b7cbebbbf89642d545dd341497f622bb2 Remove extra state variables to track what has loaded
- b7b754175dcbcbb33d3b1f7cc1e0db5d5ef7c049 Delay editing new tournaments until the group has loaded
- 30251cbfbb24749f6fc32e13972232ab881627d9 Allow tournament administrators to clone tournaments

### DIFF
--- a/src/views/Tournament/Tournament.tsx
+++ b/src/views/Tournament/Tournament.tsx
@@ -88,7 +88,7 @@ interface TournamentPlayers {
 }
 
 interface TournamentInterface {
-    id: number;
+    id?: number;
     name: string;
     director: player_cache.PlayerCacheEntry;
     time_start: string;
@@ -141,13 +141,8 @@ export function Tournament(): JSX.Element {
     const ref_max_players = React.useRef<HTMLInputElement>(null);
 
     const [edit_save_state, setEditSaveState] = React.useState<EditSaveState>("none");
-    const [tournament_loaded, setTournamentLoaded] = React.useState(false);
-    const [rounds_loaded, setRoundsLoaded] = React.useState(false);
-    const [players_loaded, setPlayersLoaded] = React.useState(false);
-    const loading = !rounds_loaded || !players_loaded || !tournament_loaded;
 
     const [tournament, setTournament] = React.useState<TournamentInterface>({
-        id: tournament_id,
         name: "",
         // TODO: replace {} with something that makes type sense. -bpj
         director: tournament_id === 0 ? user : ({} as any),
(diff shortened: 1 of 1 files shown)


src/views/Tournament/Tournament.tsx
```diff
@@ -242,15 +255,17 @@ export function Tournament(): JSX.Element {
...
+        raw_rounds.length > selected_round_idx
             ? raw_rounds[selected_round_idx]
             : null;
 
     // this is so anoek (user id 1) can quickly test tournaments
     React.useEffect(() => {
         if (user.id === 1 && tournament_id === 0) {
             setTournament({
-                ...tournament,
+                ...default_tournament,
```
dexonsmith, line 268: @anoek, for obvious reasons, I haven't been able to test this part of the code. I also suspect it blocks "Clone Tournament" from working for you.

Two options to consider:

- Delete this block entirely now that we have "Clone Tournament"? It seems like the only thing you're gaining vs. clicking "Clone Tournament" is not having to set `time_start`.
- Lift this up, changing how `default_tournament` gets set based on `user.id === 1` instead of modifying the tournament later via `useEffect`. This would be easier to reason about.

dexonsmith: I implemented and pushed the second option in 80e5e8f5. But it might be nice to delete this entirely, or limit it just to a `time_start` override, if you can use "Clone Tournament" instead.

dexonsmith:
Oh, forgot to add screenshots.

Here's what the button looks like for tournaments that have started:
Image

and for tournaments that haven't started:
Image 2

(The edit interface looks exactly like it does for other new tournaments... just that the details have been filled in... so not going to post a screenshot of that.)

dexonsmith:
One other consideration is what to do for the tournament title. If the tournament-to-clone's name was `"abc #7"`, here are some options for the default name for the new tournament:
- `"abc #7"` (this is what the patch does)
- `"abc #8"`
- `"abc #7 (" + _("cloned") + ")"`

I like the simplicity of just taking the old title verbatim. The user can fix it.

dexonsmith:
Merged in the fix for #2580 and resolved the conflict in 6cd46298a09df68182f9f27621ba4103d34586fa.

anoek:
Slick, thanks!

PR #2582, opened 2024-02-14
Center the board on mobile.

benjaminpjones:
The margins were uneven (left: 0, right: .5rem).  It was introduced in https://github.com/online-go/online-go.com/pull/2355 to address display jitter on desktop, but I don't believe the change should have been applied to portrait layout as well.  The wider margin and un-centered board recently drew criticism in [the forums](https://forums.online-go.com/t/give-max-size-to-the-board/50375/6)

## Screenshots

Taken in Firefox's Responsive Design tool (device: Samsung S20, display width: 360px)

Before and After

PR #2585, opened 2024-02-18
Game.styl: remove unnecessary 8px right margin on "center-col", so board can get larger

pdg137:
I did this before I noticed @benjaminpjones's https://github.com/online-go/online-go.com/pull/2582. My main goal was to fix mobile, but still I think we can get rid of the 8px margin under all circumstances.  The right-col already has its own left margin so I don't see why we should also have a right margin on center-col.

Before:

image

After:

image

Without the highlighting, so you can see it's not ridiculous:

image

A larger window size showing the closest the board ever gets to the stuff on the right:

image

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -317,7 +317,6 @@ goban-view-bar-width=400px
         flex-direction: column;
         align-items: stretch;
         width: 100%;
-        margin-right: 0.5rem;
     }
     &.portrait .center-col {
         .cancel-button {
@@ -326,7 +325,6 @@ goban-view-bar-width=400px
             position: relative;
             top: 25px;
         }
-        margin-right: unset;
     }
 
 


anoek:
Does the "dancing screen" aka flickering resize issue happen at all as you slowly resize the window at any point?

I am not sure why the margin fixed that issue, but the margin was introduced to resolve that issue and it was reported that it fixed it at the time.

pdg137:
Okay, I checked the following:

* ec0687c8253814479646deed2b57e497334aa867 - the commit before https://github.com/online-go/online-go.com/pull/2355/ -  has the bug, related to an unnecessary vertical scrollbar that appears
* d5ac8cefcc22fd9452c5b0e4190c80cf7ad8dc7f - the commit of https://github.com/online-go/online-go.com/pull/2355/ - does not have the bug or any unnecessary vertical scrollbars
* 2cd453b6a09068c095b74f582233bd311d02d33e - this branch - does not have the bug or any unnecessary vertical scrollbars

I did a `git bisect` checking for versions where the 8px margin is required and found e970a2c33d415aceb19418c0848ec557e571bb88 which is clearly what fixed the problem for good, though it's kind of sad that we needed to do it like that.

More generally, I think removing things that change at particular sizes - like I'm doing in this PR - is going to help the layout be more stable and prevent problems like this from popping up mysteriously under certain conditions.

anoek:
Very cool, thanks for doing that research, glad to be rid of the hack!

benjaminpjones:
Awesome! Thanks for looking into that!

PR #2586, opened 2024-02-18
Moderator pref to control report quota

GreenAsJade:
Fixes folk complaining if they can, or if they cannot, see how many reports are remaining.

## Proposed Changes

  - Make the `DAILY_REPORTS` a preference, controlled on Moderator Settings
  - Let CMs use this setting
  - Fix up "available reports" logic so that you don't see any if you have done your quota


anoek:
Awesome, good change I think - thanks!

PR #2589, opened 2024-02-20
(some) Tournament.tsx typing

benjaminpjones:
After reviewing some of the Tournament PRs, I realized the lack of types on core objects such as `rounds` and `sorted_players` hurts readability.  This PR reduces the number of `any`s in this file from 85 to 74.

## Proposed Changes

  - Add `Round` type, and use it where possible
  - Fix some other types where possible
  - I believe all of these changes are types only, so this should be pretty safe.


anoek:
Awesome, thanks!

PR #2592, opened 2024-02-20
comment - different component on mobile

GreenAsJade:
Fixes uninitiated folk overlooking that you have to change this in two places 😅

## Proposed Changes

  - Comment about the fact that mobile has a different component for the Settings menu

### DIFF
--- a/src/views/Settings/Settings.tsx
+++ b/src/views/Settings/Settings.tsx
@@ -209,6 +209,7 @@ export function Settings(): JSX.Element {
             </h2>
 
             <div id="SettingsContainer">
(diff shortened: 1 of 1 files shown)


anoek:
> Fixes uninitiated folk overlooking that you have to change this in two places

Oh man, how many times have I missed my own comments saying "IF YOU DO THINGS HERE YOU NEED TO DO THIS THING OVER THERE".

Better than nothing though :D If nothing else it provides that confirmation after I finish debugging and tracing things back only to finally notice the comment I left for myself :sweat_smile:

PR #2595, opened 2024-02-24
Wrap the <input> in chat boxes in a form and provide hint

GreenAsJade:
to try to ensure that phone browsers give users an enter button

Fixes ... hopefully... https://forums.online-go.com/t/unable-to-reply-to-group-chat/49861?u=greenasjade

## Proposed Changes

  - Wrap the `input` in a non-submit `form` to provide a hint to phone browsers that this input needs an enter button

### DIFF
--- a/src/components/TabCompleteInput/TabCompleteInput.tsx
+++ b/src/components/TabCompleteInput/TabCompleteInput.tsx
@@ -35,7 +35,8 @@ export const TabCompleteInput = React.forwardRef<HTMLInputElement, TabCompleteIn
         }, [(ref as any).current]);
 
         // The input is wrapped in a form so that it presents a send button
-        // properly on mobile.
+        // properly on mobile, avoiding Smart Go Next kind of problems:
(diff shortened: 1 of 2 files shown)


src/components/TabCompleteInput/TabCompleteInput.tsx
```diff
@@ -17,23 +17,30 @@
 
 import * as React from "react";
 
-interface TabCompleteInputProperties {
+interface TabCompleteInputProperties extends React.HTMLProps<HTMLInputElement> {
```
benjaminpjones, line 20: tbh this whole `TabCompleteInputProperties` type is unnecessary.  Everything defined here is already defined in `React.HTMLProps<HTMLInputElement>`

benjaminpjones (approved):
I think this will do it! (also, TIL we have tab completion in the chat fields.  Do people actually use that??)

anoek:
Seems good on my devices, it's up on beta to smoke test it

src/components/TabCompleteInput/TabCompleteInput.tsx
GreenAsJade: Hah - 😜

PR #2605, opened 2024-02-29
add an enter key hint to TabCompleteInput

GreenAsJade:
The minimal change suggested to fix https://forums.online-go.com/t/unable-to-reply-to-group-chat/49861?u=greenasjade

## Proposed Changes

  - add an `enterKeyHint` to the   `input` of `TabCompleteInput`
  
  This doesn't break group chat or game chat for me, at various screen widths, but I can't easily test whether it fixes group chat as requested (since installing an android simulator doesn't count as "easy" 😝 )


anoek:
I've never experienced the original bug so not sure if this fixes or not but seems safe to try at any rate.

PR #2607, opened 2024-02-29
Update setExtraActionCallback whenever the tournament changes

dexonsmith:
After f0a617aa009781d1fcbb7074df04e0ac7b68edb2, tournament is immutable state (new copy on changes), not a mutable ref, so the renderExtraPlayerActions callback wasn't being kept up-to-date when the tournament changed.  Split it out to its own `useEffect` that updates when the relevant tournament fields do.

Tested:

- Director of started tournament (disqualify and adjust points)
- Director of not-started tournament (kick)
- Non-director (no actions)
- Switching between them via the search box, which reuses the same component

Fixes #2606

PR #2608, opened 2024-02-29
Add missing cleanup function from 9a3f6fec611f965a531939e38241adaeace2284a

dexonsmith:
Follow-up to #2607. I missed the cleanup function before. Couldn't find anything broken without it, but it should probably be there.


dexonsmith:
(Cleanup was removed from the other effect in 9a3f6fec611f965a531939e38241adaeace2284a but not added to the new one.)

PR #2614, opened 2024-03-09
Respect variation stone opacity in SGF library game view

JDNdeveloper:
Previously the variation stone opacity user preference was only respected in game reviews, but not when viewing SGFs. It looks like `goban.current.on("gamedata", ...)` isn't triggered in the SGF view. Since this is a static setting for the lifetime of the game view it makes sense to set this in the initial goban opts.


anoek:
Thanks!

PR #2615, opened 2024-03-09
Show CMs a list of related reports that they are allowed to see

GreenAsJade:
Fixes CMs not having good insight into duplicate reports

## Proposed Changes

  - Show the 'related reports" list to CMs (filtered by "can they see that type")


anoek:
Thanks!

PR #2617, opened 2024-03-09
Optimize vertical space in view report

GreenAsJade:
Fixes  extra scrolling by mods and CMs

## Proposed Changes

  - Put "created by" alongside "Game heading"
  - Put "reported on turn" into game info, which is less tall than the goban so has space for it
 
** This PR is on top of #2616.   It doesn't _have to be_, but they operate on the same code, so it's easier to manage if it is.


anoek:
Thanks!

PR #2621, opened 2024-03-12
Change escaping title to "Stopped Playing", 

GreenAsJade:
and require an explanation for "stalling"

... here's a novel idea:  let's make the title for "my opponent stopped playing" say "stopped playing" 

Fixes ongoing stream of escaping reported as stalling (which is a hassle for CM's, because they can only vote for stalling outcomes in this case).

## Proposed Changes
 - Change wording of "escaping" to "Stopped Playing"
 - Require a description for stalling (so folk have to be more sure that this is what they want to report, and they hopefully tell us where to look)

### DIFF
--- a/src/components/Report/Report.tsx
+++ b/src/components/Report/Report.tsx
@@ -54,10 +54,10 @@ interface ReportProperties {
 export const report_categories: ReportDescription[] = [
     {
         type: "escaping",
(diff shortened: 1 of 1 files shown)


anoek:
Good change I think, thanks!

PR #2623, opened 2024-03-13
Modified rankString() so as to produce more understandable decimal ra…

nicolasrochette:
Note: This is small code-wise but cosmetically impacts many/most users.

Background discussion for this proposal is at https://forums.online-go.com/t/what-rating-value-can-be-considered-the-middle-of-the-rank/47569. My TL:DR is that for kyus the currently displayed decimal ranks (e.g., 3.2k) are very confusing and that a more understandable system would be welcome.

The main point is that the decimal rank should always transparently match the integer rank. The decimal rank of a 3k player should be between 3.0k and 3.9k (with 3.0k being the strong end of 3k and the closest to becoming 2.9k/2k).

This is not the case in the current system, hence the confusion. A weak 3k is 3.0k but a strong 3k is actually displayed as 2.1k, and perhaps even more confusingly a strong 1k is displayed as 0.1k. The 3.2k above indicates a 4-kyu player. The issue is with kyu ranks only: dan ranks already have a custom adjustment so that's it's not possible to be say 0.4d. Essentially, the proposed commit applies this already present custom adjustment to kyu ranks also.

Now, those ranks are not wrong per se, as there's no standard, and there may be room to apply the "it's not a bug, it's a feature" logic. Though I find it hard to justify adjusting the dan ranks but not the kyu ones. And obvisouly, I believe the proposed change would be worth it despite the initial impact on users.

~~As for the changes themselves: the core of the commit is to avoid the `floor()`/`ceil()` shenanigans and the related issues in by inserting the expression `1 + abs(r - 30)` where `r` is the OGS rank (where 0 is the minimal possible go rank of 30k, and 30 is the boundary between kyu and dan, so that between 29 and 30 is 1k and between 30 and 31 is 1d). This is all in `rankString()` in `rank_utils.ts`~~

The commit doesn't affect integer ranks.

### DIFF
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -152,7 +152,7 @@ test("getUserRating", () => {
             professional: false,
             provisional: false,
             rank: 23,
-            rank_deviation_labels: ["8.1k", "6.2k"],
+            rank_deviation_labels: ["8.2k", "6.2k"],
             rank_label: "7k",
             rating: 1465,
             unset: false,
--- a/src/lib/rank_utils.test.ts
+++ b/src/lib/rank_utils.test.ts
@@ -147,12 +147,12 @@ test("getUserRating", () => {
             bounded_rank: 23,
             bounded_rank_label: "7k",
             deviation: 62,
-            partial_bounded_rank_label: "6.3k",
-            partial_rank_label: "6.3k",
+            partial_bounded_rank_label: "7.2k",
+            partial_rank_label: "7.2k",
             professional: false,
             provisional: false,
             rank: 23,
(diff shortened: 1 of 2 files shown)


benjaminpjones:
Feel free to modify tests if the change in behavior is intentional :) You can run the tests locally with `yarn test`

Although, players may not be too happy to suddenly drop one stone in their display rank.  Did this solution have some support in the thread you linked?

anoek:
I think I like this on the surface. I would like to see some community discussion and buy-in before merging this though.

anoek:
Yep: https://forums.online-go.com/c/ogs-development/20

PR #2630, opened 2024-03-20
Cm basic activity graph

GreenAsJade:
## Proposed Changes

  -  Put a plot by day of votes that a CM made that were in consensus (which is what we hope they do) on their Activity Card.


anoek:
If you fork from `devel` you won't get the repeated commits coming in from other feature branches FYI :)

PR #2635, opened 2024-03-22
Cm breakdown of vote outcomes

GreenAsJade:
Fixes wondering how we voted vs our peers in different report

## Proposed Changes

  - Show a pie of each report type


anoek:
Very cool, thanks for all work on the CM stuff @GreenAsJade !

PR #2647, opened 2024-04-01
Show headers even if they can't be used for sorting

dexonsmith:
Headers that couldn't be used for sorting got dropped in 4e677869059e414c3c7c368d4f2997f2ac7c7f92. Put them back.


anoek:
Whoops, definitely an oversight - thanks for finding and fixing!

PR #2648, opened 2024-04-01
Show minimum players to start tournaments

dexonsmith:
Only affects display of tournaments that haven't started.

Maybe would be better to only show the minimum if it hasn't been reached...

### DIFF
--- a/src/views/TournamentList/TournamentList.tsx
+++ b/src/views/TournamentList/TournamentList.tsx
@@ -387,7 +387,10 @@ export function TournamentList(props: TournamentListProperties) {
                     {
                         header: _("Players"),
                         className: "nobr",
-                        render: (tournament) => tournament.player_count,
+                        render: (tournament) =>
+                            tournament.started
+                                ? tournament.player_count
+                                : `${tournament.player_count}/${tournament.players_start}`,
                     },
                     {
                         header: _("Ranks"),


dexonsmith:
old:
Image

new:
Image 2

Note the "73/4" on the first tournament. Would it be better to show "73", since the minimum player count of "4" has been reached?

anoek:
I think this is an improvement, the only thing giving me pause is it potentially incorrectly conveying that the minimum is actually the maximum number of players in a tournament. I'm not sure that that's that big of a concern or anything, but I thought I'd mention it in case you or anyone else has any thoughts on presentation that might avoid that potential confusion.

dexonsmith:
> I think this is an improvement, the only thing giving me pause is it potentially incorrectly conveying that the minimum is actually the maximum number of players in a tournament. I'm not sure that that's that big of a concern or anything, but I thought I'd mention it in case you or anyone else has any thoughts on presentation that might avoid that potential confusion.

I agree the display is ambiguous in that case, but I think it's the minimum that's most relevant for figuring out if the tournament will actually happen. But maybe I'm biased.

One option for showing both would be "<count>/<min>-<max>" when max is different from min. It's a bit awkward because order of operations makes my eyes group the `-` and `/` incorrectly though.

> Note the "73/4" on the first tournament. Would it be better to show "73", since the minimum player count of "4" has been reached?

Thoughts on this? It would become "73/4-100" in the above case.

dexonsmith:
> > Note the "73/4" on the first tournament. Would it be better to show "73", since the minimum player count of "4" has been reached?
> 
> Thoughts on this? It would become "73/4-100" in the above case.

Another option: `..` instead of `-`. So this would be `73/4..100`.

dexonsmith:
Another option: use an en-dash `–` instead of hyphen `-`. En-dash is the typographically correct way to do ranges, and maybe doesn't parse as much like a minus sign. Options so far:
- `73/4`
- `73` (but `3/4` if not full)
- `73/4-100`
- `73/4..100`
- `73/4–100`

dexonsmith:
(I can ask for input on the forums tonight or tomorrow if you don't have a strong enough opinion... just let me know...)

anoek:
Yeah agreed. Let's stick with what you have and leave off the max, if it's actually confusing we can revisit. I do like the idea of omitting the minimum if you're passed it, and maybe that conveys enough information once you're used to it anyways.

So,

2/4 like you have when you're less than the minimum (like you have it)
73 instead of 73/4 when you've exceeded the minimum (like you mentioned in the OP)

dexonsmith:
836f85cff35cb7c13620559828528217744179f9:

Image 3

BHydden:
>  73/4–100

For future reference, if it is needed I like endash best out of the provided options to also show max

PR #2649, opened 2024-04-02
Share code between MyTournaments and TournamentList

dexonsmith:
Add TournamentListImpl to share code between MyTournaments and TournamentList, and take the best of both.

Improvements for TournamentList:

- Show actual tournament start times if started.
- Show tournemtn end times if ended.

Improvement for MyTournaments:

- Show the minimum number of players needed for not-yet-full tournaments to start.

### DIFF
--- a/src/views/TournamentList/TournamentList.tsx
+++ b/src/views/TournamentList/TournamentList.tsx
@@ -171,87 +171,10 @@ export function TournamentListMainView(): JSX.Element {
 
 function MyTournaments(): JSX.Element {
     return (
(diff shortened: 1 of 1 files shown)


dexonsmith:
old My Tournaments:
Image 4

new My Tournaments:
Image 3

old Tournament List:
Image
Note the date from January 1st in that list.

new Tournament List:
Image 2
Note the better precision on times and the fix for the January 1st date.

anoek:
Thanks!

PR #2650, opened 2024-04-02
Change game reviews to use AIReview instead of AIDemoReview

ecyrb:


anoek:
That's awesome that it mostly just works like that, thanks for looking into it!

The only thing that I've found so far is that if you load a review that already has some variations explored, if you click on one of the variations it doesn't trigger any analysis for that move. In the current system it will.

ecyrb:
Apologies, I didn't catch that. I'll give it a look.

anoek:
When it's all ready feel free to switch back to ready for merge, and thanks for working on it!

Issue #1031, opened 2020-01-04
Additional puzzle features


RobertT53:
I would like for the puzzle section to have some additional settings options that would help with study or turn the puzzle section into a game similar to the joseki section.

Puzzle auto advance - Once a puzzle is solved or failed, it auto advances to the next problem.

Randomize puzzle order - Increase puzzle replay-ability by not just memorizing the specific puzzle order.

Running tally of solves vs failures - This would be keeping track of how many success/failures for just this attempt at the collection and could be an overall success/failure rate over all attempts at the puzzle collection.

A timer - either a timer for the overall collection (solved x problems in y time) or a time per problem (solved x number of problems before getting y number of problems incorrect)

anoek:
This issue has been marked stale and will be closed soon without further activity. To keep the issue open, please respond to the comment to keep the discussion going.

PR #2653, opened 2024-04-04
Button for Random Next Puzzle

shinuito:
Related to #1031 

Also requested in

https://forums.online-go.com/t/resource-for-making-custom-puzzles-with-randomised-order/50994?u=shinuito

## Proposed Changes

  - Add a yellow (resuing danger colour) "Random" button beside the next puzzle button, to jump to a random other puzzle in the same collection.
  - Doesn't show up if there's only one puzzle in the collection.
  - If the random puzzle id chosen is the current id, then go to either the puzzle preceding this one if available or the next one in case of it being the first puzzle. (Less likely when there's more puzzles in the collection anyway).
 
image

image

### DIFF
--- a/src/views/Puzzle/Puzzle.tsx
+++ b/src/views/Puzzle/Puzzle.tsx
@@ -907,28 +907,16 @@ export class _Puzzle extends React.Component<PuzzleProperties, PuzzleState> {
         const show_correct = this.state.show_correct;
 
         let next_id = 0;
+        const ids = this.state.puzzle_collection_summary
+            .map((p) => p.id)
+            .filter((id) => id !== puzzle.id);
         let random_id = 0;
-        let temp_id = 0;
+        if (ids.length > 0) {
+            random_id = ids[Math.floor(Math.random() * ids.length)];
+        }
         for (let i = 0; i < this.state.puzzle_collection_summary.length - 1; ++i) {
             if (this.state.puzzle_collection_summary[i].id === puzzle.id) {
                 next_id = this.state.puzzle_collection_summary[i + 1].id;
-            }
-        }
-        if (this.state.puzzle_collection_summary.length > 1) {
-            temp_id = Math.floor(Math.random() * this.state.puzzle_collection_summary.length);
-            random_id = this.state.puzzle_collection_summary[temp_id].id;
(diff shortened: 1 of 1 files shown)


src/views/Puzzle/Puzzle.tsx
```diff
@@ -907,11 +907,30 @@ export class _Puzzle extends React.Component<PuzzleProperties, PuzzleState> {
...
+        let temp_id = 0;
         for (let i = 0; i < this.state.puzzle_collection_summary.length - 1; ++i) {
             if (this.state.puzzle_collection_summary[i].id === puzzle.id) {
                 next_id = this.state.puzzle_collection_summary[i + 1].id;
             }
         }
+        if (this.state.puzzle_collection_summary.length > 1) {
+            temp_id = Math.floor(Math.random() * this.state.puzzle_collection_summary.length);
+            random_id = this.state.puzzle_collection_summary[temp_id].id;
+            if (random_id === puzzle.id) {
```
benjaminpjones, line 920: If we remove the current puzzle *before* rolling the dice, I think the code will become simpler, and we will keep randomness in more cases.

```
const ids = this.state.puzzle_collection_summary
    .map((p) => p.id)
    .filter((id) => id !== puzzle.id);
const random_id = ids[Math.floor(Math.random()*ids.length)];
```

---

Logically, this here is what I am suggesting - instead of:

1) Select a puzzle at random from the whole group
2) If current puzzle was selected:
   a) If first puzzle, go next
   b) If last puzzle, go back. I'm not sure why this has to be handled separately from (c)
   c) otherwise, go back one

We do:

1) Remove the current puzzle from the list
2) Select a puzzle at random

BHydden: Or keep it truly random and occasionally test whether they remember the solution they just found 😅

shinuito: Got a bit busy - updated this now. Yeah it seems like a nice way to do it :)

shinuito: Feel free to have another look also :)

anoek:
Nice addition, thanks!

PR #2659, opened 2024-04-18
Fix move number links in review/demo boards

dexonsmith:
Fixes issue reported in the forums:
https://forums.online-go.com/t/clickable-move-numbers-in-chat/51503

> But with the normal game moves (red arrow), that doesn’t seem to work. It would be very handy if that would work the same way.
> 
> Thanks.
> 
> Example: [FMfirst vs. Holon 4](https://online-go.com/review/1254540)
> 8be3627e2e3946c1b8945368558dcad19fdc86b4

### DIFF
--- a/src/views/Game/GameChat.tsx
+++ b/src/views/Game/GameChat.tsx
@@ -51,7 +51,7 @@ interface ChatLine {
         | protocol.GameChatReviewMessage
         | protocol.GameChatTranslatedMessage;
     date: number;
-    move_number: number;
+    move_number?: number;
     from?: number;
     moves?: string;
     channel: string;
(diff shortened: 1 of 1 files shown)


dexonsmith:
@anoek, I didn't dig up where the `ChatLine` gets created. Maybe it's better to fix it so that `move_number` is always set (not sure if that can be done for historical reviews/demos though).

(I see there are some linter checks failing... I can fix those up...)

dexonsmith:
I got curious. This snippet from `sendChat` in https://github.com/online-go/goban/blob/main/src/GobanCore.ts#L1796 appears to be why `move_number` isn't set for reviews/demos:
```ts
        const msg: any = {
            body: msg_body,
        };

        if (this.game_id) {
            msg["type"] = type;
            msg["game_id"] = this.game_id;
            msg["move_number"] = this.engine.getCurrentMoveNumber();
            this.socket.send("game/chat", msg);
        } else {
            const diff = this.engine.getMoveDiff();
            msg["review_id"] = this.review_id;
            msg["from"] = diff.from;
            msg["moves"] = diff.moves;
            this.socket.send("review/chat", msg);
        }
    }
```

anoek:
Ah thanks for finding and fixing that. That's some pretty old decisions made regarding `move_number` vs `from`+`moves`, it's not entirely clear to me why I left out `move_number` when `from`/`moves` were there, possibly because it wasn't an official move or something, possibly because I just did it at a later time and didn't really think things through. 

I'm not opinionated about it these days, though retroactively updating it for existing chat records would be a notable undertaking so probably isn't worth it.

PR #2661, opened 2024-04-18
Make move number URLs work for reviews and demos

dexonsmith:
Per this forum suggestion: <https://forums.online-go.com/t/clickable-move-numbers-in-chat/51503/5>

> Somewhat related, appending the move number to the url of a review doesn’t work to navigate there like it does in a game. Having to type that isn’t as convenient, but being a real url the plus is it is shareable externally.

Tested with <https://online-go.com/review/1254540/3> (well, via http://localhost:8080).


anoek:
Looks good, merging but you know what they're going to ask for next right? Link to the position in the tree :grin:

PR #2663, opened 2024-04-27
Tell the user how many more characters they have to type

GreenAsJade:
Fixes people choosing a different type  of report that they are allowed to submit

## Proposed Changes

  - Show the remaining character count above the greyed submit button


anoek:
Nice, thanks!

PR #2668, opened 2024-05-04
Modelled the moderator score cheating templates on the newer version approved by the forumites for CMs.

GreenAsJade:
Fixes templates encouraging moderators to use less-good wording

## Proposed Changes

  - Add beginner score cheating template
  - Word both score cheating warning templates based on the CM score cheating template


anoek:
Indeed, thanks for the revision!

PR #2672, opened 2024-05-09
Add `/reload` route to fix the default view for social logins

dexonsmith:
Before this commit: when social logins succeed after Register or SignIn views, they currently show the content from the ObserveGames view instead of from the Overview view. This is because `config.user.anonymous` is true when the Default component is loaded.

Username/password signins don't have this problem because they force a reload by explicitly setting `window.location.pathname`.

This is a speculative fix (I can't test the OAuth stuff), adding a `/reload` route that uses the same trick for social logins.

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -193,12 +193,22 @@ function SettingsRedirect(): JSX.Element {
     return <Navigate to={`/settings/${last_settings_page}`} replace />;
 }
 
+function Reload(): JSX.Element | null {
+    if (window.location.hash && window.location.hash[1] === "/") {
+        window.location.pathname = window.location.hash.substring(1);
+    } else {
+        window.location.pathname = "/";
+    }
+    return null;
+}
+
 export const routes = (
     <Router history={browserHistory}>
         <Main>
             <Routes>
                 <Route path="/sign-in" element={<SignIn />} />
                 <Route path="/register" element={<Register />} />
+                <Route path="/reload" element={<Reload />} />
                 <Route path="/welcome/*" element={<ChallengeLinkLanding />} />
                 <Route path="/appeal/:player_id" element={<Appeal />} />
                 <Route path="/appeal" element={<Appeal />} />
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -193,12 +193,17 @@ function SettingsRedirect(): JSX.Element {
     return <Navigate to={`/settings/${last_settings_page}`} replace />;
 }
 
-function Reload(): JSX.Element | null {
-    if (window.location.hash && window.location.hash[1] === "/") {
-        window.location.pathname = window.location.hash.substring(1);
-    } else {
-        window.location.pathname = "/";
-    }
+function WaitForUser(): JSX.Element | null {
+    data.watch("config.user", (user) => {
+        if (user.anonymous) {
+            return;
+        }
+        if (window.location.hash && window.location.hash[1] === "/") {
+            window.location.pathname = window.location.hash.substring(1);
+        } else {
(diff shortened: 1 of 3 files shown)


dexonsmith:
@anoek, for additional context, if this fix works, it will also make it so that social logins get the rank-chooser screen. The reason they don't right now is that, without a reload, Default defers to ObserveGames instead of Overview.

anoek:
Hmm, I saw it hit the `/reload` and redirect to `/` but it's still loading `ObserveGames`. It's up on beta if you want to check it out.

dexonsmith:
> Hmm, I saw it hit the `/reload` and redirect to `/` but it's still loading `ObserveGames`. It's up on beta if you want to check it out.

Yeah, seeing the same thing. I guess `config.user` still has `anonymous` set when the Default view is rendered after the redirect. I think you may as well revert since it seems like this won't work.

Another idea @GreenAsJade and I had was to add a state variable in the Default view for `anonymous` that updates like this:
```
const [anonymous, setAnonymous] = useState(...);
data.watch("config.user", (user) => {
    if (user.anonymous != anonymous) {
        setAnonymous(user.anonymous);
    }
};
```
I'll send a competing PR for that fix.

dexonsmith:
On the other hand, that could end up looking glitchy. Maybe the `Default` view should just be blocking on something before rendering when we've just come from a social sign-in...

anoek:
> On the other hand, that could end up looking glitchy. Maybe the `Default` view should just be blocking on something before rendering when we've just come from a social sign-in...

Could that `/reload` path wait until `anonymous` is `false` and then redirect? That might do it?

anoek:
Awesome, that seemed to do the trick I'll be merging that in a few unless y'all find something that doesn't work.

I *think* you just won't get redirected back if you fail to log in to whatever social service you're doing, but I'm not entirely sure. I'm not sure how to test that myself, I think I've given permission to login on all my social auth accounts so it doesn't even ask me anymore, it just redirects along the happy path.

dexonsmith:
> Awesome, that seemed to do the trick I'll be merging that in a few unless y'all find something that doesn't work.
> 
> I _think_ you just won't get redirected back if you fail to log in to whatever social service you're doing, but I'm not entirely sure. I'm not sure how to test that myself, I think I've given permission to login on all my social auth accounts so it doesn't even ask me anymore, it just redirects along the happy path.

Let me try to trigger it now that it's live.

dexonsmith:
- Confirmed (like you) that logging in redirects as we hoped.
- Refusing to give credentials sends us to <https://beta.online-go.com/sign-in?login-error>

By the way, I tried sign-in with apple and <https://beta.online-go.com/login/apple/?next=/wait-for-user#> showed me this:
Image 2

Same thing happens with the production server so it seems unrelated. I filed #2673.

dexonsmith:
> Awesome, that seemed to do the trick I'll be merging that in a few unless y'all find something that doesn't work.

Yeah, found something. The "wait-for-user" shows up in the history. So you click "back" it'll just hang there.

anoek:
Whoops, clicked merge before that came through sorry about that

dexonsmith:
> Whoops, clicked merge before that came through sorry about that

Probably an improvement on status quo anyway

dexonsmith:
Not sure how to avoid adding to history in this case.

dexonsmith:
Actually, I think we can just use `navigate()`, since I don't think we need a reload once we have a user...

dexonsmith:
> Actually, I think we can just use `navigate()`, since I don't think we need a reload once we have a user...

#2675 does this.

PR #2675, opened 2024-05-09
Replace '/wait-for-user' in history

dexonsmith:
Rather than directly manipulating window.location.pathname in WaitForUser, switch to useNavigate to avoid leaving `/wait-for-user` on the history stack.

Fixes a corner case in #2672.

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -18,7 +18,13 @@
 /* cspell: words groupadmin cotsen */
 
 import * as React from "react";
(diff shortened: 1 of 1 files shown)


dexonsmith:
I haven't been able to test that the OAuth sign-in still does the right thing for the default view... we should check this out on beta.online-go.com.

anoek:
Up on beta, I think it works (back button heads to sign in, which redirects back to /, so I think it's good?)

anoek:
Awesome, thanks for fixing this one up!

PR #2676, opened 2024-05-09
Fix apple id login url

### LINKED ISSUE #2673: Sign-in with Apple gets a "resource not found"
Tried sign-in with Apple as part of testing #2672. Sends me to <https://online-go.com/login/apple/?next=/>, which shows me this content:
Image 2

anoek:
Should fix #2673

Not sure if this broke somewhere along the way or if it never worked. Either way, been broken for awhile - just goes to show just how not popular login with apple is though..

### DIFF
--- a/src/components/SocialLoginButtons/SocialLoginButtons.tsx
+++ b/src/components/SocialLoginButtons/SocialLoginButtons.tsx
@@ -37,7 +37,7 @@ export function SocialLoginButtons(props: SocialLoginButtonsProps): JSX.Element 
                 <i className="twitter twitter-icon fa fa-twitter" />
                 {_("Sign in with Twitter")}
             </a>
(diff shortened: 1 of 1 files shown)


dexonsmith:
> Not sure if this broke somewhere along the way or if it never worked. Either way, been broken for awhile - just goes to show just how not popular login with apple is though..

FWIW, if it never worked, we might not have data on how popular it is.

If I were a new user and I tried sign-in-with-Apple and it didn't work, I'd either:
- make a username and password (with an autogenerated password from whatever browser I'm using) or
- give up and not use the site.

I probably wouldn't work too hard to report the issue, since I have no investment in the community, and I've either moved on or gotten it working another way (possibly blaming Apple for it not working).

But that's just a guess... maybe there is a category of people who would somehow figure out how to report it.

PR #2681, opened 2024-05-13
Add missing change from d60dc056ff0efb7a0f1f133224610f12e5d11e72

dexonsmith:
I just found an embarrassingly unsaved buffer... the Register view has been going to the now defunct `/reload` instead of `/wait-for-user`.


dexonsmith:
This is a fix-up for #2672.

anoek:
Whoops, double fail I should caught that too - thanks for noticing and fixing!

PR #2684, opened 2024-05-14
cms can see canned messages in mod log 

GreenAsJade:
(factored out ModLog to make this easier)

Fixes CMs not being able to tell if it's a repeat offender

## Proposed Changes

  - Show CM's the "Mod Log" (backend will send filtered version)
  - Factor out ModLog to make this easier

### DIFF
--- a/src/views/User/ModTools.tsx
+++ b/src/views/User/ModTools.tsx
@@ -23,12 +23,11 @@ import { Player } from "Player";
 import { openModal } from "Modal";
 import { NotesModal } from "NotesModal";
 import { _ } from "translate";
-import { Link } from "react-router-dom";
-import { chat_markup } from "Chat";
 import { errorAlerter } from "misc";
 import { put, get } from "requests";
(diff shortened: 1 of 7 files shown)


src/views/User/ModTools.tsx
```diff
@@ -180,66 +179,7 @@ export function ModTools(props: ModToolsProps): JSX.Element | null {
...
-                                        <i> {X.incident_report.moderator_note}</i>
-                                    </div>
-                                )}
-                                <pre>{chat_markup(X.note, undefined, 1024 * 128)}</pre>
-                            </div>
-                        ),
-                    },
-                ]}
-            />
+            <ModLog user_id={props.user_id} />
```
benjaminpjones, line 182: so clean 😍

GreenAsJade: LoL shucks thanks, that made that big ole refactor all worth it :)

PR #2690, opened 2024-05-18
Better canned message format

GreenAsJade:
Fixes canned messages collapsing into one blob of text

## Proposed Changes

  - Format canned messages with `white-space: pre-line`, and adjust the input text to suit.
  
  
Note: this is on top of/inclusive of https://github.com/online-go/online-go.com/pull/2689  so that the new message in that PR gets this treatment as well.

### DIFF
--- a/src/components/AccountWarning/AccountWarning.styl
+++ b/src/components/AccountWarning/AccountWarning.styl
@@ -25,6 +25,11 @@
     z-index: z.account-warning;
     background-color: rgba(0, 0, 0, 0.5);
 }
+
+.canned-message {
+    white-space: pre-line;
+}
+
 .AccountWarning {
     position: fixed;
     top: auto;
@@ -50,7 +55,6 @@
     flex-direction: column;
     justify-content: center;
     align-items: center;
-
 
     .AutoTranslate {
         font-size: 1.5rem;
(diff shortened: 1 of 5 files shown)


anoek:
If we merge this, it's going to obliterate all of the manual translations for those - would it be feasible to just clean the input strings to remove leading/trailing non-new line white space so we don't have to do that?  It's not the end of the world if we can't, but it'd be nicer on our volunteer translators

GreenAsJade:
> If we merge this, it's going to obliterate all of the manual translations for those - would it be feasible to just clean the input strings to remove leading/trailing non-new line white space so we don't have to do that? It's not the end of the world if we can't, but it'd be nicer on our volunteer translators

I wonder if we take the hit now?  The reason is that at the same time I tidied up a few paragraphs, put breaks in better places etc.

Of course, I could do only the "necessary" tidying, and add a clean up step, which will make it more consistent if folk don't properly follow the format in the future as well.

Happy to do whatever...

anoek:
Ok, we'll bite it now, I just want to be thoughtful for those fine folk

GreenAsJade:
Is there a way of communicating with them?   I'd happily say "ooo - sorry, we thought of you, but..."

anoek:
Eh, not really, there are many many that come and go, and it's all managed by the stand alone pootle server

PR #2691, opened 2024-05-20
Show number of rounds in Swiss tournament

dexonsmith:
Once a Swiss tournament starts, it has a fixed number of rounds.  Show them. This addresses one of the [forum complaints][1] about Swiss tournaments.

[1]: https://forums.online-go.com/t/how-many-rounds-should-a-swiss-tournament-have/51168


dexonsmith:
E.g., <https://online-go.com/tournament/96607> looks like this in my testing:

Image

PR #2695, opened 2024-05-27
Add "Copy SGF" Button to Game Dock

pavelwatson:
## Proposed Changes
This pull request adds a new "Copy SGF" button to the Game Dock (side panel of the game page), enhancing the user experience by allowing users to copy the SGF content of a game directly to the clipboard. Below are the key changes and features implemented:

## New Functionality:
1. Added a new button labeled "Copy SGF" to the side panel dock.
2. Implemented functionality to fetch SGF content and copy it to the user's clipboard using the existing sgf_url.
3. Utilized the existing notification system to alert users when the SGF content is successfully copied.

## Component Update:
1. Modified the GameDock.tsx component to include the new "Copy SGF" button.

## Images:
new_button_image
notification-example

### DIFF
--- a/src/views/Game/GameDock.tsx
+++ b/src/views/Game/GameDock.tsx
@@ -129,6 +129,31 @@ export function GameDock({
         ? api1(`reviews/${review_id}/sgf`)
         : null;
 
+    // Function to copy SGF content to clipboard
+    const copySGFToClipboard = () => {
+        fetch(sgf_url)
+            .then((response) => {
+                if (!response.ok) {
+                    throw new Error("Network response was not ok");
+                }
+                return response.text();
+            })
+            .then((sgfData) => {
+                navigator.clipboard
+                    .writeText(sgfData)
+                    .then(() => {
+                        toast(<div>{_("SGF content copied to clipboard!")}</div>, 2000);
+                    })
+                    .catch((err) => {
+                        console.error("Failed to copy text to clipboard", err);
+                    });
+            })
+            .catch((error) => {
+                console.error("Fetch error:", error);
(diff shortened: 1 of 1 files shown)


anoek:
I'm a little hesitant to merge this as is only because I question whether there is enough demand for this functionality to justify putting it on the already crowded dock. I'm wondering if putting this within the share dialog box or something would be more appropriate?

pavelwatson:
> I'm a little hesitant to merge this as is only because I question whether there is enough demand for this functionality to justify putting it on the already crowded dock. I'm wondering if putting this within the share dialog box or something would be more appropriate?

I get your concern. Personally, I think the 'Copy SGF' button would prove it fairly useful. Most Go editors support copy-pasting SGF files directly (pressing Ctrl+V inside an editor). This feature would make it super convenient to just quickly copy the SGF and paste it into any editor without cluttering up your drive with downloaded SGF files. As to your suggestion, having to dive into the share dialog after every single game would just add unnecessary clicks, making the whole addition a bit pointless. Another possible approach could be implementing a CTRL+C hotkey on the end game page that copies that SGF to clipboard. I can work on that if you like the idea.

anoek:
This is probably worth some forum discussion on the matter. I'll mark this as draft for now, feel free to open up a thread on the forums and we can proceed based on the discussions had there.

benjaminpjones:
Happy to comment the same in a Forum discussion, but I personally feel this belongs in the Download SGF modal.

yuki2006:
I also suggested this on a forum a while ago.
https://forums.online-go.com/t/display-sgf-as-on-screen-text/51594/16

I would definitely like to use this.

PR #2699, opened 2024-05-31
Defer warning till end of live game

GreenAsJade:
Fixes players having to wait the dismiss-delay during a live game.

## Proposed Changes

  - Have AccountWarning check if they're on the Game page, and if so then wait till a live game is not in progress to show the warning, by polling `gobal_goban`.

### DIFF
--- a/src/components/AccountWarning/AccountWarning.tsx
+++ b/src/components/AccountWarning/AccountWarning.tsx
@@ -25,7 +25,6 @@ import { get, patch } from "requests";
 import { useUser } from "hooks";
 import { AutoTranslate } from "AutoTranslate";
 import { useLocation } from "react-router";
-
 import { CANNED_MESSAGES } from "./CannedMessages";
 
 const BUTTON_COUNTDOWN_TIME = 10000; // ms;
@@ -34,6 +33,7 @@ export function AccountWarning() {
     const user = useUser();
     const location = useLocation();
     const [warning, setWarning] = React.useState<rest_api.warnings.Warning | null>(null);
+    const [liveGameInProgress, setLiveGameInProgress] = React.useState<boolean>(false);
 
     React.useEffect(() => {
         if (user && !user.anonymous && user.has_active_warning_flag) {
@@ -53,6 +53,25 @@ export function AccountWarning() {
             setWarning(null);
         }
     }, [user, user?.has_active_warning_flag]);
+
+    const mainGoban = (window as any)["global_goban"];
+
+    if (location.pathname.indexOf("game/") > 0) {
+        // On the Game page, we need to delay displaying the warning until the game is over
+        if (liveGameInProgress) {
+            return null;
+        } else if (mainGoban?.engine.phase === "play" && mainGoban.engine.game_id) {
+            setLiveGameInProgress(true);
+            const checking = setInterval(() => {
+                if (mainGoban?.engine.phase !== "play" || !mainGoban?.engine.game_id) {
+                    setLiveGameInProgress(false);
+                    clearInterval(checking);
+                }
+            }, 1000);
+
+            return null;
+        }
+    }
 
     if (!user || user.anonymous || !user.has_active_warning_flag) {
         return null;
@@ -79,11 +98,7 @@ export function AccountWarning() {
 
     const MessageRenderer = Renderers[warning.severity];
 
-    return (
-        <>
-            <MessageRenderer warning={warning} accept={ok} />
-        </>
-    );
+    return <MessageRenderer warning={warning} accept={ok} />;
 }
 
 // Support warnings that carry messages either as a reference to a a canned message, or explicit text...
@@ -92,7 +107,6 @@ interface MessageTextRenderProps {
     warning: rest_api.warnings.Warning;
 }
 function MessageTextRender(props: MessageTextRenderProps): JSX.Element {
-    console.log("rendering", props);
     if (props.warning.message_id) {
         return (
             <div className="canned-message">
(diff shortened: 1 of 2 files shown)


src/components/AccountWarning/AccountWarning.tsx
```diff
@@ -146,7 +160,7 @@ function WarningModal(props: WarningModalProps): JSX.Element {
                 if (Date.now() - now > BUTTON_COUNTDOWN_TIME) {
                     clearInterval(interval);
                 }
-            }, 100);
+            }, 1000);
         }
```
benjaminpjones, line 164: I realize this this timer wasn't added in this PR, but you'll almost always want a cleanup function in `useEffect()`.  Otherwise, if props.warning changes, you'll have multiple timers running at the same time that interfere with eachother.

src/components/AccountWarning/AccountWarning.tsx
```diff
@@ -54,6 +54,25 @@ export function AccountWarning() {
...
     }, [user, user?.has_active_warning_flag]);
 
+    const mainGoban = (window as any)["global_goban"];
+
+    if (location.pathname.indexOf("game/") > 0) {
+        // On the Game page, we need to delay displaying the warning until the game is over
+        if (liveGameInProgress) {
+            return null;
+        } else if (mainGoban?.engine.phase === "play" && mainGoban.engine.game_id) {
+            setLiveGameInProgress(true);
```
benjaminpjones, line 64: Should we check whether it's a correspondence game?

src/components/AccountWarning/AccountWarning.tsx
```diff
@@ -54,6 +54,25 @@ export function AccountWarning() {
         }
     }, [user, user?.has_active_warning_flag]);
 
+    const mainGoban = (window as any)["global_goban"];
+
+    if (location.pathname.indexOf("game/") > 0) {
+        // On the Game page, we need to delay displaying the warning until the game is over
+        if (liveGameInProgress) {
+            return null;
+        } else if (mainGoban?.engine.phase === "play" && mainGoban.engine.game_id) {
```
benjaminpjones, line 63: Should this be `phase !== "finished"`?  Do you want the message to show up during `"stone removal"`?

src/components/AccountWarning/AccountWarning.tsx
```diff
@@ -54,6 +54,25 @@ export function AccountWarning() {
         }
     }, [user, user?.has_active_warning_flag]);
 
+    const mainGoban = (window as any)["global_goban"];
```
benjaminpjones, line 57:
```suggestion
    const mainGoban: GobanCore? = (window as any)["global_goban"];
```

src/components/AccountWarning/AccountWarning.tsx
```diff
@@ -54,6 +54,25 @@ export function AccountWarning() {
...
+
+    if (location.pathname.indexOf("game/") > 0) {
+        // On the Game page, we need to delay displaying the warning until the game is over
+        if (liveGameInProgress) {
+            return null;
+        } else if (mainGoban?.engine.phase === "play" && mainGoban.engine.game_id) {
+            setLiveGameInProgress(true);
+            const checking = setInterval(() => {
```
benjaminpjones, line 65: I think this works, but generally more idiomatic to call `setInterval()` inside a `useEffect`.

See Dan Abramov's comment on this [SO post](https://stackoverflow.com/a/64274431/5001502):

> Note that the correct way to do this is still `useEffect`. I'm clarifying for the next reader who might assume that this is the correct way to do intervals (it's not).

src/components/AccountWarning/AccountWarning.tsx
GreenAsJade: Doh, I swear I had the correspondence check, must have lost it along the way.   I guess we could tolerate observers not being warned 😝    But yeah not the intent either.

src/components/AccountWarning/AccountWarning.tsx
GreenAsJade: I pondered that.    I felt being specific was safer, at the "expense" of showing in stone removal.   I don't know what other phases are possible (it's not typed, right?)  so allowing it to block on some phase like "paused" for example seems "not robust".    And ... stone removal is not time-pressured.  That was my thinking.

src/components/AccountWarning/AccountWarning.tsx
GreenAsJade: 😝👍

src/components/AccountWarning/AccountWarning.tsx
GreenAsJade: 👍 I'll have to read that post about "why  is this so".   But far be it from me to buck the idiom!

src/components/AccountWarning/AccountWarning.tsx
benjaminpjones: > it's not typed, right?

https://github.com/online-go/goban/blob/9d83b03f00090f1bbf73b23577ede26225d5122c/src/GoEngine.ts#L43

> And ... stone removal is not time-pressured.

There is a timer on stone removal, but I agree it's not the same at in game.  Not including it seems reasonable, I just wanted to check!

src/lib/hooks.ts
```diff
@@ -64,3 +65,7 @@ export function useRefresh(): () => void {
     const [, refresh] = React.useState(0);
     return React.useCallback(() => refresh(() => Math.random()), [refresh]);
 }
+
+export function useMainGoban(): GobanCore | null {
```
benjaminpjones, line 69: This isn't really a hook yet.  As in, it won't trigger an update when global_goban changes.  Notice that the other hooks in this file are composing other hooks.

https://react.dev/learn/reusing-logic-with-custom-hooks

PR #2701, opened 2024-06-03
Include opponent's name in turn notification

LForchini:
Fixes #

## Proposed Changes

  - Change the format of the notification when it's your move in correspondence games to remove the game id and instead include the username of the opponent.

## Motivation
I feel that the game id does not have any use in the notification, whereas I think the opponents name would be more useful.

I am new to this codebase so any advice would be useful, specifically if there's an established method for finding the name of the opponent.

Screenshot from 2024-06-03 09-47-38


anoek:
Thanks!

PR #2704, opened 2024-06-03
Added a slack invite link to dev server startup message.

GreenAsJade:
Fixes me having to keep an eye on who's committing and find a way to contact them

## Proposed Changes

  - Put a slack invite link into dev server startup.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -196,11 +196,12 @@ function dev_server(done) {
     dev_server.use(body_parser.text());
 
     http.createServer(dev_server).listen(port, null, function () {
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Note: I'm not sure if the permalink will work or not, it saved successfully, and worked for me, but the message on the dialog is contradictory:

Screenshot 2024-06-04 at 8 26 35 AM (2)

I'm thinking "I'll just keep an eye on it, it's worth a shot".

But if this all seems like a  bad idea feel free to reject, I won't be offended 😝

anoek:
good idea :)

PR #2713, opened 2024-06-11
Add pass to conditional planner

### LINKED ISSUE #251: Cannot pass in Conditional Move Planner
Being able to respond to a pass automatically would be nice, and being able to respond with a pass would also be nice.

LForchini:
Fixes #251 

## Proposed Changes
  - Add pass to the conditional move planner

Is there a nicer way to localise the pass message than the check I used? I considered having the prettyCoord function return undefined when given invalid moves instead of the string `pass`.

I'd also like some feedback on the UI here:
sc

### DIFF
--- a/src/views/Game/ConditionalMoveTreeDisplay.tsx
+++ b/src/views/Game/ConditionalMoveTreeDisplay.tsx
@@ -19,6 +19,7 @@ import * as React from "react";
 import { GoConditionalMove } from "goban";
 import { useGoban } from "./goban_context";
 import { generateGobanHook } from "./GameHooks";
+import { _ } from "translate";
 
 interface ConditionalMoveTreeDisplayProps {
     tree: GoConditionalMove;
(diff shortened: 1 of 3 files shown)


benjaminpjones:
Looks good to me!

> I considered having the `prettyCoord` function return `undefined` when given invalid moves instead of the string pass.

I wouldn't change it, or at least I think the current behavior (and the way you handle it) is okay.  It doesn't make a huge difference what we return in the "invalid" case, but at least now it prints out something reasonable in cases where we haven't explicitly handled it.  E.g. the Joseki tool

Screenshot 2024-06-11 at 5 56 50 PM



> I'd also like some feedback on the UI here

This one's tough, I can't think of anything better than this.  Maybe other people will have feedback...

BHydden:
I think it looks tidy. OGS is renowned for having noone on the team especially gifted at UI/UX 😅 so tidy and consistent seem adequate goals to aim for...

anoek:
Thanks!

PR #2714, opened 2024-06-12
Allow CMs to cancel their own reports

LForchini:
## Proposed Changes

  - Allows community moderators to cancel their own reports in the report view

I still have to test and style the changes

### DIFF
--- a/src/components/IncidentReportTracker/IncidentReportTracker.tsx
+++ b/src/components/IncidentReportTracker/IncidentReportTracker.tsx
@@ -245,7 +245,7 @@ export function IncidentReportTracker(): JSX.Element | null {
                                 )}
                             </div>
                         )}
-                        {filtered_reports.map((report, index) => (
+                        {filtered_reports.map((report: Report, index) => (
                             <div className="incident" key={report.id}>
                                 <div className="report-header">
                                     <div
--- a/src/lib/report_util.ts
+++ b/src/lib/report_util.ts
@@ -36,8 +36,8 @@ export interface Report {
     retyped: boolean;
     source: string;
     report_type: ReportType;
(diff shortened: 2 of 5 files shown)


LForchini:
I've styled the action buttons somewhat. The red button doesn't appear unless the report was created by the user; should I disable it instead of hiding it?
 
sc

GreenAsJade:
I prefer it to be hidden in this instance, if the report does not belong to the user, because "does not belong to the user" is by far the prevalent case, so it's clutter to have a disabled button hanging around.

There's no confirmation dialog on a normal user's "Cancel" report function, so I guess that applies here too.   I imagine the thinking is that it's not a lot lost if it's an accident :)

I have a feeling that we should be calling this "Cancel", because the intent is that this is the same functionality as the "Cancel" button that normal users have.   The result of pressing that Cancel button is that the report is set to the closed state _and_ has an annotation "closed by user".

I have an urge to move the button up underneath the reporting user's name.   I can't quite say why, other than it "feels more like the normal user case", and also emphasises that the reason it's here is because this is the reporting user's report.

It could be on the same row as "Actions"?

LForchini:
> I prefer the button to be hidden if the report does not belong to the user, because "the report does not belong to the user" is by far the prevalent case, so it's clutter to have a disabled button hanging around :)

I was thinking the same, so that's the current behavior.

> I have an urge to move the button up underneath the reporting user's name. I can't quite say why, other than it "feels more like the normal user case", and also emphasises that the reason it's here is because this is the reporting user's report.
> 
> It could be on the same row as "Actions"?

I personally prefer an interactivity area over scattering buttons all over the UI :sweat_smile:

anoek:
Thanks!

PR #2721, opened 2024-06-19
Add GoTV component

RubyMineshaft:
Depends on https://github.com/online-go/ogs/pull/1966

## Proposed Changes

- Adds GoTV component
- If there are active Go twitch streams an indicator will appear in the nav bar:
    Screenshot 2024-06-19 at 3 35 10 PM

- Clicking the indicator will open the GoTV page
Screenshot 2024-06-19 at 3 05 37 PM

- Streams can be filtered by language

- Each stream has an OGS chat panel.
 I initially had trouble embedding twitch chat, but now that I've figured out what was going on with the selector specificity in the stylesheet, I'm thinking it may have something to do with that. I'll give that another shot. 
Screenshot 2024-06-19 at 3 54 35 PM

- Stream list and chat panels are collapsible

Screenshot 2024-06-19 at 3 07 23 PM
Screenshot 2024-06-19 at 4 01 34 PM


BHydden:
Should there also be a link to this page under the "Watch" menu?

BHydden:
Also, I wonder if maybe it makes sense to have a general chat room on the page even when no streams are happening? 🤔

RubyMineshaft:
> Should there also be a link to this page under the "Watch" menu?

I actually just put this on my todo list about an hour ago :) 

> Also, I wonder if maybe it makes sense to have a general chat room on the page even when no streams are happening? 🤔

This is probably not a bad idea. I was thinking of making a YouTube playlist for Go educational content and play randomly from that when there's no stream available. Still thinking on that one.  I should at least display some sort of message when there's no stream available for now though, so thanks for that reminder!

BHydden:
It might be out of scope, but could the TV page as one of its "streams" show the strongest live game currently happening?

LForchini:
Hello, this looks pretty cool, but I was wondering what the use-case is? I like the TV icon that displays how many streams are active, but it could also just link to twitch instead of OGS's re-implementation of twitch. I suppose it's useful if someone doesn't have a twitch account? 
I really like the idea of incorporating more elements of the online go community more broadly!

anoek:
Well, if we just dumped folks on to twitch they get bombarded with a bunch of other stuff that's not really Go related, with this it's a curated list of Go content that unifies the Go category along with select streams that might be announced here but listed on Twitch under different larger categories (Board Games for instance), and nothing else. So, a much more focused view of Go content I'd say.

LForchini:
Alright, I can understand that, especially if it goes beyond just the twitch go category. I'd like to second the idea of watching the strongest ongoing game, I think that would be quite valuable.

BHydden:
Batts is live rn but I don't see his stream (or any yet) listed in the uffizzi build, just FYI

anoek:
> Batts is live rn but I don't see his stream (or any yet) listed in the uffizzi build, just FYI

The backend isn't up on beta yet so that's not expected to work just yet.

BHydden:
> > Batts is live rn but I don't see his stream (or any yet) listed in the uffizzi build, just FYI
> 
> The backend isn't up on beta yet so that's not expected to work just yet.

Cool I thought it might have been something simple like that, thanks for clarifying

anoek:
FYI I've pushed the backend up to the beta site so you can now check it out via uffizzi

BHydden:
While testing this feature my phone had a kernel crash and the whole device restarted. Hopefully unrelated and have reached out to the OS team to enquire about it, but thought I would let you know here too since it has never happened before and I was actively testing this feature when it happened.

anoek:
image

anoek:
(In all seriousness though, there is nothing that a website should be able to do that causes a kernel panic or whatever happened there, that's a whole stack of fail that is out of our control)

shinuito:
I think it looks great on beta right now :) Looking at the NAGF pro qualifiers on beta :)

shinuito:
If it is the case that in a future update a game or demo board could be shown in a similar way, I could imagine say the top ranked live game in progress, or specifically flagged demo boards or games like some pro games being played on OGS might be popular to watch also :)

Anything in particular to test out on beta? I suppose I could log into a few accounts and test out the chat - will do that later.

RubyMineshaft:
Twitch chat should be working now for all themes. I also just pushed up an update to make the layout responsive, so it shouldn't be too crazy on mobile any more. :)

> If it is the case that in a future update a game or demo board could be shown in a similar way, I could imagine say the top ranked live game in progress, or specifically flagged demo boards or games like some pro games being played on OGS might be popular to watch also :)

I still have a few things I need to get around to before I can play around with this, but that's what I'm kind of thinking too.  Top ranked games for all three sizes and maybe throw some top ranked blitz games in too. Pro games didn't cross my mind, but it's a good idea

shinuito:
>Pro games didn't cross my mind, but it's a good idea

I think normally they would also be the top ranked, but I have seen situations where they show up as [?] in the thumbnails. Not sure if that would affect it, or whether top ranked would be calculated as a kind of average between the two players etc.

The layout on mobile looks great! One possible issue that I’m not if you can actually fix, with the iframe for the chat, I can’t see the button to accept/reject the cookies popup.

image

Maybe if the iframe/chat window can be scrollable it’ll help with that?

BHydden:
> normally [pros] would also be the top ranked

Recognised pro accounts on OGS have their games listed at the top of the "Watch" page, but that doesn't help for relays of IRL tournaments on demo boards, which I would definitely want to see on the GoTV page, if not initially, at least at some point.

BHydden:
This is looking super sleek ❤️ love it! Huge kudos!

PR #2726, opened 2024-06-23
No escaping reports for resignations

GreenAsJade:
Fixes a stream of annoying escaping reports when the person resigned.

## Proposed Changes

  - Check if the accused resigned, and if so don't accept an escaping report

### DIFF
--- a/src/components/Report/Report.tsx
+++ b/src/components/Report/Report.tsx
@@ -79,7 +79,12 @@ function checkEscapingForResignation(
 ): Promise<string | null> {
     return get(`/termination-api/game/${game_id}`).then((gamedata: Gamedata) => {
         if (gamedata?.outcome.includes("Resignation") && gamedata.winner !== reported_user_id) {
-            return "That player resigned, so 'stopped playing' doesn't make sense. Do you mean 'sandbagging'?";
+            return pgettext(
+                "A message when trying to create a report that doesn't make sense",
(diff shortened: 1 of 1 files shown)


src/components/Report/Report.tsx
```diff
@@ -65,6 +67,25 @@ interface ReportProperties {
...
+    winner: number;
+};
+
+function checkEscapingForResignation(
+    game_id?: number,
+    reported_user_id?: number,
+): Promise<string | null> {
+    return get(`/termination-api/game/${game_id}`).then((gamedata: Gamedata) => {
+        if (gamedata?.outcome.includes("Resignation") && gamedata.winner !== reported_user_id) {
+            return "That player resigned, so 'stopped playing' doesn't make sense. Do you mean 'sandbagging'?";
```
anoek, line 82: Missing a `_(..)`

But also, would it be better to just leave it as a statement and not a sort of hint that they should just file their grief as a sandbagging report?

GreenAsJade: Doh, oops!   😜

I put the bit about sandbagging, because that has often seemed to be what it is, but you're probably right, I'll get rid of that with the `_()`

anoek: Either way about the sandbagging, I trust your intuition more than my own on these things these days.

PR #2728, opened 2024-06-24
Translation for settings switch for SVG renderer

GreenAsJade:
Fixes untranslated label

## Proposed Changes

  - `_()`
  
Note: this may have been left untranslated due to being transitory?


GreenAsJade:
I turned on Debug language, to see if the multiline resignation explanation looked like it would translate OK, and spotted this in passing :)

anoek:
> Note: this may have been left untranslated due to being transitory?

This is exactly why, after the scoring stuff is done I plan on cycling back to the SVG stuff and getting it fully released, probably this week if I can swing it. So, on account of that, gonna close and keep it untranslated to avoid the churn.

PR #2732, opened 2024-06-25
GoTV Updates

RubyMineshaft:
## Proposed Changes

  - Default to Twitch chat, but remember which chat tab was active
  - Remember if chat pane was expanded or collapsed
  - Add GoTV Preferences to Settings menu, with:
    - Setting to hide active streams indicator
    - Setting for disabling autoplay on load
    - Setting to filter mature streams, default: filtered
  - Clicking TV indicator while on GoTV page will return user to previous location
  - Immediately show indicator on site load if streams are available, rather than waiting for next push from backend
  - Other visual bug fixes


anoek:
Thanks for the quick improvements!

PR #2737, opened 2024-06-29
GoTV Updates

RubyMineshaft:
## Proposed Changes
  - Move language filter to GoTV preferences
  - Make language filter multi-select and selections persistent 
  - Add quick access to preferences through gear icon in the list pane
  - Add a `StreamManager` singleton class to handle stream fetches and updates


src/views/GoTV/StreamManager.ts
```diff
@@ -0,0 +1,104 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { get } from "requests";
+import { Stream } from "GoTV";
+import * as preferences from "preferences";
+import { EventEmitter } from "events";
+import { socket } from "sockets";
+
+class StreamManager extends EventEmitter {
```
anoek, line 24: We can make EventEmitter type safe like this:

```
interface Events {
     update: (streams: Stream[]) => void;
}

class StreamManager extends EventEmitter<Events> {
```

That way, if you're using it and you do something like `stream_manager.on("uplate", () => {})` it'll complain at you and let you know you meant `"update"` :)

src/views/GoTV/StreamManager.ts
```diff
@@ -0,0 +1,104 @@
...
+        socket.on("connect", () => {
+            socket.send("ui-pushes/subscribe", { channel: "gotv" });
+        });
+    }
+
+    public getStreams(): Stream[] {
+        return this.streams;
+    }
+
+    public subscribe(callback: (streams: Stream[]) => void) {
```
anoek, line 95: Elsewhere in the code we just let folks use the event emitter interface directly, so for consistency let's get rid of the `subscribe` / `unsubscribe` functions and have folks just do `streamManager.on("update", (streams) => {...})`

src/views/GoTV/StreamManager.ts
```diff
@@ -0,0 +1,104 @@
...
+    private setupWebSocket() {
+        socket.on("ui-push", (msg) => {
+            if (msg.event === "update_streams") {
+                const updatedStreams = JSON.parse(msg.data);
+                this.streams = this.filterStreams(updatedStreams);
+                this.emit("update", this.streams);
+            }
+        });
+
+        socket.on("connect", () => {
```
anoek, line 86: This is *probably* fine stand alone simply because we construct this right away and so it should always get called before the socket is actually connected. If that ever changes, you'll want to have this *and* something like `if (socket.connected) { socket.send(...) }`

src/views/GoTV/StreamManager.ts
RubyMineshaft: I tried that and Typescript yelled at me:

```
Type 'StreamEvents' does not satisfy the constraint 'EventMap<StreamEvents>'.
  Type 'StreamEvents' is not assignable to type 'Record<"update", any[]>'.
    Types of property 'update' are incompatible.
      Type '(streams: Stream[]) => void' is not assignable to type 'any[]'.ts(2344)
```

RubyMineshaft: Ah, looks like I need `TypedEventEmitter<>` 👌🏻👌🏻

anoek: Pure `EventEmitter<Events>` is preferred over `TypedEventEmitter`, typed event emitter was our own creation prior to the eventemitter3 library adding types, so now new code should just use that, and over time we'll refactor and remove the old `TypedEventEmitter` usage

anoek:
Thanks!

PR #2742, opened 2024-06-30
bugfix: don't set isLoading to true after fetching empty array of streams

RubyMineshaft:
Fixes bug introduced in most recent GoTV update where fetching an empty array of streams causes the loading message to display forever


anoek:
Thanks for the quick fix!

PR #2745, opened 2024-07-07
Better group tournament presentation

GreenAsJade:
Fixes dubious layout, and not being able to clean up "tournament records"

## Proposed Changes

  - Put Active Tournaments at the top of the Card
  - Allow moderators and admins to delete tournament records
  - Put non-render functions above the render routine, so that "render is last" in the file.


GreenAsJade:
Ironically, I think the better thing to do is remove the TournamentRecords feature, so I'll set about that.

GreenAsJade:
Blow me down, I stumbled almost immediately into a Group using Tournament Record ... so maybe we implement this as-is eh!

PR #2749, opened 2024-07-09
Reformatting automatch finder box

LForchini:
[Based on this forum mock-up](https://forums.online-go.com/t/revisiting-automatch-time-settings-data-backed-proposal-for-new-automatch-settings-on-ogs/52303/22?u=elforo)
This PR depends on [these changes in the goban repo](https://github.com/online-go/goban/pull/170)

## Proposed Changes

  - Expand front-end for a potential `rapid` automatch speed
  - Enable functionality for dynamic-ish speed displays (the text under 'Blitz', 'Rapid', 'Live'). This could change depending on byo-yomi/fischer options in the settings menu
  - Add a (current disabled) 'Play a Friend' button. I haven't gotten round to the functionality yet, but I'll work on that soon.

sc

Feedback is very much appreciated, even on small styling issues :sweat_smile:
Also I've probably missed out something obvious in adding the `rapid` time setting, so I would appreciate a double-check!

### DIFF
--- a/src/components/AutomatchSettings/AutomatchSettings.tsx
+++ b/src/components/AutomatchSettings/AutomatchSettings.tsx
@@ -162,6 +162,8 @@ export class AutomatchSettings extends Modal<
         switch (this.state.tab) {
             case "blitz":
                 return dup(this.state.blitz_settings);
+            case "rapid":
+                return dup(this.state.rapid_settings);
             case "live":
                 return dup(this.state.live_settings);
             case "correspondence":
@@ -175,6 +177,10 @@ export class AutomatchSettings extends Modal<
             case "blitz":
                 data.set("automatch.blitz", settings);
                 this.setState({ blitz_settings: settings });
+                break;
+            case "rapid":
+                data.set("automatch.rapid", settings);
+                this.setState({ rapid_settings: settings });
                 break;
             case "live":
                 data.set("automatch.live", settings);
@@ -260,6 +266,12 @@ export class AutomatchSettings extends Modal<
                             onClick={() => this.setTab("blitz")}
                         >
                             {_("Blitz")}
+                        </button>
+                        <button
+                            className={this.state.tab === "rapid" ? "primary active sm" : "sm"}
+                            onClick={() => this.setTab("rapid")}
+                        >
+                            {_("Rapid")}
                         </button>
                         <button
                             className={this.state.tab === "live" ? "primary active sm" : "sm"}
--- a/src/components/AutomatchSettings/AutomatchSettings.tsx
+++ b/src/components/AutomatchSettings/AutomatchSettings.tsx
@@ -32,6 +32,7 @@ export type AutomatchPreferences = AutomatchPreferencesBase & { size_options: Si
 interface AutomatchSettingsState {
     tab: Speed;
     blitz_settings: AutomatchPreferences;
+    rapid_settings: AutomatchPreferences;
     live_settings: AutomatchPreferences;
     correspondence_settings: AutomatchPreferences;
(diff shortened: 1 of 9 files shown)


anoek:
Aesthetically I like this. Organizationally, Correspondence is a speed setting and it would be weird to have to click that instead of the big Play button to <b>play</b> a correspondence game, so not sure I'm a fan of that.

I also don't want to merge a PR of a disabled button with the expectation that someone else will come along and implement a new interface for "Play a Friend" and enable it, but I do think that feature would be good so if it was implemented I'd be happy with that. Alternatively we can remove it for the time being until someone has the time or inclination to implement that feature end to end.


---

As for the backend, there are several things that need to happen to fully add a new rating band for Rapid, but I think we can dove-tail the long PITA stuff in with our upcoming ratings update that's tentatively slated for Q3/Q4. In the meantime I think we can get the automatch pool in relatively easily and just have those games rated as Blitz with an asterisk that we'll properly re-rate those in their own band during the next rating update.

pavelwatson:
Hey, that looks really nice! I'm a big fan of this design concept. I have a few thoughts and suggestions if you don't mind:

1. Right now, the "Play" button and the speed/board size switches don’t stand out enough from each other. It feels like everything is a "Play" button, if that makes sense. Maybe consider making the switches a bit more transparent, or just outline them in blue instead of filling them? Not sure
2. The layout looks great, but it feels a bit cramped to me. Maybe experiment with adding a bit more white space? It almost feels like I'm looking at a mobile version on desktop.
3. One thing I'm curious about - how would this design work with the current feature of searching for multiple board sizes at once? What time setting would it show when multiple sizes are selected? I largely agree with @spaceraven's point made on the forum that it would be much easier and less confusing to search just one board size at a time.

These are just some thoughts, of course. You've already done an amazing job with this redesign, and I really appreciate the work you're putting in!

While we are at it, maybe we should consider updating the Custom Request/Challenge menu too? It's not very user-friendly right now - too many settings thrown at you at once. Most options can be hidden behind an "Advanced settings" dropdown, similar to what Chess.com or Lichess do. I'd be happy to help with something too.

Also, @anoek, I don't want to be "pushy", but it would be great if you could share some thoughts on the ongoing forum discussion. The community is very engaged and excited about potential updates, any information would be much appreciated!

anoek:
>>    Aesthetically I like this. Organizationally, Correspondence is a speed setting and it would be weird to have to click that instead of the big Play button to play a correspondence game, so not sure I'm a fan of that.
>Thanks! Correspondence is already handled differently in that it you can have multiple requests going and it doesn't instantly throw you into a game. I can see the argument for putting with the rest, but the word 'Correspondence' might be a bit too long 😛

I"m second guessing putting correspondence up there. It kind of feels like a second class citizen down below the giant Play button, but I'm not sure it belongs up there with the live game controls. I'm not sure what the right solution is.

anoek:
> While we are at it, maybe we should consider updating the Custom Request/Challenge menu too? It's not very user-friendly right now - too many settings thrown at you at once. Most options can be hidden behind an "Advanced settings" dropdown, similar to what Chess.com or Lichess do. I'd be happy to help with something too. 

Revamping the custom challenge would be potentially welcome, but feels like probably that's a separate PR.

LForchini:
I've given the buttons some breathing room. I still have to give the buttons below the break some margin space as well, I'll be doing that soon.

@pavelwatson To clarify point 3, it's admittedly some inconsistent behaviour, but you can search for multiple board sizes at once, but not multiple speeds. This mainly because there are speed specific options that I don't know how to deal with.

sc

PR #2753, opened 2024-07-13
Prevent report dupes

GreenAsJade:
Fixes CMs having hassles knowing how to vote to give sensible messages about duplicate reports

## Proposed Changes

  - Instead of opening the Report form, open the IncidentList, if the user already has opened a report on that game
  
* : built on top of #2752, because it doesn't really make sense to do this if they can't edit their reports.


anoek:
Yeah that's great, thanks!

PR #2763, opened 2024-07-29
Add better support for IRL game recording

anoek:
This patch adds a new button to the SGF library

image

This creates a self-play game where the player names can be provided like a demo board. The primary difference between this and a demo board is this is a "real game" in the sense that it is not a review itself, so can be reviewed itself by third party accounts.


Skwidder:
Any Idea when this will be added to live? We are doing a tournament in late march and would love to be able to use this!

PR #2771, opened 2024-08-07
Wrap tournament steps when there are too many

cassidoo:
It's rare, but sometimes when you have a long tournament [like this one](https://online-go.com/tournament/95975) with many rounds, the page gets too wide from all of the rounds:

image

This fixes that, and wraps the steps after they hit the container width:

image


anoek:
That's a lot of rounds! Thanks for the fix.

PR #2778, opened 2024-08-13
add warning-fg color to ogs.styl

GreenAsJade:
Fixes  warnings being issued about undefined color

## Proposed Changes

  - I put the light and dark `warning-bg` into a "contrasting color generator" online, and created `warning-fg` with what it gave me

### DIFF
--- a/src/ogs.styl
+++ b/src/ogs.styl
@@ -167,6 +167,7 @@ light.chart-line                        = #184A6B
 light.error-boundary-bg                 = lighten(light.reject, 50%)
 light.error-boundary-fg                 = darken(light.reject, 50%)
 light.warning-bg                        = #FFAE2C
+light.warning-fg                        = #0051d3 // generated to contrast with warning-bg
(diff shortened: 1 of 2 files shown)


anoek:
:+1:  for fixing the missing variable, not sure how I didn't see that

I've learned over the years though, that you pretty much never want to simply pick the opposite color on the color wheel for text on a background (I've also come to appreciate the difficulty in pairing good colors and creating a good color pallet). For things like this though I've personally started having a lot better luck these days defining the background color in HSL and then adjusting the lightness (L) for contrast for a decent pairing for text, or just going with white/black/some gray. Also the blue looks a lot like a link color:

image

so what I'd do these days is convert that color (`#FFAE2C`) to HSL: `hsl(37, 83, 58.6)` and then drop that third component down to make it darker:

`hsl(37, 83, 30)`

image

`hsl(37, 83, 20)`
image

`hsl(37, 83, 15)`
image

`hsl(37, 83, 12)`
image

Or just go with black

`hsl(37, 83, 0)`
image

PR #2783, opened 2024-08-16
Fix updated route

GreenAsJade:
Fixes # 404

## Proposed Changes

  - Point at the new route name


BHydden:
> Fixes # 404

Broken link

PR #2785, opened 2024-08-17
Move cm pies to cm dash

GreenAsJade:
Fixes slow loading elements on a person's Profile page

## Proposed Changes

  -  Move the pie charts to the CM Dashboard
  - Make sure we keep data that we've fetched for the graphs
    (pies could do with this too at some point)


GreenAsJade:
Note:

is this right:

```
import UserVoteActionSummary from "User/UserVoteActionSummary";
import UserVoteActivityGraph from "User/VoteActivityGraph";
```

I don't recall using subdirectory imports like that before 🧐

anoek:
> Note:
> 
> is this right:
> 
> ```
> import UserVoteActionSummary from "User/UserVoteActionSummary";
> import UserVoteActivityGraph from "User/VoteActivityGraph";
> ```
> 
> I don't recall using subdirectory imports like that before 🧐

I tend to import those files into the `User/index.ts` and then can just do `import { UserVoteActionSummary, UserVoteActivityGraph } from "User";`, that might be what you're thinking of

GreenAsJade:
🤦‍♀️  I knew that.  Brainfade!  😝

PR #2793, opened 2024-08-20
Remove current (incomplete) period from vote count  graph

GreenAsJade:
Fixes ugly looking low data point at the end of the curve :)

## Proposed Changes

  - Put the existing "remove last period" function into misc, and call it from VoteActivityGraph

### DIFF
--- a/src/lib/misc.ts
+++ b/src/lib/misc.ts
@@ -655,6 +655,16 @@ export function getCurrentGameId() {
     return null;
 }
 
+// x is a  date, y is data for that date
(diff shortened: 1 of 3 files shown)


anoek:
Yeah sorry that was my fault, I got distracted fixing it - did that merge in the web editor

GreenAsJade:
A  "con" of doing separate branches for different changes, instead of just plonking each change on top of the other 😝

anoek:
I think if it's all touching the same code working towards a similar goal, one branch is good. It's when there are very different features all in the same branch, that's when separation is good I think.

PR #2799, opened 2024-08-24
Annotate `Players` based on role in a report

GreenAsJade:
Fixes moderators losing track of who reported and who is accused.

## Proposed Changes

  - Have `ViewReport` provide context to `Player` so that `Player` can annotate according to role in a report.


GreenAsJade:
Something messed up there, will resubmit

PR #2800, opened 2024-08-24
Annotate `Players` in ViewReport

GreenAsJade:
Fixes Moderators losing track of who is who

## Proposed Changes

  - Have  `Player` provide a context that it can be rendered in to enable it to annotate participants in a report.
  - Have `ViewReport` use that context.

### DIFF
--- a/src/components/Player/Player.tsx
+++ b/src/components/Player/Player.tsx
@@ -73,6 +73,13 @@ export interface PlayerProperties {
     forceShowRank?: boolean;
 }
 
+type ViewReportContextType = {
+    reporter: player_cache.PlayerCacheEntry;
+    reported: player_cache.PlayerCacheEntry;
+};
+
+export const ViewReportContext = React.createContext<ViewReportContextType | null>(null);
+
 export function Player(props: PlayerProperties): JSX.Element {
     const user = data.get("user");
     const player_id: number =
@@ -98,6 +105,8 @@ export function Player(props: PlayerProperties): JSX.Element {
 
     const base = player || historical;
     const combined = base ? Object.assign({}, base, historical ? historical : {}) : null;
+
+    const viewReportContext = React.useContext(ViewReportContext);
 
     React.useEffect(() => {
         if (!props.disableCacheUpdate) {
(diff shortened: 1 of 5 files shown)


GreenAsJade:
Screenshot 2024-08-24 at 9 34 30 PM (2)

src/components/Player/Player.tsx
```diff
@@ -283,6 +292,14 @@ export function Player(props: PlayerProperties): JSX.Element {
         main_attrs.className += " " + combined.ui_class;
     }
 
+    if (viewReportContext && viewReportContext.reported.id === player_id) {
```
anoek, line 295: This feels wrong, we shouldn't be spreading the context logic over to a generic component like `Player` I don't think. How about instead we put that context stuff up in the report code and then pass `"reported"` in for the `ui_class` for `Player`?

GreenAsJade:
Here's my case for doing it this way :)

"Player" is a component that knows how to render Players in various contexts.

One of those is "in a report context".  It makes a context provider available to components that feel they are "in a report context", and lets those components describe which two players have the context-specific roles.

Once that's done neither component has to worry anymore about remembering to (or even being able to) think about what is the current' player's role.  Instead _every instance_ of `Player` within that context is "magically" annotated.

Note how the `ReportedGame` is made of a hiearachy of components, many of which have `Player` components instantiated in them, and every single one is correctly annotated, in the screenshot above, without editing any code in any of them...

anoek:
> "Player" is a component that knows how to render Players in various contexts.

Does it? I always viewed it as a generic component that didn't care where it was rendered, it just rendered the bits of a player name that you told it to.

That changes fundamentally with this. Now part of the rendering code for a view is pushed down into a generic component, feels like that is a pattern that could get out of hand very quickly.

However, I could see how getting the data to all the right places might get painful. Let me ponder, will probably merge as is but want to digest if there's a cleaner way.

anoek:
> >    Now part of the rendering code for a view is pushed down into a generic component

> I'm not seeing this. A Player component is responsible for rendering the Player - it only has Player rendering code.

ehh, the `Player.styl` now has code that only takes effect when in the `#ViewReport` element and we have a special `ViewReportContextType` type that is, as indicated by the variable name, clearly only applicable for the `ViewReport` view.

I guess though it probably wouldn't have felt so wrong if it was spun as a more generic thing, as opposed to something that only works when in `ViewReport`, so in that light maybe it's fine, and maybe if we find another place where that pattern is a good fit we make it more generic or something.

GreenAsJade:
I think that there should be a contract about the styling.  The component that does instantiates the context

`  <ViewReportContext.Provider...`

should also keep the contract by putting a className that `Player` is expecting around the context as well, rather than `Player` looking for `#ViewReport`  (which is a specific special component).

I'll see about making it feel more like a generic pattern and then we'll see how it strikes you...

... and in the mean time if you can think of a different way of ensuring that "all `Player`s within this page get annotated" that'd be cool!

GreenAsJade:
It felt safer to have a deal with the context consumer to specifically contain effects of the annotation styling.

PR #2811, opened 2024-09-07
Defend against GameLog being used in a context without help

GreenAsJade:
Fixes GameLog modal not working on Game page

## Proposed Changes

  - detect when registerTargetItem isn't available, which is the case in an OGS Modal


GreenAsJade:
(this needs chicanery because eslint doesn't understand that 

`    const { registerTargetItem } = React.useContext(DynamicHelp.Api);`

can result in registerTargetItem being null, since the type of DynamicHelp.Api says it is a function.

PR #2815, opened 2024-09-08
Migrate from gulp and webpack,  to vite

pygeek:
- Simplify and speed up the build/dev process by replacing Gulp and Webpack with Vite.
- Additional changes in this PR need to be made to enable HMR — an additional lint rule has been added to support this.


src/vite.config.ts
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C)  Online-Go.com
+ * Copyright (C)  Ben Jones
```
anoek, line 3: Probably don't need to assign copyright to Ben Jones, I reckon you copied this from something that he was doing that he felt he wanted to be able to safely copy out and assign his own license to and all of that.. but that's not you so can probably just remove that from this  :)

anoek:
Is there a particular reason for the `vite.config.js` to reside in `src/`? Our other config files are up a directory, would it be fine to reside there too?

pygeek:
I initially settled on that approach, because it was the fastest to prototype the migration—I was having issues with the existing typescript setup. I've updated the code with my preferred setup (`index.html` + `vite.config.js` moved outside of `src`) now that I've worked out the bugs. 

Context: Vite has a constraint where the [`vite.config.js` must live alongside the `index.html`](https://vitejs.dev/guide/#index-html-and-project-root).

index.html
```diff
@@ -254,66 +252,33 @@
...
             try {
                 new WebAssembly.Memory({ initial: 0 });
                 console.log("WASM support detected, score estimator will be fast");
-                script.src = "/OGSScoreEstimator/OGSScoreEstimator-0.7.0.js";
+                script.src = "//localhost:8080/OGSScoreEstimator/OGSScoreEstimator-0.7.0.js";
             } catch (error) {
                 console.warn(error);
                 console.log("No WASM support detected, score estimator falling back to ASM.js mode");
-                script.src = "/OGSScoreEstimator/OGSScoreEstimator-0.7.0.asm.js";
+                script.src = "//localhost:8080/OGSScoreEstimator/OGSScoreEstimator-0.7.0.asm.js";
```
benjaminpjones, line 266: Will this work in other environments? (i.e. user clients)

index.html
```diff
@@ -1,19 +1,18 @@
 <!doctype html>
-<html lang="{{PAGE_LANGUAGE}}">
```
benjaminpjones, line 2: Doesn't Vite also have the ability to use variables in html?

https://vitejs.dev/guide/env-and-mode.html#html-env-replacement

package.json
```diff
@@ -9,19 +9,15 @@
         "npm": ">=8.5.0"
     },
     "scripts": {
-        "dev": "supervisor -w Gulpfile.js,webpack.config.js,tsconfig.json supervisor -w Gulpfile.js -x gulp --",
+        "dev": "npx tsx server.ts",
```
benjaminpjones, line 12: Reload on change is definitely an important feature for devs. (Is this what you're referring to with the HMR updates needed?)

benjaminpjones (commented):
Overall, I think a switch to Vite would be exciting. It seems to be the preferred modern build tool for React and other frameworks :)

anoek (commented):
I'm pretty excited for this, I've looked into it in the past but it was a bit daunting, thanks for taking it on.

FTR there's some webpack usage within our `Makefile.production` that will need to be replaced with the appropriate vite equivalents, probably most notably is the production builds: 

https://github.com/online-go/online-go.com/blob/4b40715f0b9c065321154e06e9a6ee45a86a4684/Makefile.production#L125-L132

index.html
pygeek: Values are temporarily hardcoded for testing.

index.html
pygeek: Yes, this was only done for rapid prototyping.

package.json
pygeek: Vite supports live reload by default. However, with HMR (Hot Module Replacement) the specific module which was updated is reloaded, rather than the entire app. [https://webpack.js.org/concepts/hot-module-replacement/](https://github.com/ArnaudBarre/eslint-plugin-react-refresh)

In order to get this working, I have to make updates to a few components. Once this is done, Vite will automatically enable HMR. Related eslint rule: [https://github.com/ArnaudBarre/eslint-plugin-react-refresh](https://github.com/ArnaudBarre/eslint-plugin-react-refresh)

pygeek:
Still working on this

pygeek:
> This pull request has been marked stale and will be closed soon without further activity. Please update the PR or respond to this comment if you're still interested in working on this.

still working on this

PR #2816, opened 2024-09-08
Fix PaginatedTable numbering

pygeek:
Fix pagination numbering on PaginatedTable component when user has played 0 games.

Before my change:
Pagination would display "1 / 0"

After my change:
Pagination displays "1 / 1"

This change also fixes the misaligned page numbering.


anoek:
Thanks!

PR #2817, opened 2024-09-09
Guard GoThemePicker against trying to persist null elements.

GreenAsJade:
Fixes a stream of PersistentElement warnings

## Proposed Changes

  - Don't instantiate PersistentElement with null target in GoThemePicker.
 

FWIW ChatGPT thinks that PersistentElement should `throw` in that case, since it feels the caller is clearly in error.  I wasn't sure enough about it to do that :)


anoek:
Yuck, not sure how I missed that. Thanks for the fix!

ryanbishopj:
518f602624c8463cbd16693001f2ee93a827dc19

PR #2822, opened 2024-09-09
Show CM the current week's vote count

GreenAsJade:
Fixes disappointment at the end of the week when the final number comes in lower than hoped for.

## Proposed Changes

  - Put a dot on the weekly vote count graph showing the current week so CM can see how they are doing.


GreenAsJade:
(branching oopsie)

PR #2833, opened 2024-09-13
Restore dynamic help for mobile (with layout fix)

GreenAsJade:
Fixes having dynamic help on phones

## Proposed Changes

  - Leave RDH popups at the correct place vertically, just push them on the screen for phones, where the narrow width otherwise has them fall off the edge.


anoek:
Nice!

PR #2834, opened 2024-09-14
"Stone font scale" preference

GreenAsJade:
Fixes struggles getting used to the new big numbers and last move indicator :)

## Proposed Changes

  - Provide a "stone font scale" preference, use it to set Goban stone font scale.

** Needs  https://github.com/online-go/goban/pull/177


GreenAsJade:
(webpack fails because of goban update dependency)

anoek:
Nice, thanks!

PR #2839, opened 2024-10-02
Add typings for window.

benjaminpjones:
`(window as any).some_field` is abundant across the codebase.  This diff applies some types using [interface merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).  There are two benefits:

- More consistency when storing and accessing fields.
- Centralized documentation of the global variables.
- `window.some_field` just looks a bit nicer

Tested: `yarn webpack` builds fine, `yarn dev` has no errors and looks okay when I visit `localhost:8080`


anoek:
Nice, thanks!

PR #2844, opened 2024-10-05
Support for CMs voting for suspend

GreenAsJade:
Fixes Community Moderation being reliant on full moderators to process escalated reports that need user-suspension action.

## Proposed Changes

  - CMs can get vote-to-suspend power
  - CMs with vote-to-suspend can  vote to suspend users when the report about the user is escalated
  - Full moderators no longer get "CM-class" reports in their queues
     - if they want to see a CM-class report they need to access it via History or URL (likely from Mod Log)

 
Needs:  https://github.com/online-go/ogs-node/pull/66
Needs: https://github.com/online-go/ogs/pull/2003

### DIFF
--- a/src/components/AccountWarning/CannedMessages.ts
+++ b/src/components/AccountWarning/CannedMessages.ts
@@ -294,7 +294,23 @@ moderator to let them know.
 Thanks for your recent report about {{bot}}.
 
 We've notified the owner of that bot.
-            `),
+`),
             { bot },
+        ),
+    ack_suspended: (reported) =>
+        interpolate(
+            _(`
+Thank you for your report.  {{reported}} is a repeat offender, their account has been suspended.
+`),
+            { reported },
+        ),
+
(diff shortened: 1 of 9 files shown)


src/components/AccountWarning/CannedMessages.ts
```diff
@@ -294,7 +294,23 @@ moderator to let them know.
...
+        interpolate(
+            _(`
+Thank you for your report.  {{reported}} is a repeat offender, their account has been suspended.
+`),
+            { reported },
+        ),
+
+    ack_suspended_and_annul: (reported) =>
+        interpolate(
+            _(`
```
anoek, line 310: Good candidates for `llm_pgettext` maybe?

GreenAsJade: (It felt simple and stable to me, but hey why not :) )

PR #2845, opened 2024-10-05
POC - Add tooltip to modal, Modal provider (for computer challenge, language picker, game fork)

andrewjcasal:
Fixes #

https://github.com/user-attachments/assets/611d3cdc-0b2a-4419-a74b-1ac3d95eccf6

## Background
- We had been unable to add tooltips on modals because modals were never a nested component under the main tree containing the DynamicHelp provider.
- Feel free to try this out by following the video!

## Proposed Changes
  - Infrastructure: add the modal provider as a component under the OGSHelpProvider to consumer tooltip information.
  - ActivateTooltip - React component to make it easy to trigger tooltips anywhere. Happy to update if I'm not handling all cases.
  - Modal Context
  - Modal Provider - A context for modals to be rendered at the top of the tree, right under the DynamicHelp functionality. CURRENTLY used only for creating a computer challenge
  - Modal Registry - This will include the list of modals that will use this logic - currently there's only three. Other modals can be added to the registry over time, or can use the register and unregister functions to do this dynamically.
  - ForkModal and LanguagePickerModal switched to functional components to better utilize useContext.
  - Backwards compatibility with other existing modals.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,2 +1,5 @@
 {
+    "[ruby]": {
+        "editor.formatOnSave": false
+    }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,2 @@
 {
-    "[ruby]": {
-        "editor.formatOnSave": false
-    }
 }
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -25,6 +25,9 @@ module.exports = {
         "src",
         "node_modules"
     ],
+    "moduleNameMapper": {
+        '^@/(.*)': '<rootDir>/src/$1',
+    },
     "setupFiles": ["./setup-jest.js"],
-    "setupFilesAfterEnv": ["jest-chain"],
+    "setupFilesAfterEnv": ["jest-chain", "@testing-library/jest-dom"],
 }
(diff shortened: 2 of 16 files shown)


anoek:
I'm switching this to draft until while you work on it, feel free to change it back to ready for merge when you're ready

GreenAsJade:
@andrewjcasal   I realised an even better way of testing this is to get rid of the TBD in GameLog :)

I thought "oh, I will quickly do that right now", but of course that requires GameLogModal to be a new modal, which is not a 2 minute job.

I can do it when I have a moment soonish, or if you add it to the PR, I'll test it out.

GreenAsJade:
See above PR: I made GameLogModal use this, which lets it have RDH about autoscore (🎉).  I took out the "test case" RDH, and did a slight refactor as per comment on that commit.

anoek:
Marking as draft while stuff is being worked on, mark as ready to merge when everything is set, and thanks again!

GreenAsJade:
@anoek   I wonder if things are stable enough to incorporate this?

anoek:
Seemed like it was close from your comments but there's been a lot of silence since then..

GreenAsJade:
I think it actually was "hey, I've made this all working and nice, what do you all think?" 😝

GreenAsJade:
I've created at PR for the changes I suggested:  https://github.com/online-go/online-go.com/pull/2901

@andrewjcasal  might prefer to pull them onto this PR and have this one merged - I'm fine with either way!

Issue #2075, opened 2022-12-01
Add mousewheel support for game-move scrolling


christiancodes:
**Is your feature request related to a problem? Please describe.**
We would like to add mousewheel scrolling to move through the game.

**Describe the solution you'd like**
Discussed here: https://forums.online-go.com/t/mousewheel-to-navigate-moves-in-games/45007/7

**Additional context**
Maybe also adding a setting so players can opt into this behaviour.

anoek:
This has been implemented, not stale!

PR #2855, opened 2024-10-20
add scroll to navigate functionality

TransientError:
Fixes #
https://github.com/online-go/online-go.com/issues/2075

## Proposed Changes
  - Add the ability to navigate the game by scrolling the mouse wheel behind a preference that is off by default

I am normally a backend engineer, so very open to feedback on my React 😄
I tested this on the local environment on a demo board and in puzzles, but was not sure if this could potentially break any other views.


anoek:
Very cool, thanks!

PR #2856, opened 2024-10-25
Added handicap filter

### LINKED ISSUE #2358: Add filter for handicap games
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I enter a game from the game list and find out it's a handicap game (It's hard to always remember to check). I then cancel it and I imagine the other person to be frustrated as well as they have to recreate the challenge.

Being able to filter in/out handicap games from the game list.

**Current filters**
Current state

I've poked a bit in the code and managed to make it work. 
The following are different layouts i have implemented :

- **The crude way**
proposal 1 - crude

- **Grouped by rows**
proposal 2 - rows

- **Grouped by columns**
proposal 3 - columns

My personal favorite is the third one which I think makes more sense.
I don't know what to do about the icon however, it's a copy of the "ineligible" icon for now, I'd love to know your thoughts on what color and shape to use.

A cleaner alternative would be to have drop-down filters on each of the columns in the list, something like the following (not implemented yet) :
Clean way

What do you guys think ?

fnoaman:
Fixes #2358

## Proposed Changes

  - Added Handicap filter, see #2358

Apologies for the very late PR

### DIFF
--- a/src/components/SeekGraph/SeekGraphLegend.tsx
+++ b/src/components/SeekGraph/SeekGraphLegend.tsx
@@ -71,15 +71,6 @@ export function SeekGraphLegend(props: SeekGraphLegendProps): JSX.Element {
 
     const group1 = [
         legendItem(_("19x19"), () => BoardSizeLegendIcon(currentPalette.size19), "show19x19"),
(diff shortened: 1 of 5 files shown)


anoek:
Nice, thanks!

PR #2868, opened 2024-11-22
Put seek graph alongside create game on desktop screens

GreenAsJade:
Fixes 
 - the list that people are looking for being way down the bottom.
 - seek-graph being curiously wide and space-consuming

## Proposed Changes

  - Put `seek-graph` alongside `create game` form for wide screens
  -- Note the deliberate choice to make the row containing these two wider than normal
  ---  So that Create Game is the focus, and can be a nice width.

### DIFF
--- a/src/views/Play/CustomGames.styl
+++ b/src/views/Play/CustomGames.styl
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
- 
+
  #CustomGames {
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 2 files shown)


anoek:
It's hanging off in a weird way over here: 

image


I'm not sure how to shrink that up nicely. I do find myself wondering if we should revert the change to have the game creation options on the left, and instead have two buttons, one to create a custom game that pops open the open challenge modal like we used to have, and another to create a custom computer game like we used to have. That'd probably free up enough space for the graph. Perhaps a third to challenge a friend with a search field, it seems doing that isn't the easiest right now, it might be worth turning that space into space to launch the various challenge modals we have.

I'm also not sure how useful the graph is anymore these days, perhaps we just get rid of it.

GreenAsJade:
I couldn't quite come at getting rid of it, so I gave it a toggle in #2871 😝

PR #2871, opened 2024-11-24
Provide a toggle to turn on display of seek graph

GreenAsJade:
Fixes seek graph taking up unwelcome space on Custom Games page

## Proposed Changes

  - Provide a toggle to turn it on for people who want to see it.
  

Note: some shenanigans are required to keep seek graph rendered, so we don't loose the ref to it.

### DIFF
--- a/src/views/Play/CustomGames.tsx
+++ b/src/views/Play/CustomGames.tsx
@@ -74,6 +74,12 @@ export function CustomGames(): JSX.Element {
     const canvas: HTMLCanvasElement = React.useMemo(() => allocateCanvasOrError(), []);
     const seekgraph = React.useRef<SeekGraph>();
 
+    const [seekGraphVisible, setSeekGraphVisible] = React.useState(true);
+
+    const toggleSeekGraph = () => {
+        setSeekGraphVisible((prev) => !prev);
+    };
+
     // Used to not change the challenge list while they are trying to point the mouse at it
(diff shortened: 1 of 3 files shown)


src/views/Play/CustomGames.tsx
```diff
@@ -74,6 +74,12 @@ export function CustomGames(): JSX.Element {
     const canvas: HTMLCanvasElement = React.useMemo(() => allocateCanvasOrError(), []);
     const seekgraph = React.useRef<SeekGraph>();
 
+    const [seekGraphVisible, setSeekGraphVisible] = React.useState(false);
```
anoek, line 77: Maybe should do a `usePreference` here instead of `useState` so we remember their setting?

GreenAsJade: Doh yes!

BHydden:
Wouldn't default on be less abrasive amidst some other controversial UI changes? 😅

GreenAsJade:
I haven't seen anyone arguing in favour of the seek graph, and it's position seems to be agreed as "making it even harder to find the custom games".

BHydden:
Personally I agree, if it were me I would scrap it. But I seem to remember past forum discussions about that meeting resistance. Lemme see if I can find any.

GreenAsJade:
I think anyone desperate that we keep it should be satisfied that in one click it will remain there for ever more for them

PR #2891, opened 2024-12-13
Improve mobile menu left padding to match hamburger

tylerbarker:
# Overview

I noticed the mobile menu items left padding currently doesn't match that of the hamburger and it was bothering me a bit, hope this is welcome!

## Before

mobileMenuBefore

## After

mobileMenuAfter


anoek:
Thanks!

PR #2895, opened 2024-12-18
Move @types/* packages from runtime to dev dependencies

tylerbarker:
Spotted some `@types/*` packages that made their way into runtime dependencies 🧹


anoek:
thanks!

PR #2898, opened 2024-12-19
My own reports

GreenAsJade:
Fixes mods and cms not having a way to view the list of reports they have raised

## Proposed Changes

  - Wack the IncidentReportList into a section in the ReportsCenter

This is on top of #2897


anoek:
Thanks!

PR #2901, opened 2024-12-21
PR2845

GreenAsJade:
This is my tweaks to #2854


GreenAsJade:
Obviously, I need to fix these recent conflicts 😭

GreenAsJade:
I think this is good to go - it makes Language Picker Modal and Fork Game modal use the new system, as examples.

I'll probably gulp and look at Challenge Model when I remember what RDH I wanted to put on it :)

anoek:
Awesome, thanks for getting it back to a mergable state. I do think we should probably port over the remaining modals and get rid of the class based system so we're not maintaining two ways of accomplishing the same thing.

GreenAsJade:
Yeah fo sho.   Deep breath needed for that one :)   But at least this isn't going stale now :)

PR #2905, opened 2025-01-01
buxfix: "urgent" game notification handling

benjaminpjones:
`time_to_move` does not exist on the boards in notification manager.  Yay TypeScript :partying_face:

---

In case you don't believe typescript, one can also verify in the console:

```
> Object.values(notification_manager.active_boards)[0].time_per_move
88449
> Object.values(notification_manager.active_boards)[0].time_to_move
undefined 
```

### DIFF
--- a/src/components/Notifications/NotificationManager.tsx
+++ b/src/components/Notifications/NotificationManager.tsx
@@ -191,8 +191,8 @@ export class NotificationManager {
     notifications: { [k: string]: Notification };
     ordered_notifications: Notification[];
     unread_notification_count: number = 0;
-    boards_to_move_on: { [k: string]: any };
-    active_boards: { [k: string]: any };
+    boards_to_move_on: { [k: string]: GameListEntry };
+    active_boards: { [k: string]: GameListEntry };
     advances: number = 0;
     auth: string | undefined;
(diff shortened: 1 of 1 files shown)


benjaminpjones:
Requesting @GreenAsJade review, as I'm not totally clear on how the urgent game logic (#1802) is supposed to work.

GreenAsJade (approved):
Looks like the right fix to me.

Back when this bug was created, we didn't even have ClientToServer types :)

benjaminpjones:
> Back when this bug was created (by me), we didn't even have ClientToServer types :)

I figured! It may not have even been a bug at that point in time 🤷‍♂️

> The effect of fixing it should be that when you play a move in a live game, and you have more than two lives game going and you have auto-advance on, then you should arrive at the most urgent game rather than the next in the sequence from where you were up to circling through them. Subtle :)

Ah, that makes sense! I can see why no one complained about the bug - very few simul players and urgency may not often be an issue. But this seems like the right behavior for the tight cases 👍

GreenAsJade:
I had a little dig in the back end - it looks like it was really a bug.

anoek:
Thanks for the fix!

PR #2923, opened 2025-01-15
fix: replace history to allow back button

### LINKED ISSUE #2922: Report center back button doesn't work for CMs
When I press the back button in my browser in the 'Reports Center' the page flashes and doesn't go back.

**To reproduce**
1. Go to [the OGS Reports Center](https://online-go.com/reports-center)
2. Wait for the redirection to a specific incident
3. Press the back button
4. See the same thing as before

**Expected behavior**
I expect to go back to the page I was on before

**Additional context**
I assume this is due to the redirect from the `/reports-center` to the specific report. This seems to only happen to CMs.

LForchini:
Fixes #2922

## Proposed Changes

  - Allow for use of back button when navigating to reports center


anoek:
Thanks for the fix!

PR #2928, opened 2025-01-20
Re-add variation stone opacity setting

### LINKED ISSUE #2910: Variation stone opacity setting missing
Just noticed I can't find this setting anymore after logging in on a new computer, it seems to have been commented out here as part of revamping the settings page but was never added back: https://github.com/online-go/online-go.com/commit/9f6612036100a803bd3837580fbb79b429420899

JDNdeveloper:
Fixes #2910

Adds the variation stone opacity setting back to the settings page. Incorporated MiniGoban to preview the settings change in keeping with similar settings.

image


anoek:
Thanks! I can't recall why I had disabled that, seems to work just fine though so thanks for getting that back in working order and adding the preview :)

PR #2957, opened 2025-02-14
Add current turn to the condition to show Pass button

### LINKED ISSUE #2943: Pass button shows on previous moves (analysis off)
**To Reproduce**
1. Play correspondence game with analysis off
2. Press back button to view move history
3. Observe Pass button shows up on your turn

**Screenshots**

https://github.com/user-attachments/assets/f0228b35-1841-4a24-a774-16b2ceb5ca6c

**Smartphone:** Android + Chrome

miguelython:
Fixes #2943

## Proposed Changes

  - It add a condition (current move is last move) to show the Pass Button

I verified also that the pass button still shows during a Live game

### DIFF
--- a/src/views/Game/PlayButtons.tsx
+++ b/src/views/Game/PlayButtons.tsx
@@ -22,9 +22,19 @@ import * as preferences from "@/lib/preferences";
 import * as data from "@/lib/data";
 import { game_control } from "./game_control";
 import { alert } from "@/lib/swal_config";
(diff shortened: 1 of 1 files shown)


anoek:
That test may or may not be broken, considering what you're fixing.

miguelython:
I think it was not broken, just a rare case for the pass button be shown in the first turn.

In the first try I didn't want to copy the function from PlayControls.tsx. I found it ugly to duplicate the code.

And I don't think import is a good Idea for import recursion.

Maybe you could refactor into a better solution

anoek:
Thanks for the fix!

PR #2958, opened 2025-02-15
Updated game sound logic for case when game ends in a Tie

### LINKED ISSUE #2791: Incorrect "You have won!" announcement for a tied game
**Describe the bug**
If you tie a game, the automated announcement sound says "You have won!". But it's a tie.

**To Reproduce**
1. Turn on whatever sound settings cause loud announcements after the game. (Resetting all settings to the default should do it.)
2. Play a game that ends in a tie. Integer komi required (such as New Zealand rules).
3. Hear "You have won!" announcement when scoring phase completes.

**Expected behavior**
"You have tied!" or "Tie game!" or "Game over!" or silence.

**Additional context**
Drawing a game used to require moderator assistance, but it's more common in the last six months. I only just noticed now because I happened to score a tied game while logged in on a friend's computer (where I hadn't bothered to turn off sound).

mister-dl:
Fixes #2791

## Proposed Changes

  -Add a case when game ends in a tie (returns `0` instead of a player number)
  -updates `winner` interface prop with `tie` as an option
  
### Concerns
- should `winner` interface prop be number or string? Where would it be set to be a string? I only saw numbers returned (for player id) but not sure when a string is set. 
- I didn't see any jest tests for this. If we don't need it its fine, just want to make sure i'm not missing it somewhere.


anoek:
Since this change is in https://github.com/online-go/goban/ , you'll need to fork that repo and push up a pull request over there, sorry for the trouble.

mister-dl:
makes sense, not a problem!

mister-dl:
PR is actually : https://github.com/online-go/goban/pull/183

PR #2973, opened 2025-02-25
Ai review to functional

GreenAsJade:
Fixes need to wrap AIReview for context, and other React.Component limitations

## Proposed Changes

  -  Followed the steps to convert to functional
  
  BUT

I tried to set up tests for this and it proved too hard to mock.  I had intended to at least make sure the same tests pass on before and after.

It has a lot of conditional stuff in it, so it's hard to be sure some use-case is not broken.

### DIFF
--- a/src/views/Game/AIReview.tsx
+++ b/src/views/Game/AIReview.tsx
@@ -23,7 +23,7 @@ import Select, { components } from "react-select";
 import { UIPush } from "@/components/UIPush";
 import { AIReviewStream, ai_request_variation_analysis } from "@/components/AIReviewStream";
 import { openBecomeASiteSupporterModal } from "@/views/Supporter";
-import { errorAlerter, errorLogger, Timeout } from "@/lib/misc";
+import { errorAlerter, errorLogger } from "@/lib/misc";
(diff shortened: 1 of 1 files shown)


anoek:
Nice, this was one of the big ones that was still a class :)

First thing I notice is when loading an existing review, instead of the graph popping up instantly it seems to be waiting for some UI update. If I reload and hit an arrow key to go back a move, it pops up instantly then, so I think it just needs a refresh at some appropriate time, I'd assume after loading but unsure.

The other thing I notice is when exploring variations I'm not seeing any updates on the board to show what the AI is thinking, the graph also seems delayed. Going back a move and then forward again does show the updates, so I think it's a similar issue as above, but maybe not the same thing.

anoek:
It's up on beta, might replicate better on there for you if you were seeing everything working as expected on your local dev system.

GreenAsJade:
Darn - I think I'll need to solve "why doesn't AI work in my Dev environment" first.

And, darn, the fragile ViewReport page has become unworkable, so I think I need to deal with that first.

Hence parking this.

GreenAsJade:
Nah we really need to do this :)

PR #2975, opened 2025-02-25
Hide user's empty associations and servers

### LINKED ISSUE #2805: Can't clear other server ranks once I selected one previously
**Describe the bug**
Description above. I selected GoQuest and KGS before in the rank selection and set it back to select a rank, and it displays a ? instead of removing it from the list.

**To Reproduce**
1. Go to Account Linking
2. Select a rank on any association
3. Unselect it and navigate to profile
4. See error - rank displays a ? instead of nothing

**Expected behavior**
Rank displays nothing

**Screenshots**

image

 - OS: Windows 10
 - Browser: Chrome
 - Version:

**Additional context**

bodrych:
Fixes #2805

### DIFF
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -654,12 +654,13 @@ function SelfReportedAccountLinkages({
 }): React.ReactElement {
     const associations = [1, 2, 3] as const;
     const userAssociations = [];
-    for (const i of associations) {
-        const country = links[`org${i}`];
-        const id = links[`org${i}_id`];
-        const rank = links[`org${i}_rank`];
+    for (const num of associations) {
+        const country = links[`org${num}`];
+        const id = links[`org${num}_id`];
+        const rank = links[`org${num}_rank`];
         if (country && (id || (rank && rank > 0))) {
             userAssociations.push({
+                num,
                 country,
                 id,
                 rank,
@@ -686,6 +687,7 @@ function SelfReportedAccountLinkages({
         const rank = links[`${server}_rank`];
         if (id || (rank && rank > 0)) {
             userServers.push({
+                server,
                 name,
                 id,
                 rank,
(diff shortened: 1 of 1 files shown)


src/views/User/User.tsx
```diff
@@ -652,46 +652,58 @@ function SelfReportedAccountLinkages({
...
-            <ServerLink name={_("Yike Weiqi")} id={links.yike_username} rank={links.yike_rank} />
-            <ServerLink name={_("GoQuest")} id={links.goquest_username} rank={links.goquest_rank} />
-            <ServerLink
-                name={_("BadukPop")}
-                id={links.badukpop_username}
-                rank={links.badukpop_rank}
-            />
+            {userAssociations.length > 0 && <h3>{_("Associations")}</h3>}
+            {userAssociations.map((props) => (
+                <AssociationLink {...props} />
```
anoek, line 700: Need to have a `key` prop on this to avoid React complaining at us on the console, either the index or name I reckon

src/views/User/User.tsx
```diff
@@ -652,46 +652,58 @@ function SelfReportedAccountLinkages({
...
+            ))}
+
+            {userServers.length > 0 && <h3>{_("Servers")}</h3>}
+            {userServers.map((props) => (
+                <ServerLink {...props} />
```
anoek, line 705: Same here

anoek (changes requested):
Thanks for working on this!

src/views/User/User.tsx
bodrych: I used the spread syntax. The properties are the same as they were (country, id, rank), so react doesn't complain about it.
Should I explicitly define the properties in order to make code more readable maybe?

GreenAsJade: I think

`{userAssociations.map((props) => (
    <AssociationLink key={props.id} {...props} />
))}`

is needed, which is OK since those associations do have unique ids

anoek:
I see the done message but no new commits - did you forget to push by chance?

bodrych:
Should be okay

anoek (approved):

anoek:
Thanks!

PR #2976, opened 2025-02-26
Refactor ViewReport

GreenAsJade:
... removing report-navigation to a parent, and ensuring it fetches its own data from Django when told what report we're looking at.

Fixes intermittent bugs/fragility due to different sources of info and timing.

## Proposed Changes

  - Move "Report Navigation" out to a parent (new component, ReportsViewer)
  - Fetch the latest Report data when the selected report changes, or when reports_manager says there's new info

### DIFF
--- a/src/lib/report_manager.tsx
+++ b/src/lib/report_manager.tsx
@@ -88,7 +88,6 @@ class ReportManager extends EventEmitter<Events> {
         const user = data.get("user");
         report.id = parseInt(report.id as unknown as string);
 
-        //console.log("updateIncidentReport", report);
         if (!(report.id in this.active_incident_reports)) {
             if (
                 data.get("user").is_moderator &&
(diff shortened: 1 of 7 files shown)


GreenAsJade:
Hah pfff non-moderators, who thinks to test that 🤪

anoek:
I'm still seeing it, to reproduce create a report by a normal user, then click on the blue triangle. When directed to the reports center to view your report the error pops up.

anoek:
Upon looking more, it looks like I have 3 reports open in my system, two by other users and 1 by my normal user. For some reason my normal user has a list of all the report ID's, but cant' load them (expected they can't load them, but they shouldn't see the ids in any case). If I select the one I filed, I can in fact view it, it's just that the triangle was landing on the page which tried to load one of the other ones by default instead of just the ones that the user has access to.

GreenAsJade:
"If I select the one I filed, I can in fact view it, it's just that the triangle was landing on the page which tried to load one of the other ones by default instead of just the ones that the user has access to."

Do you think that "triangle landing on the page" is in fact "triangle opening the IncidentReportList dropdown"?   That would then make sense if you have ids you can't see, because it would try to show them all.

GreenAsJade:
Let's work with #2967 instead.    We might as well find out the issues at the end of the process.

PR #2985, opened 2025-03-08
Add a web server back end choice: DOCKER.

GreenAsJade:
## Proposed Changes

  - Add DOCKER as a webserver back end choice: lets docker-compose set this to point to it's frontend instance in e2e test config.
  - This points things direct at loadbalancer ... seems to work :o

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -29,19 +29,26 @@ import autoprefixer from "autoprefixer";
 import { nodePolyfills } from "vite-plugin-node-polyfills";
 
 const OGS_I18N_BUILD_MODE = (process.env.OGS_I18N_BUILD_MODE || "false").toLowerCase() === "true";
-let OGS_BACKEND = process.env.OGS_BACKEND || "BETA";
-OGS_BACKEND = OGS_BACKEND.toUpperCase();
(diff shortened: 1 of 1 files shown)


anoek:
Can't say I fully understand how `loadbalancer` is getting resolved yet but I trust that it will or will get worked out :)

GreenAsJade:
This is AI's innovation.  It says:

"When your frontend code uses http://loadbalancer, Docker's internal DNS resolver will automatically resolve "loadbalancer" to the IP address of the container running your loadbalancer service, as long as:
 - The loadbalancer service is defined in your docker-compose file 
 - Both services are connected to the same network (which they are - the bridge network)"

the "DOCKER" backend option that specifies `http://loadbalancer` as the backend URL is used this way in the e2e compose:


> frontend:
>         container_name: ogs_frontend
>         build: 
>             context: ../ogs-ui
>         ports:
>             - "8080:8080"
>         volumes:
>             - "../ogs-ui/:/usr/src/app"
>             - "/usr/src/app/node_modules"
>         networks:
>             - bridge
>         depends_on:
>             - loadbalancer
>             - django
>         environment:
>             - OGS_BACKEND=DOCKER
>

PR #2987, opened 2025-03-08
Add screen reader support for initial loading indicator

gsouf:
That's a first PR of a series a changes I'll propose to improve accessibility of OGS for visually impaired users. 
I'll keep each PR minimal to make it easy to review and move forward with individual pieces of work.

## Proposed Changes

  - Add a class `.sr-only` (screen-reader only), that will hide content from screen but make it available for screen readers. Added directly in the html to make it available from the very beginning as it's used for the loading message.
  - Add a "loading" message for screen readers, using `.sr-only`.
  - Add aria- properties [`role status` + `aria-live polite`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role) to make it a low priority information
  - mark the svg loader as hidden for screen readers

Note: I don't think there is a good way to translate the loading message as the index.html file is served as a static file, right?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -251,6 +251,17 @@
             body.accessible {
                 background-color: #1a1a1a;
             }
+            .sr-only {
+                position: absolute;
+                width: 1px;
+                height: 1px;
+                padding: 0;
+                margin: -1px;
+                overflow: hidden;
+                clip: rect(0, 0, 0, 0);
+                white-space: nowrap;
+                border-width: 0;
+            }
         </style>
         {{EXTRA_CONFIG}}
     </head>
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for the patch!

Regarding translating the loading message in index, it could certainly be done but it'd be a stand alone solution. The two ways that come to mind would be to simply have a client side dictionary of the different translations and pick one based on `navigator.languages`, and the other would be to put a placeholder `{{LOADING}}` that gets replaced by the server. That said, it doesn't usually stay on the screen very long (~250-750ms) so it seems like it might be more trouble than it's worth, but if you want it in there, those are the two ways that come to my mind.

gsouf:
@anoek if you process the index.html in a backend and have an easy access to the translation at render time, I think that the placeholder is a good approach. The server is not open source, right? Would that require a lot of efforts to implement?

Even if the loading is fast, I believe it's good to translate it if possible. For any reasons the network might get slower and some users might not be comfortable with english, even for a simple word as "loading"

anoek:
I put it on my todo list

PR #2988, opened 2025-03-09
RFC: Accesibility - integrate with main section

gsouf:
This PR intends to add a main section to each route for accessibility. Importance of a such a thing is outlined here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/main_role

 I faced some challenges related to how the app is architectured. This PR is a draft to discuss of the best implemention.

----

### Approach 1

At first I experimented with wrapping all routes contents directly in [routes.tsx](https://github.com/online-go/online-go.com/blob/b5244f99e617ac6069c935ffb8ba922c1f192d6d/src/routes.tsx#L143)  as per:

```diff
-                    <ErrorBoundary>{props.children}</ErrorBoundary>
+                    <ErrorBoundary>
+                        <main id="app-content" role="main">
+                            {props.children}
+                        </main>
+                    </ErrorBoundary>
```

**Why is this approach preferred? **
 - Including the main region as part of the layout requires less maintenance and ensures no route will get missed in the future.

However a few routes are using a fullscreen view via absolute positioning. Leading the main region to collapse to a height of 0px and that might have side effects on screen readers. 

**Why is this approach hard to implement?**
- because the layout is wrapping the router, the main element unaware of the route's display. There is no easy way to have a specific layout for a given route. It's complicated to implement a one fits all solution and even if we did it might cause more problems in the future.

----

### Approach 2

An alternative that implies very few structural changes is to have a `<MainContent>` component, and make sure that each route use it as a wrapper, as per the changes in this PR. It requires updating all routes components one by one to implement it so I just demonstrated it with 2 components: Game and SignIn.

Pros of this approach:
- No constraint in how each route is rendered

Cons:
- More maintenance
- Routes might potentially forget to add it

-----

### Note on usage of `id` for styling

I also noticed that [some routes are using an `id` property](https://github.com/online-go/online-go.com/blob/b5244f99e617ac6069c935ffb8ba922c1f192d6d/src/views/SignIn/SignIn.tsx#L202) to apply styling. They seem to be old ones, while recent ones seem to use a class name that starts with an uppercase letter. ie: `id="SignIn"` vs `className="SignIn"`. 

Because we want to have an `id` on the main element too for navigation purpose, it conflicts, forcing to nest the route one level deeper in the <main>.

However as it seems to be an old practice, if I go with approach 2 I'll update all routes individually, leaving a chance to change usage if `id` with `className`. Is it a desired change?

## Proposed Changes

  - [ ] wrap all routes content within a <main> element
  - [ ] provide accessibility shortcut to navigate to <main>

Note: the common practice for the id of the <main> element is to use `main-content`, but this id is already taken by the react root element

### DIFF
--- a/src/components/MainContent/MainContent.tsx
+++ b/src/components/MainContent/MainContent.tsx
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+
+interface MainContentProps {
+    className?: string;
+    children: any;
+}
+
+export function MainContent(props: MainContentProps): React.ReactElement {
+    return (
+        <main id="page-content" className={props.className} role="main">
+            {props.children}
+        </main>
+    );
+}
--- a/src/components/MainContent/index.ts
+++ b/src/components/MainContent/index.ts
@@ -1,0 +1,18 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


gsouf:
@anoek Do you have any concerns or feedback if I proceed with updating all the routes in the same way I did for the SignIn and Game components?

Also, what are your thoughts on replacing the `id` property with `className` ? Do you anticipate any potential side effects?

anoek:
> However a few routes are using a fullscreen view via absolute positioning. Leading the main region to collapse to a height of 0px and that might have side effects on screen readers.

If it makes it any easier, it'd probably work to have a minimum size for that main, maybe even 100vw/100vh

It's also possible we'd be able to revamp those displays that are using absolute positioning to use flexbox nowadays, but that'd be a bit of an undertaking most likely.


> Also, what are your thoughts on replacing the id property with className ? Do you anticipate any potential side effects? 

That should be fine. I think I initially started out doing that because technically it's a smidge faster, but I don't think we'll notice in practice, the difference is very minor.


> Note: the common practice for the id of the element is to use main-content, but this id is already taken by the react root element

Feel free to change that if you'd like

gsouf:
@anoek I think that setting a min height would work too, I was worried that it leads screen readers to misbehave, but I have done a bit of researches and I haven't found anything saying it would cause issues on screen readers.

gsouf:
Replaced by #2990

PR #2990, opened 2025-03-09
Wrap main content in <main> and add an accessibility menu.

gsouf:
## Proposed Changes

  - wrap all route content into <main>
  - add an accessibility menu to jump to <main> directly
  - in routes.tsx, renamed `Main` to `AppLayout` to avoid confusion with the new `MainSection` component


anoek:
Thanks!

PR #2992, opened 2025-03-10
A11y navbar

gsouf:
Quite a large update to the navbar to make it usable with the keyboard and accessible to screen readers.

## Proposed Changes

  - updated the markup to make it accessible (<ul>, <button>, etc...)
  - added labels, and aria- properties
  - updated the styles to accommodate with the new markup
  - added keyboard navigation on each menu
  - fixed "theme" heading in profile menu (note below)
  
## New keyboard navigation:

The keyboard navigation is built after the [Disclosure Navigation Menu with Top-Level Links patter](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation-hybrid/#mythical-page-content).

You can use tab to navigate between menus. As there is a link on the menu (top level link) we need to add an additional button to open the menu. When the menu is open we can navigate with tab or up/down arrow keys, and press escape to exit the menu. Note that unlike the recommended approach, we cannot move between menus using arrow keys (only tab). The only reason is that it's requiring more work, and time is limited. I might implement it later.

keyboardnav

## CSS changes

That's a lot of changes, I won't comment everything as I feel it is not necessary. The styles are not encapsulated and therefore a bit hard to maintain, but overall I made it so everything looks exactly the same as before, while  trying to make the style slightly more maintainable.Tested on various screen sizes and hamburger menu too.
I added a new jsx file/stylesheet for the Menu component as many things are going on here. 
Feel free to ask if you have question about the reason behind each change.

## Also fixed the theme heading in the setting menu 

in the settings menu (right of the topbar), the "theme" heading  was broken due to global styles on classes `.left` and `.right`. I fixed it with this PR.

image

## Things I haven't done:
- omnisearch navigation
- language and theme selection when logged out
- goban selection  
image

- chat, announcement, etc.. 
image
 
Note that I keep a todo list of these things and I plan to do them later. I give priority to navigating and playing a game.

### DIFF
--- a/src/components/GobanThemePicker/GobanThemePicker.tsx
+++ b/src/components/GobanThemePicker/GobanThemePicker.tsx
@@ -97,9 +97,10 @@ export function GobanBoardThemePicker(props: GobanThemePickerProperties): React.
         <div className="GobanThemePicker">
             <div className="theme-set">
                 {standard_themes.map((theme, idx) => (
-                    <div
+                    <button
                         key={theme.theme_name}
                         title={_(theme.theme_name)}
+                        aria-label={_(theme.theme_name)}
                         className={
                             "selector" + (selected.board === theme.theme_name ? " active" : "")
                         }
@@ -107,7 +108,7 @@ export function GobanBoardThemePicker(props: GobanThemePickerProperties): React.
                         onClick={selectTheme.current[theme.theme_name]}
                     >
                         {canvases.current[idx] && <PersistentElement elt={canvases.current[idx]} />}
-                    </div>
+                    </button>
                 ))}
             </div>
         </div>
--- a/src/components/GobanThemePicker/GobanThemePicker.tsx
+++ b/src/components/GobanThemePicker/GobanThemePicker.tsx
@@ -97,7 +97,7 @@ export function GobanBoardThemePicker(props: GobanThemePickerProperties): React.
         <div className="GobanThemePicker">
             <div className="theme-set">
                 {standard_themes.map((theme, idx) => (
-                    <button
+                    <div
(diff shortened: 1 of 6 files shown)


anoek:
Very cool, nicely executed!

anoek:
FYI: I didn't catch this at first but I'm seeing this on the console now:

```
In HTML, <li> cannot be a descendant of <li>.
This will cause a hydration error.
 <AppLayout>
      <Experiment name="v6">
        <Default>
          <div id="default-va...">
            <ErrorBoundary>
            <ErrorBoundary>
              <NavBar>
                <header className="NavBar">
                  <KBShortcut>
                  <span>
                  <nav>
                  <section>
                  <section className={"right "}>
                    <div>
                    <IncidentReportIndicator>
                    <ChatIndicator>
                    <TournamentIndicator>
                    <FriendIndicator>
                    <NotificationIndicator>
                    <GoTVIndicator>
                    <TurnIndicator>
                    <span>
                    <Menu menuId="settings" title={<span>} to="/user/view/1" className="profile de..." as="nav" ...>
                      <nav className="Menu profi..." aria-label="Profile">
                        <LinkWithRef>
                        <button>
                        <ul className="Menu-children" id=":r7:" ref={{current:null}}>
                          <ProfileAndQuickSettingsBits settingsNavLink={function ref}>
                            <_c2>
                            <_c2>
                            <_c2>
                            <li>
>                           <li>
                              <_c2 title="More options" centered={true} to="/settings/...">
>                               <li>
                  ...


```

gsouf:
woops, guilty of not checking the console 🙋 fixed in #2997

PR #2993, opened 2025-03-11
Setup Playwright with basic smoke test.  

GreenAsJade:
`yarn test:e2e` to run 

(will fire up frontend dev server locally if it doesn't find one at FRONTEND_URL, default localhost:8080).

Fixes not having e2e smoke test.

## Proposed Changes

  - Add Playwright infrastructure to support e2e tests under e2e-test/
  - Add an e2e smoke test to check registration, logout, login.
  
Change the front end it's looking at by setting FRONTEND_URL

 - Works for me unset (default to localhost:8080) and FRONTEND_URL=https://beta.online-go.com

Change the backend the usual way - OGS_BACKEND controls the backend choice of the dev server front end.

 - Works for me pointing backend at LOCAL and BETA.

Seed data not required, it creates a new user each time it runs.


gsouf:
That's very nice - I was thinking of having cypress or playwright to test the accessibility features I implemented. Will definitely be useful!

e2e-tests/smoke/smoketests.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        // Fill in registration form
+        await page.getByLabel("Username").fill(testInfo.newUsername);
+        await page.getByLabel("Password").fill("test");
+        const registerButton = await expectOGSClickableByName(page, /Register$/);
+        await registerButton.click();
+
+        // Verify successful registration
+        await expect(page.getByText("Welcome")).toBeVisible();
+
+        const userDropdown = page.locator(".username").getByText(testInfo.newUsername);
```
gsouf, line 52: @GreenAsJade I'd recommend against using class names to target items in e2e tests. They are intended for styling and might change over time. That can lead to difficulties to maintain tests. 
A more common approach is to have `data-` properties, or to use labels.

GreenAsJade: I agonised over this.

In the initial setup I really wanted to base the testing purely on _what the user sees_ (which is the PW approach anyhow) _and_ I wanted to see if it would work _without changing any client code_.

The answer is embedded in a comment near here:

```
        // Look for fake-link sign out button within the profile menu
        // First wait for the specific menu container...

        const profileMenu = page.locator(".Menu.profile .Menu-children");
        await expect(profileMenu).toBeVisible();

        // Kinda icky implementation dependent, but this "fake link" thing is not test friendly!
        const logoutButton = profileMenu.locator('span.fakelink:has-text("Sign out")');
        await expect(logoutButton).toBeVisible();
```

The challenges of finding this element were many :)

The username one I spent less time on, just wanting to get to the end of this round by that point :)

Maybe the goal of "no code changes" should be scrubbed, and we learn what sensible "put this here for testability" things are required.

I'd love to see a proposal (PR even?) with the client change and accompanying test...

GreenAsJade:
> That's very nice - I was thinking of having cypress or playwright to test the accessibility features I implemented. I see that playwright has component testing (but still experimental). Will definitely be useful!

(Note that we do already have jest doing some component level testing).

gsouf:
@GreenAsJade thanks for the note. I'll consider it, but I think it's better to rely on browser rendering for that.

e2e-tests/smoke/smoketests.spec.ts
gsouf: @GreenAsJade you mentioned it is difficult to find items such as the fakelink signout button. Just so you know, after #2992 it became a button rather than a span:

```html
<button class="MenuLink">
  <span aria-hidden="true"><i class="fa fa-power-off"></i></span>
  <span class="MenuLinkTitle">Sign out</span>
</button>
```

Would that make things easier?  I'm not familiar with playwright (I'm working with cypress usually), but after a quick research I found that playwright has this: `page.getByRole('button', { name: 'Sign out' })`. That seems to be even better, as we're looking for a role rather than exactly an element (for example a span could have the role button:  `<span role="button">`.  And everyday has something to teach us, I found that cypress also has such a getByRole, but via an additional package - that I'll start using from today.


From experience I found that when we implement accessibility for screen readers, it also makes e2e testing easier.

benjaminpjones: +1 one to the things said here, specifically:

- Good  to use what is available to the user, not "implementation details" like class names or IDs
- Good to avoid test-specific code
    - Caveat: accessibility labels often aid testability

I think the aria label approach is the best approach in this case

```
nav[aria-label="Profile"]
```

That way it's at least exercising a path that screen readers are making use of.

As for finding the Sign out button, I'm a little surprised `page.getByText("Sign out")` isn't sufficient.  Playwright just returns `null`?

GreenAsJade: PW returns multiple hits on that, and bombs because the context is expecting one.

The signout button implementation is "funky" and unsuprisingly confusing for PW, though I don't yet have an explanation for why it doesn't understand that the other instance is not actually visible.   I don't blame it though: I'm looking forward to merging gsouf's fix :)

PR #2994, opened 2025-03-11
E2e in ci

GreenAsJade:
Fixes not having e2e testing in CI

## Proposed Changes

  - Add playwright to the test target, see what happens.


GreenAsJade:
Just messing around really, would have been amusing if it worked, but it's not the right way to do it in anycase :)

PR #2997, opened 2025-03-11
remove nested <li> in navbar > settings > more settings

gsouf:
follow up of #2992

## Proposed Changes

  - remove `<li>` nested in `<li>`


anoek:
Thanks for the quick fix!

PR #2998, opened 2025-03-11
Fix navbar link to translations

gsouf:
in #2992 I haven't used `external` for the link to the translation (in tools dropdown) 

## Proposed Changes

  - Fix link to translation


anoek:
Thanks for the fix!

Issue #2431, opened 2023-11-25
Rank histogram


SamAmco:
As a player it would be really nice to know what percentile I am in the global distribution of ranks on OGS. This has been raised before in the forum here : https://forums.online-go.com/t/rank-distribution-2021/33010

This could be as simple as a number e.g. "you are in the top 35% of players" or as comprehensive as an interactive histogram like the one found on lichess here : https://lichess.org/stat/rating/distribution/classical

viczommers:
@anoek is there a better API socket than [v1 Django REST](https://apidocs.online-go.com/)? + doc/api_v1_doc.yaml

The best endpoint I found is: https://online-go.com/api/v1/players/?ordering=-rank&page=1&page_size=100 but I cant filter results by field without getting the whole result payload first.

I can write a simple scheduled job that will be traversing players list (once a day/week), fetching all rankings & plotting into frontend plot similar to `RatingsChart` inside `ProfileCard`. 

I just need either: 
(A.) an endpoint to store the json dump if we put it into the main client or 
(B.) I can just create a standalone FastAPI/Flask app, call it OGS stats...

viczommers:
@anoek Sure, I digged around the Lichess repo and replicated their .js chart + more or less followed their histogram logic.
It fetches`ratings.jsonl`, transforms it into 2 arrays: freq, ratings with `createRatingDistribution()` and plots results. There is also a variable to display user's rating.

- Chart https://viczommers.github.io/Go-Rating-Distribution/  
- Code Repo https://github.com/viczommers/Go-Rating-Distribution/

The only major modification I made is that each frequency/count represents sum of all players with rating in range between the current Bin label & excludes previous Bin label (Upper Boundary). For example, the 1600 point shows all players rated between 1501-1600.

Lichess logic was other way around (Lower Boundary) - Bin labeled 1600 shows all players rated in [1600, 1699] range.

Below are some assumptions for MVP:
- Glicko2 Ratings below 100 & above 3000 are excluded from Hist (extreme values)
- Provisional, high Deviation ratings are excluded from Hist as well
- User's rating will be plotted even if it is provisional/ high Deviation or outside of [100,3000] range, but the rendering is clipped at axis min,max.
- Histogram only for 'Overall' category (inline with current `ratings-chart`) + possible Kuy/Dan label conversion toggle.

Ideas for Next Step:
[...]

PR #3009, opened 2025-03-22
New Feature Request: #2431 Player Rating Histogram

viczommers:
Feature Request #2431 
## Source
- Borrowed from Lichess repo based on user feature request: https://lichess.org/stat/rating/distribution/classical
## Proposed Changes
- adds RatingsChartDistribution.tsx to Profile Section (User.tsx) below Ratings above current Game Chart
### To Discuss
- UX/UI experience? Best location/div to put in
- React component mounting raises warnings see browser log when testing
- adds chart.js to dependencies
- fetch() for /termination-api/ reading? vs get?

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -134,6 +134,7 @@
         "Seki",
         "semeai",
         "sepa",
+        "serie",
         "shadowban",
         "shadowbanned",
         "shownotesindicator",
--- a/src/components/RatingsChartDistribution/RatingsChartDistribution.styl
+++ b/src/components/RatingsChartDistribution/RatingsChartDistribution.styl
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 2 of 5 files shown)


viczommers:
Screenshot 2025-03-22 at 06 52 40
Screenshot 2025-03-22 at 06 54 19

anoek:
Nice!

* I think it's way too big right now, how would it look to the right of the rating grid?

* On mobile the labels make the graph look pretty funky, probably need to smart scaling, or maybe it'll be a non issue if we get a good looking one that's smaller to the right of the grid:
image


* I think the console warning is real, that should be taken care of I think: `Failed to initialize chart: Error: Canvas is already in use. Chart with ID '0' must be destroyed before the canvas with ID 'rating_distribution_chart' can be reused.` I think that means there's some cleanup that needs to be done, possibly within the chart.js logic

* Re chart.js, I'm not too wild about adding another charting library, we have nivo being used elsewhere already - how hard of a switch would it be to use that for consistency? https://nivo.rocks/line/ has some examples and "recipes" (examples) of how to do the marker, shaded area, and labels.

* fetch() for /termination-api/ reading? vs get? `get` for consistency

viczommers:
- Let's try Nivo! importing a whole new module for one chart feels like overkill...
- Nivo doesn't support dual axis, but there seems to be a workaround https://github.com/plouc/nivo/issues/2666
- Labels look just fine, 45-angled x-axis ticks is a standard behaviour when there is not much horizontal width, same in Plotly, etc
- tbh I prefer tighter fits, where distribution chart is vertically-biased (bigger height-to-width ratio/ exaggerated y-axis). It makes Kurtosis / Central tendency more obvious, compared to 16:9 Desktop view where tails appear much fatter
- I feel we need to put the chart into a collapsible button inside the Ratings. that would be the least disruptive tot he current workflow

viczommers:
- reworked for Nivo js chart 
- added Button collapse
- added Rating-to-Rank conversion, depending on toggle
Screenshot 2025-03-28 at 02 47 47

Screenshot 2025-03-28 at 02 50 18

anoek:
Nice, thanks!

PR #3017, opened 2025-03-26
Preferred settings -  rank restriction support

GreenAsJade:
Fixes not taking into account rank-restriction in "preferred settings" of ChallengeModal

## Proposed Changes

  - Detect when rank-restriction has changed, and allow new setting to be saved based on that
  - Extend the challenge-text-description to incorporate rank restriction information
  -- Change the name of the working variable in that function, because it's not HTML.
  -- Add utilities and comments to help understand what the units of "max and min rank" actually are.
 - Add an e2e test for this scenario


GreenAsJade:
Whoa - something wierd ^^ it says I added these commits 13 hours ago, but I just committed them now!

src/lib/rank_utils.ts
```diff
@@ -452,3 +452,55 @@ export function effective_outcome(
...
+    38: "9 Dan+",
+};
+
+/**
+ * Converts a rank selector index to its text representation
+ * @param index The rank selector index (5-38)
+ * @returns The text representation of the rank
+ */
+export function rankSelectorIndexToText(index: number): string {
+    const text = rankIndexToText[index];
```
anoek, line 501: Reckon we should have these translated, I *think* adding a `+` suffix to the translation of 9 Dan should suffice for any translations, so I think this table could be eliminated and then just call `rankString` with `{ rank: index}`, and if rank >= 38 append a `+` at the end?

GreenAsJade: I wonder why rankString doesn't do the + itself? 🧐

debug-screenshot.png
anoek, line 1: Is this `debug-screenshot.png` used? I'm thinking it was added by mistake. If not, we should probably add it via git lfs

anoek:
Merged in a squash commit of all of these to get rid of the png: https://github.com/online-go/online-go.com/pull/3033

PR #3021, opened 2025-03-28
A11y quickplay view

gsouf:
## Proposed Changes

  - overall improved accessibility for play page 
  - button to cancel a search has the text "search for game". Reverted to how it was months ago to have a separated text for "search for game" while the button content should be "cancel"
 
    - before:
    -  image
    - after: 
    - image
  - Cancel button will receive focus automatically to help screen readers

Notes that the accessibility work is not finished yet but remaining items will take so time and efforts:
- the overall accessibility of this page is not satisfying yet:
  - the games settings' "<select>" should be improved
  - the interaction with the matchmaking status is to be improved, ideal solution would be to improve the accessibility of the toasts and make them the first source of source here
- custom games list should be improved

### DIFF
--- a/src/components/LoadingButton/LoadingButton.tsx
+++ b/src/components/LoadingButton/LoadingButton.tsx
@@ -16,6 +16,7 @@
  */
 
 import * as React from "react";
+import { useEffect, useRef } from "react";
 
 interface LoadingButtonProps extends React.ComponentProps<"button"> {
     loading?: boolean;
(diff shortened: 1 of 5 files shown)


gsouf:
@viczommers I simply re-applied how it was done before as I'm more focusing on the accessibility and I don't want to shift the efforts to another direction. But happy to apply anything that makes sense. I also think it's doesn't look very natural that way. 

So unless someone feels like implementing something more fancy, we could go that:

image

PR #3029, opened 2025-03-30
Make it a toggle-button like "Show Plot".

GreenAsJade:
Fixes intrusive wide button with graphic for rarely used feature.

## Proposed Changes

  - Make it consistent with "show plot"

Screenshot 2025-03-30 at 5 28 31 pm (2)

### DIFF
--- a/src/views/User/User.tsx
+++ b/src/views/User/User.tsx
@@ -396,20 +396,27 @@ export function User(props: { user_id?: number }): React.ReactElement {
                                                 />
                                             </h3>
                                             {renderRatingGrid(show_ratings_in_rating_grid)}
-                                            <button
-                                                className="btn-group sm toggle-chart"
+                                            <div
+                                                className="toggle-container"
                                                 onClick={() =>
                                                     setShowDistributionChart(!showDistributionChart)
                                                 }
                                             >
-                                                <i
-                                                    className="speed-icon fa fa-bar-chart"
-                                                    title={_("Global Distribution")}
-                                                />
-                                                {showDistributionChart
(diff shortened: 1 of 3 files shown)


viczommers:
I quite like the small [hist chart icon](https://fontawesome.com/v4/icon/bar-chart) ngl - it really communicates the feature quite well

anoek:
If we're going to make it consistent with the show plot, should the state be persistent too, like show plot?

I do like the hist chart icon too, would it look good at the end of the string here (and maybe at show plot for consistency?) Might not look right, but something worth trying maybe.

src/views/User/User.tsx
```diff
@@ -396,20 +396,27 @@ export function User(props: { user_id?: number }): React.ReactElement {
...
-                                                    : _(" Compare to Global Distribution")}
-                                            </button>
+                                                <div className="toggle-indicator">
+                                                    {showDistributionChart ? "▼" : "▶"}
+                                                </div>
+                                                <span className="toggle-label">
+                                                    {showDistributionChart
+                                                        ? pgettext(
+                                                              "label for button to hide the global distribution chart",
+                                                              "Hide distribution",
```
viczommers, line 412: "Distribution" should be capitalised?

GreenAsJade: I copied the capitalisation of the original  😝

As I did so I thought to myself "the author must have felt that Global Distribution is a proper noun, wheras distribution is just any old thing that we're closing" ;)

viczommers: Lol, I read it in Catch22 Yossarian’s inner monologue voice

PR #3030, opened 2025-03-30
Initial mocked up ChallengeModalBody with a few AI generated tests.

GreenAsJade:
Fixes # being too scared to touch ChallengeModal

## Proposed Changes

  -  Have some unit tests.
  - ALSO make some html-for tweak that appeared broken in ChallengeModal (!)


src/components/ChallengeModal/ChallengeModal.test.tsx
```diff
@@ -0,0 +1,543 @@
...
+
+    it("renders challenger color selection", () => {
+        render(<ChallengeModalBody {...defaultProps} modal={mockModal} />);
+        const colorSelect = screen.getByLabelText(/your color/i);
+        expect(colorSelect).toBeInTheDocument();
+        expect(colorSelect).toHaveValue("automatic");
+    });
+
+    it("renders rank restrictions when enabled", async () => {
+        const openModeProps = {
```
benjaminpjones, line 363: Instead of explicitly annotating each property, you can mark the entire object `as const`. This eliminates the need to use `as SomeUnionType` everywhere.  

For example:  

```typescript
const myFoo = {
    someLetter: "a" as LetterType,
    someDigit: 1 as DigitType,
};
```  

can be simplified to:  

```typescript
const myFoo = {
    someLetter: "a",
    someDigit: 1,
} as const;
```  

This approach remains compatible with the type `{ someLetter: LetterType, someDigit: DigitType }`.

src/components/ChallengeModal/ChallengeModal.test.tsx
```diff
@@ -0,0 +1,543 @@
...
+ */
+
+import * as React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
+import { ChallengeModalBody, ChallengeModalProperties } from "./ChallengeModal";
+import { JGOFTimeControl, JGOFTimeControlSpeed, JGOFTimeControlSystem } from "goban";
+import { RuleSet } from "@/lib/types";
+import { post } from "@/lib/requests";
+
+type ChallengeModes = "open" | "computer" | "player" | "demo";
```
benjaminpjones, line 25: I was thinking export the original type, but I realize this type isn't necessary at all if you use `as const` below.

src/components/ChallengeModal/ChallengeModal.test.tsx
```diff
@@ -0,0 +1,543 @@
...
+
+// Mock preferences first to prevent data.setDefault error
+jest.mock("@/lib/preferences", () => ({
+    get: jest.fn(() => false),
+    set: jest.fn(),
+    defaults: {},
+}));
+
+// Mock misc module to prevent infinite recursion
```
benjaminpjones, line 34: Hmm.. this shouldn't be necessary.  I think you can just remove it, but if the tests do trigger a loop, there's probably a bug in `dup`.

Also, we now have [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone), so maybe OGS no longer needs to implement custom clone logic.

GreenAsJade: The ended up here because ChatGPT got itself into an infinite loop trying to mock something related 😝

I didn't dig into whether there is actually a runtime loop.

PR #3032, opened 2025-03-31
RatingsChartDistribution Fix

viczommers:
## Fixes
- left axis tick value overflow (players) + toLocaleString formatting
- fixed kuy-dan rank tick label precision (bottom axis) by setting true in boundedRankString(value, true) from rank_util
- added translation support
## Proposed Changes
  - added h4 heading with percent calc
  - current wording is "You are better than x% of players"
  - tried "You are in the Top (1-percentile)% of players", but slack consensus  is it'll be confusing for "bottom" % of players. 
    - explicitly saying "you are in the Bottom%" seems not inclusive/demotivating...
    - removing h4 completely for bottom half of players seems wrong, so settled on "better than x%"
Screenshot 2025-03-31 at 01 02 11
Screenshot 2025-03-31 at 01 01 38


merowin:
If I may comment, I think a more neutral title would be "\<name\> has higher rank than \<percent\> of players".

GreenAsJade:
Will we get "my red line" on "other people's chart" soon? :)

anoek:
Nice work, thanks! I agree with @merowin about the more neutral title, I'll make that change post merge.

viczommers:
@anoek @merowin Im afraid it'll raise issues re rank vs rating context translation but I declare "creative bankruptcy" & we cant please everyone...

PR #3035, opened 2025-04-01
Css sanity smoke test

GreenAsJade:
Fixes missing some css-inflict layout problems

## Proposed Changes

  - Screenshot-compare settings and profile page for a new user

** Note, adds esnext to package.json


anoek:
I'm *pretty sure* the "module": "esnext" in the tsconfig won't hurt anything since I don't think that'll change any of the bundling behavior.

PR #3041, opened 2025-04-05
Challenge Modal State Rework

merowin:
Update the state management of `ChallengeModal` (part 1), + general improvements

## Proposed Changes

  - Add explicit type for the state (constructed s.t. it compiles with minimal changes)
  - replace usages of `upstate` by `setState` for update bindings for: `bot_id, demo_name, game_name, handicap`
  - add interface `ChallengeModalInput` to reduce code duplication
  - convert `string` to `int` before handicap input values get into the state
  - remove unused module exports
  - fix input type for function `updateSystem` in `TimeControlUpdates.ts`
  - sanitize handicap when coming from local storage (currently it can have type number or string)


src/components/ChallengeModal/ChallengeModal.types.ts
```diff
@@ -120,3 +131,40 @@ interface TimeControlConfig {
...
+export type ChallengeModalChallengeSettings = ChallengeDetails & {
+    invite_only?: boolean;
+    boardWidth?: number;
+    boardHeight?: number;
+    game: ChallengeDetails["game"] & { speed?: JGOFTimeControlSpeed };
+};
+
+export type ChallengeModalDemoSettings = DataSchema["demo.settings"];
+
+export type ChallengeModalState = {
```
merowin, line 152: I tried to make this type such that it compiles with minimal changes.

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -1749,7 +1758,7 @@ export class ChallengeModalBody extends React.Component<
                 width: this.state.challenge.game.width,
                 height: this.state.challenge.game.height,
                 ranked: true,
-                handicap: this.state.challenge.game.handicap !== "0",
+                handicap: this.state.challenge.game.handicap.toString() !== "0",
```
merowin, line 1761: Typescript complained that a comparison between number and string is superfluous, so I added this conversion. I'm not sure what's going on though, did I get the type wrong, or is this comparison needed for legacy reasons?

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -907,8 +914,8 @@ export class ChallengeModalBody extends React.Component<
             const tc = updateSystem(
                 this.state.time_control,
                 "simple",
-                this.state.challenge.boardWidth,
-                this.state.challenge.boardHeight,
+                this.state.challenge.boardWidth ?? 19,
```
merowin, line 917: I suspect that these are supposed to have an actual value when the `forceTimeControlSystemIfNecessary` is called, but not 100% sure.

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -752,10 +742,27 @@ export class ChallengeModalBody extends React.Component<
...
+        this.setState((prev) => ({ ...prev, conf: update_fn(prev.conf) }));
+    update_challenge_settings = (update_fn: UpdateFn<ChallengeModalChallengeSettings>): void =>
+        this.setState((prev) => ({ ...prev, challenge: update_fn(prev.challenge) }));
+    update_demo_settings = (update_fn: UpdateFn<ChallengeModalDemoSettings>): void =>
+        this.setState((prev) => ({ ...prev, demo: update_fn(prev.demo) }));
+    update_game_settings = (update_fn: UpdateFn<ChallengeModalGameSettings>): void =>
+        this.update_challenge_settings((prev) => ({ ...prev, game: update_fn(prev.game) }));
+
+    /* direct fn updates */
+    update_bot_id = (id: number) => this.update_conf((prev) => ({ ...prev, bot_id: id }));
```
benjaminpjones, line 757: Doesn't have to be in this PR, but part of the pain, I'm sure you noticed, is the heavy nesting of `state`.  It may be worthwhile to flatten the state object in the future to make this easier.

https://react.dev/learn/choosing-the-state-structure#avoid-deeply-nested-state

src/components/ChallengeModal/ChallengeModal.tsx
benjaminpjones: If you're sure this is a number, then the original code looks like a bug to me.  I'd probably just do numeric comparison:

```suggestion
                handicap: this.state.challenge.game.handicap !== 0,
```

merowin: I see now that string values can get into state because the input is a <select>. I intend to fix it by converting the string to number before it gets into the state.

src/components/ChallengeModal/ChallengeModal.tsx
merowin: Also maybe the nesting can be reduced by extracting (or splitting) parts as individual components. I feel like the readability suffers a lot from the many different use-cases, and the giant file size.

merowin:
Converting to draft again, because I believe that I need to handle string values for handicap that come from local storage.

anoek:
Love where this is going, thanks!

PR #3042, opened 2025-04-05
Change config for jest so we don't have to use --experimental-vm-modules 

GreenAsJade:
... which is unix only.

Fixes jest bomb on Windows (hopefully)

## Proposed Changes

  - follow AI's suggestion to use ungap/structured_clone polyfil instead.


benjaminpjones (approved):
I am a little surprised that the diagnosed issue is that the flag is windows-only, since Jest documents windows support [here](https://jestjs.io/docs/ecmascript-modules).

Aside from windows support, this also seems like a solid improvement to the `package.json` file :D

cc: @merowin could you confirm Windows works now?

merowin:
Can confirm that it works on my PC now (windows 10), thanks

PR #3043, opened 2025-04-05
Fixes #3025 AIReview when current move index is out of range of Scores

### LINKED ISSUE #3025: Ai Score showing NaN and undefined
**Describe the bug**
The score estimate of the ai in certain games and languages is showing NaN and undefined in certain cases.

**To Reproduce**
Not sure about reproducing so I’ll link some example games

https://forums.online-go.com/t/scoring-system-error/56013?u=shinuito

https://online-go.com/game/72795712

W+NaN in the final position

Image

For https://online-go.com/game/73301136

it seemed like according to the forum thread having the language option 繁體中文 could be causing the undefined 

I can confirm that this does happen with that language choice.

viczommers:
Fixes https://github.com/online-go/online-go.com/issues/3025
when `move_number` index is out of range of `scores` array in `score = scores[move_number]`
## Proposed Changes
  - If score is "falsy"/undefined/nan, reverts to the last numeric value in `scores.slice(0, move_number + 1)` array 
  - If no numeric elements in `scores` reverts first to 0 with `|| 0 ` and then to its initial **"falsy" val (pass)**
```
  if (last_score && last_score !== 0) {
      score = last_score;
  }
```
  - This is a typescript workaround: because `score` type is set to number only (Type 'undefined' is not assignable to type 'number')
  - Tradeoff: will ignore Edge Cases where score was "falsy"/undefined/nan, but the closest available value from `scores.slice(0, move_number + 1)` (i.e. past scores) is 0; should be fine because:
    - (a.) regardless this is not the latest score (of the current move on the board) and the counter-factual for Player is to display "undefined/nan", which is IMO **"more truthful"** than just to assume that the move value is 0;
    - (b.) How often 0 score moves are expected at the end of a go game, probabilistically speaking? 

I would like @GreenAsJade to sense check


anoek:
Thanks for the fix!

PR #3049, opened 2025-04-12
fix detected of saved restrict rank prefs

GreenAsJade:
Fixes not detecting that a current no-restrict-ranks setting matches an old saved one

## Proposed Changes

  - Detect rank restriction during match consistently with the way it is loaded


anoek:
Thanks!

PR #3050, opened 2025-04-13
Challenge modal refactor

merowin:
Fixes #

## Proposed Changes

  - refactor update bindings for private, invite_only, `rengo, rengo_casual, rengo_auto_start`
  - remove `input_value_warning` from state and turn it into a derived quantity
 
## Notes
  - Updating `private` flag in demo mode does not write `ranked: false` into `state.demo` anymore (because I believe that demo boards don't use a ranked setting).
  - `update_rengo_auto_start` does not write empty string into the state anymore, instead the value in the input element is adjusted on render.


src/components/ChallengeModal/ChallengeModal.utils.ts
```diff
@@ -207,7 +207,7 @@ export function rejectionDetailsToMessage(details: RejectionDetails): string | u
 
 // For legacy reasons, handicap can have a string value
 // when these challenge details come from local storage.
-export function sanitizeChallengeDetails(challengeDetails: ChallengeDetails): ChallengeDetails {
+export function sanitizeChallengeDetails<T extends ChallengeDetails>(challengeDetails: T): T {
```
merowin, line 210: This change occured to me during the process. It is not necessary here, but I believe it's worth keeping.

src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -1093,8 +1105,15 @@ export class ChallengeModalBody extends React.Component<ChallengeModalInput, Cha
                                 <div className={"checkbox"}>
                                     <input
                                         type="number"
-                                        value={this.state.challenge.rengo_auto_start}
-                                        onChange={this.update_rengo_auto_start}
+                                        // It's clearer to display blank ("") if there is no auto-start.  Blank means no autostart, the same as zero.
```
merowin, line 1108: moved this comment here from the update function

anoek:
Thanks!

PR #3057, opened 2025-04-19
extract utils function getDefaultKomi, refactor update binding of komi op…

merowin:
Refactoring challenge modal continued

## Proposed Changes

  - extract utils function `getDefaultKomi`
  - add type guard for komi option
  - refactor update bindings for `komi_auto` (and remove string values for komi)
  - sanitize legacy komi values from local storage
  - add unit test for sanitize functions
  - rename type


anoek:
Thanks!

PR #3059, opened 2025-04-21
Add a colon to display directive

GreenAsJade:
Fixes wondering whether that line is broken or not :)

## Proposed Changes

  - Add a colon to a display stylus that didn't have one.


anoek:
hah, yeah I wish stylus wasn't so accepting. Thanks for the catch :)

PR #3062, opened 2025-04-21
321Go to Fundamentals in learn-to-play-go

TCGWim:
Fixes #

## Proposed Changes
16 lessons in learn-to-play-go Fundamentals


anoek:
FTR you don't need to create a new PR when changes are made, you can just push up the new changes and it'll re-run the tests and just append your changes to the pull request. Also, if you're still working on a PR you can mark it as a draft PR so folks can see it but know it's not ready for merging yet.

TCGWim:
Ah, OK, thanks.
I had already started a new PR and this passes all checks.

anoek:
I'm open to adding some select topics to the Fundamentals section as needed, but I don't think we should dump everything there. We don't want it to be an overwhelming to beginners, we want the fundamentals to run them quickly through the absolute basics, then if they want they can progress through the rest. 

I did like your idea in your screenshot you had sent earlier:
321GoOGS



One other thing to note, we need to avoid the use of the word "Suicide", we should refer to that as Self Capture. Similarly, kill should just be capture.

viczommers:
"Self-capture" can be confusing for non-native speakers. Sacrifice/illegal moves might be better?

anoek:
> "Self-capture" can be confusing for non-native speakers 

That's why we have translations :)

> Sacrifice/illegal moves might be better? 

Sacrifice suffers a similar problem to the problematic words, and illegal move is too broad I'd say.

anoek:
Great work!

PR #3069, opened 2025-04-27
Use the warning_system_messages route to get and show players their warning system messages

GreenAsJade:
Fixes traceability of these things.

## Proposed Changes

  -  Show outstanding warning system messages on Home (pointless if there is no bug, only used if the message isn't popping up)
  - Show warning system messages on Profile in Activity (selectable all or just outstanding)
  - e2e test to show that warning ack still works after all this.

Needs https://github.com/online-go/ogs/pull/2071


src/views/Overview/EXV6Overview.tsx
```diff
@@ -172,6 +173,7 @@ export class EXV6Overview extends React.Component<{}, OverviewState> {
                                     offered_moderator_powers={user.offered_moderator_powers}
                                 />
                             )}
+                            <WarningSystemMessagesPane onlyOutstanding={false} />
```
anoek, line 176: We need this on the Not EXV6 version too right? (... I think it's far past time we finish that "experiment up", maybe just add an option to have the stuff on the right or not - but that's something for a different day/pr :) )

GreenAsJade: I'm thinking "if you have trouble with Warning Sytem Message, turn on this interface and take a look" ;)

anoek: Oh okie dokie

PR #3071, opened 2025-04-28
Say "Report System Messages" for the section that contains these.

GreenAsJade:
Fixes confusing language about what is an "acknowledgement"

## Proposed Changes

  - Use "Report System Messages" as the title of the thing that contains warnings and acks, rather than "Warning System Messages"


GreenAsJade:
Hah this is going to bomb smoke _again_!

PR #3090, opened 2025-05-19
fix: append seek graph modal to container to ensure correct text color

mreinhardt:
## Proposed Changes

  - append seek graph modals to seek graph container to ensure correct text color
  - currently text color in **dark** _and_ **accessible** mode is black but should be white for readability
  - could just add styling, but appending to the seek graph container seems like the correct approach so that it inherits existing `body.dark #main-content` and `body.accessible #main-content` text styling

## Optional TODO

  - This implementation currently falls back to the existing implementation of appending the modal to the body for robustness if for whatever reason the seek graph container is unavailable. Now that I'm thinking about it, it's probably better to add `SeekGraphConfig.modalContainer` to specify the container element to add the modal to, and use the implemented defaults if the config is not set. May submit a follow-up commit on my own or definitely if requested during review.

### Incorrect modal text color in *CURRENT* implementation

Screenshot 2025-05-19 at 11 27 49

### Correct modal text color in *PROPOSED* implementation

Screenshot 2025-05-19 at 11 28 13

### DIFF
--- a/src/components/SeekGraph/SeekGraph.tsx
+++ b/src/components/SeekGraph/SeekGraph.tsx
@@ -894,7 +894,7 @@ export class SeekGraph extends TypedEventEmitter<Events> {
             list.appendChild(e);
         }
 
-        document.body.appendChild(list);
+        appendModalElement(list);
         this.moveChallengeList(ev);
     }
     closeChallengeList() {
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for hunting that one down and figuring out a fix!

PR #3094, opened 2025-05-23
Store AID filter params in URL for ease of sharing and refresh page

GreenAsJade:
Fixes having to retype the parameters all the time

## Proposed Changes

  - Write filter updates to the URL params, and load them from there


anoek:
Nice

PR #3095, opened 2025-05-24
Challenge modal update bindings

merowin:
## Proposed Changes

  - refactor update bindings for width, height, rules, komi and challenger color
  - add type guards for some of these updates
  - change the state type a bit


src/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -1817,8 +1847,8 @@ export class ChallengeModalBody extends React.Component<ChallengeModalInput, Cha
 
             const settings = {
                 rank: user.ranking,
-                width: this.state.challenge.game.width,
-                height: this.state.challenge.game.height,
+                width: this.state.challenge.game.width ?? -1,
```
merowin, line 1850: This change was necessary due to changing width & height to being nullable. With -1 values, no bot is available (as far as I can tell)

src/components/ChallengeModal/ChallengeModal.types.ts
```diff
@@ -140,21 +139,56 @@ export type ChallengeModalConf = {
     restrict_rank: boolean;
 };
 
-export type ChallengeModalGameSettings = ChallengeDetails["game"] & {
+export type GameInput = {
```
merowin, line 142: copied type over so I can modify it more easily.

PR #3101, opened 2025-06-03
Challenge modal state updates

merowin:
## Proposed Changes

  - update bindings: board size, disable analysis, rank restrictions, demo board players
  - remove upstate functions


anoek:
Glad to see that upstate function gone, thanks for all of your hard work!

PR #3111, opened 2025-06-09
Show move categorisation

GreenAsJade:
Fixes wondering which moves were categorized which way.

## Proposed Changes

  - Add a dropdown to show the moves of each category
  - Show these moves on the graph
  - Make them clickable to get to them.


anoek:
Slick :)

PR #3116, opened 2025-06-19
Avoid mutating ai_review branch contents while updating highlights and heatmaps

GreenAsJade:
Fixes # AI review summary changing as you click through the game

## Proposed Changes

  - Take copies of data needed (not much) instead of mutating.

### DIFF
--- a/src/lib/ai_review_move_categories.ts
+++ b/src/lib/ai_review_move_categories.ts
@@ -259,6 +259,7 @@ function categorizeFullReviewNew(
         console.log(
             ` (2) score after blue move would be played -> \nai_review.moves[${move_index}].branches[0].score: `,
             score_after_blue_move.toFixed(3),
+            `(${ai_review.moves[move_index].branches[0].moves[0]["x"]},${ai_review.moves[move_index].branches[0].moves[0]["y"]})`,
         );
         console.log(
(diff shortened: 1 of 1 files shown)


anoek (commented):
Isn't this the code that makes it so our branch estimates aligns with the observed next move estimate? If so, I don't think we want to change that, despite MrMtal's many complaints about it - it's the correct thing to do and we already went down that path 5 years ago, without having those synchronized it's a confusing user experience, also synchronizing it is more accurate.

GreenAsJade:
I dunno - it totally breaks the AI review summary experience.    The actual number of moves in the ai_review data structure appears to change.

It may be that with recent changes the AI review summary is updating more often than it did before, so we're only noticing now, but "something needs to be fixed: the net result is that we're holding a broken ai_review in memory".

To see the problem point a dev front end without this change at beta game 18465 and click back from the end of the game watching the summary

I don't really see how it makes sense that an ai_review outcome would change just by clicking through the game?

That's what's happening.   We're saying "oh, now that you click there, I see that I got that wrong before".   Why didn't we just get it right in the first place?

anoek:
That does sound pretty broken, was there a game this was particularly noticeable in?

GreenAsJade:
I pushed up an tweak here just to confirm that the move played was what we think it was.  Which it was.

GreenAsJade:
We won't do this.

PR #3119, opened 2025-06-23
new section Beginner Level 1

TCGWim:
Fixes #

## Proposed Changes

  - new section: Beginner Level 1 with 37 new lessons
  - numbered the sections and the lessons in this new section
  - minor changes in earlier lessons

### DIFF
--- a/src/views/LearningHub/LearningHub.styl
+++ b/src/views/LearningHub/LearningHub.styl
@@ -71,6 +71,10 @@
                 text-transform: uppercase;
                 themed(color, shade2);
                 padding-right: 3rem;
+
+                .section-number {
+                    padding-right: 1.5rem;
+                }
             }
 
             > .contents {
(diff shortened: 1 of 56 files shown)


anoek:
Is the naming convention update intentional? Previously we haven't prefixed the titles with "4.1", "4.12", etc. My gut says we should probably leave that off the actual title, and if we want/need them then we should generate them automatically so if we move stuff around they remain accurate (and so that it's consistent throughout the learn to play page).

TCGWim:
Yes, the numbering of items is intentional. Since there are so many lessons and the titles of lessons re-occur in different sections, it helps the students if the lessons are numbered (it was a request from one of the students).
And yes, I agree that the numbering should be generated, but I did not know how to do this.
I have a similar issue with the structure of the lessons, the list of LearningPages: sometimes I combine two sets of exercises into one lesson and in many cases I add an instructional exercise as first exercise in a lesson, to replace the 321Go instructions. In both situations, I have to re-number the pages (Page01, Page02, etc.) and add the added pages to the array of LearningPages.

anoek:
Heya @TCGWim , I've taken a stab at automatic numbering for the lesson sections, see #3126 to see how you like it.

PR #3121, opened 2025-06-24
Moved mod tools into moderator-ui

GreenAsJade:
Fixes mod tools dangling out here.

## Proposed Changes

  - Move ModTools into moderator-ui
  - Tweak tsconfig.js to pick up submoduled moderator-ui first - seemed to be necessary to get props properly identified.


anoek:
> But I left it in this way around because it seemed to make sense: maybe check? 

Yeah I think you're right

PR #3127, opened 2025-07-04
Added a "utillity test", which is there to let us run games 

GreenAsJade:
…rather than really trying to test stuff.

## Proposed Changes

  -  Adds the e2e-utils group of tests, with run-game in it.
  
To use: edit the sequence of moves in run-game.ts, then run that specific test.

I find the UI the most convenient:

`yarn test:e2e:ui`

Screenshot 2025-07-04 at 12 45 13 pm (2)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,4 +5,5 @@
     "[stylus]": {
         "editor.formatOnSave": false
     }
+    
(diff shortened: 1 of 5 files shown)


anoek:
Pretty cool!

PR #3129, opened 2025-07-05
Provide a checkbox to say if you want to notify your friend request decline.

GreenAsJade:
Fixes people feeling bad about declining.

Slightly controversial

https://forums.online-go.com/t/feature-request-not-to-send-declined-friend-request-notification/56862/22?u=greenasjade

## Proposed Changes

  -  Checkbox to opt into decline friend request notification
  

Needs https://github.com/online-go/ogs/pull/2098

### DIFF
--- a/e2e-tests/friends/decline.ts
+++ b/e2e-tests/friends/decline.ts
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


GreenAsJade:
Good ole smoke test - stopped me from messing up the pretty home page.

I probably would have found this before committing if I had followed my own rule of "always write e2e for new features" :D

anoek:
The other option is to add an "ignore" button which would remove the pending request from your displayed list (maybe have a page under settings to show all that you've ignored), that'd prevent the same user from sending you another request too. 

Also the *follow* person mentioned in that thread might help to side step this problem a bit, it's something that has seemed like a good idea for many years now I just haven't gotten around to it.

(FTR I've no objection to this reject-without-notification thing either, though I find myself wondering if that shouldn't just be the always case, not sure why we tell people they've been rejected in the first place, that's an *old* decision I made years ago probably without much thought.)

GreenAsJade:
Yeah, I submitted this as kind of "the least thing we could do given the feedback in that thread".

It doesn't take away anything, and it lets people be comfortable declining.

The whole "what should this feature _really_ be, and is that actually a "follow" feature" is ... a valid consideration ... for another day :)

anoek:
Do you think we should make this the default behavior instead of a checkbox? Just not notify the other person when they are rejected? I'm kinda thinking maybe so, but I haven't thought through it beyond a current gut check.

GreenAsJade:
I made it the default of the checkbox, to be unchecked.  Not providing the option would be "taking something away", which I thought to avoid right now ... if we're taking something away, it'd probably be time to do the more systematic reappraisal eh?

PR #3132, opened 2025-07-19
Add `data` means to persist debug setting

GreenAsJade:
Fixes having to set the debug on every time

## Proposed Changes

  -  Look in `data` as well as window for debug enabled.

### DIFF
--- a/src/lib/data_schema.ts
+++ b/src/lib/data_schema.ts
@@ -189,6 +189,10 @@ interface DemoSettings {
     komi?: number;
 }
 
+interface DebugSchema {
+    [module: string]: boolean;
+}
+
 /**
  * Prefixes every member of a type.
  *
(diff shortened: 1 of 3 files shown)


anoek:
Since we have a beta server would it be feasible to use that instead of mocking? The https://github.com/online-go/online-go.com/pull/3133 that it also suggested had code in it to clean up sockets and whatnot, so if that was necessary it'd seem that at least some socket stuff is working.

I'm generally a fan of not-mocking when feasible as with mocks you can get into situations where your tests are not actually testing what you want them to test, and you have to maintain the mocks as whatever you're mocking is updated which somewhat defeats the purpose half the time.

That said, if we need to mock we need to mock. I just don't want to do it because an AI is taking the easy route route like they tend to do, I want to do it because we've analyzed the situation and determined that we can't actually use real sockets because of `<reason>`.

GreenAsJade:
It's jest that's failing, so ... we have to mock, right ... we can't talk to beta from jest?

I'm more worried about "why did this start failing now".   I will pursue that ... but note that it started failing across the board, not just this PR.

... ooh, actually it's a @benjaminpjones  test, so he'll almost certainly have good insight.

anoek:
yeah I was thinking this was a playwright thing, I reckon jest we have to mock stuff.

GreenAsJade:
mmm ... still the wierd "but why now" ?   🤷‍♀️

GreenAsJade:
Update:  @benjaminpjones took a look and said it seems fair as well.

PR #3146, opened 2025-07-31
"Add to Library" button to directly add a game to user's collection

csessh:
I created a feature request issue a long while back and I said I'd give it a try. I never got around to it until now. I hope it's still something relevant.

## Proposed Changes

  - An additional tool in the in-game screen to allow user to directly add the game to their SGF collection. Currently, one may either download the SGF file to re-upload to a collection or one may track a link to the game elsewhere.
  - SGF filename is customizable and defaults to game ID if none provided.
  - At the moment, it doesn't preserve chats. I don't think it matters but happy to add if it's preferable.

## Screenshots

The added button  in the in-game tool section:

Screenshot From 2025-07-31 12-47-42

The pop up modal:

Screenshot From 2025-07-31 12-48-07

Screenshot From 2025-07-31 12-48-12

Screenshot From 2025-07-31 12-48-15

Notification:

Screenshot From 2025-07-31 12-49-10

The game:
Screenshot From 2025-07-31 12-49-28

image


csessh:
Here's the stale issue: https://github.com/online-go/online-go.com/issues/2210

anoek:
I think it would be much better if instead of making an SGF clone a link to the game is added to the library, which previously wasn't possible but I just threw together an endpoint for you that's now available on the beta server. 

```
post("library/<ID>", {game_id, collection_id, name})
```

should now create an entry within the library that points directly to the game.

there's some issues currently with displaying the player name information on the library page, and I haven't fully vetted things, but what's up there should be enough to update this PR.

I'm leaving on a trip here so it'll be a bit over a week before I can get back to all of this, so sorry in advance for the slow reactions and merging of this, but thanks for undertaking the feature, it's definitely useful!

csessh:
`post("library/<ID>", {game_id, collection_id, name})`

This endpoint may not be functional yet. The POST request I send to this endpoint returns 200, but nothing shows up on the SGF Library.

```

    addToCollection = () => {
        if (!this.state.collections || this.state.uploading) {
            return;
        }

        this.setState({ uploading: true });

        const user = data.get("user");
        console.log("DEBUG: Adding game to library", {
            user_id: user.id,
            game_id: this.props.gameId,
            collection_id: parseInt(this.state.selectedCollectionId),
            name: this.state.fileName,
            url: `library/${user.id}`,
        });

        post(`library/${user.id}`, {
            game_id: this.props.gameId,
            collection_id: parseInt(this.state.selectedCollectionId),
            name: this.state.fileName,
        })
            .then(() => {
                if (this.props.onSuccess) {
                    this.props.onSuccess();
                }
                this.close();
            })
            .catch((error) => {
                errorAlerter(error);
                this.setState({ uploading: false });
            });
    };

```

I believe this request is constructed correctly.

anoek:
Working great over here, sorry there were troubles with the beta server I'll verify again once some dust settles over here.

Thanks for all your work on this!

PR #3151, opened 2025-08-04
simplify challenge modal

merowin:
I tried to remove function `next()` but that created too many type errors at once. While working on it, I found some simplifications though. I also made some changes to the type of the state, but I'm not sure where I want to go with this.


anoek:
Thanks!

PR #3160, opened 2025-08-11
Opponent rank selector : actual rank range

GreenAsJade:
Fixes confusion about what is the user really selecting.

## Proposed Changes

  - Show the effective rank range (actual) that they are selecting.
  
https://forums.online-go.com/t/opponent-rank-settings-bug/55714/30?u=greenasjade


anoek:
Yeah that's better

PR #3161, opened 2025-08-12
Anshulxyz/fix wonky private game info

### LINKED ISSUE #2102: Wonky game info on private games
**Describe the bug**
When game is set as private and you're not one of the players playing, the game info looks like this

Screenshot 2023-01-07 at 17 06 39

**To Reproduce**
1. Go to some game that is set as private
2. Click on 'Game information" on right side pop-up menu

**Expected behavior**
Have the game info look like it normally does but have the details hidden. 

**Additional context**
Very minor thing, only happens when game is set as private so won't cause any harm. Just looks weird.

anshulxyz:
Fixes #2102 

https://github.com/online-go/online-go.com/issues/2102

## Proposed Changes

  - The change makes the "Creator" label in the table always visible, just like all the other labels are.

Demo:
This is the view os someone who should have access to private game looks like. As expected.
Screenshot 2025-08-13 at 12 11 24 AM

And this is the view now for someone other than the invitee:
Screenshot 2025-08-12 at 11 46 10 PM

### DIFF
--- a/src/views/Game/GameInfoModal.tsx
+++ b/src/views/Game/GameInfoModal.tsx
@@ -236,7 +236,7 @@ export class GameInfoModal extends Modal<Events, GameInfoModalProperties, GameIn
                         </dd>
                         <dt>{_("Creator")}</dt>
                         <dd>
-                            {this.props.creatorId && (
+                            {!!this.props.creatorId && (
                                 <Player icon rank user={this.props.creatorId} />
                             )}
                         </dd>


GreenAsJade:
How about explaining why it's  "not normal".   You could put  `(Private Game)` in the `Game` field.

It'd be nice to hav Black vs White layout symmetrical, also.

I suspect that we should _not_ be trying to render a `<Player>` at all in this instance   If we have to, then maybe a `<Player>` with the icon turned off would do the trick (IIRC there's a prop for that?)

It _could actually_ be nice to have an option on `<Player>` that is "just show their color".   I bumped into this very recently: did a hacky "if we have the actual player then render the `<Player>` if not render a string with their color".

I was too lazy, and it seemed too niche, to modify `<Player>` at the time.  But now we have two cases...

I'd also be inclined _not_ to render anything in the creator field.  I'm not a fan of that `...` that `<Player>` renders if the player is not valid, because it looks like a link, which tempts me to click on it.

anoek:
I think the ... and the non aligned player names are coming from the `Player` class, I think we'd need to null protect and just use `""` instead of `<Player..` for those. 

I do second changing the title to "Private Game" would help clarify what's going on a lot in any case.

anshulxyz:
Any reason why the playright screenshot test failed here.
But it passed when I ran it on my repo? Refer: https://github.com/anshulxyz/online-go.com/actions/runs/16949439971

I've made a change for Game Title, I figured that easied way for me to find if the User has access to view the game or not is to use the `creatorId` being equals to `0`.

Result:
Screenshot 2025-08-14 at 2 36 49 AM

anoek:
The failures look completely unrelated to your stuff so no worries

anshulxyz:
Screenshot 2025-08-14 at 4 43 38 PM

I've updated the creator field to not get rendered when the `creatorId` is `0`. Looks much cleaner now.

1. Let me know if I am in the right direction.
2. Is there a better way to write that ternary operator?

GreenAsJade:
I think that ternary is exactly how we'd do it, though perhaps it doesn't need parens, and would be more compact without?

Now there's just the big space in front of `White [?]`

anoek:
> Now there's just the big space in front of `White [?]`

I think if the `icon` is removed from the `Player` tag when there's nothing to display will fix that

anshulxyz:
> perhaps it doesn't need parens

These parens? They are there because of _prettier_.
Screenshot 2025-08-15 at 12 28 01 AM

>I think if the icon is removed from the Player tag when there's nothing to display will fix that

I found that `icon` is a boolean, so I went ahead and used the `creatorId` for it. Looks more aligned now. Here's the result:

Screenshot 2025-08-15 at 2 13 36 AM

Why the double exclamation mark `!!`? Because I am trying to convert a number to a boolean. Are there better ways?
```tsx
                                            icon={!!this.props.creatorId}
```

anshulxyz:
I refactored the code. I believe this addresses the parens thing, and code looks more compact now.

anoek:
Thanks @anshulxyz !

GreenAsJade:
Hah darn prettier :)   Nice solution.

I wonder why there is still space inserted for the player icon?

I was actually meaning that 

`Black [?] vs                  White [?]`

would nice look like

`Black [?]    vs    White [?]`

Same space each (1 space?) side of `vs`  😅

PR #3164, opened 2025-08-14
Shadow theme selection support

anoek:
Adds a drop down to our theme settings to support different shadow styles:

image

Low: (what we used to have a few days ago)

image

Mid: 
image

High: 
image

And of course custom because we love options here:

image

image


kevincoleman:
@anoek thanks for making this adjustment. It probably should have been my task as I stirred this pot, but I appreciate your taking the time to help ease the conversation. 🙏🏻

PR #3171, opened 2025-08-19
Add a modal that asks announcers not to spam

benjaminpjones:
Hopefully, this encourages tournament announcers to create a more organized central post, rather than spam the announcements feature

If the user tries to create more than one post, they are shown this message.  It won't completely block them, just aims to educate.

Screenshot_20250818_214102-2

### DIFF
--- a/src/components/AnnouncementSpamModal/AnnouncementSpamModal.tsx
+++ b/src/components/AnnouncementSpamModal/AnnouncementSpamModal.tsx
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as React from "react";
+import { _ } from "@/lib/translate";
+import { Modal } from "@/components/Modal";
+import { Markdown } from "@/components/Markdown";
+
+interface Events {}
+
+interface AnnouncementSpamModalProperties {
+    onProceed: () => void;
+}
+
+const MODAL_CONTENT = `## 📢 Rule: One Announcement Per Event
+
+To keep the community clear and avoid disruption:
+
+- **Post only one announcement per event.**  
+  Do not create multiple announcements for individual games, streams, or related items.  
+
+- **Consolidate all details and links** into a single, well-organized hub.  
+  Acceptable places include:  
+  - a **forum post** (recommended)  
+  - an **external event website** (if you maintain one)  
+
+- **Use your hub post to direct users** to all related content.  
+
+This policy reduces pop-ups, prevents redundant alerts, and keeps the community space accessible for everyone.`;
+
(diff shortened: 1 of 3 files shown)


benjaminpjones:
Related discussions:

- https://forums.online-go.com/t/announcement-center-guidelines/43300
- https://forums.online-go.com/t/too-many-pop-ups-in-the-initial-page/23620/123

benjaminpjones:
@GreenAsJade is that signal legit?  I did build/run these changes locally, so I think it shouldn't have broken a "smoke" test.

GreenAsJade:
> @GreenAsJade is that signal legit? I did build/run these changes locally, so I think it shouldn't have broken a "smoke" test.

I'll take a look.  Note that you have to very deliberately run smoke tests locally - they need to be run in a docker that has the right browser config so that screenshots match.

They are quite a small and basic set of "do things look right?" tests, so we've just been letting the CI catch them, and it's normally been solid.

That said, it's still a bit annoyingly breaky-sometimes.

This was my quick reply ... off to look at it now :)

GreenAsJade:
OK - I can't tell whether this is "changed behaviour" or just "someone didn't leave Beta in a clean state, and/or we need to tidy up the test to cover it".

The problem is that the logged-out-screenshot has an announcement in it, where the reference one does not

logged-out-initial-page-expected

logged-out-initial-page-actual

If announcements _are_ supposed to appear when logged out, then the solution is we need to update the screen-mask to not-look in the announcement area, so that we are somewhat robust against there being an announcement, or not, on Beta when this test runs.

(Even better would be a feature that lets us clear announcements, which we could run before running this test !)

If announcements are _not_ supposed to appear when logged out, then this is a bug.

I'll write up how I found these: it's basically "click on the failed test link, look in the details for the step that says 'upload e2e results', click on that and then the download link for the screenshots"

GreenAsJade:
Note:  this fail can be avoided by making sure that there are no annoucements active at the time of pushing the commit when the CI runs :)

benjaminpjones:
That helps, thanks! I had looked at the "Run" tab, but missed the "Upload results" tab.

So in summary, that failure is not caused by my code changes. (Only my manual tests 😂)

GreenAsJade:
> (Only my manual tests 😂)

Cool - if we're expecting announcements to be there for logged out, then I better robustify the test 🤪

anoek:
@kevincoleman has some ideas in the works that might obviate this by collapsing multiple announcements down (for the case where there's legitimate need for multiple links, one per game kind of thing). The other common use case I see is repeated announcements for events and whatnot, this is out fault because we limit the maximum duration to 6 hours, but with the evolution coming for the announcement UI I think part of that should be to lift that limit and display those differently (as events, not sent as an pop up notification, that sort of thing).

anoek:
Marking as draft for now while the announcement updates cooks for a bit

benjaminpjones:
nice! If those announcement updates are POR, feel free to close this one. I wasnt sure if that was a longer term project or not

GreenAsJade:
#3173 should fix the CI here.

GreenAsJade:
LOL I was just yesterday looking at an announcement of a game thinking "hah, we need this".

What's the current plan/status?

benjaminpjones:
Yeah I think it would be helpful, but my understanding is there's another solution in the pipes (which seem fine too)

GreenAsJade:
Maybe it's time to wack this in.   In line with our vision that OGS be a relaxing spam-free place.

A replacement can come...

anoek:
I think I'm going to riff off of what @kevincoleman proposed in terms of notification interfaces and see if the AI can put something together in the background while I work on some goban code.

I think we need to better support the need for multiple notifications in terms of linking multiple boards and whatnot, not display a message telling them they shouldn't do that, so I think it's on us not them, so I don't want to merge this. That said, I'll mark it as draft so it nags me to get a solution moved forward, I'll close it when things are better :)

anoek:
New announcement updates are in effect which I think addresses the announcement problems and better enables using one announcement per event/stream. See https://github.com/online-go/online-go.com/pull/3220 and https://github.com/online-go/online-go.com/commit/a426743d506064d151de37a3b414224d249038a7

PR #3177, opened 2025-08-20
add demo board modal

merowin:
## Proposed Changes

  - Extract Demo Board Modal into its own component.

PR #3180, opened 2025-08-23
Add e2e test for demo board

anshulxyz:
As mentioned in #3177 Here's a basic test with structure for creating more e2e tests for demo board.

This refers to the comment from slack:
>I think a test of the basic functionality is useful, like "open dialog, enter some values, click create demo board and assert that we navigate to the new demo board".

cc: @merowin, @GreenAsJade 

Let me know if I should change anything.


anshulxyz:
Two things i have faced issue with:

1. The name and rank for player fields don't have "id", so can't pick them to fill them.

2. Once I am on the Demo board page, board is created, how do I assert that the Goban is of size 19x19 (or something else), I didn't find anything relevant to that in the HTML.

anoek:
Thanks!

PR #3188, opened 2025-08-28
Make a safe use goban controller for when there might be no context

GreenAsJade:
Fixes red screens on joseki and puzzle

## Proposed Changes

  - have a useGobanControllerSafe for Game to use which has context and the others to use which pass goban


anoek:
Whoops, thanks!

PR #3189, opened 2025-08-28
Add a smoke test for joseki page basic render.

GreenAsJade:
Fixes not having something that makes sure Joseki page is kinda live.

## Proposed Changes

  - Add a basic screenshot comparison
  
(get rid of unused old png)


GreenAsJade:
Note: this is in the CI, so deploy at a time when we can handle CI breakages :)

PR #3190, opened 2025-08-29
Update e2e test for demo board, make it data driven

anshulxyz:
This is a continuation from #3180 

@GreenAsJade I've made the tests table-based since I realised that the two tests were almost identical. This makes creating new tests with different settings much easier.

**NOTE**: While working I found that the Rules is not correctly reflected. This issue is also faced by Users, as documented here https://forums.online-go.com/t/demo-board-with-chinese-rules-shows-japanese-rules-instead/57801

I've left a failing assertion statement (for "Chinese") in the code.


GreenAsJade:
It's a work of art.

Ship it.

GreenAsJade (approved):

PR #3204, opened 2025-09-07
Update Turkish association name to match the original

pektezol:
Turkey's go association recently updated their official name from 'Turkish Go Players Association' to 'Turkish Go Association'.

Source: https://www.tgod.org.tr/


anoek:
Thanks for the update!

PR #3223, opened 2025-09-20
VIBE Share private collection

GreenAsJade:
Fixes  not being able to share libraries of sgf with select folk.

## Proposed Changes

  - Use the new backend ACL to add people to share private collections with
  - Fix up the gnarly interface for making new collections that are private
  -- Hide the selector until they start creating a new collection
  -- Use a toggling icon and label to be clear what is happening

Needs https://github.com/online-go/ogs/pull/2140

### DIFF
--- a/src/components/CollectionSharingModal/CollectionSharingModal.styl
+++ b/src/components/CollectionSharingModal/CollectionSharingModal.styl
@@ -1,120 +1,144 @@
-.CollectionSharingModal
-    .header
-        display: flex
-        justify-content: space-between
-        align-items: center
-        padding: 1rem
-        themed(border-bottom-color, shade3)
-        border-bottom: 1px solid
+.CollectionSharingModal {
+    .header {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        padding: 1rem;
+        themed(border-bottom-color, shade3);
+        border-bottom: 1px solid;
 
-        h3
-            margin: 0
-            font-size: 1.2rem
+        h3 {
+            margin: 0;
+            font-size: 1.2rem;
+        }
 
-        .close
-            background: none
-            border: none
-            font-size: 1.5rem
-            cursor: pointer
-            padding: 0
-            width: 2rem
-            height: 2rem
-            display: flex
-            align-items: center
-            justify-content: center
+        .close {
+            background: none;
+            border: none;
+            font-size: 1.5rem;
+            cursor: pointer;
+            padding: 0;
+            width: 2rem;
+            height: 2rem;
+            display: flex;
+            align-items: center;
+            justify-content: center;
 
-            &:hover
-                themed(background-color, shade5)
+            &:hover {
+                themed(background-color, shade5);
+            }
+        }
+    }
 
-    .body
-        padding: 1rem
-        max-height: 60vh
-        overflow-y: auto
+    .body {
+        padding: 1rem;
+        max-height: 60vh;
+        overflow-y: auto;
 
-        .share-with
-            margin-bottom: 2rem
+        .share-with {
+            margin-bottom: 2rem;
 
-            h4
-                margin: 0 0 1rem 0
-                font-size: 1.1rem
+            h4 {
+                margin: 0 0 1rem 0;
+                font-size: 1.1rem;
+            }
 
-            .loading
-                text-align: center
-                padding: 1rem
-                themed(color, shade2)
+            .loading {
+                text-align: center;
+                padding: 1rem;
+                themed(color, shade2);
+            }
 
-            .empty
-                text-align: center
-                padding: 1rem
-                themed(color, shade2)
-                font-style: italic
+            .empty {
+                text-align: center;
+                padding: 1rem;
+                themed(color, shade2);
+                font-style: italic;
+            }
 
-            .shared-items
-                .shared-item
-                    display: flex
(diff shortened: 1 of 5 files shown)


src/components/CollectionSharingModal/CollectionSharingModal.styl
```diff
@@ -0,0 +1,115 @@
+.CollectionSharingModal
```
anoek, line 1: Let's have it refactor this to use curly braces like the rest of the stylus files

Also have it use our themed variables not put in it's own color codes, it messed up at least the dark themed disabled color. For the buttons, it shouldn't be coloring them anyways, it should be using the css classes like other buttons.

src/views/LibraryPlayer/LibraryPlayer.styl
```diff
@@ -33,9 +33,46 @@
...
+                font-size: font-size-big;
+                color: text-color;
+            }
+
+            .share-collection-btn {
+                display: flex;
+                align-items: center;
+                margin-left: 1rem;
+                gap: 0.5rem;
+                background: primary;
```
anoek, line 60: I think it tried here, but it missed the themed call

src/views/LibraryPlayer/LibraryPlayer.styl
```diff
@@ -55,6 +92,43 @@
...
+            flex: 1;
+            cursor: pointer;
+
+            span {
+                padding-right: 1rem;
+            }
+        }
+
+        .share-collection-btn {
+            background: #007bff;
```
anoek, line 113: needs to be using theme

anoek (commented):
image

Let's have it fix that alignment, I realize it was just using what we had, but what we had looks awful next to that button, vertically aligning should make it look fine though.

Also, I think this icon: 
image
is the kind of icon for share these days



---
And for bonus points while we're there, if you wanna tell it to make this not suck maybe it'll come up with something better (I imagine this is my fault from however long ago, but looking at it today, it's bad)

image

GreenAsJade:
> image
> Let's have it fix that alignment, I realize it was just using what we had, but what we had looks awful next to that button, vertically aligning should make it look fine though.

I can't reproduce this ^^

Screenshot 2025-09-21 at 2 15 17 pm (2)

It gets pretty ugly any narrower, but that's how it already was.

I was surprised to see your screenshot, because I already had to take care of this post-VIBE - that div now has `align-items anchor-center`.

src/components/CollectionSharingModal/CollectionSharingModal.styl
GreenAsJade: Done - at least it doesn't move around now

GreenAsJade: (I made both the input and the button stick to the top, and the open/private selector takes the same space whether visible or not, and moved over "indented" just a bit")

anoek:
> I can't reproduce this ^^ 

Looks like it happens on firefox, not on chrome

It also looks like that whole element is new, we might just have it move the share button over with the other buttons and get rid of the redundant collection name

image

anoek:
When viewing a shared collection the game date is messed up:

image


vs the owners view:

image

anoek:
Marking as draft so it's not nebulous when we should re-review, let me know when you're ready

GreenAsJade:
Yea, lets see if you're OK with this.  I left the header in, for "reasons" - it just needed `center` instead of `anchor-center`

(The header is kind of a solid place to hang both the private marker and the share button - it's not massively in conflict with the "breadcrumb" thing at the top )

anoek:
Oh gotcha, yeah I'm still seeing this on firefox:

image

GreenAsJade:
Interesting - although that is pretty much exactly the same as I see, it looks worse here because your Blue buttons are ove closer.  🧐

The way they are, it's as if the share button should be much closer to the lock.

I wonder if we care?

GreenAsJade:
Screenshot 2025-09-24 at 9 50 43 pm (2)


That's what I was looking at :)

anoek:
> I wonder if we care?

Yep, plenty of folks are going to want to have that aligned. Or the button moved over with the other buttons and the duplicate label removed.

GreenAsJade:
What do you mean by "aligned"?    It's center-line aligned with the header and the lock :p

The thing is: the blue buttons are about games in the collection

The header is about the collection, which is both locked and shareable.

anoek:
I mean if you have stuff that close in what looks to be a row, they should all be vertically aligned - so everything on the left and everything on the right. Or just get rid of the stuff on the left and move the button to the right, either way.

GreenAsJade:
Yeah its painful because for 99.999% of people, it looks nothing like one row, because they don't have the AI buttons extending over from the right

Darn incremental UI design 🧐

anoek:
image

Dude, I'm tellin ya, that right there, that's grounds for war with any designers we have on the site 😜

anoek:
Also any agent worth it's weight in salt should be able to fix that in quick order, is it giving you fits? Try a different one?

anoek:
Much better, war averted, you have my thanks!

GreenAsJade:
You detected my cunning plan for flushing out designers to come and fix our site :p 

rocky-and-bullwinkle-moose-and-squirrel

Issue #2026, opened 2022-10-07
Parts of nicknames recognized as coordinates


Sofia-mal:
https://online-go.com/game/44531856

![immagine](https://user-images.githubusercontent.com/100384001/194497262-e6f6e88c-c140-4167-a554-e4df53b09d63.png)

Not an important issue at all, but I think [P1.04](https://online-go.com/player/1039370/P1.04)'s nickname is not a reference to P1 on the board and it should not be recognized as a coordinate.

kobakitty:
Same thing happens when mentioning users with usernames like "player 55" or "user 1213" - it got turned into link for players who have the corresponding user-id

Sofia-mal:
I don't think it's the same thing and what you are saying is probably intended :)

Jlll1:
I'd like to work on this. How would we like to see this resolved? I'd go with making the user be implicit about embedding anything into his message. E. g. instead of converting `P1` or `player 55` into a link by default, require the user to wrap the link, like `[[P1]]`, `((P1))`, `[[player 55]]` etc.

benjaminpjones:
TBH I don't think this is a great issue to start with - I think the system messages are generated on the backend which isn't open source.

As far as ideas go - there are actually player escapes already. See [chat_markup.tsx](https://github.com/online-go/online-go.com/blob/421b711821d74817cbbcba39b22dda5ee8a7596a/src/components/Chat/chat_markup.tsx#L122-L161).  It has a few ways to add a player such as `%%%PLAYER-12345%%%` and `@"benjito"`.  We could probably have the server bot send one of these instead of the plaintext username.

@GreenAsJade any thoughts?

anoek:
I think this should be all front end, and so a good first issue to work on?

I believe the relevant code is here: https://github.com/online-go/online-go.com/blob/devel/src/views/Game/GameChat.tsx#L595-L616

benjaminpjones:
I thought the solution was to change what `system [9d]` adds to chat. But I think I may have read that wrong.

I see now that the proposed solution is to wrap moves like `[[M1]]`.  It seems like that will reduce discoverability, but I'm not against it.

GreenAsJade:
As a side note, we can of course easily change what `system [9d]` says, though it is indeed on the back end.

An also side note:

> Same thing happens when mentioning users with usernames like "player 55" or "user 1213" - it gets turned into link for players who have the corresponding user-id

[...]

PR #3226, opened 2025-09-24
Change chat coordinate matcher to require whitespace or line start/end

### LINKED ISSUE #1321: URI Fragment not properly parsed for github line links.
**Describe the bug**

The part after the # specifying line number is not included in the link when shared in chat. I have two guesses on what is breaking this:

- L78 is parsed as a board coordinate
- numeric anchors aren't handled correctly

**To Reproduce**

Paste the following in chat:
 https://github.com/online-go/online-go.com/blob/b3fde06d24cb33e62d68ecb091baa48a366e528e/src/components/Chat/chat_markup.tsx#L78

**Expected behavior**
The whole link should be hyperlinked.

**Screenshots**

lazerwalker:
Fixes #1321, partial fix for #2026

The initial problem here was that URLs like `https://github.com/online-go/online-go.com/blob/b3fde06d24cb33e62d68ecb091baa48a366e528e/src/components/Chat/chat_markup.tsx#L78` get mangled in chat, such that the "L78" doesn't become part of the URL.

The URL regex is fine. It turns out, because the coordinate matcher fires before the URL matcher, the "L78" gets flagged as a potential coordinate, then rejected for not being a valid coordinate. This results in it getting wrapped in a `<span>` tag, which causes it to not be included in the URL regex.

The ultimate fix, as @bhostetler18 describes in #2026 discussion, is probably a smarter system for prioritizing different regex matchers and making sure a given substring only gets targeted by a single matcher. However, for now, modifying the coordinate regex to only accept coordinates that are surrounded by whitespace on either end (or the start/end of a line) would effectively fix this issue, plus several of the usecases presented in #2026, including the one that triggered opening the issue.

Didn't see any obvious easy path to writing a regression test (the existing chat_markup tests just test the URL regexes in isolation, and it would require a bunch more setup to test the full "comment to parsed comment" path, which currently doesn't have any existing tests to crib setup from), but manually tested. Happy to write that test if folks think it's high-impact, but it strikes me the most likely next time someone will touch this code is if someone wants to do the bigger refactor.

In case it's unfamiliar, the regex syntax here is negative lookahead/lookback, which will match the whitespace/line-start/line-end characters without consuming them. We could also write this as `(^|\s)([a-zA-Z][0-9]{1,2})(\s|$)`, which might be slightly more performant (in a way that's almost certainly negligible for this usecase), but would require us to then properly handle the lookahead/lookbehind capture groups since we'd be consuming those tokens that we otherwise want to print as whitespace. IMO that would make things more complicated than just using maybe slightly more obtuse regex syntax. Open to suggestion!

### DIFF
--- a/src/views/Game/GameChat.tsx
+++ b/src/views/Game/GameChat.tsx
@@ -598,8 +598,8 @@ function MarkupChatLine({ line }: { line: ChatLine }): React.ReactElement {
             <React.Fragment>
                 {chat_markup(body, [
                     {
-                        split: /(\b[a-zA-Z][0-9]{1,2}\b)/gm,
-                        pattern: /\b([a-zA-Z][0-9]{1,2})\b/gm,
+                        split: /((?<=^|\s)\b[a-zA-Z][0-9]{1,2}\b(?=\s|$))/gm,
+                        pattern: /(?<=^|\s)\b([a-zA-Z][0-9]{1,2})\b(?=\s|$)/gm,
                         replacement: (m, idx) => {
                             const pos = m[1];
                             if (parsePosition(pos, goban).i < 0) {


lazerwalker:
Hmm, is this a known flaky playwright test? I wouldn't expect a 2-regex change in the chat component to affect smoke tests for the logged-out landing page.

GreenAsJade:
It's typically not flakey, but it does rely on the Beta server being up and clean.

I'll take a look (you can too - there are instructions in /e2e/README to find out what went wrong)

The easy part of these is something like "download the result screenshots, it's usually easy to see whats wrong in those"

GreenAsJade:
logged-out-initial-page-actual

FWIW E2E is not expecting this Register button to be there.

@anoek   ?

lazerwalker:
Ah, cool, thanks! I couldn't figure out how to download the screenshots, but those docs are great.

Looks like the issue is the new 'register' button added to the navbar, which looks like the last commit to hit main. So a valid (or, rather, expected and intentional) test failure, where someone who isn't me in a PR that isn't this one just needs to update the screenshots.

(Ah, see you beat me to it :joy:)

anoek:
Awesome, thanks!

PR #3227, opened 2025-09-24
New screenshots for logged out

GreenAsJade:
Fixes CI failing

## Proposed Changes

  - Update screenshots with new ones for logged out... that include the new Register button.
  
(Is it wierd having two Register buttons on the same page?  🧐 )


anoek:
> (Is it wierd having two Register buttons on the same page? 🧐 ) 

Yeah maybe..

PR #3240, opened 2025-09-29
Attempt at making the landing page for learn to play go not so overwhelming

ogsai:
Our current layout was made when we only had a handful of lessons. Now that the collection has grown significantly the original layout is no longer sufficient. This proposed layout organizes our view into collapsible sections and shows up to 3 lesson cards per row, along with a progress bar through that section.

image

Would be interested in how you feel about it @TCGWim 

It's also up on https://beta.online-go.com/learn-to-play-go to play around with

PR #3246, opened 2025-09-30
Dynamic loading of LearningHub

ogsai:
Built on top of the learning hub patch

This patch splits the LearningHub code out into a separate bundle which is loaded on demand.


anoek:
Merged into #3240

PR #3252, opened 2025-10-02
Fix for issue #3031

### LINKED ISSUE #3031: Difficulties navigating the puzzle page on the phone
**Describe the bug**
On my iPhone, after creating a puzzle, it is difficult to navigate to the puzzle that I created because of the below issues.

1. It’s hard to scroll down to the puzzle because a simple touch of the screen would cause the puzzles to be moved up and down instead of scrolling the page.
2. The edit button is cut out from the page and it’s difficult to scroll right because of problem 1.
3. Clicking the edit button on the puzzle doesn't do anything and it doesn't bring me to the puzzle page.

**To Reproduce**
1. Create a new puzzle in your puzzle collection
2. Try to navigate to the puzzle on your phone

**Expected behavior**
Able to navigate to the puzzle and open the page successfully.

**Screenshots**
Screenshots will probably not be very helpful so I'll add a video if needed.

 - Device: iPhone 15 Pro Max
 - OS: iOS 18.3.2
 - Browser: Chrome, Safari
 - Version: Chrome 135.0.7049.35

**Additional context**
Because it's so hard to scroll, once you scroll down in the puzzle page it's hard to do anything else anymore.
For the edit button, I could make it work on the Safari after clicking many times.

TimaGribanov:
Fixes #3031

## Proposed Changes

  - [SortablePuzzleList.tsx] Added an activation constraint to `TouchSensor`. This constraint has a delay of 100 ms and a 0 px tolerance. With this, it's allowed to scroll through the list of the puzzles without unintentionally moving them. To move an element, one must press the item for 100 ms.

### DIFF
--- a/src/views/PuzzleCollection/SortablePuzzleList.tsx
+++ b/src/views/PuzzleCollection/SortablePuzzleList.tsx
@@ -179,7 +179,7 @@ function SortablePuzzleListContainer({
         // Without this activation constraint, the "Edit" button doesn't work
         // because dnd swallows click events.
         useSensor(MouseSensor, { activationConstraint: { distance: 5 } }),
-        useSensor(TouchSensor),
+        useSensor(TouchSensor, { activationConstraint: { delay: 100, tolerance: 0 } }),
     );
     return (
         <DndContext onDragEnd={onDragEnd} sensors={sensors}>


anoek:
Awesome, thanks for digging into that and getting it fixed!

PR #3254, opened 2025-10-02
Automatically focus submit move button

JonKo314:
When playing on my notebook with just the touch pad, I find moving the mouse to the submit button and back to the board the whole (live) game quite cumbersome.

That's why this pull request
- adds a setting to automatically focus the submit move button,
- adds a symbol to the submit move button when focused (if the setting is turned on), to show that it can be activated by pressing the enter key.

---

There's a small issue though:
- If the setting is turned on and the button loses focus (e.g. by focusing the chat before submitting), once the button is clicked it gets focused (by the browser) and briefly shows the symbol too, before the whole button disappears.

I'm not happy about that, but I don't know how to prevent it neatly.

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -38,6 +38,12 @@
     .submit-button {
         white-space: nowrap;
     }
+    .autofocus-submit-button:focus::after {
+        content: ' ⏎';
+    }
+    .autofocus-submit-button:active::after {
+        content: none;
+    }
 }
 
 .condensed-game-information {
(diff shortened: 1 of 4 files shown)


src/views/Game/Game.styl
```diff
@@ -38,6 +38,9 @@
     .submit-button {
         white-space: nowrap;
     }
+    .autofocus-submit-button:focus::after {
+        content: ' ⏎';
+    }
```
anoek, line 43: I think if you add this after the above rule:

```
    .autofocus-submit-button:active::after {
        content: none;
    }
```

it'll fix what you're seeing with clicking the submit button when it doesn't already have focus (as it will have focus but it'll also be active during the click, and the active wins since the rule is later).

anoek (commented):
This looks great. I think it's fine as is, but I also think we might be able to address your nit about the symbol briefly showing up when clicking after having lost focus too with the below code

PR #3256, opened 2025-10-04
Incorporate auto-suspension e2e test

GreenAsJade:
Fixes not having a test for this.

## Proposed Changes

  - Point to the test in moderator-ui
  
Needs 

https://github.com/online-go/ogs/pull/2147  for the actual functionality
and moderator-ui deployment for the actual test.

(oops I pushed onto main of moderator-ui ! )


GreenAsJade:
The build failure is because this needs the updated moderator-ui component.

PR #3259, opened 2025-10-06
Fix for issue #3012

### LINKED ISSUE #3012: Chat setting can't be changed when watching games
**Describe the bug**

I play a lot of correspondence games at the moment, so I switched my default chat setting to personal.
In my own games, I can change the chat setting by clicking next to the input text line.

But when watching other games, this button is not there.
Since my default setting is personal, I'm isolated and can only talk to myself when watching live games.. which is a bit strange =)

**To Reproduce**
1. Visit https://online-go.com/settings/game
2. Set default game chat mode to Personal
3. Visit https://online-go.com/observe-games
4. Open live game, chat is stuck in personal mode

**Expected behavior**
The chat mode button should be there, as in my own games.

**Screenshots**

Image

 - OS: Debian
 - Browser Firefox
 - Version 137

TimaGribanov:
Now, the chat in the watched games is automatically set to main, so that a user won't be locked in Malkovich or private mode if one of those is set in the user's settings.

Fixes #3012

## Proposed Changes
  - Now, the chat in the watched games is automatically set to main, so that a user won't be locked in Malkovich or private mode if one of those is set in the user's settings. The new `useEffect` which checks that the current user is not the one playing in the opened match and that the user is not a moderator; in that case, the log is set to `main`.
  - **NB!** I haven't checked how it behaves if the user is moderator because I tested it running my local instance against the production backend and using my own user.


src/views/Game/GameChat.tsx
```diff
@@ -87,6 +87,12 @@ export function GameChat(props: GameChatProperties): React.ReactElement {
         };
     }, [goban_controller]);
 
+    React.useEffect(() => {
+        if (!userIsPlayer && !data.get("user").is_moderator) {
+            setSelectedChatLog("main");
```
anoek, line 92: I think this needs to be

 ```
     goban_controller.setSelectedChatLog("main");
```

right?

File: src/views/Game/GameChat.tsx [hunk 2]
```diff
@@ -87,6 +87,12 @@ export function GameChat(props: GameChatProperties): React.ReactElement {
...
+        }
+    }, []);
```
anoek, line 94: I think this needs to be

```
 }, [userIsPlayer, goban_controller]);
```

src/views/Game/GameChat.tsx
TimaGribanov: Yes, it seems more reasonable. And I've added an else branch because I noticed that sometimes on my games, it could show me the main chat instead of the default one

anoek:
Thanks!

PR #3260, opened 2025-10-07
Fix for issue #3219

### LINKED ISSUE #3219: Resignation button in zen mode
There is no way to resign in zen mode.

After passing two times in a row, a button allowing the player to resign would be nice.

Instead you have to pass until the opponent appetite for stones cease or AI suggests resignation or leave the board out of boredom.

Thanks in advance !

TimaGribanov:
Allowing the resignation button in zen mode

Fixes #3219

## Proposed Changes
  - Removed `&& !zen_mode` from the `show_cancel` initialisation in `PlayControls.tsx`


anoek:
Thanks!

PR #3264, opened 2025-10-07
Put suspend test where it can be found

GreenAsJade:
Fixes breaking e2e not finding the test

## Proposed Changes

  - Relocate it into the moderator test section


GreenAsJade:
Feedback on review: not bad, though I think it misinterpretted the comment.

#4 is a good observation, I wonder why it didn't use that helper when it wrote the test 😅

#1 may be an improvement, though I've had strife waiting for things to appear.

I'd still merge this though, no showstoppers there.

PR #3270, opened 2025-10-14
Add buttons to support sending an "appeal approved" message

GreenAsJade:
Fixes not being able to send a message to a person accompanying the restoration of their account

## Proposed Changes

  - A button to send a system PM with the message while restoring the account.


GreenAsJade:
Wierd eh - critical bug but seems to work 🧐

I think we're supposed to be checking the message visibility checkbox actually - if it's hidden, then there's no point in having a "Restore account" button showing.

So - wrong review in this case.

I guess to be fair it should have been provided the proper statement of requirements :)

PR #3271, opened 2025-10-14
Resing/Cancel button is now available in zen mode for portrait layout too

### LINKED ISSUE #3185: Option in preferences or confirmation dialog for auto-resignation
Hitting the back button during a game could trigger a resignation confirmation dialog and bring us back to /play to start a new game when we do resign.

I zen mode it is unpleasant having to leave zen mode to hit the resign button. The notification message after hitting the back button is unclear and it would be nice for the opponent if it was an immediate resignation.

TimaGribanov:
Fixes #3185

## Proposed Changes

  - Dynamic styling for portrait + zen (in this setting, the button is visible in portrait in zen mode the same way as Undo button)

### DIFF
--- a/src/views/Game/Game.styl
+++ b/src/views/Game/Game.styl
@@ -333,6 +333,12 @@ goban-view-bar-width = 400px;
             position: relative;
             top: 25px;
         }
+
+        .cancel-button-zen {
+          align-self: center;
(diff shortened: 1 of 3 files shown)


anoek:
Thanks!

PR #3276, opened 2025-10-15
Fix scoring thumbnails

GreenAsJade:
Fixes thumbnails not showing scoring properly, regression from changes in Aug.

## Proposed Changes

  - Fix to match changes, add e2e test


GreenAsJade:
... my Claude thought that your Claude raised a good point, so it agreed to implement the change 😝

GreenAsJade:
I think I need to get this review done locally somehow, pushing to get it is not ideal.

It's interesting that the latter review comments were about unchanged code in this PR.

(And performance _is_ an issue we experience, but that's a fix for another day)

PR #3278, opened 2025-10-17
Adds e2e test for recent mod ui updated

GreenAsJade:
Fixes not having a test for https://github.com/online-go/moderator-ui/pull/10

## Proposed Changes

  - Add a test for clicking on a player to filter the page for them
  
Needs: https://github.com/online-go/moderator-ui/pull/10 or it won't pass (but it's not in the CI suite)


GreenAsJade:
Nice in theory, but actually sometimes it's just easier to do it this way.

We'll keep this.

PR #3279, opened 2025-10-17
Change e2e test titles to be just descriptive

GreenAsJade:
Fixes unnecessaary verbiage

## Proposed Changes

  - Change titles to state scenario instead of say what should be possible


GreenAsJade:
Yeah, nah, it's good to go.

Issue #819, opened 2019-06-22
Feature Request: Rotate Go Board


BHydden:
### Summary

We desire the ability to rotate the go board to all 4 positions of the 90 deg. axis. to help in game review.

### Environment

Review / demo board.

### Additional Information:

As per https://forums.online-go.com/t/is-it-better-to-review-games-from-start-to-finish-or-begin-at-the-end-and-work-backwards/21865/12

anoek:
This issue has been marked stale and will be closed soon without further activity. To keep the issue open, please respond to the comment to keep the discussion going.

felleg:
:+1: on this feature request

BHydden:
Yeah this would be good, reopening.

BHydden:
@GreenAsJade you mentioned possibly wanting this for your joseki feature. Is this still something you might consider picking up?

GreenAsJade:
It's definitely something that is worth "keeping as a request".  I'm pretty far from being able to do it right now, due to rather a long list...

BHydden:
Yeah fair enough. Depending on how OGS maps their boards I wasn't sure if it was fairly easy (basically a matrix manipulation) or painfully complicated haha

GreenAsJade:
I think it's probably both.  At least for me, as soon as you start throwing around maths then it's already painfully complicated  :D :D

anoek:
We do already have some board flipping/rotation stuff that's used in the puzzles, the main complication with the feature that I see is how to present the controls in a way that it's not junking up the interface for the vast majority of people who don't care about the feature, but still handy enough for those that do.

BHydden:
Feel like it should only be a review tool, not on standard game boards 🤔

felleg:
Personally, I would be happy to have the ability to rotate even in standard games. When I played a physical game with a friend, we had to stop it quickly because he had to leave, so we imported it in OGS via a .sgf file we recreated from a photograph. Because we were sitting in front of each other, currently there is no way on OGS to display the board in a way that both players will recognize. 

I realize this is probably a niche usecase, but I hope you see the value it would bring to people in hybrid mode between physical and virtual games :)

BHydden:
Yep, 100%... and if you cam find a sleek way to incorporate the function into the UI that our playerbase supports, all power to you 😉 i sincerely wish you the best of luck because ui/ux is not any of our strong suits 😅

GreenAsJade:
[...]

PR #3280, opened 2025-10-17
Goban rotation

TimaGribanov:
Some changes related to #819. I've made it possible to rotate the board for separate games.

The logic behind this solution is the following:
- we get a new button in the game dock called ‘Rotate goban CW’
- every click on that button rotates the goban div once CW (90 deg)
- moreover, with every click an object is added to the new local storage setting
- on every render of goban (for Game instances), we try to get the `goban-transform` value from local storage and find an angle we need to rotate to by the game id, so every time you open a game, goban of that game will rotate to the last set value

It looks like this now:
image

I haven't set it to fix the issue, because I'm not sure it's the ideal solution and that it covers all desires.

### DIFF
--- a/src/components/GobanContainer/GobanContainer.tsx
+++ b/src/components/GobanContainer/GobanContainer.tsx
@@ -138,10 +138,13 @@ export function GobanContainer({
         if (!goban || !goban_div || !ref_goban_container.current) {
             return;
         }
+        if (!goban.engine.game_id) {
+            return;
+        }
         const gobanTransformArray: GobanTransformSetting[] | undefined =
             data.get("goban-transform");
         if (gobanTransformArray) {
-            const transformValue = gobanTransformArray.find((e) => e.game === goban.game_id)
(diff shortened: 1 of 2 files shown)


TimaGribanov:
I addressed some of the issues pointed out by the action, but I'm not sure about this one

>5. Unbounded Storage Growth
>The goban-transform array will grow indefinitely as users play more games. Consider implementing a cleanup mechanism to >prevent localStorage bloat.

I certainly see the concern, but I'm not sure what would be the best approach here. From which games to clean up? With that I may need some guidance

anoek:
So I think this method of rotation isn't what we want, having the letters and the shadows oriented "wrong" (despite that's how it might look in real life) is a bit of a jarring experience and isn't expected on an online platform. If we're going to do board rotation stuff, which would be nice, it should probably be handled down in the goban library and deal with things like flipping around the drawn labels and stones and whatnot so the contents are flipped, but everything is rendered right side up for the player.

PR #3284, opened 2025-10-19
Notify submodule deploys

GreenAsJade:
Fixes not knowing when submodule updates are deployed

## Proposed Changes

  - Have a check for each submodule during deploy that sends a notification if it has changed.
  
Note: tested with the dry-run target, obviously not with the real target.

Note: this is on top if https://github.com/online-go/online-go.com/pull/3285 'cause I kinda need that.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -168,6 +168,7 @@
         "strs",
         "styl",
         "sublicensable",
+        "submod",
         "subprops",
         "subselect",
         "subtool",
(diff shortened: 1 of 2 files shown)


anoek:
I think we need to have it read the last commit from the submodules and display those, if we want to get fancy, write down what was last announced and don't announce it again if there hasn't been any changes?

anoek:
https://github.com/online-go/online-go.com/commit/bf18fe8daf41d3fa7903a846030dfe7225e97e97

PR #3288, opened 2025-10-20
Show gamelog timestamps in UTC

GreenAsJade:
Fixes CMs seeing different timestamps for the same game.

## Proposed Changes

  - Show timestamps in GameLog in UTC
  
Note: they are still shown in local timezone in Game Info.   Potentially confusing but... I think I'd want to know when it was "my time" that I played the game 🤔

### DIFF
--- a/src/views/Game/GameLog.tsx
+++ b/src/views/Game/GameLog.tsx
@@ -135,7 +135,7 @@ export function GameLog({
                                         }
                                     >
                                         <td className="timestamp">
-                                            {moment(entry.timestamp).format("L LTS")}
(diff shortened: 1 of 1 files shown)


GreenAsJade:
It's almost tempting to have a "time display component" that we can click on to change the TZ of 😝

anoek:
A time display component would be cool. Honestly I'd love to move away from moment all together too, it's a huge library and from a time when it was the only sane choice, but browsers now have built in Intl.DateTimeFormat and there's a number of much lighter weight libraries out there that take advantage of it (or we just use that directly).

PR #3293, opened 2025-10-22
E2E report track robust upgrade

GreenAsJade:
Fixes fragile E2E tests that expect no pending reports at the outset

## Proposed Changes

  - Instead of assuming the report count is zero, actually read it and track it does what it is supposed to do


GreenAsJade:
The review feedback for this one is implemented in #3294

PR #3294, opened 2025-10-22
E2e for aid player filter clear

GreenAsJade:
Fixes not having a test for https://github.com/online-go/moderator-ui/pull/15

## Proposed Changes

  - Add a test that makes sure the other filters are cleared when the user clicks a player name on the AI D page.
  
Note: this is on top of https://github.com/online-go/online-go.com/pull/3293 to avoid having to merge with that.


GreenAsJade:
Currently we're just using the lock to detect unexpected conflict between tests, not really to serialize them so we just want to fail fast.

Good catch on the test regression - my "run all" just found that fail.

GreenAsJade:
I should comment that lock code to reassure claude we don't care.

This is good to go.

PR #3296, opened 2025-10-24
Clearer dev startup banner

GreenAsJade:
Fixes thinking that the server is up OK, only to find that lints are stopping it

## Proposed Changes

  - Print a "lints are still pending" message before starting them


GreenAsJade:
Tried those.  A delay is yucky, and the first option put the message too early.

PR #3301, opened 2025-10-30
Add good sport rating feature to game play controls

anoek:
image


anoek:
(note it's in a phase === "finished" block so non issue)

PR #3308, opened 2025-11-06
Fix React import in LearningHub

anoek:
Fixes "Invalid hook call"'s caused by differing ways React is imported in the LearningHub modules. This caused the build to include two copies of React which is the real root cause of the hook error.

FYI @TCGWim basically we just need to import react like this:

```ts
```

(as opposed to `import React from "react";`)


anoek:
Actually @TCGWim I thought this fixed it but it did not, so while a harmless change I think, so far it's not addressing the problem. Will tag you on another PR when I figure it out.

PR #3328, opened 2025-11-25
Admin anullment settings support

GreenAsJade:
Fixes not being able to tweak bulk annulment criteria

## Proposed Changes

  - Enable the moderator UI to give Admin a table to do this.
  
Needs https://github.com/online-go/moderator-ui/pull/24


GreenAsJade:
Actually, this is how it is supposed to be: the settings we're talking about are used when the moderator does the auto annul.

PR #3337, opened 2025-12-09
Unranked tournaments

anoek:
Add support for marking a tournament as unranked


anoek:
We backfill so wrong claude

PR #3347, opened 2025-12-13
Remove network idle waits, fix report-button-click popover handling

GreenAsJade:
Fixes instability in running e2e


GreenAsJade:
Note: if the button is not enabled, the test will fail when we go to click it.

PR #3354, opened 2025-12-22
Adjust Player username margin in ProfileCard styles

cassidoo:
Currently on OGS, there is a slight leading margin on the player's username in `.Player .Player-username` (.25rem to be exact). On the homepage, it shows slightly off compared to other elements:

image

It *may* be worth removing the margin entirely from `src/components/Player/Player.styl`. I went through a few parts of the site to check out if it was necessary and couldn't find any use cases, but it's possible I missed an edge case I didn't think of.


anoek:
Thanks!

PR #3355, opened 2025-12-23
Cm use voter notes instead of dissenter note

GreenAsJade:
Fixes CMs feeling sad that they can't comment on important things they can see about a report.

## Proposed Changes

  - Let them all make a comment, instead of using dissenter_note only for dissention.

( Could do with https://github.com/online-go/ogs/pull/2238 , although does work without. )


GreenAsJade:
@claude I'm not getting typescript errors, automated checkin tests passed.  I wonder what is going on?

PR #3362, opened 2026-01-01
Show move categories and breakdown for all users

anoek:
image

image


anoek:
@claude you were looking at old code somehow, we don't differentiate between fast and full categories anymore

Issue #3377, opened 2026-01-17
In challenge modal, the Delete button in the preferred settings bar can overflow the modal


pt8o:
**Describe the bug**
Delete button overflows the challenge modal when the selected preferred game setting has a very long description.

**To Reproduce**
1. In the challenge modal, create a "preferred setting" that will have a long description that goes to a 2nd line (e.g. min/max rankings, analysis disabled, verbose time settings)
2. When that setting is selected, the Delete button to the right will overflow the modal

**Expected behavior**
Delete button should not overflow the modal

**Screenshots**
<img width="888" height="602" alt="Image" src="https://github.com/user-attachments/assets/3a3e518c-a09a-46f7-9e92-0d77f18491ac" />

**Desktop (please complete the following information):**
 - OS: MacOS and Linux
 - Browser: Firefox
 - Version: 146.0.1

PR #3378, opened 2026-01-17
fix Delete button overflowing challenge modal

pt8o:
Fixes https://github.com/online-go/online-go.com/issues/3377

## Proposed Changes

In a flexbox, `minWidth: 0` on a child component makes the parent properly calculate that child's width based on its box bounds. Without it, the width is calculated based on child content, i.e. the long text. Even if the text is cut off with an `overflow: hidden`, its non-truncated width bubbles up to the flex parent and is used to calculate the container width.

Before:
Screenshot_20260116_212427

After:
Screenshot_20260116_212440

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.tsx
@@ -1476,7 +1476,7 @@ export class ChallengeModalBody extends React.Component<ChallengeModalInput, Cha
                 <hr />
                 <div className="preferred-settings-container">
                     <div style={{ display: "flex", gap: "1em", alignItems: "center" }}>
-                        <div style={{ flex: 1 }}>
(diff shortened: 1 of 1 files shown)


GreenAsJade:
Any chance of an e2e test for this?   There are existing ChallengeModel tests.

https://github.com/online-go/online-go.com/blob/main/e2e-tests/README.md

pt8o:
I'm not sure... what would the test case be? That the styling rule is applied?

GreenAsJade:
I guess that since the button still appeared (but outside where it should be) that's a great question!

Typically the tests are for "existence of UI elements".

So maybe it just isn't a sensible thing to test 🧐

anoek:
Thanks for the fix!

PR #3394, opened 2026-02-02
Switch from stylus to postcss

anoek:
This PR is an automated attempt to convert us from using stylus to postcss to align the project better with what most developers are used to these days, and from using stylus variables to CSS variables, which results in a 18% CSS bundle size reduction in our case. It also shaves about 10% off of our total build time, and when compiling just css the results are even more pronounced.


anoek:
@claude the nesting plugin is in there, not sure why you're complaining about it. Similarly the main.tsx file doesn't use the classList anymore, it uses the datatheme.

anoek:
@claude everything builds fine and the theme stuff is working. Please validate the plugin order as I can't possibly be bothered to look into that, and I'm pretty it's right - but you go double check so you don't bring it up again.

PR #3403, opened 2026-02-08
Make textareas visible

GreenAsJade:
Fixes unselected textarea elements being invisible (indistinguishable from the background)

## Proposed Changes

  - Make them comply with --input-bg

### DIFF
--- a/src/global_styl/global.css
+++ b/src/global_styl/global.css
@@ -188,7 +188,8 @@ input, select, textarea {
 }
 
 textarea {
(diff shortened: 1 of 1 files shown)


GreenAsJade:
@claude your comment implies you think input.boxed would already have been doing the same thing as the changes here.   And yet, this change was necessary to see the textareas in ViewReport.   Why is that?

anoek:
I didn't see this one before I pushed up another commit that addresses it, it's in though

PR #3407, opened 2026-02-14
Web worker socket

anoek:
Moves the websocket connection handling out to a web worker to avoid connection timeouts in chromium based browsers when the tab is moved to the background for more than 5 minutes.


anoek:
Re point 1 - these are WebWorkers holding a websocket, we never want to get rid of it while we're on the page, and it will be automatically removed when the tab is closed.

anoek:
double check those @claude I thing some of those have been addressed

PR #3416, opened 2026-02-22
e2e test to make sure that suspended users can log in

GreenAsJade:
Fixes not having a test to prove suspended users can log in to appeal

## Proposed Changes

  - Add a test for that
 

Needs: https://github.com/online-go/online-go.com/pull/3415   ( test fails successfully without that PR )


anoek:
awesome thanks for adding a test for that one

PR #3418, opened 2026-02-24
tighten up sgf download

GreenAsJade:
Fixes players not realising they shouldn't download SGF during play, because that will tempt them to use assistance in game.

## Proposed Changes

  - Disable SGF download for players in game
  - Disable SGF download for anonymous users in game

### DIFF
--- a/src/views/Game/GameDock.tsx
+++ b/src/views/Game/GameDock.tsx
@@ -536,9 +536,11 @@ export function GameDock({
             </Tooltip>
             <Tooltip tooltipRequired={tooltipRequired} title={_("Download SGF")}>
                 <a
+                    href={sgf_url}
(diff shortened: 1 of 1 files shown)


GreenAsJade:
It seems dubious to me to fire an alert when someone clicks a _disabled_ button.   There are no other examples (that Claude could find) so I've left it off.

Test fixed.

GreenAsJade:
I think I already mentioned: no other element has an alert for clicking it when it is _disabled_.

I considered putting the info in the tooltip, but we generally dont even have tooltips, and these ones don't pop up if/when the dock opens, so they can't be relied on.

GreenAsJade:
If someone save-ases the link, the backend change above blocks it.

PR #3421, opened 2026-02-28
Have a "Next" button for lessons

GreenAsJade:
Fixes the solution disappearing before you have a chance to absorb the lesson.

As suggested here:   https://forums.online-go.com/t/idea-add-next-button-to-learn-exercises/59464?u=greenasjade

## Proposed Changes

  - Have a "Next" button in the same place as "Retry" for failed lessons

(I think we don't have lessons in dev - I don't seem to - so I tested this pointing at Prod instead of e2e.  I like the feel of it).

### DIFF
--- a/src/views/LearningHub/LearningPage.tsx
+++ b/src/views/LearningHub/LearningPage.tsx
@@ -41,6 +41,7 @@ export abstract class LearningPage extends React.Component<LearningPagePropertie
     correct_answer_triggered: boolean = false;
     wrong_answer_triggered: boolean = false;
     error_triggered: boolean = false;
+    auto_advance_scheduled: boolean = false;
 
     static underConstruction(): boolean {
         return false;
@@ -78,6 +79,7 @@ export abstract class LearningPage extends React.Component<LearningPagePropertie
         };
     }
     next = () => {
+        this.auto_advance_scheduled = false;
         setSectionPageCompleted(this.props.section, this.props.curPage);
 
         this.correct_answer_triggered = false;
@@ -111,7 +113,8 @@ export abstract class LearningPage extends React.Component<LearningPagePropertie
         if (this.complete()) {
             sfx.play("tutorial-pass");
             this.instructional_goban?.goban?.disableStonePlacement();
-            if (preferences.get("learning-hub-auto-advance")) {
+            if (preferences.get("learning-hub-auto-advance") && !this.auto_advance_scheduled) {
+                this.auto_advance_scheduled = true;
                 this.setState({ show_reset: this.showReset(), show_next: true });
                 setTimeout(() => this.next(), 500);
                 return;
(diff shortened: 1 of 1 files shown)


anoek:
So I borrowed this behavior, and a lot about the lessons really, from lichess. They default to this auto-next, and I think for the initial stuff it's probably mostly desired to not have to do two clicks to progress through, but I also get wanting to take a moment especially once you get out of the first lessons.

Lichess solves this by having a toggle

image

I'm wondering if we should do something similar, default to the auto advance, and then when you're ready to slow down a little, toggle that option off. We could put it maybe either at the bottom of the list on the left, or at the bottom of the contents on the right, or have a gear icon that has that option in it to auto-advance or not. What do you guys think?

denishowe:
An option would be perfect. I didn't want to complicate the dev work but I agree absolutely that you want to be able to go fast or slow depending on how easy you're finding it.

GreenAsJade:
I too thought that a choice would be nice, but the UI of it wasn't clear, and I personally found that the click-to-proceed behaviour is a better default than jumping to the next straight away.

I think this because it encourages reflection and learning, and because the button is right where it should be.

For the questions with answers (like how many ataris) you just clicked the radio button, your cursor is right there over the "next" button that appears under it.  Maybe you don't need to look back at the board to see that "yes I'm glad I was right", but you can glance before clicking.

For puzzles, your cursor is where you clicked, over the board, and you _should_ reflect on the resulting board position when it is correct, to let it sink in.

Of course, we all know I'm not a UI designer, but that was my thinking and experience with it.

(And this is a desktop-oriented line of thinking!  🤔   Mobile might be another thing altogether...)

I did wonder whether "radio button lessons" should auto advance, and puzzles should not, but that becomes jarring inconsistent behaviour.

I'm totally happy to add a toggle at a UI design-rapproved place ... personally I'd default it to "don't proceed", and let impatient "tick the box and move on" people change it ;)

anoek:
Let's put it as a toggle under the stuff on the right, that'll flow to be at the bottom of the screen on mobile, and I think we might want to (in the future) make the puzzle more'less full screen and that content on the right can maybe expand a bit better, and that'd naturally push the toggle to the bottom right which would be a fine evolution of things I reckon.

GreenAsJade:
IIUI item 2 only happens if you already solved and are staring at Next and then change the auto-toggle to true.

I'm not convinced that's even a bug.

GreenAsJade:
@claude   Puzzle-based lessons use move_tree with
  onCorrectAnswer/onWrongAnswer, while non-puzzle lessons use complete() checked in onUpdate. But
  crucially, both handlers fire on the goban's update event and puzzle-correct-answer event. For a puzzle
   lesson, when the correct answer is played:

  1. puzzle-correct-answer fires → onCorrectAnswer → schedules setTimeout(next, 500)
  2. update fires → onUpdate → this.complete() returns false (base class returns false), so no second
  timeout

  For most puzzle lessons, complete() is not overridden (base returns false), so onUpdate won't trigger a
   second timeout. The double-fire would only happen if a subclass overrides both complete() to return
  true and uses a move_tree with correct answers — which looks unlikely 
  (puzzle lessons rely on move_tree, non-puzzle ones override complete()).

  So it's a theoretically real but practically unlikely issue. The two mechanisms are mostly mutually
  exclusive

GreenAsJade:
Claude bombed but I decided to put the guard in anyhow.

GreenAsJade:
^^ yep, not a problem.

anoek:
Nice, thanks!

PR #3426, opened 2026-03-06
Fix move coordinate chat highlight not matching when followed by punctuation

JDNdeveloper:
## Proposed Changes

- Move coordinates followed by punctuation (e.g. `R17?`, `R17.`, `R17,`) are now highlighted in chat and show the position on hover/click.
- The coordinate and trailing punctuation are rendered in separate spans so only the bare coordinate is passed to `parsePosition`.

Before:
image

After:
image

### DIFF
--- a/src/views/Game/GameChat.css
+++ b/src/views/Game/GameChat.css
@@ -131,6 +131,16 @@
 
     .position {
         cursor: crosshair;
+
+        &.tight-right {
+            margin-right: 0;
+            padding-right: 0;
+        }
+    }
+
+    .position-trailing {
+        margin-right: 0.2em;
+        padding-right: 0.2em;
     }
 
     .chat-input-player-list-toggle {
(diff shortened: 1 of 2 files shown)


JDNdeveloper:
The test failures look unrelated to my changes.

JDNdeveloper:
> The approach here looks sound — using `data-position` instead of `innerText` is the right call given the new structure. One thing worth double-checking:
> 
> **Visual spacing between coordinate span and punctuation span**
> 
> The `.position` CSS class currently applies `margin-right: 0.2em` and `padding-right: 0.2em`. With the punctuation rendered as a sibling span immediately after the position span, this spacing would push the two apart visually — e.g. `R17 ?` instead of `R17?`. The PR does touch `GameChat.css`, but if that right margin/padding hasn't been removed or compensated for, the rendered output will look disconnected.

Not an issue; this is already handled. When punctuation follows a coordinate, the position span gets the `tight-right` class, which zeroes out `margin-right` and `padding-right`. The punctuation is then rendered in a sibling `position-trailing` span that carries the spacing after it. So `R17?` renders without any gap between the coordinate and the punctuation.

anoek:
Thanks for the fix!

anshulxyz:
Thanks for doing this, I really wanted this one.

PR #3427, opened 2026-03-07
POC: Refactor tests, increase speed, use API methods to create user

anshulxyz:
This is a **Proof-of-Concept**. Fixes slowness of the e2e tests to some degree.

## Proposed Changes

In every test we don't have to test the user registration flow. That is taking a lot of time of the test. For e2e tests we can register users directly with the API instead of the UI. For speed.

We can have a single separate test for user registration through UI.

I have tested this with the4 demo board related tests, and have found that the newer tests are 60% faster.

Results:
```text
With new user registration
Time: 11 + 9.6 = 20.6

With old user registration
Time: 18.5 + 14.1 = 32.6
```

I am all ears. Open to all suggestions.

### DIFF
--- a/e2e-tests/helpers/user-utils.ts
+++ b/e2e-tests/helpers/user-utils.ts
@@ -262,7 +262,7 @@ export const atomicPrepareNewUser = async (
         // Prevent desktop notification prompts
         localStorage.setItem("ogs.preferences.asked-to-enable-desktop-notifications", "true");
         // Disable dynamic help popups (RDH system)
-        localStorage.setItem("ogs.rdh-system-state", '{"enabled":false}');
+        localStorage.setItem("ogs.rdh-system-state", JSON.stringify('{"enabled":false}'));
     });
 
     // Final navigation to ensure we are logged in and ready


anoek:
> Conflicts with stated e2e test principles
> 
> e2e-tests/CLAUDE.md explicitly says:
> 
>     Avoid direct API calls - the intent of e2e testing is to test by driving the system as a user does.

I feel like if we're doing something that makes the testing process notably faster, so long as we've exercised those parts (such as user registration) with an actual e2e test, we don't need to *always* drive it through the whole browser method necessarily. That said, I defer to @GreenAsJade , if he feels it's better to always go through that whole process then I'm good with it too.

GreenAsJade:
Yeah I mentioned to Anshul - it's cool to deliberately to it.  I suggested we make sure that we have one test in each "class" of test where the users are "real" not "pre-fab", just in case something depends on the signup flow.

GreenAsJade:
Maybe we should tell CLAUDE that we've decided this, in the README/CLAUDE, so it doesn't complain all the time, and checks that it's true!

anshulxyz:
> so long as we've exercised those parts (such as user registration) with an actual e2e test, we don't need to always drive it through the whole browser method necessarily

Yes, that was my precisely intention. Since User registration is already getting tested in other tests, I don't want it to be tested again with Demo Board.

Other reason was improvement in speed. Faster test suite == faster feedback. But yes, speed should not come at the cost of confidence in the test. So, if doing what I am doing in this PRs brings down confidence, then we throw this PR away. (I have no attachment to this.)

> I suggested we make sure that we have one test in each "class"

I agree. I have now added a option of `setupType: "ui" | "atomic";` so we can now easily switch between the two types. Based on this option, tests generates user accordingly.

anshulxyz:
I am losing hope in this PR, because I feel if I keep adding more if/else, I am missing my goal of simplification.

GreenAsJade:
Are there any more if/else on the table?

I was imagining just manually making sure that one test of each category uses the "full registration" process to get its users ... the rest use this... ?

anshulxyz:
> Are there any more if/else on the table?

@GreenAsJade Nope. This is all. 

> Direct API calls conflict with documented project policy

Would just have to update the e2e policy to allow this.

> atomicPrepareNewUser — CSRF token may be stale after registration

Expected.

> login verification is weaker than the established pattern

This is expected.

GreenAsJade:
Great - yep, let's do that.     I think the policy gets an exception along the lines of "API use for creating test users is allowed, as long as there is one test in each category of tests that does it via the UI".

anshulxyz:
@GreenAsJade that should do it, thoughts?

GreenAsJade:
Looks good… I’m rebuilding my server against latest main… will test there then merge

GreenAsJade:
Update: this ran fine on the "new" backend.  I'm setting up e2etesting server to use the new backend, then will merge this, which should trigger a new e2e test run

PR #3430, opened 2026-03-09
Hide puzzle collection edit controls for non-owners

JDNdeveloper:
## Proposed Changes

This change fixes puzzle collection permissions in the collection page UI so non-owners only see actions they can actually use.

For collections the current user does not own, the page now hides collection management controls like save, delete, new puzzle, and collection settings fields, while keeping the collection name displayed in the existing underlined input style as disabled for visual consistency. In the puzzle list, non-owners now see a Solve action instead of Edit, drag-and-drop reordering is fully disabled, and the row cursor no longer shows a grab affordance.

Before fix, for puzzle collection you don't own (or for logged out users):
ogs_puzzle_collection_not_owner_before_fix

After fix, for puzzle collection you don't own (or for logged out users):
ogs_puzzle_collection_not_owner_after_fix

After fix, for puzzle collection you do own (no changes):
ogs_puzzle_collection_owner

### DIFF
--- a/e2e-tests/cm/cm-escape-rate-display.ts
+++ b/e2e-tests/cm/cm-escape-rate-display.ts
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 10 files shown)


anoek:
Thanks!

PR #3434, opened 2026-03-14
Fix zero padding

GreenAsJade:
Fixes 1:03 appearing as 1:3

## Proposed Changes

  - format properly


GreenAsJade:
This is used in moderator-ui, works fine.

PR #3436, opened 2026-03-15
Bump to vite 8.0

anoek:


anoek:
Looks like we need to wait for https://www.npmjs.com/package/vite-plugin-node-polyfills to be updated first

PR #3441, opened 2026-03-16
Undo atomic user registration

anshulxyz:
Reverts the changes we made to Demo Board e2e tests. We were making direct API calls to register the user.

Initially I Wanted to do this to speed up the tests.

What I have found is that it did speed it up, but has added more complexity to the test code base. Which is not worth the speed bump. With two ways to sign up test user, it was double to task to make sure everything went alright.

I have however split big function that was doing a lot into smaller more understandable ones.

cc: @GreenAsJade

### DIFF
--- a/e2e-tests/helpers/demo-board-utils.ts
+++ b/e2e-tests/helpers/demo-board-utils.ts
@@ -86,7 +86,6 @@ export const fillOutDemoBoardCreationForm = async (
     }
 
     if (final_settings.komi !== undefined) {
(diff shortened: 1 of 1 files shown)


anshulxyz:
> e2e-tests/CLAUDE.md not updated

I disagree with the PR bot. Since `CLAUDE.md` is just a symlink of `AGENTS.md`

GreenAsJade:
Yay thanks for helping keep this in shape!

PR #3443, opened 2026-03-17
Add a test to show that in progress games are not assessed as thrown

GreenAsJade:
Fixes not having a test for this common corner case

## Proposed Changes

  - Add a test
  
  
  Needs https://github.com/online-go/ogs/pull/2306


GreenAsJade:
(timing issue noted, if it takes 270s to sign up that's a problem!)

Issue #3366, opened 2026-01-06
Lessons in The "Learn to play Go" Section Currently Don't Sync Between Machines (or Browsers)


DimensionalDot:
**Is your feature request related to a problem? Please describe.**
I am new to Go (and loving it!) and was completing lessons from the "Learn to play Go" section of the site. Later I logged onto my account on a different machine and saw that none of my lesson progress transferred over. 

**Describe the solution you'd like**
It would be great if this was able to persist between different machines. It appears that currently this data is stored in local storage and I recognize that adding this feature would likely result in increased database costs. That being said, if you are open to this, I would be happy to work on/help with an implementation.

**Describe alternatives you've considered**
A middle ground could be a system that stores full lesson completion (e.g., "lesson 3.6 finished") remotely. While still storing example completion within a lesson (e.g., "lesson 3.6 - examples 1+3+4 complete") within local storage.

anshulxyz:
More people have requested this.

<img width="618" height="109" alt="Image" src="https://github.com/user-attachments/assets/593198d2-6d40-4d58-99d6-68dbed64c74f" />

I've recently worked on a sync issue (refer #3402), so I am willing to take this on if this is something you want in OGS.
cc: @anoek @GreenAsJade

anoek:
That would be great, thanks!

anshulxyz:
This has been fixed by #3390

GreenAsJade:
Yay thanks for following up!

PR #3450, opened 2026-03-22
Update LearningHub to sync legacy data

anshulxyz:
Further fix for #3366

## Proposed Changes

- Updated lib/data.ts because it was triggering "remote_sync" on page load, which was causing the remote data to override the local
- Updated views/LearningHub/LearningHub.tsx to have a one time migration flag for merging and pushing legacy data to remote servers.

@GreenAsJade please check. Open to feedback.

### DIFF
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -327,6 +327,8 @@ function safeLocalStorageRemove(key: string) {
     }
 }
 
+export const legacy_learning_hub_data: Record<string, Record<number, true>> = {};
+
 /* Load previously saved data from localStorage */
 try {
     for (let i = 0; i < localStorage.length; ++i) {
@@ -337,6 +339,9 @@ try {
                 const item = localStorage.getItem(`ogs.${key}`);
                 if (item) {
                     store[key as keyof DataSchema] = JSON.parse(item);
+                    if (key.startsWith("learning-hub.")) {
+                        legacy_learning_hub_data[key] = store[key as keyof DataSchema];
+                    }
                 }
             } catch (e) {
                 console.error(
--- a/src/lib/data.ts
+++ b/src/lib/data.ts
@@ -658,7 +658,10 @@ socket.on("remote_storage/update", (row) => {
     }
 });
 
+export let initial_sync_complete = false;
+
 socket.on("remote_storage/sync_complete", () => {
+    initial_sync_complete = true;
     events.emit("remote_data_sync_complete");
 });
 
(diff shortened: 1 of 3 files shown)


anshulxyz:
@GreenAsJade I've addressed all the Claude issues.

To fix the the issue of the sync my strategy has been, that we will maintain a flag, which tells us wether the historical data has been synced or not.

If not, then we merge the local and the remote data and sync it.

Else, we ignore.

But we also need to consider wether adding this complexity to the code base is really _that_ important.

anshulxyz:
Also I want to head if there are more cleaner way to do this.

anoek:
So I guess my first question is how necessary is this migration anymore? Is my understanding correct that without this people who were going through the lessons would have their local progress erased, but going forward everything is synchronized and no progress will be lost?

It seems to me like yes we should have done this initially, but we didn't, so now it's been out there for a month and a half and I'd wager that the large majority of the damage has likely been done. If that's correct the question is, is there much benefit to injecting this migration complexity at this point? It seems to me like this would only benefit a potentially hypothetical few people that had started through the learning hub lessons more than a month and a half ago, left it entirely alone for the past month and a half, and then at some point in the future will decide to resume their lessons. If that assessment is accurate, I think maybe it's not worth carrying the complexity, it's not *that* big of a deal to either do the lessons again or skip to where you remember leaving off. 

If we do decide it's necessary, I'm not really a fan of having stuff in data.ts that deals with migration stuff directly. My gut says you could probably have a migration file that loads from localStorage when the main app starts and keeps that data around for whatever work the LearningHub needed to do, thus keeping that outside of data.ts. Even better would be if it did the work that LearningHub does so that way the migration is completely self contained, but I haven't thought through how that might work and interact with the LearningHub as things are loading, so maybe that part is in fact better handled in the LearningHub like how you are doing it now, but something to consider if you think it's still necessary to implement the migration.

anshulxyz:
> Is my understanding correct that without this people who were going through the lessons would have their local progress erased, but going forward everything is synchronized and no progress will be lost?

Local progress would not be **erased**. But it will stay local, and only on that _specific_ device. As is, that data will not be sync-ed across.

And yes, going forward all the new data will be sync-ed.

> the large majority of the damage has likely been done. If that's correct the question is, is there much benefit to injecting this migration complexity at this point?

Yeah, as mentioned in my previous messages I also think its not worth adding this complexity. I am personally in favour of closing this PR. We could just tell the affected users (they mentioned this on OGS forum) that we are sorry for this, but going forward all your progress will be synced.

PR #3452, opened 2026-03-23
Update AGENTS.md to point to CONTRIBUTING.md

JDNdeveloper:
There were some inconsistencies between the two files, so I cleaned that up. I believe the `Use yarn, not npm` rule was just referring to package management, not run scripts. Also, `npm run build` is pretty slow, so I updated `AGENTS.md` guidance to remove this from the main development loop, and just do it before pushing up a final commit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,15 +28,15 @@ React/TypeScript frontend for online-go.com. Uses Vite, PostCSS, `yarn`.
 
 ## Before Committing or Considering a Change Complete
 
-Follow [CONTRIBUTING.md](CONTRIBUTING.md) before marking any change as done. Run these checks during development after each meaningful change:
+Follow [CONTRIBUTING.md](CONTRIBUTING.md) before marking any change as done. In particular:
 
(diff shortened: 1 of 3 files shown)


JDNdeveloper:
> ## Review
> **CLAUDE.md vs AGENTS.md inconsistency**
> 
> This PR updates AGENTS.md to defer `yarn build` to "before final push only," replacing it with `yarn type-check` in the normal development loop. However, CLAUDE.md (which AI agents like Claude Code load as authoritative project instructions) still lists `yarn build` in its pre-commit checklist without qualification:
> 
> ```
> ## Before Committing or Considering a Change Complete
> - Run `yarn build` to verify the build succeeds.
> ```
> 
> An AI agent reading both files will receive conflicting guidance. The optimization of skipping slow builds in the normal dev loop won't take effect for agents that load CLAUDE.md. This PR should update CLAUDE.md consistently with the new guidance.

`CLAUDE.md` is just a symlink to `AGENTS.md`...this comment makes no sense.

anoek:
Thanks!

PR #3454, opened 2026-03-23
Update the way we identify Home page, hopefully more resilient this time

GreenAsJade:
Fixes mass fail

## Proposed Changes

  - Look for the home page container as confirmation we navigated and rendered the page


GreenAsJade:
I'm not wild about the other timeouts.   Ironically the first run with this change in place seemed noticably _less_ flakey!  🤷‍♀️

PR #3455, opened 2026-03-24
Support auto-vacation notify and opt out

GreenAsJade:
Fixes folk complaining about not knowing, not wanting.

## Proposed Changes

  - Support notifications for when auto-vacation kicks in
  - Provide an opt-out
 
Won't do anything till https://github.com/online-go/ogs/pull/2343 is in.

### DIFF
--- a/e2e-tests/users/new-user-vacation-balance.ts
+++ b/e2e-tests/users/new-user-vacation-balance.ts
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


GreenAsJade:
@claude can you check about your bug report?   It is supposed to be the case that auto-vacation is tied to a specific game. The game server triggers it when a specific correspondence game is about to time out, and it always sends the game_id. The notification data includes game_id because that's what triggered the vacation activation.

  Linking to the game that caused the auto-vacation is the right UX — the user wants to know "which game almost timed out?" and clicking takes them there.
  
  Where do we lose the game_id?

GreenAsJade:
(The new test for zero initial vacation will fail until the back-end is in).

PR #3460, opened 2026-03-26
Add an ErrorBoundary around our ScoringEventThumbnails for when they crash

anoek:


anoek:
well.. it catches it. The try catch solution was tried but hides the error before the error boundary, better to have a big red box pronouncing the error instead of an empty space.

PR #3478, opened 2026-04-11
Use Lichess hamburger

Seth-Rothschild:
## Proposed Changes

This PR replaces the fontawesome times/bars icons used for the hamburger with what lichess does, which is a nifty bit of CSS that draws out the 3 lines in `hbg__in`, `hbg__in:before` and `hbg__in:after` and then spins and collapses them to make an x icon. I've tried my best to keep the size and spacing as similar as possible to the existing icons: the only difference users should notice should be the animation on click.

Claude was used for some of this code but I've read through it and take responsibility for it. The larger contribution candidly is lichess: it's their component and I would not have been able to write the transitions/animation myself. Since this is my first contribution here, please let me know if there are things that could be done better here.

### DIFF
--- a/src/components/NavBar/Hamburger.css
+++ b/src/components/NavBar/Hamburger.css
@@ -9,6 +9,7 @@
   }
 
   .hamburger {
+    position: relative;
     display: inline-flex;
     align-items: center;
     justify-content: center;
--- a/src/components/NavBar/Hamburger.css
+++ b/src/components/NavBar/Hamburger.css
@@ -8,13 +8,19 @@
     align-items: center;
   }
 
-  .hamburger {
+  button.hamburger {
     position: relative;
     display: inline-flex;
     align-items: center;
     justify-content: center;
     width: 2rem;
     height: var(--navbar-height);
+    background: none;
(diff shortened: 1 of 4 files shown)


anoek:
Pretty slick!

@claude the hamburger is primarily on mobile. Does that affect your recommendation?

lazerwalker:
> Pretty slick!
> 
> @claude the hamburger is primarily on mobile. Does that affect your recommendation?

Looks like Claude didn't explicitly reply, but it's important for a11y this was switched to a button. For this purpose, mobile screen-reader users functionally count as "keyboard-only users".

Seth-Rothschild:
Yeah by my count Claude had two bits of feedback that didn't seem particularly important and then the two a11y catches seemed solid. I adjusted all 4 anyways. 

I think this should be pretty unobjectionable at this point. Most questionable thing here I think site-wise is using those particular aria-labels, but both have existing translations and this is how the other Home icon was marked too.

anoek:
Thanks @Seth-Rothschild  for working through everything and getting this really nice!

PR #3479, opened 2026-04-11
added default.nix for building reproducible dev environment

pdg137:
This PR proposes adding a `default.nix` to the top level, using the [Nix build system](https://nixos.org/) to define a precise dev environment. This can be useful for developers familiar with Nix and also casual/infrequent contributors since (after Nix is installed) they only have to run one command to enter a dev shell:

```sh
nix-shell
```

No more reading docs to figure out exactly what you need to install (and what versions) to get it going! I don't know much about NodeJS environments so for me at least this satisfies a desire to keep them as encapsulated and well-defined as possible.

It works like this:
1. The standard `fetchYarnDeps` function downloads the packages specified in `yarn.lock` and stores them in a cache folder in the Nix store.  This cache is locked with a hash that will need to be updated from time to time, when `yarn.lock` and `package.json` change.
2. The script builds a `node_modules` folder in `--offline` mode, using yarn.lock, package.json, and the cache as inputs.
3. Finally a shell starts, running a startup hook that links `node_modules` into the right place.

The `node_modules` directory is read-only, which is a benefit since it's not possible for files in there to be corrupted. This required a change to Vite configuration to make it save its cache to the top level `.vite` instead of there. I made this configurable with an environment variable, so it still defaults to the standard location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ i18n/*.tar.gz
 # Playwright
 /playwright/.cache/
 /playwright/.auth/
+
+.vite
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ old-*
 i18n/*.tar.gz
 *.old
 .playwright-mcp
+.vite
 
 # Playwright
 /playwright/.cache/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,6 +35,8 @@ yarn run dev
 
 If you're on Linux, you can simply type `make` and it will do all this for you as well.
 
+[Nix](https://nixos.org/) users can run `nix-shell` to open a temporary shell with a dev environment, or `nix-build -o devshell` to build a persistent script that does it. This method might be especially useful for casual/infrequent contributors who don't want to install all the NodeJS dependencies permanently on their system (but are willing to at least install Nix).
+
 If you're on Windows and need specific help getting tools installed and the repo cloned, see [Detailed Setup Steps](#detailed-setup-steps-windows-but-applicable-mostly-to-others) below.
 
 (And ... those detailed steps may even be a useful pointer about how to get started under Linux/macOS: they're broadly applicable, even if details differ slightly.)
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,123 @@
+# Nix build that defines a reproducible environment for running the
+# server.  You can use this as an easy way to jump into the dev
+# environment without depending on any OS-level software packages.
+#
+# Requirements: nix must be installed; use "apt install nix-bin" or
+# see https://nixos.org/ for more detailed options.
+#
+# Usage:
+#
+# To jump into a dev shell:
+#   nix-shell
+#
+# To create a permanent shell for later use:
+#   nix-build -o devshell
+#
+# Updating:
+#   If yarn.lock or package.json get updated, the build will fail due
+#   to missing packages or a mismatched hash in the defnition of
+#   offlineCache below.  You can force it to display the new hash by
+#   blanking out the existing one, then update the hash to the new
+#   value to proceed. (Please submit a PR.)
+#
+# Note: this includes a dependency on a personal GitHub repository for
+# the simple mkBuildableShell utility. If that URL ever breaks we
+# could replace it with the less-functional mkShell or copy in the
+# code from a backup somewhere.
+
+let
+  offlineCache = pkgs.fetchYarnDeps {
+    inherit yarnLock;
+    # Update this hash for a new yarn.lock.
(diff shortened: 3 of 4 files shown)


pdg137:
I think the added complexity of path traversal for something that's only used in one place isn't worth it. We can fix it later if it becomes a problem.

As for my little [mkBuildableShell](https://github.com/pdg137/mkBuildableShell/) utility, I'm not planning to delete it and thought we'd prefer having just one nix file in the OGS project. If I ever do delete it, we could switch to mkShell or find a fork/backup of the (simple) code somewhere.

pdg137:
I think the latest comment by Claude is wrong. That's what it says in [the docs](https://vite.dev/config/shared-options#cachedir) but not how it actually works out. My change should not be a change for non-nix users. Someone else should try and confirm.

PS not knowing really anything about his, I'd support putting .vite at the project level for everyone. Shouldn't node_modules be reserved for storing node modules, not random data? But I wanted this PR to be a minimal change.

anoek:
Interesting, can you update the CONTRIBUTING.md to layout how this is used? How does this compare to something like docker in terms of spinning up a quick environment on linux/mac/windows ?

pdg137:
Okay, I added a bit of description.

Nix has a similar goal to Docker - specifying exactly how to build and install everything, including each of its dependencies. With Nix however you are not in any kind of virtualized environment, it's just accomplished using normal environment variables like PATH and by installing each dependency in a unique location. So you can use Nix applications on your system like normal programs. You can even get NixOS, which applies this method to the entire OS.

I think Yarn achieves something similar, in a more limited scope, using the hashes in yarn.lock to lock down exact versions of all the JavaScript packages.  But Nix is general and it also goes all the way down. The instructions for building the specific version of NodeJS are referenced in there, as well as the exact compiler versions, Bash, etc., with no requirement for a binary blob at any point.

anoek:
I'm familiar with Nix as a full install or for a base image for vms or docker instances, but I haven't seen used for a development environment per-say. 

What's the proposed flow here? Say a contributor comes along, they're running Mac OsX, what do they do to get up and running quickly?

pdg137:
If it's someone who is already familiar with Nix and has it installed, it should be trivially easy: just `nix-shell` should automatically do all the work and pop them into a working dev shell.

However that's still a very niche case and for a typical person, say on macOS, I'm hesitant to suggest it since maybe your normal CONTRIBUTING instructions are already good enough. But if that has been a pain point, and assuming the Nix setup on this branch *does* actually work on macOS, this might offer an easier path:

1. install Nix
2. run `nix-shell`

By the way, as an example of how it can be useful, last time I was running things in this code base, we used `npm run dev` to start the server. I didn't know to switch `yarn run dev` until I read through CONTRIBUTING again. Nix provides a way to get those instructions recorded programmatically so you aren't counting on people to read. That's kind of an easy example, but if it starts to get more complicated, like if you require specific versions of things to be installed in a certain way, you can put all of that info into `default.nix`.

Speaking of more complicated things, I started looking at the e2e-tests and ran into this:
```
Error: 
╔═══════════════════════════════════════════════════╗
║ No chromium-based browser found on the system.    ║
║ Please run the following command to download one: ║
║                                                   ║
║     yarn playwright install chromium              ║
║                                                   ║
║ <3 Playwright Team                                ║
╚═══════════════════════════════════════════════════╝
```

Looks like that's another thing to add to the Nix config if we can figure out how to make it happen ahead of time. It's a bit troublesome that it seems to want to do a network download - is there a way to just point it at a pre-installed system chromium?

anoek:
Ok, so this is strictly for those running Nix. My main concern is maintaining it in this case. Neither I nor any regular contributors use Nix so I worry this is going to rot over time and no one will notice until someone comes along and spends a few hours banging their head against outdated instructions.

So I think to get this merged we need to address that, I'm not sure if we can do that with a github action to add a build test or not, but if we can that would be great. Further we need some way of updating that default.nix to track things as they change, that's not manually editing and updating stuff - I see a bunch of sha hashes for versions and whatnot, so if we get updating that automated and some way of testing this so it's regularly exercised I think we're good to go.

pdg137:
For automated testing, you could run e.g.

```sh
nix-shell --command 'yarn test'
```

which will break if someone updates yarn.lock without updating the hash correspondingly, or if the project starts to require a newer version of node/yarn than what's included presently (looks like I got v24.14.0/1.22.22).

But it's fine to just leave this alone until there's more interest. I can keep using and improving it no matter what, and it's easy for others to find in the search. Maybe one of the regulars will get curious eventually, try it out, and see the benefits.

anoek:
When I say automated test I mean hooking it up into the CI, so github actions in our case. That might be very possible as you can specify base images and whatnot so it might just be a matter of selecting a nix base image and then running your above command.

pdg137:
Updated it for the latest yarn.lock and made the update process a little smoother.

PR #3499, opened 2026-04-16
Change lightmode bg color

Seth-Rothschild:
This PR adds a border to the nav dropdowns as shown in these pictures. The border color was actually already set and the border itself was commented out. Also removed the box shadow below the nav and made the shadow on the dropdown darker.

More dangerously: this PR **also** changes the background color of lightmode from #fff to a light beige #edebe9. This gives us the opportunity to use white card color as the highlight color. Personally I think this looks quite nice (the idea is once again stolen from lichess), but I'm not a designer. I would be happy to open up a forum post if you think there's risk to this change.

image
image

And stock OGS lightmode for comparison
image


Seth-Rothschild:
(existing smoke test screenshots are going to look different and fail the test. I can go correct them depending on your preference for how to proceed but would prefer to not correct them multiple times)

Seth-Rothschild:
Actually, it's very easy to update the snapshots with the (very clear) documentation @GreenAsJade put in the e2e readme. I put in the updated images now but can still create a forum post if that's useful.

anshulxyz:
> changes the background color of lightmode from #fff to a light beige #edebe9

I love this change.

anoek:
I like the blue border to the nav elements, those I think we can merge.

I'm open to exploring changing the background color but that'd have to come with a larger theme change set, as is it reduces the contrast for several elements throughout the site and and flips contrast on some of the wells and whatnot - the later of which aren't necessarily a problem but is worth considering if we want that or not. Additionally, dropping the nav bar distinction for one theme and not the other (in this case light has no distinction and dark does) is a weird inconsistency. I'm not convinced we want to drop the distinction for the nav bar, while there's some visual appeal at a glance, it makes those elements blend in with the content on the various pages, I'm not totally sold on that being a good thing, worth some discussion but not a quick merge.

Seth-Rothschild:
Sounds good, I've created #3504 which has just those changes.

Looking at the difference I do think it would be worth having that longer discussion on the lightmode background change: it feels quite a bit softer to not have the bright white. I'll rebase this PR on top of main and use it for working through the background color issues.

> Additionally, dropping the nav bar distinction for one theme and not the other (in this case light has no distinction and dark does) 

I don't feel particularly strongly about the box shadow... but I think this distinction in color is pre-existing. Definitely agree it would be nice for them to both be the same.

Seth-Rothschild:
Associated forum post for BG changes: https://forums.online-go.com/t/lightmode-theme/59822

Seth-Rothschild:
Alright, force pushed over the previous branch to deal with only color concerns:
1. Changed the background-color and nav colors as before
2. Bumped the shade colors down by 22 hex (`#abcdef` -> `#89abcd` etc). This should mitigate any contrast concerns since it's higher contrast to the background than previously.
3. Fixed some but not all of the CSS badness on the learninghub page. In OGS it uses a shade instead of card-color, progress bars disappear on hover. I can keep those changes out of this but it was more jarring as shades got darker.

That being said, there's still the community aspect to these changes. While initial feedback seems to indicate that a majority would like them, there's (what I expect to be a minority) that likes things exactly as they are. With more voters I think it would be reasonable to force the issue but as it stands I would not recommend merging this with what currently stands at a 5-2 vote split for/against.

Per the claude feedback to proceed with this I would change bg and check disabled, but the learninghub page likely needs some actual CSS attention and I wouldn't go out of my way to preserve the previous formatting. Smoke test error is unrelated to these changes.

Seth-Rothschild:
Closing as noted here https://forums.online-go.com/t/lightmode-theme/59822/28?u=sjrothsc.

Insufficient consensus to throw this in as is. I think it _would_ look nice if more pages had white cards to lift them off a background. This is fairly common across the web. But this change alone would not do this, so I think worth holding off without a bit more redesign work.

PR #3504, opened 2026-04-16
Add border to dropdown

Seth-Rothschild:
As discussed in #3499 this PR just adds the blue border to the dropdown. No lightmode background color change and box shadow between nav and body is preserved.

image

### DIFF
--- a/e2e-tests/smoke/smoketests.spec.ts-snapshots/initial-page-chromium-linux.png
+++ b/e2e-tests/smoke/smoketests.spec.ts-snapshots/initial-page-chromium-linux.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fff2e4e605772f1be3a63ec42213c7d5dae7b993e509f73e51ce38d2e609e564
-size 27601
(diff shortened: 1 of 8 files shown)


Seth-Rothschild:
Something odd is going on with the smoke tests where a local run is consistently getting me a different joseki page. Happy to fix it or rerun as needed, but candidly nothing here should have changed the screenshots at all or necessitated a rerun of smoke tests.

Seth-Rothschild:
Ok so:

1. Nav items are off by about 2px because of the border-inline-start. That's what requires the smoke test screenshots to be re-run.
2. Something locally is showing the joseki help page during the screenshot on my local device. I got a clean screenshot by removing that helper code locally but it would be nice to know why this happens to save other people from the same thing.
3. I checked a few RTL languages given how many comments Claude has been putting about it. I wouldn't say it looks any better or worse than before in the navbar. As it is the nav renders in the same direction so we would have been a-ok to just use "physical properties" instead of going down the logical properties rabbit hole.  It would certainly be overkill to start playing more with the profile css at this point.

anoek:
Thanks!

PR #3508, opened 2026-04-18
fix: paginated tables now bring you to page 1 when filters change

DimensionalDot:
Aims to fix a bug introduced in https://github.com/online-go/online-go.com/commit/0679486fba0b2ad7d19c317c02ce0315ceb4f616 which causes the `PaginatedTable` component (used in a player profile's `GameHistoryTable` for example) to break when selecting a filter that doesn't result in enough pages to accommodate your current page selection.

## Cause

The commit noted above introduced the `onJson` function which brought with it a meaning change to behavior. Specifically, when the response status is not in the 2xx range, the parsed JSON is still returned. This is different to how it worked before this where the `Response` would be returned. In the case of (at least) `PaginatedTable` this resulted in said broken behavior. This is because, in it's `refresh()` function, it has fallback code to run `SetPage(1)` if the promise returned from `ajax_loader()` rejects. The issue is that this is conditional on `err.status === 404` which is never true given that `err` is the JSON body of the response so `err.status` is `undefined`.

## Changes

> **Note:** Much of the below section is outdated, see my follow up comment

My tentative change is to revert the behavior in `onJson` to return the `Response` when the it doesn't have a 2xx status. I recognize however that the entire point of https://github.com/online-go/online-go.com/commit/0679486fba0b2ad7d19c317c02ce0315ceb4f616 was to pass along readable error messages instead of just an error code. An approach could be to return an object which includes both the readable error as well as the raw response. For example we could (exclusively in the error status code case) do `data.status = res.status` or something similar. This would allow us to check the status code within `PaginatedTable`. I'm a bit hesitant with both approaches here given that `lib/requests` is used in so many files meaning it may be hard to audit for breakages. Of course, an alternative would be to make the fix locally in `PaginatedTable` but that might be tricky to do in a clean way... Please let me know how you'd recommend proceeding.

## Reproduction:

1. Navigate to some player's profile.
2. Apply some filter to their game history.
3. Note how many pages are available (let's call this amount N).
4. Remove all filters.
5. Navigate to page N+1 or greater and apply those same filters.
6. Note the filters don't correctly apply until you navigate to page N or earlier.

**For example:** If you only have two pages of 13x13 games then, without filters applied, navigate to page 3 and filter to 13x13.

### DIFF
--- a/src/components/PaginatedTable/PaginatedTable.tsx
+++ b/src/components/PaginatedTable/PaginatedTable.tsx
@@ -127,6 +127,11 @@ function PaginatedTableImpl<RawEntryT = any, GroomedEntryT = RawEntryT>(
         if (!force && last.length && softEquals(last, cur)) {
             return;
         }
+        const last_filter = last[3];
+        if (!softEquals(last_filter, filter)) {
+            setPage(1);
+        }
+
         last_loaded.current = cur;
 
         if (loading) {
@@ -174,9 +179,6 @@ function PaginatedTableImpl<RawEntryT = any, GroomedEntryT = RawEntryT>(
             })
             .catch((err) => {
                 console.error(err);
-                if (err.status === 404) {
-                    setPage(1);
-                }
                 setLoading(false);
                 if (load_again.current) {
                     load_again.current = false;
--- a/src/components/PaginatedTable/PaginatedTable.tsx
+++ b/src/components/PaginatedTable/PaginatedTable.tsx
@@ -128,11 +128,14 @@ function PaginatedTableImpl<RawEntryT = any, GroomedEntryT = RawEntryT>(
             return;
         }
         const last_filter = last[3];
-        if (!softEquals(last_filter, filter)) {
-            setPage(1);
-        }
-
         last_loaded.current = cur;
+
+        if (last_filter && !softEquals(last_filter, filter)) {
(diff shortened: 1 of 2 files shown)


DimensionalDot:
Just had the realization that the easiest way to solve this would be to call `SetPage(1)` whenever a filter is changed in `PaginatedTable`. This would result in your page number getting lost when changing filters but at that point the meaning of page N is changing anyways so it doesn't seem like a major loss to me. I'll probably switch the fix to just that tomorrow.

DimensionalDot:
Claude's most recent review is incorrect, as filters not being set (on the user's end) doesn't correlate with `last_filter = undefined` (or being falsy in general). The only times that would be the case is when (1) the page just loaded, (2) the filter prop was never set (in which case we wouldn't want this to ever run anyways), or (3) the filter prop was set to something falsy, but that would almost certainly be unintended.

Also, Claude's secondary note at the bottom is technically correct, but unlikely to happen. It's also probably out of scope for this PR as handling that would potentially be complex for reasons discussed in my original comment.

DimensionalDot:
Also, let me know if this is the kind of thing you'd like an e2e test created for. I don't have experience with Playwright but I can try my hand.

anoek:
Thanks!

PR #3513, opened 2026-04-20
feat(TabCompleteInput): support multiline text input

boonespace:
This PR adds multi-line input support for the chat input box, improving the experience when composing longer messages.

My changes mainly include the following three points:

1. Replace <input> with <textarea>
2. Auto-height adjustment (max 150px, min 1 line)
3. CSS layout fixes to prevent chat log from covering input box

image

This is an early draft. I'd like to confirm if this direction is worth pursuing.

### DIFF
--- a/src/components/TabCompleteInput/TabCompleteInput.tsx
+++ b/src/components/TabCompleteInput/TabCompleteInput.tsx
@@ -23,6 +23,7 @@ interface TabCompleteInputProperties extends React.HTMLProps<HTMLTextAreaElement
     id?: string;
     placeholder?: string;
     disabled?: boolean;
+    maxMessageLength?: number; // optional, undefined means unlimited.
     onKeyPress?: React.KeyboardEventHandler<HTMLTextAreaElement>;
     className?: string;
     onFocus?: (event: React.FocusEvent<HTMLTextAreaElement>) => void;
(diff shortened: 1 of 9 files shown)


anoek:
As part of your experimentation here you can increase the default message split length too if you need. I think it's either 200 or 300 right now, feel free to bump it up to 1000-2000 whatever seems appropriate.

boonespace:
Got it. I wasn't aware of the limit before — 300 has been fine for me. But with multi-line input, longer messages make sense. I think 1000 should be enough.

Currently, the input box doesn't prevent typing beyond the limit — it shows a "Message too long" error when you try to send. Perhaps we could also disable sending when the limit is exceeded, along with a visual counter. That way, users can adjust before sending instead of hitting the "Message too long" error.

Like this:
image

boonespace:
I noticed the server enforces a maximum message length of 1024 characters (confirmed via testing).

For GameChat this is straightforward — there's no split logic involved, so the input cap is simply set to 1024. Other contexts (group, chatview, etc.) are slightly different since the split mechanism is still wired up there.

I went back and forth between two approaches:
- Split messages at 1024 characters per chunk
- Hard-cap input at 1024 characters and reject anything over

For now I've gone with a hard 1024-character limit across all contexts (leaving the existing split logic in place but effectively unreachable). This keeps behavior consistent, and realistically very few users will ever hit this limit anyway.

@anoek  what's your preference here? If you'd rather keep split alive with per-context limits, I can expose maxMessageLength as a prop on TabCompleteInput.

boonespace:
This PR support multi-line input with auto-resize, char counter, and bottom-anchored log

- Add auto-resizing textarea with max height limit
- Add character counter with warning near limit
- Anchor chat log to bottom, sync with input height changes
- Support Shift+Enter for line breaks
- Redesign GameChat layout to two rows for better space usage

In GameChat
image

In ChatView
image

In Group
image

anoek:
I like the shift-enter capability but the new layout takes up two lines instead of one for normal quick chat, can we get that back to being compact? If it switches after a shift enter that could be fine too.

Also, let's fix the font and spacing so it matches the chat font, for example: 
image

compared to the text area which has what looks like increased font sized and spacing
image

boonespace:
I've reverted to the compact single-line layout, added distinct warning styles for approaching/exceeding the character limit, and fixed autoscroll detection under column-reverse layout.

P.S. The anti-flood system may need follow-up adjustments to ensure line breaks don't bypass the limits. Perhaps we should limit the number of line breaks allowed, but that's a separate issue.

src/components/TabCompleteInput/TabCompleteInput.tsx
```diff
@@ -105,17 +106,54 @@ function matchFullName(input: string, nicknames: string[]): MatchResult {
...
 }
 
-function setCaretPosition(input: HTMLInputElement, position: number) {
+function setCaretPosition(input: HTMLTextAreaElement, position: number) {
     input.setSelectionRange(position, position);
 }
 
-export const TabCompleteInput = React.forwardRef<HTMLInputElement, TabCompleteInputProperties>(
+// max length support in server is 1024
+const maxMessageLength = 1024;
```
anoek: Probably best to define that in the lib chat_manager and import it here

src/lib/chat_manager.ts
```diff
@@ -588,7 +588,7 @@ class ChatChannel extends TypedEventEmitter<Events> {
     }
 
     public send(text: string): void {
-        if (text.length > 300) {
+        if (text.length > 1024) {
```
anoek: better to use the constant you're moving into this file

src/lib/misc.ts
```diff
@@ -386,7 +386,8 @@ export function errorLogger(...args: any[]) {
     }
     console.error(err);
 }
-export function string_splitter(str: string, max_length: number = 200): Array<string> {
+// max length support in server is 1024
+export function string_splitter(str: string, max_length: number = 1024): Array<string> {
```
anoek: use that constant here too

anoek (changes requested):
Looking pretty great, got a few things to fix up:

For the colors, we use CSS variables so the colors are consistent throughout the site and can be changed dependent on theme, so instead of grays use `var(--shade3)` or whatever, instead of warning colors manually specified here, use `var(--warning-bg)` and `var(--warning-fg)`, or the error equivalents, etc. You can find a definition of all of them in https://github.com/online-go/online-go.com/blob/main/src/global_styl/01_variables.css

Currently, if someone holds down shift enter they can enter in a ton of empty space and clear out chat essentially, let's trim up successive white space so only two new lines are entered. Make sure that they can't just do <enter>, <space>, <enter>, <tab>, <enter> ... etc and copy paste that too, whatever pattern matching should detect and collapse that too.

boonespace:
Done! Updated all colors to use CSS variables and added input normalization to collapse excessive whitespace — specifically, zero-width characters are removed, consecutive spaces are collapsed to one, and newline sequences are collapsed at two.

anoek (approved):

anoek:
Very cool thanks for doing this!

PR #3519, opened 2026-04-23
fix(joseki): explorer now properly removes placed stones when resetting

DimensionalDot:
Fixes a bug introduced in https://github.com/online-go/online-go.com/commit/b402605bd0ab6a9b7f766215285a358c8c54c1c3 which caused stones to not be cleared from the Joseki Explorer when using the "fast backwards" button to reset the board.

## Cause

The old version avoided destroying the Goban on first mount by checking if the `initial_position` wasn't unset. However, this check generates a false-positive _whenever_ the Goban is being reset as those are also cases where we don't specify an initial position. This means that it is not properly destroyed when one tries to manually reset the board.

## Proposed Changes

Switched to checking if `last_click.current` is set (instead of `initial_position`) as the only time that could be `undefined` is when the component has just mounted.


anoek:
Thanks for the fix!

PR #3527, opened 2026-04-25
Adding Smart Variations (Photoshop-like: Toggleable and Color-coded)

Johnniedarkoo:
The big UX change: 
- Added the new concept of variations that can be toggled on/off by the user for their pleasure, in a similar way of turning on/off layer visibility in Photoshop. 

Other UI tweaks:
1. desktop layout of the variation pane is clean and well space. 
2. presence list gets tiny avatars. 
3. Chat UI is cleaned up: 
- 1. lost styling reintroduced (as Claude flagged), 
- 2. removed the duplicate announcement of "user posted variation" and then a nearly identical variation button underneath it
- 3. chat input now gives a clean textbox highlight when active.

### DIFF
--- a/e2e-tests/kibitz/kibitz-visual.spec.ts
+++ b/e2e-tests/kibitz/kibitz-visual.spec.ts
@@ -15,6 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { writeFileSync } from "fs";
 import { expect, Page, TestInfo } from "@playwright/test";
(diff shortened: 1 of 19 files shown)


Johnniedarkoo:
Just a few minutes too late to include in the merge, but all review issues are addressed. :) 

Performance: Full board reload on every incoming variation 
- The active variations are now listed and the board only reloads when the listed variations change. 

The other three were trivial. 

They aren't crucial to continue development (front or backend) so they'll make it into a next PR.

PR #3528, opened 2026-04-26
Css challengelist

Seth-Rothschild:
This PR:

1. Makes css changes to the challengelist to make it look a bit more like a table and give it breathing room
2. Splits Rengo into "Rengo Live" and "Rengo Correspondence" visually
3. Fixes minor other styling problems (line under header was the wrong length, one cell was missing the cell css class)

Side by side with ordinary ogs. Individual cells have a bit more padding
Screenshot 2026-04-26 at 7 22 46 AM

Rengo update: split it into two visual tables without changing the underlying js.
Screenshot 2026-04-26 at 7 25 14 AM

Mobile screenshot. Counterintuitively giving it a bit more space might make it easier to read and easier to click buttons on mobile.

Screenshot 2026-04-26 at 7 24 05 AM

### DIFF
--- a/src/views/Play/ChallengeLists.tsx
+++ b/src/views/Play/ChallengeLists.tsx
@@ -224,7 +224,7 @@ export function RengoList(props: RengoComponentProps): React.ReactElement {
 
     if (!anyChallengesToShow(filter, list)) {
         return (
-            <table id="rengo-table">
+            <table className="rengo-table">
                 <tbody>
                     <tr key="none-available">
                         <td colSpan={11}>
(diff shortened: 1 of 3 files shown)


Seth-Rothschild:
This change doesn't do anything about this weird formatting, but I would like to. Happy to add it to this PR or make a different one. I'm not quite sure the best fix for it yet.

image

Seth-Rothschild:
This change now also increases the join button size in rengo challernges, which I think is the least complex way of fixing alignment issues.
image

Seth-Rothschild:
I also looked more deeply into the join button and got some debug rendering for admin and other buttons. I had added the line

```
        .btn.success {
            min-width: 6.0em;
        }
```

Which fixes the particular variant of the alignment issue I see. But it turns out there are *a lot* of buttons which are supposed to fit in that cell and can conditionally render. I've very confident this doesn't fix all of them, but I'm hoping it at least makes it a bit better for one of the standard views and doesn't make anything worse.

anoek:
Thanks!

PR #3532, opened 2026-04-29
fix: correct textarea baseline alignment in chat input

boonespace:
This PR fixes the textarea alignment issue in the game chat.

After feedback from godeon on the [forum](https://forums.online-go.com/t/how-about-adding-multi-line-input-support-for-chat/59854/7), I noticed that the chat input displays correctly in Firefox but is misaligned in Chrome.

This is caused by differences in how browsers calculate the textarea baseline. I added `vertical-align: bottom` to force the baseline to align at the bottom of the element, restoring proper alignment across browsers.

In Chrome before:
image

after:
image

### DIFF
--- a/src/components/TabCompleteInput/TabCompleteInput.css
+++ b/src/components/TabCompleteInput/TabCompleteInput.css
@@ -60,4 +60,5 @@
     border-top: none;
     padding: 3px 3px;
     outline: none;
+    vertical-align: bottom;
 }


anoek:
> Firefox but is misaligned in Chrome.

Huh, didn't know people still used Chrome, weird. :sweat_smile: 


Thanks for getting it fixed up!


For a possible future fix, I do notice there is (always was) a small gap in firefox:

image


vs chrome

image

Issue #3536, opened 2026-04-30
Filter annulled games from game history


jufda:
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to filter annulled games from the list.

**Describe the solution you'd like**
I'd like to filter the annulled games to only see the actually played games.

**Describe alternatives you've considered**
Hiding all annulled games without filter would lose the possibility to check if a player frequently disappears from games.

**Additional context**
Screenshot of the relevant area.

<img width="834" height="681" alt="gamehistory" src="https://github.com/user-attachments/assets/4c5c34d6-fe67-4e70-a4d6-13c61d65a1cc" />

jufda:
https://github.com/online-go/online-go.com/pull/3537

PR #3537, opened 2026-04-30
Add "Hide annulled games" games button to filters

jufda:
Fixes #
Implemented the proposed feature: https://github.com/online-go/online-go.com/issues/3536

## Proposed Changes

  - Add the button for filterin annulled games.
  - From my point of view, it seems like default behaviour could be hiding the annulled.

### DIFF
--- a/src/lib/preferences.ts
+++ b/src/lib/preferences.ts
@@ -209,7 +209,7 @@ export const defaults = {
     "game-history-size-filter": "all",
     "game-history-ranked-filter": "all",
     "game-history-bot-filter": "humans" as "humans" | "bots",
-    "game-history-annulled-filter": "hide" as "all" | "hide",
+    "game-history-annulled-filter": "all" as "all" | "hide",
 
     "help-system-enabled": true,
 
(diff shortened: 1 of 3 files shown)


anoek (changes requested):
This is a good filter I think, however for a technical reason (hiding them is slow for players with a lot of games, especially bots, because we don't yet have the indices built for that query), hiding should not be the default at this point.

jufda:
> This is a good filter I think, however for a technical reason (hiding them is slow for players with a lot of games, especially bots, because we don't yet have the indices built for that query), hiding should not be the default at this point.

Alright, so the only necessary change would be to switch the default in preferences.ts?

    "game-history-annulled-filter": "**all**" as "all" | "hide",

anoek:
Functionally yes but I'd also invert the toggle logic for the button to go along with it so it's like "Hide annulled games" or something, lit up blue when it's active

anoek (approved):

anoek:
Thanks!

PR #3539, opened 2026-05-01
Joseki page overhaul to use GobanView

anoek:
image

image

image

image

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -47,6 +47,7 @@
         "deepwiki",
         "deinit",
         "deinitialize",
+        "desync",
         "devel",
         "Dismissable",
         "dndkit",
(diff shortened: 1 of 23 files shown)


anoek:
(1) Category enum keys vs values — not actually a bug, no change. The reviewer assumed the server returns title-case "Ideal", but it returns uppercase (ColorMap[current_move_category] and
  bad_moves.includes(move.category) only work with uppercase, and saveNewPositionInfo always .toUpperCase()s before send/PUT). Storing the uppercase key as the option value keeps state and
  option values in lockstep. Switching to values would create a mismatch on initial render after a server fetch — the option list would have "Ideal", state would be "IDEAL", find(o.value ===
   ...) would return undefined, and the fallback wouldn't catch it because "IDEAL" is a valid MoveCategory key. Net regression. The misleading enum comment is about TS string-enum
  discriminants in TS comparisons, not the wire format.

PR #3540, opened 2026-05-02
fix: update border and radius of learning section

Seth-Rothschild:
Two changes in this PR

1. remove the bottom border from learninghub sections
2. change the border radius to rounded when section is collapsed

Before
image

After
image

Dark
image


anoek:
Thanks!

PR #3542, opened 2026-05-04
Scroll chat on keypress

Seth-Rothschild:
People in chat were talking about sending messages and realizing the conversation had moved on from the message they were looking at. This is *a* fix to the issue and one proposed by the chat, but I'm not sure if it's the best fix.

In this code we add a few lines to `onKeyPress` for the chat input to scroll to the bottom. This means as the user starts typing it will move them, but since that component is autofocused it means that any keypress will move them. I think it would probably be ok but worth thinking about since it would catch like... them switching to another tab using a browser hotkey.

The other options to consider are adding it to send, or being more restrictive about which keyboard events we want to scroll for. Another proposal in the chat is some sort of visual indicator that they've gotten new messages. If chat had that, I think the standard is to click it to scroll the new messages into view. Adding that feature would be a lot more work than this though.

Putting this PR up as is for discussion, can modify it to whatever people think would be best.

### DIFF
--- a/src/components/Chat/ChatLog.tsx
+++ b/src/components/Chat/ChatLog.tsx
@@ -563,6 +563,12 @@ function ChatInput({
 
     const onKeyPress = useCallback(
         (event: React.KeyboardEvent<HTMLTextAreaElement>): boolean | undefined => {
+            const lines = document.querySelector(".chat-lines") as HTMLElement | null;
+            if (lines) {
+                lines.scrollTop = 0;
+                scrolled_to_bottom = true;
+            }
+
             if (!event.shiftKey && event.key === "Enter") {
                 const input = event.target as HTMLTextAreaElement;
                 if (!socket.connected) {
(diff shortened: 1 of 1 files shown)


anoek:
The new kibitz page being worked on has two chat windows frames and both have a `.chat-lines` element, so the note about not using a global dom query is going to be immediately applicable on that page pretty soon.

The other note is if you want to only trigger the scrolling for printable characters you can do something like 

```
if (event.key.length === 1) {
    scrollTobottom();
}
```

that won't trigger for things like shift, control, function keys, etc.

Seth-Rothschild:
Nothing to do about the scroll position hijacking, that's just what this PR does. I would imagine a more standard use case is the one people were talking about in chat: a user is responding to a message that they don't know is old and this is a cheap way of letting them know.

The desync described in 2 seems unlikely to me but I'll remove it since removing shouldn't do much harm.

Seth-Rothschild:
I did a bit of playing with hotkeys for browser shortcuts (Firefox, OSX). Seems like *not all* browser shortcuts trigger the reset but some do, and since it's not consistent within one browser and one operating system I wouldn't expect it to be consistent for everyone. Safest to assume it will just trigger for most of them.

I generally think this would be ok to go in. I suspect it would do more good than harm. But I *would not* be offended if you'd like to throw away this code and wait for someone to fix the root cause of the problem which is that people don't know when they're scrolled up in chat.

anoek:
Thanks!

PR #3543, opened 2026-05-05
Allow custom games buttons to expand in other languages

### LINKED ISSUE #3541: Translated button labels don't fit the buttons
**Describe the bug**
Buttons are sized to the text they need to contain. Translated languages (at least Dutch, probably German too) can result in a longer text which no longer fits the buttons. This then overflows and looks messy. 

image

Expected is that:
Text fits. If not feasible it is expected to be cut-off with ellipses.

Possible extra ways to mitigate:
Have the font size dynamically size down one step when necessary. 

Alternatively a test can be created to flag ill-fitting text, to manually retranslate it.

Seth-Rothschild:
Fixes #3541. I don't know if this should close that issue since I'm sure there are other places this happens. 

The buttons in question have fixed height. Giving them a min-height instead allows them to expand if the translated text is longer than would fit. The `height: auto` is necessary because otherwise `button, a.btn, .btn` sets a height (2em) which is also too small.

image

### DIFF
--- a/src/views/LearningHub/LearningHub.css
+++ b/src/views/LearningHub/LearningHub.css
@@ -254,7 +254,7 @@
                 text-decoration: none;
                 width: 100%; /* Make cards responsive to grid cell width */
                 max-width: 23rem; /* But don't exceed original max size */
-                height: 5rem;
+                min-height: 5rem;
                 color: #f9f9f9;
                 border-radius: 0.3rem;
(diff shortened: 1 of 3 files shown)


Seth-Rothschild:
From slack this now handles all the known issues with buttons overflowing.

anoek:
Thanks!

PR #3549, opened 2026-05-09
Kibitz

GreenAsJade:
This is a new major feature.

## Proposed Changes

Kibitz brings a view where users can observe a chosen game and comment on it, showing variations.

For reference, backend implementation notes follow.

---

# Kibitz implementation briefing

This is a self-contained briefing for implementing the Kibitz backend. It captures what we learned about the existing OGS system, the design decisions made after rejecting parts of the original handoff, the Phase 1 plan, and the resolved product decisions.

The original handoff document is at `Kibitz backend handoff.md` in this directory. **This briefing supersedes it for the Phase 1 backend implementation.** The handoff described a much larger scope (9 endpoints, new variation/stream tables, Option A/B/C variation-ownership debate) which — after investigating the existing OGS infrastructure — we determined is unnecessary. Almost all of what the handoff wanted can be assembled from primitives that already exist.

The frontend state that this supports lives in `online-go/online-go.com` on the `kibitz_main` branch (locally checked out as `ogs-ui/`), specifically:

- `src/models/kibitz.d.ts`
- `src/views/Kibitz/KibitzController.ts`
- `src/views/Kibitz/KibitzGamePickerOverlay.tsx`
- `src/views/Kibitz/KibitzRoomStage.tsx`
- `src/views/Kibitz/KibitzVariationList.tsx`
- `src/views/Kibitz/KibitzPresence.tsx`
- Other files in `src/views/Kibitz/`

`src/lib/KibitzMockService.ts` (the Phase 1A/B mock backend) has been deleted; the controller talks to the live endpoints described below.

---

## 1. What OGS already provides (key investigation findings)

Before planning Kibitz, we investigated how existing OGS primitives work. Several findings directly shape the design.

### 1.1 Two chat systems, deliberately split

OGS has two chat backends, split by architectural principle:

**Scylla (`game_chats` table)** — chat that is _about_ a specific game record. Handled by game-server.

- Schema: `PRIMARY KEY ((game_id, channel), chat_id)`, `chat_id` is a TimeUUID
- TTL: **none on real games**, 90 days on ephemeral games (`game_id < 0`)
- Channels coexist in the same table: `main`, `spectator`, `malkovich`, `shadowban`, etc.
- Load API: `load_game_chat_lines(channel, line_count)` in `ogs/services/_shared/lib/ScyllaModels.ts:378-403`
- Buffer sizes on join (hardcoded): main=1000, spectator=1500, malkovich=2000 (see `ogs/services/_shared/lib/config.ts`)
- **No pagination.** The query is `SELECT ... LIMIT line_count`; there is no `WHERE chat_id < ?` cursor anywhere in the codebase. Client gets a fixed buffer on join and that's all it sees.

**Redis (`chat-<channel>` lists)** — standalone conversations, not about a game. Handled by comm-server.

- Used by: group chat (`group-<id>`), tournament chat (`tournament-<id>`), global chat (`global-<language>`), PMs (`pm-<id1>:<id2>`), shadowban channel
- TTL: 30 days for channels, 7 days for PMs — **TTL resets on every write** (so active channels never expire; idle ones evict after 30d)
- Retention within the list: **`LTRIM` to last 50 messages on every write** (`MemChat.ts:506-508`) — this is the binding retention constraint, not the TTL
- No pagination API beyond the list contents
- Persistence: Redis RDB/AOF, configured by ops (confirmed by ops to be solid)

Net practical difference: game chat gives a ~1000-msg buffer on join with no decay; Redis chat gives a 50-msg buffer with 30d idle-decay.

### 1.2 Variations are already a primitive — just not a first-class one

OGS already supports sharing an analysis variation as a chat message. Investigation path: a user in game chat opens the analysis viewer, enters a variation, names it, hits "share" → a chat line with a typed body is sent and later clickable-reopened.

- **Payload shape** (`ogs-ui/src/lib/GobanController.ts:71-78`):
  ```ts
  interface SerializedAnalysisChatLineBody {
    type: "analysis";
    from: number; // move number the variation branches from
    moves: string; // encoded move sequence
    name: string;
    marks?: Record<string, string>;
    pen_marks?: MoveTreePenMarks;
  }
  ```
- **Send**: `GobanController.shareAnalysis()` at `GobanController.ts:867` → `goban.sendChat(analysis, chatLog)`
- **Transport**: WebSocket, `game/chat` event
- **Storage**: Scylla `game_chats.line` column (JSON string containing the body)
- **Addressable ID**: `chat_id` TimeUUID — stable but not designed for cross-feature reference
- **Rendering** (game chat only today): `ogs-ui/src/views/Game/GameChat.tsx:649-741` dispatches on `body.type === "analysis"` and calls `goban.engine.followPath()` to render
- **No standalone model, no standalone endpoint, no cross-chat-channel use**
- **UX when it TTLs**: the entire chat line (including "X shared variation Y" text) is deleted together — user sees no ghost link, no broken reference, no placeholder

### 1.3 The generic chat line renderer does not handle typed bodies

Group/tournament/global chat uses `ogs-ui/src/components/Chat/ChatLine.tsx:74`, which assumes `body` is a string. It does not dispatch on `body.type`. Typed bodies would render as `[object Object]`. This is an asymmetry with `GameChat.tsx:649-741`, which does dispatch.

A TODO comment at `MemChat.ts:31` acknowledges the gap: `type MemChatMessage = string` with `/* I think this can be an object as well */`.

### 1.4 comm-server validation is ambiguous for object bodies

`comm-server.ts:385-386` validates message length:

```ts
if (config.message.length === 0 || config.message.length > 8192) {
  throw new Error("Invalid Message length for chat message");
}
```

The check is written string-shaped. When `config.message` is an object, `.length` is `undefined`, both comparisons are false, and the message passes through. Whether this is a deliberate "don't length-check objects" policy or an oversight isn't clear from the code — there's no `typeof` guard and no comment either way. The adjacent TODO at `MemChat.ts:31` (`type MemChatMessage = string; /* I think this can be an object as well */`) suggests the author was aware objects could arrive but hadn't settled the contract.

Regardless of intent, the effect prior to Phase 1B was that object bodies round-tripped through comm-server and Redis intact but unvalidated. The Phase 1B change made the contract explicit: string bodies length-checked at 8 KB, object bodies required to have `typeof body.type === "string"` and capped at 32 KB JSON-stringified (see §2.4 item 1).

### 1.5 UIPush — Django's standard fire-and-forget push to clients

`pushToUI(channel_or_player, event, data)` in `ogs/services/django/ogs/notifications.py:29-46`. Publishes to Redis pub/sub; existing socket servers (termination-server, game-server) relay to connected clients.

**Established naming convention:**

- **Per-entity channels**: `tournament-<id>`, `group-<id>`, `ladder-<id>`, `appeal-<id>` (e.g., `api/views/tournament.py:332,385,585,...`; `api/views/group.py:144,178,185,304`; `api/views/ladders.py:236,254`)
- **Broadcast channels**: `announcements` (`go_app/models/announcements.py:46,68`), `moderators` (`api/views/login.py:314`, `go_app/models/moderator_log.py:53`), `gotv` (`go_app/tasks/fetch_twitch_streams.py:104`)
- **Events are descriptive verbs**: `reload-tournament`, `players-updated`, `update-streams`, `new-user`

**Important detail**: `group-<id>` is used as _both_ the chat channel name and the UIPush channel name. Same for `tournament-<id>`. So the entity name is a shared namespace across both mechanisms — not a collision, a convention.

### 1.6 Presence is already derivable from chat channel subscription

The platform does not have a separate room-membership model. "In a group chat" = subscribed to the chat channel.

`MemChat.ts:226-291` — when a client calls `chat/join`:

1. Existing subscribers get a `chat-join` event with the new user's `MemChatUserObject`.
2. The joining client gets a `chat-join` event containing the **full current roster** (all existing members' user objects).
3. Last 50 chat lines are replayed.
4. Topic (if any) is sent.

On `part`, everyone gets `chat-part` with the leaving user.

So viewer count, users list, active chatters, friends-in-room are all derivable client-side from the roster + recent chat buffer. No backend-side presence model needed.

---

## 2. Design decisions (with rationale)

These decisions were made after several rounds of discussion that narrowed the design.

### 2.1 "Variation" means "a branch of a game-tree"

Semantics pinned: a variation is analysis that branches from some move of some game. It belongs to a game by definition.

In a Kibitz room, variations become visible via stream items (chat messages of type `variation_posted` / embedded analysis body). The room's "shared variations" list is **derived** client-side from its chat history filtered to `body.type === "analysis"` — it is not a stored collection.

This resolves the original handoff's Option A/B/C variation-ownership debate as a non-issue. The ownership debate was conflating three separate things:

- Analysis content (data — belongs to the game)
- Posting event (room-scoped — a stream item / chat line)
- Visibility in a room's list (derived from stream items)

Once the terminology is right, the cross-room-visibility concern (Option B/C) evaporates.

### 2.2 Kibitz chat runs on comm-server / Redis, not Scylla

Redis chat (`kibitz-<roomId>` channel) is the right home for Kibitz conversation:

- Matches group/tournament/global conventions — Kibitz is a social channel
- No new table, no new Scylla model
- Retention inherited from comm-server defaults: 30d TTL (resets on activity) + LTRIM cap (raised for Kibitz — see 2.3)

We considered Scylla (either piggyback on `game_chats` with a `kibitz-<roomId>` channel, or a new `kibitz_chats` table) but rejected it: Scylla's "durability" advantage over Redis is smaller than it seems given that game chat doesn't paginate either. The practical difference is a ~50-msg buffer vs. a ~1000-msg buffer on join. Once the LTRIM cap is raised, Redis matches the experience.

### 2.3 Variations as typed chat bodies — reuse existing primitive

A Kibitz room's "post variation" action is: send a chat message with body `SerializedAnalysisChatLineBody` to the `kibitz-<roomId>` channel. Same format as today's game-chat variations. Stored and relayed by comm-server with no new plumbing.

Benefits: no new `Variation` model, no new endpoint, no new storage, and the goban client already knows how to render these bodies.

The handoff's idea of returning a `board_snapshot` payload in variation responses is unnecessary — the goban renders from `game_id` + `moves` + marks, which is exactly what's in the body.

### 2.4 Small comm-server / frontend cleanup (delivered)

Four small changes were needed to make typed-body chat work end-to-end on non-game channels (all co-located, all improvements anyway). All four have landed:

1. **`comm-server.ts:385-403`** — message validation split into two branches:
   - `typeof config.message === "string"`: original 8 KB length cap applies.
   - `config.message` is a non-null object (typed body): requires `typeof config.message.type === "string"` and caps `JSON.stringify(config.message).length` at 32 KB. Anything else is rejected.
   - The 32 KB cap is generous for an analysis variation (move list + marks + pen marks) and well above any plausible legitimate payload, but bounds what an authenticated user can flood Redis with.
2. **`MemChat.ts:36`** — `type MemChatMessage = string | TypedChatBody`. TODO removed.
3. **`MemChat.ts` retention** — `chatRetentionLimit(channel)` (`kibitz-*` → 1000, else 50) used at all three callsites: the LTRIM on write (line 569), the `lrange` in `join()` (line 325, otherwise users joining a kibitz room only replay the last 50 of up to 1000 messages), and the `lrange` in `removeAllChatsByPlayer` (otherwise a moderator purge of a banned user only scans 50 messages per channel and can leave up to 950 behind).
4. **`ogs-ui/src/components/Chat/ChatLine.tsx:147`** — the generic chat-line renderer now dispatches on `body.type` for the analysis case, matching what `GameChat.tsx` already does. The shared per-line component that backs both call sites is `src/components/Chat/GameChatLine.tsx`; see §2.11 for the rationale of the per-line (vs. per-log) extraction.

### 2.5 UIPush for room-state events; chat socket for chat

Two existing primitives used for what each is good at:

- **Chat messages** (including `variation_posted` and `system` types) → comm-server `kibitz-<roomId>` channel
- **Room-state events** (created, removed, updated, board-changed, viewer-count-changed) → `pushToUI`

Naming follows the established convention:

- Per-room UIPush channel: `kibitz-<roomId>` (same name as the chat channel, matching how `group-<id>` and `tournament-<id>` operate). Events: `room-updated`, `board-changed`.
- Broadcast UIPush channel: `kibitz-rooms` (matches `announcements`/`gotv`/`moderators` pattern). Events: `room-created`, `room-removed`, `viewer-count-changed`.

**Viewer counts are pushed by comm-server, not polled.** comm-server's `MemChat.join()` and `MemChat.part()` publish to `ui-pushes/kibitz-rooms` with `{event: "viewer-count-changed", data: {channel, viewer_count}}` whenever any `kibitz-*` channel's unique-player count changes. Every directory-subscribed client updates its `_rooms[*].viewer_count` in real time — the only active-room-specific path. The backend's GET /rooms snapshot is only authoritative for the initial list; post-load the UIPush stream is the source of truth. This was added after an earlier iteration tried three hacks in parallel (local chat-proxy mirror into rooms, optimistic −1 on part, and stale backend snapshots) that all disagreed with each other — see commits `b7d96ae84` (comm-server publisher) and `88cfee0b0` (frontend subscriber, hacks removed).

**Seed mechanism for the directory's first paint.** Django needs viewer counts for every room in the directory response. Naively this was N synchronous urlopen calls (one per room) into comm-server, blocking the Django worker for up to N × timeout — a worker-pool exhaustion risk under any comm-server hiccup. comm-server therefore exposes a batch endpoint, `POST /api/channels/viewerCounts` with `{channels: [...]}` in the JSON body, returning a `{channel: count}` map from in-memory state in one round-trip; Django uses it via `_get_viewer_counts()` from `KibitzRoomList.get`. (POST rather than GET-with-querystring so the channel list is not capped by nginx's URL length limit (~4–8 KB) — at higher room counts a querystring-based list would 414 and the except-block fallback would silently zero out every viewer count.) The single-room paths (`KibitzRoomDetail.get`, post-create / post-change-board stamps) keep the existing single-channel `_get_viewer_count` against `GET /api/channel/users`. Failure mode degrades to "all 0s" until UIPush updates land, rather than failing the directory response.

### 2.6 "In the room" is a single fact

"Being in a Kibitz room" = being subscribed to the `kibitz-<roomId>` chat channel. Everything else — viewer count, user roster, active chatters, friends-in-room — derives from that.

Entering the room view triggers three concurrent actions: join chat channel, subscribe to per-room UIPush channel, fetch the hydration endpoint. Leaving reverses all three. The directory UIPush channel (`kibitz-rooms`) is subscribed independently while browsing the room list.

### 2.7 Scope cuts from the original handoff

The original handoff proposed a Phase 1 with 9 endpoints. We removed:

- **Stream-history endpoint** — chat history is chat history; use the existing chat mechanism
- **Variation list/create endpoints** — variations are chat messages; no separate endpoints
- **`board_snapshot` payload** — goban renders from `game_id` + moves
- **Preset rooms (for this phase)** — no seeding, no auto-pick worker
- ~~**Delete-room endpoint**~~ — re-added post-Phase-1; see §2.12. Automatic cleanup / decay of stale rooms is still deferred (Phase 2).
- **Proposals and voting** — Phase 2

### 2.8 System chat lines are posted by the initiating client

When a user changes the watched game, the room sees both a `board-changed` UIPush event and a `system`-typed chat line in history. The chat line is posted **by the initiating client** immediately after the change-board POST returns 200, not by the backend.

Rationale:

- Phase 1 has no non-UI initiator of change-board (no admin tool, cron, or bot), so a connected client is always present.
- Avoids adding a new Django→comm-server insertion path. comm-server has no precedent for accepting injected chat — §1.4 / §2.4 only widened the contract for client-sent typed bodies.
- Game-server has Scylla precedents for system chat injection (rengo player-drop in `services/game-server/Game.ts:3493`, review-created in `services/game-server/api.ts:587-603`), but these live alongside game-server's direct Scylla access. Replicating that for comm-server would mean either a new HTTP API, a new Redis pub/sub bridge with `gen_auth`, or Redis-direct writes that bypass `MemChat.send`'s broadcast — each carries tradeoffs (no Django-side error feedback, multi-instance duplicate-publish risk, bypassed comm-server auth, Redis pub/sub being non-persistent).

Cost of this choice: if a future phase needs a non-UI change-board initiator (preset rooms / auto-pick worker per §2.7, moderator tool, etc.), the system chat line will need to be retrofitted via one of the above paths. Accepted as a deferred decision.

The same principle applies to any future room-state change initiated by a connected client; if a future event needs server-originated chat history, revisit at that time.

### 2.9 Analysis bodies carry their own `game_id`

Variations posted to a Kibitz room are `SerializedAnalysisChatLineBody` chat bodies (§2.3), but posting code (`KibitzController.postVariation`) adds an extra `game_id: number` field to the body before sending. Reason: the room's watched game can change after a variation was posted, and the variations panel groups by the game the variation branches from (the "Previous game" divider in `KibitzVariationList`). Embedding `game_id` in the body makes this grouping stable regardless of later `change-board` events.

`mapAnalysisToVariation` requires `body.game_id` and drops bodies that lack it. An earlier version fell back to `room.current_game?.game_id` for pre-1C bodies, but that fallback made variation rendering depend on the timing of `GET /games/:id` (see §2.10) and produced a race where the variation list silently disappeared on slow backends — the chat history would replay, but the fallback-id was undefined for variations that arrived before the game-lookup completed, so they were dropped and never re-derived. With the fallback gone, the variation list is independent of game-data hydration timing.

### 2.10 Rich game data is fetched client-side, not embedded in room payloads

Room payloads (directory + hydration) include `current_game_id` (int), not the full game. The client fetches game details via the existing `GET /api/v1/games/:id` when rendering room cards and when hydrating the active room's watched game. This is deliberately N+1 over the directory — Phase 1 room counts are small. The alternative (richer serializers) would couple kibitz's schema to Game's many fields and duplicate fetches for the common case where the user is only going to open one room. If room counts grow, a batch game-lookup endpoint is the natural next step.

**Cost accumulating — re-evaluation candidate.** This deferral has produced a recurring class of timing bugs and recovery scaffolding because every consumer that needs the watched game's data has to handle the window where REST-hydration has resolved but `current_game` is still undefined: `KibitzController.hydrateActiveRoomGame` exists purely as the recovery path that re-fires `setActiveRoom` once the game lookup completes; the variation-list race in §2.9 is one symptom; `KibitzRoomStage.tsx`'s `goban.on("load", ...)` deferred-apply for the secondary board is another; the "No game selected" flash on room cards is a third. Each new feature that depends on the watched game adds another instance. The lighter-weight fix is to embed a kibitz-shaped game summary on the room serializer (`{game_id, board_size, title, black, white, tournament_name?, ended?}` — six fields, the shape of `KibitzWatchedGame`) and `select_related("current_game")` on the directory query. That's roughly 30 lines of backend code; on the frontend it deletes `lookupGameForKibitz`, `hydrateActiveRoomGame`, and several "wait for game" branches. Deferred for now but worth doing before the next round of features that touch the watched game.

### 2.11 Watched-game chat in the kibitz panel: source from goban, not chat_manager

`KibitzSharedStreamPanel` has two panes — Room (the `kibitz-<roomId>` channel via `chat_manager`) and Game (the watched game's chat). The Game pane's source of truth is the **goban instance** for the watched game, not `chat_manager.join("game-<id>")`. Per §1.1, the two are completely separate chat backends: comm-server's Redis (`chat_manager`) carries kibitz/group/tournament/global chat, while game-server fronts Scylla `game_chats` and exposes lines on the goban (`goban.chat_log`, `goban.on("chat" / "chat-remove")`). Joining `chat-game-<id>` on comm-server reaches a Redis channel that is unrelated to actual game chat.

Wiring:

- `KibitzBoard` constructs its goban with `connect_to_chat: true` so the goban's `game/connect` request asks game-server to also push chat messages on the connection. This is purely a server-side flag (no client-side notification side effects). One consequence worth knowing: kibitz spectators count as chat-subscribed spectators on the watched game, which bumps the spectator count visible to the players. That's correct — kibitz watchers _are_ spectators of that game — but worth surfacing.
- `KibitzRoomStage` exposes the `mainBoardController` to its parent via an `onMainBoardControllerChange` callback (mirroring the existing `onMobileCompareControllerChange`). `KibitzInner` holds the controller in state and provides it via `<GobanControllerContext.Provider value={mainBoardController}>` around both `KibitzSharedStreamPanel` invocations (mobile and desktop).
- The panel consumes via `useGobanControllerOrNull()` (from `@/components/GobanView`) — the same hook every other view that needs the watched goban uses (`PlayerCards`, `GameChat`, `fragments`, `AntiGrief`). It then subscribes to `goban.on("chat" / "chat-remove")` and reads `goban.chat_log`.

Channel filter: kibitz surfaces `main` + `spectator` channels. The kibitz user is functionally a spectator on the watched game so both are appropriate. `malkovich` is a player-only side channel and `shadowban` is per-user; neither belongs in a kibitz pane.

**Per-line rendering reuse via `<GameChatLine>` (done).** The earlier shape of this briefing item proposed extracting a `<GameChatLog>` (a list component) from `GameChat.tsx`. The actual extraction landed one level lower — `src/components/Chat/GameChatLine.tsx` — a per-line component taking `{line: protocol.GameChatLine, lastLine?, gameId?}`. This is a better decomposition: the kibitz panel keeps its own list semantics (merged main+spectator, no channel switcher, no composer, kibitz-specific scroll / unread / split-pane behaviour) and just delegates per-line rendering. `GameChatLine` reads the goban via `useGobanControllerOrNull()` so analysis-body lines render as clickable variations with the analysis surfacing on the main board (matching what `GameChat.tsx` does in the game view), plus position links, `/me` italic, chat_markup sanitization, `<Player>` rendering, etc. — full feature parity with the game view's chat rendering, without the chrome.

Wiring in the panel: `PaneEntry` carries the original `protocol.GameChatLine` alongside the `ChatMessage`-shaped `line` field, and the render loop forks: `entry.gobanLine ? <GameChatLine .../> : <ChatLine .../>`. Room-chat entries continue to use `ChatLine` (chat_manager-shaped). The `gobanChannel` class on the wrapper still drives the spectator vs main color rule.

This pattern works because the Line-level component is purely presentational. A Log-level extraction would have implied ownership of subscription / scroll / dedupe — kibitz needs its own versions of those, so a Log component would have either fought the panel or been parameterised heavily. Per-line is the right boundary.

The room/group/tournament side (briefing §2.4 item 4) has since landed too: `ChatLine.tsx:147` dispatches on `body.type` for the analysis case, so analysis bodies in non-game channels render correctly rather than as `[object Object]`.

### 2.12 Delete-room semantics

Added post-Phase-1 to enable the existing frontend Delete affordance (which had been calling a non-existent endpoint).

**Auth.** Identical to `can_edit_room`: room creator (active, not banned/shadowbanned) or full moderator. `compute_permissions` exposes `can_delete_room` as its own field even though the rule matches `can_edit_room` today, so future divergence (a delete-only role, a creator who lost edit privileges, etc.) doesn't require touching the UI plumbing.

**Cleanup model — match `PlayerGroup`, not over-engineer.** `api/views/group.py:417` deletes the row and returns; it does not touch the group's chat channel in Redis or in comm-server's in-memory state. The chat history TTLs out within 30 days (§1.1) and connected clients drain naturally as they unsubscribe. Kibitz follows the same pattern for the same reasons:

- Doing more for kibitz than for groups would set an inconsistent bar across the codebase.
- Cross-service cleanup adds plumbing (a new comm-server endpoint, `gen_auth`, error paths) for marginal real-world benefit — the chat key is unreachable from the directory the moment the row is gone.
- If we later want stricter cleanup for both groups and kibitz rooms (e.g., GDPR-driven), it's a single coordinated change rather than a kibitz-only one-off.

**Frontend close-out is fully UIPush-driven.** Django publishes `room-removed` on `kibitz-rooms` post-COMMIT; `KibitzController.onRoomRemoved` already handled this event from before delete shipped (it removes the row from the directory, unsubscribes the chat channel, and resets the active-room view if it was this one). No new frontend code path was needed for the "someone else deleted my room" case.

**UI gating treats edit + delete + change-board as one "room management" concept.** `KibitzRoomSettingsPopover` shows the "Edit room details" entry to anyone with `canEditRoom || canDeleteRoom` — so a hypothetical delete-only privilege still has a path to its action. The "You do not have room management access yet" note (the popover's catch-all message for users with no actions) requires `!canEditRoom && !canChangeBoard && !canDeleteRoom` to display, so it stays semantically correct under any future divergence. Inside the edit-details view, Save and the title/description fields gate on `canEditRoom`; Delete gates on `canDeleteRoom`. In Phase 1 these collapse to the same value, so user-visible behavior is unchanged from before delete shipped — but the gates are now in the right places for whatever comes next.

**Concurrency.** `KibitzRoomDetail.delete` uses `KibitzRoom.objects.select_for_update()` inside `@OGSAtomic`. `KibitzRoomChangeBoard.post` also uses `select_for_update()`, but inside an explicit `with transaction.atomic():` block scoped to just the read-modify-write rather than `@OGSAtomic` covering the whole method — that change was made so the row lock is not held across the comm-server `_get_viewer_count` HTTP round-trip after the save. The two endpoints still serialise against each other on the row lock; whichever transaction commits first wins, and the loser sees a clean 404 or has its change-board write overwritten by the imminent delete. Either is acceptable.

**What stays deferred.** Automatic cleanup / decay of stale or empty rooms — Phase 2. Resolved Product Decision #1 in §6 still applies for that case.

### 2.13 Presence panel layout: Owner is a separate line, not a roster entry

`KibitzPresence.tsx` renders the owner on its own `Owner: <Player> [away]` line above the "In the room · X watching" summary, with the live roster beneath. The owner is **not** injected into the roster list — when present, they appear in the chat-channel `users` like any other subscriber, and the `[away]` marker only shows when the owner is _not_ in `users`.

This sharpens the architectural separation §2.6 already established: the room's **owner** is a Postgres fact (`KibitzRoom.creator`, hydrated client-side via `player_cache.fetch`), the room's **presence** is a comm-server fact (chat-channel roster). The earlier shape pinned the owner into the roster with an "owner" tag and an inline "away" status, which made the two facts visually indistinguishable.

The empty-state message ("No one is in the room yet.") was also removed: by the time the panel renders, the viewer is in the room they're looking at, so the only way `users` is empty is a transient pre-roster state during chat-channel join. An empty `<section>` is the right placeholder for that — no message that becomes false as soon as it's read.

---

## 3. Phase 1 data model

One new Postgres table. No new Scylla table. No new models for variations, stream items, or memberships.

### Postgres: `kibitz_rooms`

```
id                string PK          -- "user-<pk>" for user rooms; "preset-<slug>" reserved for later
channel           string unique      -- "kibitz-<id>"
title             text
kind              enum('user', 'preset', 'broadcast')  -- only 'user' created in Phase 1
description       text nullable
current_game_id   FK -> go_app.Game nullable
creator_id        FK -> Player nullable
created_at        timestamp
last_activity_at  timestamp          -- auto_now=True; bumped by change-board (see note below)
settings          jsonb              -- pinned, proposals_enabled (Phase 2), future group_id for group-private rooms
```

The `kind` column is included now to avoid a migration later. Only `user` is created in Phase 1; `preset` and future `broadcast` are reserved.

`last_activity_at` powers the directory's `-last_activity_at` sort. The first cut declared it `auto_now_add=True`, which only sets the field on INSERT — the sort silently degraded to `-created_at`. Now `auto_now=True`, which fires on every `save()`. Two Django gotchas worth knowing: (1) `auto_now` does **not** auto-inject itself into `update_fields`, so any `save(update_fields=[...])` call that should bump activity must list `last_activity_at` explicitly (see `KibitzRoomChangeBoard`); (2) chat traffic flows through comm-server straight into Redis without touching Django, so chat does not bump the field. Phase 1 activity = "someone changed the board," which is a reasonable signal. Wiring chat-driven activity would need a comm-server → Django ping; deferred.

Extensibility: when group-private rooms arrive later (see §6), an additional column `group_id FK -> PlayerGroup nullable` and a `visibility` field in `settings` handle it additively.

---

## 4. API surface

Six REST endpoints under `/api/v1/kibitz/` (DELETE was added post-Phase-1, see §2.12; `/directory` was added on top of `/rooms` to give the rail a curated list, see below).

### `GET /api/v1/kibitz/rooms` — full room list

Returns every room without filtering. Used as the unfiltered "all rooms" surface; the rail consumes `/directory` instead. Per-room payload: identity, watched game id, viewer count (queried from comm-server for each channel), and the metadata the directory/hydration views need.

### `GET /api/v1/kibitz/directory` — curated rail listing

Returns the subset of rooms suitable for the left rail. The visibility rule:

> Include the room if `kind ∈ {preset, broadcast}` **OR** `viewer_count > 0` **OR** the caller is the room's `creator`.

Anonymous callers naturally fall through to "presets/broadcasts + non-empty rooms" because they own nothing. The endpoint exists as a separate URL rather than a query-string filter on `/rooms` so the bare-collection URL retains its conventional "all rooms" meaning and per-caller filtering isn't hidden behind it.

Same response shape as `/rooms`. Filtering happens in Python after the batched comm-server count fetch — `viewer_count` lives in comm-server, not the row, so a pure DB pre-filter isn't possible without a denormalised counter. Acceptable while room counts stay small (see Phase 2 cleanup item in §6); a smarter split is sketched in §10's deferred-improvements list.

The frontend rail is wired to `/directory`. Two related fixes in `KibitzController.selectRoom` make the experience seamless when a room would otherwise be filtered out:

- If the active room is not in `_rooms` after the directory snapshot (e.g., user reloaded while watching a non-owned room with no other watchers — viewer_count was 0 at fetch time), `selectRoom` merges the room into the rail using the data already fetched from `/rooms/:id`. The rail must always show the room the user is currently viewing.
- The `setActiveRoom({...full, current_game})` call after `subscribeActiveRoom` spreads `this._active_room` (current state) rather than the original `full` payload from the REST fetch, so the chat-channel roster populated by `syncPresenceFromChat` synchronously inside `subscribeActiveRoom` isn't clobbered back to the empty `full.users`.

### `GET /api/v1/kibitz/rooms/:id` — hydration

Returns the room record (including `viewer_count` at request time) plus a `permissions` block (see §5). After this initial fetch, viewer counts are kept current via the `viewer-count-changed` UIPush on the `kibitz-rooms` channel (§2.5) rather than by re-polling. The client also subscribes to the chat channel and per-room UIPush channel on entering the view.

### `POST /api/v1/kibitz/rooms` — create

Request: `{ source_game_id, title, description? }`

Validation:

- `source_game_id` must resolve to a Game
- The game must be public
- If the game is live and analysis is disabled, reject
- **Duplicates allowed** — two independent rooms for the same game are permitted

On success: creates the row, emits `room-created` on `kibitz-rooms` UIPush (payload includes the room's current `viewer_count`), returns the room payload.

### `POST /api/v1/kibitz/rooms/:id/change-board` — change watched game

Request: `{ game_id }`

Validation: **same rules as room creation** (public, analysis-enabled if live). `game_id` is required — there is intentionally no path through this endpoint that clears `current_game` back to NULL. The column is nullable for seeded preset/broadcast rooms (§3); the directory UI has no "empty room" affordance, so user-initiated clearing is out of scope for Phase 1.

On success: updates `current_game_id` (and bumps `last_activity_at` via `auto_now`, see §3), emits `board-changed` on `kibitz-<roomId>` UIPush. The initiating client posts a `system`-typed chat line to the room channel (e.g., `"<username> changed the watched game to <name>"`) after the POST returns — see §2.8.

### `DELETE /api/v1/kibitz/rooms/:id` — delete

Auth: room creator (active, not banned/shadowbanned) or any full moderator. Response shape and chat-cleanup behavior mirror `PlayerGroup` deletion (`api/views/group.py:417`):

- `200` with `{"success": "Room removed", "id": <room_id>}` on success.
- `403` if anonymous or insufficient permissions; `404` if missing.
- The row is hard-deleted; a `room-removed` event is published on the `kibitz-rooms` UIPush channel post-COMMIT (frontends already subscribe via `KibitzController.onRoomRemoved`).
- The Redis `chat-kibitz-<roomId>` key is **not** explicitly deleted — it self-expires within 30 days under §1.1 retention once new writes stop. Comm-server in-memory `channels[<chan>]` drains as users disconnect. See §2.12 for why.

### Chat send/receive

Uses existing comm-server WebSocket on channel `kibitz-<roomId>`. No new socket code.

### Error contract

- 400 invalid payload
- 403 permission
- 404 missing room or game
- 409 state conflicts

---

## 5. Permissions

One module `kibitz/permissions.py`. The results are returned to the client in the hydration response's `permissions` block so the UI doesn't guess.

| Permission                  | Phase 1 rule                                        |
| --------------------------- | --------------------------------------------------- |
| `can_read` (board + chat)   | true for everyone, including anonymous              |
| `can_chat`                  | authenticated, not banned                           |
| `can_post_variation`        | authenticated, not banned                           |
| `can_change_board_directly` | creator; moderators (validation as §4 change-board) |
| `can_create_room`           | authenticated, not banned/shadowbanned              |
| `can_edit_room`             | creator; moderators                                 |
| `can_delete_room`           | creator; moderators (added post-Phase-1, see §2.12) |

**Not in Phase 1**: group-visibility rules, proposal-related permissions.

Future: group-private rooms add a check like "visible only to members of `settings.group_id`." The extensibility is clean because permissions are already a single-module concern.

---

## 6. Product decisions (all resolved)

Pinned before kickoff:

1. **Room durability**: rooms persist until explicitly deleted (`DELETE /rooms/:id` added post-Phase-1; see §2.12). No automatic cleanup / decay job — that remains a Phase 2 item.
2. **Preset rooms**: none in Phase 1. The `kind` enum accepts them for later.
3. **Who can change board**: only the room creator (or a moderator). Tightened during Phase 1C-a from the original "any authenticated user" — reasoning: the room is conceptually owned by its creator, and arbitrary users swapping the watched game out from under the room is hostile. Server enforces; UI hides the affordance for non-owners via `permissions.can_change_board_directly`.
4. **Anonymous access**: read-only (can view board and read chat; cannot post).
5. **Duplicate-room policy**: two users creating rooms for the same game → two independent rooms.
6. **Change-board target validation**: same rules as room creation.
7. **Future**: group-private rooms (visible only to members of an associated Group). **Not in Phase 1.** Schema is extensible.

---

## 7. Phased delivery

Ordering, not time estimates. Phases 1A and 1B are complete; Phase 1C is largely complete (mock service deleted, controller hits live endpoints, chat dispatch wired, change-board integrated). Items still in flight are noted inline.

### Phase 1A — data + directory

- `kibitz_rooms` Postgres migration
- Room serializer(s)
- `GET /rooms` and `GET /rooms/:id` endpoints
- `kibitz/permissions.py` with the Phase 1 rules from §5

### Phase 1B — room lifecycle + comm-server hardening

- `POST /rooms` and `POST /rooms/:id/change-board` (with validation from §4)
- The four comm-server / frontend changes from §2.4 (validation, type widening, per-prefix LTRIM, `ChatLine` dispatch)
- `pushToUI` calls at the right points: `room-created` on create, `board-changed` on change-board, `room-updated` on edit

### Phase 1C — frontend integration

- Delete `ogs-ui/src/lib/KibitzMockService.ts`
- `KibitzController` hits the live endpoints, subscribes to `kibitz-<roomId>` via the existing chat socket, and subscribes to both `kibitz-rooms` (directory) and `kibitz-<roomId>` (per-room) UIPush channels
- Adapt `GobanController.shareAnalysis()` (or add a parallel method) to post analysis bodies to channels other than the goban's game channel
- Extract shared analysis rendering from `GameChat.tsx:649-741` into a shared component; use from both game chat and Kibitz stream panel
- Variations list derived client-side from chat history filtered to `body.type === "analysis"` (grouping by `game_id` for the "Previous game" divider already implemented in `KibitzVariationList.tsx`)
- Change-board flow wired to `POST /:id/change-board`; the initiating client posts the `system`-typed chat line to `kibitz-<roomId>` on success (per §2.8)
- System-message rendering (`type: "system"`) in the stream panel

### Phase 2 (out of scope here)

- Postgres proposal tables, server clock, resolution, queue, vote state. Proposals are the only piece that needs genuinely new stateful backend modeling.
- Automatic room cleanup / decay job (manual delete shipped post-Phase-1; see §2.12).
- Group-private rooms and their visibility rules.

---

## 8. Open items to resolve during implementation (not blockers for kickoff)

- ~~**Extracting `GameChat.tsx` analysis rendering into a shared component**~~ — done; see `src/components/Chat/GameChatLine.tsx` and §2.11.
- ~~**Comm-server exposing viewer counts to the directory**~~ — resolved: live updates flow over the `viewer-count-changed` UIPush stream (§2.5); the directory's first paint is seeded by a single batched HTTP call to `POST /api/channels/viewerCounts` on comm-server (POST rather than GET to avoid the URL-length ceiling — see §2.5).

---

## 9. Delta vs. the original handoff

| Handoff                                             | This plan                                                                                                          |
| --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| 9 Phase 1 endpoints                                 | 4 endpoints                                                                                                        |
| Variations as new model + optional room-share table | Variations as typed chat bodies on Redis                                                                           |
| Stream items as new table                           | Same Redis chat list; typed bodies                                                                                 |
| Dedicated Kibitz socket namespace                   | Existing chat socket + existing UIPush                                                                             |
| Option A/B/C variation-ownership debate             | Resolved: variations are game-tree branches; visibility derived from stream items                                  |
| `board_snapshot` payload on variations              | Not needed; goban renders from `game_id` + moves                                                                   |
| Preset rooms shipped                                | Deferred — reserved in schema                                                                                      |
| Cross-feature variation reuse                       | Possible future retrofit of game chat's share feature to a first-class Variation primitive; explicitly not Phase 1 |

---

### Repository layout

- `/Users/mgregory/src/OGS/ogs/` — Python/Django backend. Actual Django code under `services/django/ogs/`. Kibitz work is on the `kibitz_phase_1b` branch.
- `/Users/mgregory/src/OGS/ogs-ui/` — React/TypeScript frontend (`online-go/online-go.com`, `kibitz_main` branch).
- `/Users/mgregory/src/OGS/ogs/services/comm-server/` — Node service for non-game chat and related realtime.
- `/Users/mgregory/src/OGS/ogs/services/game-server/` — Node service for game-associated chat and realtime.

### CLAUDE.md rules worth observing

- **ogs** (`ogs/CLAUDE.md`, `ogs/services/django/CLAUDE.md`): `make lint` before finishing. Never apply migrations — create them only. No emojis. 120-char line length.
- **ogs-ui** (`ogs-ui/CLAUDE.md`): no `any` types, no emojis, `yarn` not `npm`. One component per file for Vite HMR. Run `yarn type-check`, `yarn lint`, `yarn prettier:file <files>` before finishing. Run `yarn build` once before final push. All user-visible strings must be translated via `pgettext`/`llm_pgettext`/`interpolate()`.
- **Root** (`CLAUDE.md`): confirm reading the root CLAUDE.md at the start of any new session.

### Files of interest (most already touched)

Backend (`kibitz_phase_1b`):

- `ogs/services/django/ogs/kibitz/` — the kibitz Django app (models, views, urls, permissions, serializers, migrations). Add new endpoints / serializer fields here.
- `ogs/services/django/ogs/api/urls.py` — registers `/api/v1/kibitz/` routes via `include("ogs.kibitz.urls")`.
- `ogs/services/comm-server/comm-server.ts` — typed-body validation at lines 385-403; `chat/part` handler.
- `ogs/services/comm-server/MemChat.ts` — `MemChatMessage` type, `chatRetentionLimit`, `publishKibitzViewerCount`, the `join`/`part` paths.
- `ogs/services/comm-server/api.ts` — `POST /api/channels/viewerCounts` for the directory's first-paint seed.

Frontend (`kibitz_main`):

- `ogs-ui/src/views/Kibitz/KibitzController.ts` — talks to the live endpoints, owns the directory + active-room state, subscribes to UIPush + chat.
- `ogs-ui/src/views/Kibitz/KibitzPresence.tsx` — owner line + roster (see §2.13).
- `ogs-ui/src/views/Kibitz/KibitzSharedStreamPanel.tsx` — room/game stream pane; joins the chat channel via `chat_manager`.
- `ogs-ui/src/views/Kibitz/KibitzVariationList.tsx` — derives the variations list client-side from chat history.
- `ogs-ui/src/components/Chat/ChatLine.tsx` and `src/components/Chat/GameChatLine.tsx` — generic chat-line and game-chat-line renderers; the latter is the shared per-line analysis renderer (§2.11).
- `ogs-ui/src/lib/GobanController.ts` — `shareAnalysis` posts analysis bodies to chat channels.

### Where to find the existing patterns to mimic

- A new Django app structure: look at any of the existing small apps under `ogs/services/django/ogs/` (e.g., `api/views/group.py`, `api/views/tournament.py`)
- A typical `pushToUI` call site: `api/views/group.py:144,178,185,304`
- Scylla model patterns (for reference, not needed here): `ogs/services/_shared/lib/ScyllaModels.ts`
- Chat channel subscription from client: `ogs-ui/src/lib/chat_manager.ts`

---


Johnniedarkoo:
With regards to the last review item: 

> Potential issue: kibitzAnalysisPolicyCopy.ts looks like an orphaned file
> File: src/views/Kibitz/kibitzAnalysisPolicyCopy.ts
> 
> The PR includes both kibitzAnalysisPolicy.ts (with tests) and kibitzAnalysisPolicyCopy.ts (no tests). The Copy suffix strongly suggests this is a leftover duplicate. If it is unused, it should be removed to avoid confusion about which file is authoritative.

It is not a copy as in a duplicate, but it refers to Copy in the context of "text/messages". I will rename this to avoid confusion.

Johnniedarkoo:
There were 2 bugs and a potential issue in the last PR review. 

- Bug 1: this is a general (not kibitz specific) bug in the chatmanager.ts. The bug report overstated the issue, in fact for Kibitz this error could not occur because it chatmanager did have a Kibitz specific path. Seems the review looked at an older file. Either way, there is now a fix for the general bug committed. 
- Bug 2: this is about playwright e2e using windows screenshots, so it'd fail on other platforms. I suppose the screenshots should be regenerated on the dev end. 
- Issue: some confusing naming thing which has been renamed in a commit.  

There was one more bug in earlier PR reviews: 
- Bug: _GobanAnalyzeButtonBar extracted from PlayControls with no tests for the standalone usage_
This bug report was also partly stale, stating "Kibitz uses GobanAnalyzeButtonBar directly in KibitzBoardControls", which it doesn't. Nonetheless, the idea of adding smoke tests is good and a has been added in a commit.

Johnniedarkoo:
The last review comment contains 2 comments and a note. 

1. SYSTEM MESSAGES BYPASS THE TRANSLATION SYSTEM (violates CLAUDE.md rules)
Requires a change to the backend, and then a change to the frontend. 

2. N+1 GAME HYDRATION IS ACKNOWLEDGED ACCUMULATING DEBT (PR description section 2.10)
Same, though it is not critical for release, it's good to reformat before further development. 

3. MINOR: PR description section 2.9 describes a state the code has not reached
Current state is better than the description so this can be left.

PR #3550, opened 2026-05-11
Refactor Joseki comments from a takeover panel to an inline toggle below the explore pane, and remove the comments button from the actions panel

anoek:
@GreenAsJade


anoek:
image

GreenAsJade (approved):
Good improvement, thanks!

PR #3553, opened 2026-05-13
Add ogs-kibitz icon

anoek:
Adds the kibitz icon

Original had some rendering issues when converted into a font, so I cut out the center problematic area and pasted in the goban icon, let me know if it's ok or not:

image

To use in the kibitz branch, for the navbar do something like this:

```
<MenuLink
    title={_("Kibitz")}
    to="/kibitz"
    icon={<i className="fa ogs-kibitz" />}
/>
```


anoek:
Merged into https://github.com/online-go/online-go.com/pull/3554

Issue #3544, opened 2026-05-05
Tournament list clips off screen on iOS mobile UI


Johnniedarkoo:
**Describe the bug**
Tournament list clips off the left of the screen on iOS mobile UI, on iPhone 14 pro. The left part of the table is unreachable, zooming or scrolling cannot make the left side of the list visible. 

See screenshot:
![image](https://github.com/user-attachments/assets/51e2881b-22ab-4866-9d01-e32c093ef284)

Seth-Rothschild:
I started to look at this [here](https://github.com/online-go/online-go.com/pull/3555). The headers was straightforward: there was some CSS elsewhere that was overwriting css already on this page.

But I can't reproduce the no-scroll behavior for the table from devtools. Is it possible it's just picky about where you scroll on ios, or does no scrolling at all work? 

What I think was happening is that the "overflow hidden" meant that when you scrolled the top bar would move and the table is just totally cut off past a certain point. But scrolling from inside the table might still have worked. If that's the case then I think this fix should be sufficient. The top won't be scrollable so there won't be visible dead space and the place where you try to scroll will just work.

Johnniedarkoo:
Yes it's different now (even though maybe nothing has changed). 

Horizontal scrolling works often. However, sometimes the scrolling targets scrolling the page itself sideways (as it's too wide due to the header probably), rather than than the table in its window. 

I can't reproduce when that is happening. I tried all sorts of tap+slide locations, on text, on empty space, etc. Or first tap outside of the table. Start by vertical scrolling. But I can't find a pattern. 

Perhaps it's picky and I can't tap accurately enough to find it.

Seth-Rothschild:
Alright. I'll removed the "fixed" from the PR message which will stop the merge from auto-closing this issue. I don't know if the change will fix this entirely, but it will definitely make it less broken so it's worth adding.

ogsai:
Fixed by #3656 (merged), which scopes the tournament list table to allow content-width sizing with horizontal scroll on the wrapper instead of clipping the left side on narrow iOS viewports. Verified the fix is present on main (src/views/TournamentList/TournamentList.css).

PR #3555, opened 2026-05-13
Fix tournament mobile CSS

Seth-Rothschild:
#3544

This PR is potentially a fix for the above issue "Tournament list clips off screen on iOS mobile UI" which describes how the tournament header is weirdly positioned and it's hard to scroll the below table. Because the tab head was wider than the page on mobile because of the leaking CSS, the header centered above it. Worse though, the table had overflow hidden so scrolling high on the page scrolled the tabs and header section, and scrolling lower on the page either did nothing or scrolled the table depending on where.

This should at a minimum help with the listed issue, if not fix it completely.

Before:
image

After:
image

### DIFF
--- a/src/views/Play/Play.css
+++ b/src/views/Play/Play.css
@@ -149,16 +149,6 @@
         border-style: ridge;
     }
 
-    .tab-head {
-        display: flex;
-        flex-wrap: wrap;
-        align-items: center;
-        justify-content: space-between;
-    }
-
-    .tabs-container {
-        text-align: center;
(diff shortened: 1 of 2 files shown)


Seth-Rothschild:
@claude the automatic review to check this code for correctness and performance impact failed to trigger, can you just check it with a message like this? (update: nope!)

anoek:
@claude review this

anoek:
Thanks @Seth-Rothschild !

PR #3557, opened 2026-05-13
Followed Claude's advice to make the Kibitz icon consistent

GreenAsJade:
Fixes icon not appearing when deployed.

## Proposed Changes

  - Follow the same URL pattern and img location as others.


GreenAsJade:
(Note: this PR is only targettting kibitz_main at the moment.  I suggested an anoek review because I really don't have a strong understanding of "what is correct).

FWIW... it worked 😅

anoek:
Well, latest branch you've got the icon, no sense making the png thing work I don't think?

GreenAsJade:
The problem was that somehow I messed up the appearance of it when deployed.

kibitz_main no longer displays it

This branch does display it.

So... I'm not sure what I broke  😢

anoek:
let's just cherry pick the actual icon fix instead of this though maybe?

GreenAsJade:
JD found the issue, fixed on main

PR #3558, opened 2026-05-13
Kibitz main

GreenAsJade:
This is a revived branch for kibitz feature

## Proposed Changes

  - All the kibitz feature updates
  -


GreenAsJade:
The last two ai issues are non-issues - if the server is down, then this is the least of our worries, and the backend always does send the current_game.

GreenAsJade:
Both review comments say "if the payload doesn't include current game" ... but it does.

PR #3562, opened 2026-05-14
Kibitz jd updates

GreenAsJade:
Fixes various niggles

### DIFF
--- a/src/global_styl/01_variables.css
+++ b/src/global_styl/01_variables.css
@@ -255,6 +255,7 @@
     --navbar-highlight-background-color: $bg;
     --navbar-background-color: $bg;
     --logo-svg: url("../../assets/ogs_light.svg");
+    --mobile-room-bar-bg: #fff;
 
     /* Input */
     --input-bg: #f7f7f7;
@@ -437,6 +438,7 @@
     --navbar-highlight-background-color: #333;
     --navbar-background-color: #262626;
     --logo-svg: url("../../assets/ogs_dark.svg");
+    --mobile-room-bar-bg: var(--card-background-color);
 
     /* Input */
     $input-color: #5e5e5e;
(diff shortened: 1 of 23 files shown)


Johnniedarkoo:
The issues have been addressed and the commits can be pulled from a new branch: https://github.com/Johnniedarkoo/online-go.com/tree/kibitz-jd-latest

This includes the addition of features: resizable chat panes on desktop, and the "Variation Quick List" idea. 

The commits address 2 of the 3 review comments, the last one isn't correct. 
- Cleanup 1: _useEffect refactor in KibitzPresetChangePendingBanner.tsx._ >> true, this is cleaned up. 
- Dead code 1: _nonce in PendingSecondaryVariationBaseLoad is never checked (KibitzRoomStage.tsx)_ > true, this is cleaned up
- Dead code 2: _moveTreeRef/movePathRef update effect is a no-op (KibitzBoard.tsx)_ >> This isn't valid. Long explanation why below, intended for the PR Reviewing agent. 

Concrete case where the ref effect matters:

Secondary pane is already previewing the current game in small mode.
User starts a variation from the current board.
startVariationFromCurrentBoard() keeps preview_game_id as the same game id, but changes:

size to "equal"
variation_source_move_tree from undefined to a snapshot
variation_source_move_path from undefined to the current move path

The same KibitzBoard instance remains mounted in the secondaryGameId branch at KibitzRoomStage.tsx (line 1447).
interactive={secondaryPaneSize === "equal"} changes from false to true, so the initialization effect in KibitzBoard.tsx (line 71) reruns.
Because React runs effects in declaration order, the ref-sync effect at KibitzBoard.tsx (line 66) updates moveTreeRef and movePathRef before the initialization effect creates the new GobanController.
The new controller gets the latest move tree/path snapshot.

If that ref-sync effect were removed while leaving moveTreeRef/movePathRef, then the reinitialized board would read whatever values were captured when the component first mounted. In the case above, that could be undefined, causing the newly interactive variation board to initialize without the intended snapshot/path.

Johnniedarkoo:
Both review points are addressed.

PR #3566, opened 2026-05-17
GameHistoryTable: improved clicking

pdg137:
See discussion in https://forums.online-go.com/t/how-do-i-access-an-annulled-game/53686/5

## Proposed Changes

* fix bug that blocks clicking on annulled games
* make date a link to the game for improved interaction (e.g. middle-click, hover to see link preview)
* put date in span to vertically align with other elements

Why just link the date column? Ideally the whole row would be a link. Unfortunately [that is not possible](https://stackoverflow.com/questions/10245279/wrapping-html-table-rows-in-a-tags) without major changes to the HTML. The date is the most notable element that is not already a link and it is very logical and common on the Internet for dates to link to the thing that happened on that date. Also it's relatively far left so it gives people with smaller windows a clickable target that they don't have right now.

Before/after:

image

image

Note that the alignment in these tables is still not perfect; since game names have a `<span>` within an `<a>` tag, the link underline is messed up:

image

That's how it looks on the current site and I'm not fixing that yet.

### DIFF
--- a/src/views/User/GameHistoryTable.css
+++ b/src/views/User/GameHistoryTable.css
@@ -82,6 +82,12 @@
         color: var(--strong-loss);
     }
 
+    .date {
+        a {
+            color: unset;
+        }
+    }
+
     .board_size {
         padding-right: 0.5em;
     }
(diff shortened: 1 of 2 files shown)


anoek:
Nice, thanks @pdg137 !

PR #3567, opened 2026-05-17
Kibitz fix races

GreenAsJade:
Fixes at least one race condition, and some minor UI updates

### DIFF
--- a/src/lib/GobanController.ts
+++ b/src/lib/GobanController.ts
@@ -222,6 +222,16 @@ function getLineTreeEndpoint(lineTree: MoveTreeJson): MoveTreeJson {
     return cursor;
 }
 
+export function getMoveTreeTrunkTail(moveTree: MoveTree | null | undefined): MoveTree | null {
+    let cursor = moveTree ?? null;
+
+    while (cursor?.trunk_next) {
+        cursor = cursor.trunk_next;
+    }
+
+    return cursor;
+}
+
 export interface ReviewListEntry {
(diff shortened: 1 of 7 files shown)


src/lib/GobanController.ts
```diff
@@ -558,7 +568,9 @@ export class GobanController extends EventEmitter<GobanControllerEvents> {
         const last_estimate_move = this.stopEstimatingScore();
         this.stopAutoplay();
         this.checkAndEnterAnalysis(last_estimate_move);
-        if (this.goban.engine.last_official_move.move_number !== 0) {
+        const trunk_tail = getMoveTreeTrunkTail(this.goban.engine.move_tree);
```
anoek: This seems sketchy, why wouldn't the original (`O(1)`) operation be sufficient, seems like that should be fixed instead of introducing an `O(N)` scan here?

src/views/Kibitz/KibitzBoard.tsx
```diff
@@ -37,6 +37,14 @@ interface KibitzBoardProps {
     onReady?: (controller: GobanController | null) => void;
 }
 
+function getLiveTrunkTail(moveTree: MoveTree | undefined): MoveTree | undefined {
```
anoek: Dup

Johnniedarkoo:
Valid, taking a look

Johnniedarkoo:
I do have to get to sleep, continuing tomorrow

PR #3575, opened 2026-05-22
Overhaul game page

anoek:
This patch switches to using our GobanView class on the game page, completely getting rid of the pop out side bar and introducing a new layout option to have the player and clock info on the top and bottom of the board.

image

image

image

image


GreenAsJade:
We need a decent effort to get e2e passing on this - I haven't tried yet, but it seems likely it will break selectors.

AI is pretty good at reasoning about e2e - let me know if/when you need me to do that - or maybe your AI can run e2e now and handle it?

PR #3579, opened 2026-05-23
Kibitz update

GreenAsJade:
Fixes everything ;)


Johnniedarkoo:
Bug 1. canReconnectController creates a circular dependency on controller remount ...
is addressed in 2fc940af.

Bug 2. _isCurrentSecondaryLoadController compares context.gameId against ...
is superseeded by later updates with the epoch system and room-id tracking.

Johnniedarkoo:
Review point 1: _KibitzRoomStage.tsx — backoff ref not cleared on variation re-selection_ is superseeded since the better error handling commit b76539eb

Review point 2: _KibitzBoard.tsx — scheduleInitialResizeRetry silently terminates the retry chain_ is addressed in 3ff03e1

GreenAsJade:
Review item 1:    Corner case when game ends on move zero.   Not applicable.

Review item 2:   Claude reckons:

  - Performance: 300 pointer-chases is a few microseconds. Real but not
  a hot path issue.
  - Correctness: the value is consumed synchronously by
  isMainBoardSafeForReconnect (line 1279), which is itself called at
  render time. Within a single render the value is consistent with
  what's read from cur_move/last_official_move on the adjacent lines
  (1273-1278), which have the same staleness profile. The bigger
  correctness risk is if anything downstream uses this in a useEffect
  deps array — which it isn't here.
  
  The comment identifies a real architectural inconsistency (two
  derivations of the same fact, one event-driven and one not) but
  overstates the urgency. A cheaper interim fix is a useMemo keyed on
  something that does change on tree updates (e.g., mainBoardController
  epoch + last-known move count) — or just accept the cost since it's
  microseconds. Full state lift is worth doing only as part of a broader
   cleanup of the KibitzInner ↔ KibitzRoomStage controller ownership
  boundary.

No action required at this time.

Johnniedarkoo:
For bug 1:  a fix is in. The risk is tiny but so is the fix so why not :)
For bug 2: the assessment is correct, and splitting that ownership was already in the works. So we'll leave that as is.

PR #3584, opened 2026-06-01
Accessibility: last-move crosshair

superkeil:
Fixes #

> **Draft / WIP — depends on goban PR online-go/goban#247.**
> The `submodules/goban` pointer currently references the head of that PR. Once goban#247 merges, the submodule will be re-pointed to the merged commit on `goban:main` before this leaves draft.

## Proposed Changes

  - Add an opt-in **Accessibility** settings section with a "Highlight the last move with a crosshair" toggle (disabled by default), plus color picker (default `#1e6bff`) and a relative thickness slider, with a live `MiniGoban` preview beside the controls.
  - Add the preferences (`accessibility.last-move-crosshair{,-color,-thickness}`) and wire them into the goban via a new `getLastMoveCrosshair()` config callback in `configure-goban`.
  - Suppress the crosshair on `MiniGoban` thumbnails/preview boards by default (via goban's new `dont_draw_last_move_crosshair` flag); the setting's own preview opts back in.
  - Add the `--z-goban-crosshair-layer` z-index variable used by goban's crosshair layer.
  - Design spec + implementation plan under `docs/superpowers/`.

The rendering itself (Canvas + SVG, under-stone crosshair) is in goban#247.

## Notes for reviewers

- Manual testing done in both renderers: lines span the full board, pass under the stones, follow each move, erase cleanly on navigation; color/thickness work; absent on thumbnails and unrelated setting previews; disabled by default.
- Please test on mobile and desktop, logged-in and anonymous.


GreenAsJade:
(Note: anoek is away at the moment, please don't take lack of response as lack of interest.   A screenshot would help with a quick appreciation of what this does, for the curious)

superkeil:
Thanks for the review — all three addressed:

- **`...props.json` override:** `dont_draw_last_move_crosshair` is now assigned **after** the `...props.json` spread in `MiniGoban`, so a server-supplied config can't bypass the `lastMoveCrosshair` prop.
- **Translation function:** switched the static UI labels from `llm_pgettext` to `pgettext`.
- **Sample preview:** the board now ends on a **black** stone at the centre.

Also added a Playwright e2e test (`e2e-tests/accessibility/crosshair.spec.ts`) that drives the setting and asserts the crosshair renders with the chosen colour and thickness on **both** the SVG and the old Canvas renderer, and bumped the goban submodule to the updated goban#247 head.

superkeil:
> (Note: anoek is away at the moment, please don't take lack of response as lack of interest. A screenshot would help with a quick appreciation of what this does, for the curious)

@GreenAsJade in attachment you will find 2 screenshots

I hope it would make this PR better understood

Capture d’écran 2026-06-02 à 08 46 21
Capture d’écran 2026-06-02 à 08 47 08

superkeil:
Thanks! Two clarifications on the latest automated review:

**1. E2E SVG selector — not a mismatch (working as intended).** The two renderers deliberately use different class conventions, matching their existing layers:
- Canvas: `className = "CrosshairLayer"` (PascalCase, like `StoneLayer` / `ShadowLayer`)
- SVG: `class="crosshair-layer"` (kebab-case, like the SVG `shadow-layer` / `grid` layers)

So `.crosshair-layer` (SVG) and `.CrosshairLayer` (Canvas) are both correct. The SVG-phase assertions in the test (`.crosshair-layer line` stroke / stroke-width) pass on every run — if the selector didn't match, those would fail.

**2. Slider re-mount — consistent with the established pattern.** The key-based MiniGoban re-mount mirrors what `ThemePreferences` already does for its slider previews (e.g. `key={stone_font_scale + …}` for the stone-font-scale sample). The preview is a tiny 5×5 board shown only on this settings page, so the per-tick cost is negligible. Debouncing only this one setting would diverge from the existing convention; optimising slider previews would be better as a separate, codebase-wide change.

Happy to revisit if you'd prefer the debounce here regardless.

superkeil:
On the slider re-mount / per-tick preference write: this is the **established convention**, not new behaviour. ThemePreferences' "Stone font scale" range slider does exactly the same — `onChange={setStoneFontScale}` writes the preference on every tick, and its preview `MiniGoban` re-mounts via `key={stone_font_scale + …}` on every tick (ThemePreferences.tsx ~L378/L386). The custom-colour pickers there behave the same way.

So the per-tick remount + write isn't introduced by this PR; it matches the surrounding code. Moving slider previews to local controlled state + a single `redraw()`/`preferences.set` on pointer-release is a reasonable improvement, but it'd be better as a separate, codebase-wide change so all the settings previews stay consistent — happy to do that separately if the team wants it.

superkeil:
On the missing `useEffect` dependency (`props.lastMoveCrosshair`): this is intentional, and consistent with the surrounding code rather than an oversight.

`dont_draw_last_move_crosshair: !props.lastMoveCrosshair` is a **construction-time flag**, in the same object literal as `connect_to_chat: !!props.chat`, `enable_sounds`, `draw_*_labels`, `last_move_opacity`, `stone_font_scale` — all read once when the `GobanController` is built, and **none** of them are in the dependency array. That array is deliberately scoped to board identity (`[game_id, review_id, width, height]`): the controller is expensive to construct (chat/socket wiring, etc.), so it's only rebuilt when the board itself changes. By the review's logic `connect_to_chat` on the line above would be an identical 'gap'.

On the dynamic-toggle scenario: the only caller that passes the prop is `AccessibilityPreferences.tsx`, with a constant `lastMoveCrosshair={true}`, and that preview already remounts via a `key`. Every other `MiniGoban` usage omits the prop entirely. So adding it to the deps is a provable no-op for all current call sites, and `react-hooks/exhaustive-deps` (which we run in CI) does not flag the current array.

I've added a short comment at the flag making this explicit (construction-time, why it's not a dep) so the intent is clear to future readers. If the team would rather move all settings previews to controlled local state + a single `redraw()` (which would also address the slider-remount thread above), that'd be a good separate, codebase-wide change.

superkeil:
> ## Code Review
> **Summary:** Adds an opt-in accessibility setting to draw a crosshair through the last-played stone, with color/thickness controls and a live MiniGoban preview.
> 
> **Performance: preview board re-mounts on every slider step**
> 
> `src/views/Settings/AccessibilityPreferences.tsx` — the MiniGoban uses `key={`${color}-${thickness}`}` to force a remount whenever color or thickness changes. For the color picker this fires once per pick, which is fine. For the range slider (step=0.01, range 0.02–0.40) a single full drag fires ~38 separate `onChange` events, each one tearing down and re-initializing the entire goban (canvas/SVG setup, layout, event wiring). On slower devices this will cause visible flicker and may stall the UI thread mid-drag.
> 
> A simple debounce on the key — or on just the `thickness` preference write — would reduce this to a handful of re-mounts per drag rather than one per step, without requiring any goban API change. The color picker can keep the direct key since it only fires on commit.

already answered at https://github.com/online-go/online-go.com/pull/3584#issuecomment-4599831711

PR #3585, opened 2026-06-10
Add support for multiline input in private chat

boonespace:
- Adds multiline input to PrivateChat (same approach as #3513) + IME fix from #3582. 

img

img

- Also fixes a small gap in Game Chat (some older versions or ESR versions of Firefox).

Before:
img

After:
img

### DIFF
--- a/src/components/PrivateChat/PrivateChat.css
+++ b/src/components/PrivateChat/PrivateChat.css
@@ -198,6 +198,7 @@
         /* border-bottom: 1px solid #aaa; */
         padding-right: 0.3em;
         overflow-anchor: none;
+        white-space: pre-wrap;
     }
 
     .date, .timestamp {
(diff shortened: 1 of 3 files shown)


GreenAsJade:
Thanks for this!   Note that anoek is away at the moment - please don't take "inaction" as "lack of interest"!

anoek:
Looks great.

Did we have to do anything in the main chat to get multi-line display working? I notice in private chat we're stripping white space between words out I reckon, so the display isn't what we'd expect after entering a multi line input.

boonespace:
Got it — I've fixed the CSS to properly display line breaks in private chat messages, while using sanitizeMessage to limit excessive whitespace.

anoek:
Awesome, thanks!

PR #3588, opened 2026-06-14
Fix "undefined" for suspension reason when we can't get a translation.

GreenAsJade:
Fixes suspended people getting "undefined" for the suspension reason

## Proposed Changes

  - Handle failed translation better
  - More robust update (for e2e mostly)


GreenAsJade:
It's true that this will case a re-translate, but worth it I think.

The first reviewer comment is misunderstanding what was fixed.

anoek:
Thanks!

PR #3590, opened 2026-06-15
Add e2e test for warning dialog during games.

GreenAsJade:
Fixes not knowing for sure that this works.

## Proposed Changes

  - Test that dialog appears during corre game, but after live game


GreenAsJade:
That context is cleaned up later.

PR #3596, opened 2026-06-20
Feature/malicious report report type

GreenAsJade:
Fixes CMs not being able to easily respond to trollish reports

## Proposed Changes

Support the "malicious report" report type.

 - Report type raiseable only by CMs (from their ViewReport page)
 - Report type handled by suitable empowered CMs via voting
 - e2e tests for all primary activities
    - Wider ranging changes in e2e tests and infra tackling fragility

Needs:  https://github.com/online-go/ogs/pull/2416
https://github.com/online-go/moderator-ui/pull/52


GreenAsJade:
Note: yes, this is a CM-only feature.   It's raised in the context of CM-type reports only for CMs to consider whether those are malicious or note, by voting on it.

If moderators think that a report is malicious they already have the power to act immediatley, they don't need to raise a report like this, and it would be confusing if they did do that.

PR #3597, opened 2026-06-21
Fix delete button shown on root SGF collection

### LINKED ISSUE #3594: Present delete button when there's no sgf collection to delete
When opening your SGF library when there's nothing in it, you see a delete button for the root collection.

Image

teauxfu:
Fixes #3594. When navigating to the root SGF collection via the breadcrumb, `componentDidUpdate` was storing the result of `parseInt()` into `collection_id` (a `string`). This caused the root collection check `collection_id !== "0"` to pass. The fix removes the `parseInt()` call so `collection_id` stays a `string`. As a follow-on, the repeated `collection_id !== "0"` checks throughout the component could be consolidated into an `isRootCollection` helper, but I'll leave that style decision to another


anoek:
Thanks for the fix!

PR #3600, opened 2026-06-23
Escape rate predictive badge

GreenAsJade:
Fixes wrong treatment when "this" report would push the person into the "unacceptable rate".

## Proposed Changes

  - Make the badge show whether it's acceptable _including this one_.
 

Needs https://github.com/online-go/ogs/pull/2417


GreenAsJade:
These are as intended - the super-slow one has 30s quiesce waits to reduce fragility from the post processing of its multiple games.

GreenAsJade:
Nah, bot's lost the plot there, this is all good now.

PR #3609, opened 2026-06-30
Cancel in progress tests when a new merge to main is made

GreenAsJade:
Fixes wasting cycles on previous merges

## Proposed Changes

  - Cancel runs in progress when a new merge happens


GreenAsJade:
That's enough - it needs to bomb if my folders went away :)

Issue #3608, opened 2026-06-29
More expressive theme system


lykahb:
I'd like to share a few ideas about improving themes before sending PR's. I hope that fits the OGS vision. My goal is to make OGS boards look as awesome as in the screenshot from my exploration, while keeping the UX simple.

# Ideas
- Expose controls for the label color. Now it is derived from the grid color using `blendWithInverseColor`. The bright and dark colors can't be expressed.
- Custom stone variants, based on URL. Goban already supports stone variants for the default shell theme. I don't have a strong opinion if the UX should take a list of URL's or a sprite. For simplicity, best to support only one.
- Expose controls to scale the stones. If an image is larger than the grid square, allow the images overlap. This works well for the glow or smoke effects - see the screenshot.
- Simplify sharing themes - by import/export of JSON. This is inspired by VSCode settings.json `workbench.colorCustomizations` and Slack's "Share your theme".

# The endgame
This board comes from my experiments. It has many stone variants, custom board with grid implemented at [PR](https://github.com/online-go/online-go.com/pull/3593), scaled stones with overlapping - see the plasma extending from the stars at K10-L10, B4-B5.

<img width="773" height="769" alt="Image" src="https://github.com/user-attachments/assets/7d771f14-3f65-49fc-818e-5ab54e70bead" />

anoek:
Those all sound good.

For " If an image is larger than the grid square, allow the images overlap." - that's only really doable on the SVG renderer, don't bother with having the AI do the work for the canvas, it's a lot bigger of a challenge than you'd think to get right for that one. Since it's basically deprecated, just do the SVG I'd say.

lykahb:
Done!

PR #3611, opened 2026-07-01
Custom board label color

lykahb:
This adds a new setting for the label color, as I suggested at #3608 

The UX looks better when all pickers remain on a single line.

I did not keep the inherited defaults because they were designed for two color pickers: fixed input width, left margins per input, and reset-button top margin. With three pickers those defaults either overflow or need fragile hand tuning. The new rules make the layout express those constraints: one preview, three equal controls, default-size inputs when possible, responsive shrinking when needed.

Now that there are three colors, as a side improvement I added titles to the inputs. The coding is mostly done with Codex.

Tested on the desktop and mobile views.

image

Also see https://github.com/online-go/goban/pull/257

### DIFF
--- a/src/components/GobanThemePicker/GobanThemePicker.css
+++ b/src/components/GobanThemePicker/GobanThemePicker.css
@@ -109,6 +109,30 @@
 }
 
 .GobanCustomBoardPicker {
+    .custom-board-theme-set {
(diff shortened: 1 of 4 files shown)


anoek:
Thanks!

PR #3612, opened 2026-07-05
Add stone scale theme setting

lykahb:
This adds a new setting for the stone scaling, as I suggested at https://github.com/online-go/online-go.com/issues/3608

image

It also fixes a regression from ce49240bfe. The stone url edits weren't updating the stone preview. That path imperatively appended SVG nodes and tried to clean them up via `div.current`, so URL edits could leave the old SVG definition visible.

The fix is make the preview container own a single SVG. `ThemeSample` now captures the host element, rebuilds the SVG when theme inputs change, and uses `replaceChildren(svg)` so custom stone URL edits replace the stale `<defs>` image reference immediately.

See also https://github.com/online-go/goban/pull/258

### DIFF
--- a/src/components/GobanThemePicker/GobanThemePicker.tsx
+++ b/src/components/GobanThemePicker/GobanThemePicker.tsx
@@ -693,7 +693,8 @@ function ThemeSample({
     const [stone_scale] = usePreference("goban-theme-stone-scale");
 
     React.useEffect(() => {
-        if (!div.current) {
+        const host = div.current;
+        if (!host) {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


lykahb:
Clipping is expected - the scaling factor over 1 is for the cases where the stone takes a smaller part of the asset image, and the margin is either blank or has subtle visual effects that are okay to clip.

The difference of how SVG and canvas renderers handle the scale factor over 1 is intentional. Anoek recommended this at https://github.com/online-go/online-go.com/issues/3608

lykahb:
With the extreme scale factors the stone themes indeed look worse. That is consistent with the current convention where every setting applies to every preview.

This concern is similar to choosing a background that matches the custom stones well, but not the default themes.

PR #3615, opened 2026-07-11
Drop deprecated tsconfig baseUrl

GreenAsJade:
## Proposed Changes

  - Remove `baseUrl` from `tsconfig.json` (flagged as deprecated/unnecessary by the TypeScript language server under `moduleResolution: "bundler"`).
  - Prefix each `paths` entry with `./` so TypeScript resolves them relative to `tsconfig.json`'s directory (required once `baseUrl` is gone; produces `TS5090` otherwise).

Without `baseUrl`, only explicitly-declared `paths` aliases are used for module resolution — no silent shadowing of npm packages by same-named files under the project root.

Verified with `yarn type-check`, `yarn lint`, and `yarn prettier:file tsconfig.json`.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,19 +12,15 @@
         "emitDecoratorMetadata": true,
         "experimentalDecorators": true,
         "lib": ["es2020", "dom"],
-        "baseUrl": ".",
         "paths": {
-            "@/*": ["src/*"],
-            "goban": ["submodules/goban/src"],
-            "goscorer": ["submodules/goban/src/third_party/goscorer/goscorer"],
-            "react-dynamic-help": ["submodules/react-dynamic-help/src"],
-            "@helpers/*": ["e2e-tests/helpers/*"],
-            "@helpers": ["e2e-tests/helpers"],
-            "@stubs/*": ["src/stubs/*"],
-            "@moderator-ui/*": [
-                "submodules/moderator-ui/*",
-                "src/stubs/moderator-ui/*"
-            ]
+            "@/*": ["./src/*"],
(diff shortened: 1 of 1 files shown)


GreenAsJade:
This was motivated by tsconfig.json suddenly showing up with squiggles.

My Claude said:

```
  What baseUrl actually does (beyond helping paths): it makes any file under that root importable as a bare specifier.
  If you have src/react.ts and baseUrl: ".", then import React from "react" can silently resolve to your local file
  instead of node_modules/react. This is a real footgun â TS users get bitten by it periodically.

  Why paths alone is preferred: paths is a whitelist. Only the aliases you explicitly declared work. No accidental
  shadowing of npm packages, no "wait, why is this import resolving to a local file I didn't know existed."

  Why the verbosity is intentional: the ./ prefix forces you to acknowledge these are filesystem paths anchored to
  tsconfig.json's directory. Without it, TS can't tell "is this a name that should resolve to a bare import (like from
  node_modules) or a filesystem path?" â hence the TS5090 error we hit. Explicit-over-clever is TS's usual stance for
  module resolution.

  But there's a real tradeoff:

  - The "footgun" only matters if someone adds a top-level file with the same name as an npm package. In practice,
  unlikely for OGS.
  - The squiggle is a hint/suggestion, not an error. yarn type-check was clean before my change too (I only saw the LSP
  squiggle, not a compiler error).
  - The cleanup swaps one line of implicit-magic-that-works for nine lines of explicit-verbosity-that-also-works.
  ```

... so I threw this in and let you look at it @anoek :)

PR #3618, opened 2026-07-15
Add custom stone asset variants

lykahb:
This adds support for multiple custom stone image URLs in the theme picker, allowing users to enter one variation per line. I suggested this at https://github.com/online-go/online-go.com/issues/3608. Existing single-URL preferences are migrated automatically, and the legacy canvas renderer continues to use the first configured image.

The new editor includes translated guidance, reset controls, and component tests.

Default view when zero or one custom stones are set
image

Textarea view for multiple stones
image

The textarea lets user view and set the URL's in bulk. It occupies little space but can be expanded. It also avoids the complexity of building a list-based UX.

It requires https://github.com/online-go/goban/pull/260

### DIFF
--- a/src/components/GobanThemePicker/GobanCustomStoneUrlInput.css
+++ b/src/components/GobanThemePicker/GobanCustomStoneUrlInput.css
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C)  Online-Go.com
+ *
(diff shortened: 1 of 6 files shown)


lykahb:
I'll update the submodule once goban is merged. I can also squash the commits for cleaner history

PR #3622, opened 2026-07-18
fix(anti-stall): only show stalling estimate to the predicted winner

benjaminpjones:
This seems to be a relatively uncontroversial modification of the stalling system.  See discussion in https://forums.online-go.com/t/can-we-please-get-rid-of-ai-interrupting-a-game-and-ending-it/55280/115

## Proposed Changes

  - The `AntiStalling` card gated visibility on "is this user a player in this game", so **both** players saw the predicted winner and the *"Accept predicted winner and end game"* button — including the player predicted to lose. Prompting the player who is behind to end the game is effectively a resign button by another name, which isn't what the anti-stalling feature is for.
  - Gate on the viewing player's color instead: a player only sees the card when their color matches `estimate.predicted_winner`. Non-player moderators still see it for either color.
  - `any` -> `StallingScoreEstimate | null` so the new color comparison is type-checked.
  - Add `AntiGrief.test.tsx` covering the visibility matrix.

NOTE: This is a **client-side visibility change only**. The `sendPreventStalling` path is now unreachable from the UI for the losing player, but it isn't blocked server-side — this change could easily belong server side, but I'm submitting here because it's the open source repo.

## Testing

I tried to test in-browser, but couldn't trigger Anti-stall on the beta server because the AI server is currently dead. Here was my attempt: https://beta.online-go.com/game/20986

I added a jsdom test suite to simulate the anti-stall message and verify when the messages are shown.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/views/Game/AntiGrief.test.tsx
+++ b/src/views/Game/AntiGrief.test.tsx
@@ -1,0 +1,135 @@
+/*
+ * Copyright (C)  Online-Go.com
+ * Copyright (C)  Benjamin P. Jones
+ */
+
+import { AntiGrief } from "./AntiGrief";
+import { render, screen } from "@testing-library/react";
+import * as React from "react";
+import * as data from "@/lib/data";
+import { MemoryRouter as Router } from "react-router-dom";
+import { GobanControllerContext } from "./goban_context";
+import { OgsHelpProvider } from "@/components/OgsHelpProvider";
+import { GobanController } from "../../lib/GobanController";
+
+const BLACK_PLAYER_ID = 123;
+const WHITE_PLAYER_ID = 456;
+const SPECTATOR_ID = 789;
+
+const TEST_USER = {
+    anonymous: false,
+    id: BLACK_PLAYER_ID,
+    username: "test_user",
+    registration_date: "2022-05-10 11:03:24.299562+00:00",
+    ratings: {
+        version: 5,
+        overall: { rating: 1500, deviation: 350, volatility: 0.06 },
+    },
+    country: "un",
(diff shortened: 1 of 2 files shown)


anoek:
I'm going to mark this as draft as we stew on the appropriate behavior. trohde didn't realize the student clicked the button themselves when he made the initial comment, he thought it was ended by the system which it was not. To me it's clear that the student was playing and didn't know they had lost or know when to stop playing, so the nudge by the system was what they needed to say "Oh ok it's time to end the game, got it", which I think is a net positive for most beginner cases. 

What the change your proposing does is eliminates the beginner nudge, the "Hey FYI the game is over, the polite thing to do is resign" hint. I don't think that getting rid of that is a good thing. Folks coming from chess often carry over the "play to the end" courtesy that's common in chess, which in Go is received as stalling and causes reports to be generated and beginners to be banned, so I think in general having this feature when the player is a beginner to Go is overall a good thing.

I could get behind changing the messaging, and adding a way to hide the message like some in the forum have advocated for, but it'll take some good points to change the behavior away from helping beginners not get flagged as rude stallers when they're just getting started on OGS.

benjaminpjones:
Agree it sounds reasonable to nudge a beginner to finish the game.

W.r.t. messaging, I think it would be better if there were some message to that effect, and also the system should nudge these players to use the usual buttons (resign/pass) rather than the anti-grief button.  Otherwise, I think it could have the adverse effect of telling users "it's not over until OGS says it's over", and the anti-grief system is very conservative, by design.

PR #3623, opened 2026-07-25
Add customizable colors for custom stone markers

lykahb:
## Proposed Changes
- Add separate marker color preferences for Custom black and white stone themes.
- Add marker color inputs to next to the existing custom stone colors.
- Show an “A” marker on each custom stone preview. I also considered using the last move indicator, but I think that the color fit is easier to check with a marker that overlays both the edges and the center of the stone.

This preserves the existing behavior. The opponent's custom stone color is used for both marker and territory squares by default.

Depends on https://github.com/online-go/goban/pull/261. Since most changes are in that PR, I included the details and motivation for the change there.

Screenshot:
image


lykahb:
Thanks! I've merged the recent changes and updated the goban reference.

anoek:
Awesome, thanks!

PR #3625, opened 2026-07-30
 Fix puzzle editor hiding the board on mobile portrait screens

anoek:
Add a keepGobanVisible option to GobanView takeover tabs so panels that need to be used alongside the board, like the puzzle editor, render in the scroll area below the goban instead of as a full-screen takeover. Apply it to the puzzle edit tab and add an e2e test covering the mobile edit flow.


anoek:
@claude review this code

PR #3636, opened 2026-08-01
Sharing goban themes

lykahb:
This is the last part of the [More expressive theme system #3608](https://github.com/online-go/online-go.com/issues/3608).

## Summary

Adds controls to copy and import goban themes as JSON that is decoupled from the internal preferences representation. It supports everything that affects board look - except for the personal display and accessibility preferences.

The imports are fully parsed, migrated, and validated before replacing the theme settings. Clipboard fallback, translated UI, documentation, and test coverage are included.

## Motivation

The themes I developed need tweaking about ten settings to apply. I'd like to make this easy to do, but in a more portable manner than adding those themes as OGS defaults. I would expect players to share their own themes as JSON on the OGS forum.

As the settings evolve, the older snippets should be handled in a consistent manner. Most of the complexity in this PR comes from that - it includes strict runtime validation, stable defaults, versioned migrations, atomic application.

## Refactoring
Centralizes goban theme defaults in a module.  Reuses those defaults across preferences, imports, and theme-editor reset controls. It affects a lot of the existing code, but in a mechanical manner. I decided that this is better than repeating defaults in yet two more places. This improves on the existing duplication.

## Testing
1. Copy your existing settings as a backup by using Copy Theme
2. Apply theme JSON from https://github.com/lykahb/goban-decals/tree/master/themes/sand-and-pebbles.

image

Implemented with Codex, with close steering for the key technical decisions.

### DIFF
--- a/docs/goban-theme-json.md
+++ b/docs/goban-theme-json.md
@@ -17,17 +17,23 @@ opacity, variation display, undo indicators, and accessibility settings are excl
 
 ## Effective theme settings
 
-The selected `theme` is authoritative. A built-in board theme such as `Kaya` supplies its
-own background, grid, and label colors, so its exported `board` object contains only the
-selected theme name. The `custom` object is required when the selected theme is `Custom`
-and is rejected when a built-in theme is selected.
+The JSON stores only the settings that currently affect how the goban looks.
 
-Black and white stones follow the same rule independently. Custom shadow colors and
-gradients are included only when the selected shadow style is `custom`.
+For example, a built-in board theme such as `Kaya` already defines its background, grid,
+and label colors. When `Kaya` is selected, the exported `board` object therefore contains
+only the theme name. When `Custom` is selected, the `custom` object is required and
+contains the custom board settings. A `custom` object is not allowed for a built-in theme.
 
-This makes the document a portable description of the effective goban appearance rather
-than a backup of dormant theme-editor settings. Importing a built-in selection resets its
-inactive Custom settings to the documented defaults.
+Black and white stones follow the same rule separately: each stone color can use either a
+built-in theme or `Custom`. Custom shadow colors and gradients are exported only when the
+selected shadow style is `custom`.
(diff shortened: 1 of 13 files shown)


lykahb:
The [https://github.com/lykahb/goban-decals](https://github.com/lykahb/goban-decals/tree/master/themes/sand-and-pebbles) is still not published. I plan to announce it once OGS accepts this change, and Sabaki publishes the next release. A few themes need the last touch-ups too.

I'd be happy to adjust the JSON schema or code if you have any concerns

ogsai:
### Code review

Found 1 issue:

1. Custom board colors are exported without a fallback, so a legacy `""` value produces JSON that the importer rejects. `getGobanThemeConfig()` forwards `goban-theme-custom-board-background` / `-line` / `-label` straight from preferences, while the adjacent stone-color export guards with `customStoneColor(value, fallback)` (`value || fallback`). `colorAt()` validates against `/^#[0-9a-f]{6}$/i`, so `""` fails import — including a self round-trip. Empty values are reachable: `migrate_key("custom.board", "goban-theme-custom-board-background")` in [preferences.ts](https://github.com/online-go/online-go.com/blob/e4451cba92917d60575528ac10ce4c3ea063dc95/src/lib/preferences.ts#L449-L451) persists `data.get(from) || ""`, and 7db60e3ac ("Fix crash on invalid color code for custom board colors") fixed a crash caused by `""` in these exact keys. The shadow color fields at L169-L177 have the same gap.

https://github.com/online-go/online-go.com/blob/e4451cba92917d60575528ac10ce4c3ea063dc95/src/lib/goban_theme_json.ts#L112-L116

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anoek:
Not worried about that.


Thanks for all your work making this happen @lykahb !

PR #3637, opened 2026-08-02
Replace react datetime

GreenAsJade:
react-datetime is unsupported and has a typing error revealed by vite 8

Get rid of it.


anoek:
On firefox the native datetime input here doesn't support editing the time.

On firefox the widget is always "dark mode"

On chrome the widget is always "light mode"

GreenAsJade:
Summary:

## Why

`react-datetime@3.3.1` contradicts itself. Its shipped TypeScript declarations say `export = ReactDatetimeClass`, but its CJS bundle actually sets `module.exports = { __esModule: true, default: Class }`. Vite 7 papered over this by always unwrapping `.default`; Vite 8 implements Node's ESM interop faithfully and hands you the wrapper object instead.

The result is that every site rendering `<Datetime>` throws:

```
Element type is invalid: expected a string (for built-in components) or a
class/function (for composite components) but got: object.
```

This arrived with the Vite 8 bump in #3634 (`3c18ee3c4`). It was found because the moderator ban modal stopped opening, which showed up as a failing Playwright test, but tournament start date and dev/styling also render it.

**Upstream is a dead end.** The defect is [arqex/react-datetime#770](https://github.com/arqex/react-datetime/issues/770), open since January 2021 with no maintainer response, reporting the identical error string. The root cause is visible in their `config/webpack.config.build.js`, which sets `libraryExport: 'default'` on the UMD output and omits it on the CJS output that `main` points at. The project has not merged a community PR since October 2022 and has not committed since December 2024. A one-line fix exists but would be semver-breaking for anyone currently reaching for `.default`, and there is no one to merge or release it.

So rather than shim the type, this removes the dependency.

At the time of writing, production was on `5.1-10298-gb22865ec9`, which predates the Vite 8 bump — so this is pre-emptive rather than a live outage, but `main` is not deployable until it lands.

## What

Three call sites replaced with native `<input type="datetime-local">`, backed by two pure helpers in `src/lib/datetime_input.ts`:

- `toDatetimeLocalValue(date: Date): string` — local wall-clock `YYYY-MM-DDTHH:mm`, `""` for an invalid date
- `fromDatetimeLocalValue(value: string): Date | undefined` — the inverse, `undefined` for empty or unparseable

Both are local-time only by design. `<input type="datetime-local">` carries no timezone, and both call sites already convert to UTC at the point they send to the server, so the conversion boundary is unchanged.

| Site | Change |
|---|---|
| `BanModal` / `BanDetails` | Ban expiration field |
| `Tournament` | Tournament start time |
| `Styling` | Styleguide demo removed — a native `datetime-local` already sat on the line above it |

`time_start` keeps its existing `moment().format()` shape, so the save path (`Tournament.tsx:575`) and display path (`:888`) are untouched. `ban_expiration` still reaches `.toISOString()` as `Date | undefined`, and a blank field still omits the key from the PUT entirely.

Two latent bugs fixed incidentally: `<label htmlFor="start-time">` previously pointed at no element, and a missing `time_start` used to render react-datetime's literal "Invalid date" text.

## Why these other changes are here

Four changes in this PR are not about react-datetime. Three are here because **`e2e-tests.yml` runs on push to `main`** — merging with red e2e families would break the post-merge run for everyone.

**`AIReview.tsx` / `hooks.ts` — a real app bug.** `setSelectedAIReview` called `close_all_popovers()`, and it is reached both by the user picking a review *and* by the mount effect that auto-selects the initial one. So opening a player popover in the Reports Center while a reported game's AI review was still resolving destroyed the popover ~180 ms later. Confirmed with instrumentation showing the call arriving from `commitHookPassiveMountEffects`. The call now lives only in the user-initiated handler. This is moderator-facing, not just a test problem — the `@CM` failure was the symptom.

**`suspended-user-profile-updates.ts` — a pre-existing test race.** `AccountSettings` overwrites the username field when its `me/account_settings` fetch resolves; the tests typed before that landed and had their input silently discarded. In the captured trace the response returned 200 at t=18991.4 ms and `fill()` ran 21.6 ms later. Unreachable until this branch stopped the ban modal crashing earlier in the same test.

**`matchers.ts` — unbounded action.** `playwright.config.ts` sets no `actionTimeout`, so `scrollIntoViewIfNeeded` on an element that never appears blocked until the whole test timed out — turning legible failures into silent 420-second hangs with no call log, and starving the surrounding 3-attempt retry loop past attempt 1. Now bounded at 5 s.

**`close-all-pending-reports.ts`** — reliability fix for the report-clearing utility: the history table renders empty before its fetch completes, so reading it for pending rows could conclude there were none while data was still in flight.


## Notes for review

- **`jest.config.cjs`** pins `TZ=Australia/Adelaide`. The zone is arbitrary and the comment says so — it is chosen for a non-zero, half-hour offset with DST. Without a non-UTC zone, the new assertions cannot fail: under UTC a wrong `toISOString().slice(0, 16)` implementation produces byte-identical output to the correct one. It sits in `jest.config.cjs` rather than `setupFiles` because Node caches the zone on first `Date` construction, which Jest does during startup — verified that the `setupFiles` placement is inert.
- **`color-scheme` is scoped to `input[type="datetime-local"]`**, not declared at `:root`. The root-level version would make every browser-drawn control follow the theme — roughly 32 `range`/`color` inputs, all checkboxes and radios, `<select>` option lists and autofill — which is a worthwhile change but not one to bundle with a dependency removal. The theme mixins carry `--control-color-scheme` as an inert custom property, so that follow-up is a one-line addition.
- **`BanDetails` moved to its own file.** It was already a second component inside `BanModal.tsx`, which CLAUDE.md forbids; this PR made it worse by exporting it for a test, so it now has its own co-located file. No `BanDetails.css` — its only styled element is a spacing rule that belongs to the modal's layout.
- The commented-out per-round start-time block in `Tournament.tsx` was deleted. It was the last `<Datetime>` reference in the tree and would have been the first misleading hit for anyone grepping for remnants.

## Deliberately not included

- Declaring `color-scheme` at `:root` (see above).
- Reworking `Tournament.tsx` to hold a raw input string. Clearing a *segment* there wipes the in-progress edit, because the cleared value is unparseable and React re-renders from the unchanged `time_start`. `BanDetails` is immune since it holds the raw string. This matches how `react-datetime` refused invalid input previously, so it is not a regression, but it is the most likely thing to be misread as one.
- An accessibility and i18n pass over `BanModal`, whose three fields use `<h3>` headings rather than labels and are untranslated English.

anoek:
Thanks!

Issue #313, opened 2017-05-31
When refreshing the game list, it forgets the current page number


viridium:
Can the page number be remembered through page refreshes (and implicitly through browser navigation events)?

anoek:
I think this would produce some annoying behaviors, I think that if you manually refresh (which there should be no reason to do generally speaking), then it'd be nice to keep your page, whenever you are navigating there I think you generally want to start on page 1 as the lists update a lot. (Ie, if you went and played a game, went back to the watch page, it'd be weird to start on page 5)

viridium:
What if I hit the Back button because the game I just opened is not to my liking? Should I have to manually go back to page 5 to continue looking for games to watch?

anoek:
Fair enough, and encoding this on the url would mitigate what i was thinking.. or not thinking as the case may be.

anoek:
This issue has been marked stale and will be closed soon without further activity. To keep the issue open, please respond to the comment to keep the discussion going.

viridium:
Mnnnooooooooooooo!

benjaminpjones:
Linking similar issue #1398

pt8o:
I've personally wanted something like this and also thought we could just push the `page` and `page_size` into the URL params, so I started investigating. What could get messy is that there are a *lot* of parameters that get sent in the request for the list of games. Do we want those all encoded in the URL? It could get cluttered.

In addition to `page` (number) and `page_size`, there's also `alt_player` (opponent id), `height` and `width` (board size), `ranked`, `annulled`, `ordering`, `source` (not sure what this one is), and `ended__isnull`.

If we put all of these in the URL params then that adds a whole lot of clutter to the URL, which might not be so bad nowadays (happens a lot in online shopping for example).

On the other hand if we exclude some of these from the URL params then we could get into situations where the user's browser history doesn't reflect their app state. For example a user clicks through to the 3rd page of their unfiltered games, and then filters only `ranked=true` games. But then if they press back in the browser, they'll end up on the 3rd page of `ranked=true` games, because the table filters are saved in the settings. So they state they ended "back" at is different from the state they actually were at in real life. :thinking:

I'm inclined to just put all of the query params into the URL and let the URL get super long but I'm curious what you think @anoek

viridium:
user experience > URL hygiene

teauxfu:
[...]

PR #3639, opened 2026-08-03
Persist PaginatedTable page number across refreshes

cMohammadT:
## Summary

Fixes / partially addresses #313 ("When refreshing the game list, it forgets the current page number").

### What changed

`PaginatedTable` already persisted `page_size` via `data.get` / `data.set`. This change does the same for the current **page** number:

- On mount, the table restores the last page the user was on for that table name.
- When the user changes page (arrows, typing, page size change), the new page is stored.
- When filters change, the page still resets to 1 (existing behaviour preserved).
- If the restored page is past the end of the result set, it is clamped to the last valid page.

### Why this approach

The discussion on #313 prefers encoding state in the URL for full browser history support. That is a larger change. This patch is a smaller, low-risk improvement that solves the common "I refreshed and lost my place" case using the same storage mechanism the component already uses for page size.

### Testing

- [ ] Manual: open a paginated list (e.g. user game history), go to page 3+, refresh → should stay on same page.
- [ ] Manual: change filters → should reset to page 1.
- [ ] Desktop + mobile.

### Notes

AI-assisted contribution (Grok). I have reviewed the change. Happy to adjust based on maintainer feedback.

### DIFF
--- a/src/components/PaginatedTable/PaginatedTable.tsx
+++ b/src/components/PaginatedTable/PaginatedTable.tsx
@@ -85,10 +85,14 @@ function PaginatedTableImpl<RawEntryT = any, GroomedEntryT = RawEntryT>(
     ref: React.ForwardedRef<PaginatedTableRef>,
 ): React.ReactElement {
     const table_name = props.name || "default";
+    const initial_page = data.get(
+        `paginated-table.${table_name}.page`,
+        props.startingPage || 1,
+    ) as number;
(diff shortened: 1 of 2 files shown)


anoek:
This PR would cause the pages to be remembered both across tables that have the same name and through navigation (for instance, navigating from one profile to another, or just back to the same profile). During normal navigation you expect to start off on page 1, this would make that not true and keep pages sticky.

Issue #3601, opened 2026-06-24
suggestion: add back the "key moves" label in the AI Review component


teauxfu:
Some screenshots in the docs from earlier versions show this "key moves" label, I haven't done a git blame yet but it looks like it was intentionally removed from the markup at some point. I think the label was useful. I figured out what the buttons were for eventually, but as a brand-new user it was kind of confusing. 

I wanted to ask if there were any particular reasons for removing it, and to see how we felt about bringing it back?

![](https://camo.githubusercontent.com/8cba574160ce549fcf9fa57e4b0546824899a333a6ab3d591a8215a2155a5321/687474703a2f2f6164616d72757a69636b612e637a2f4f47532f67726170682e6a7067)

https://github.com/online-go/online-go.com/wiki/Understanding-the-AI-review#the-graph

https://github.com/online-go/online-go.com/blob/1c52f403bfd6b28ef1109aea8621836fdcfd9fdc/src/components/AIReview/AIReview.tsx#L563-L571

ogsai:
Fixed by #3640 (merged), which restored the "Key moves" label markup in the AI Review worst-moves list. Verified on main (src/components/AIReview/WorstMovesList.tsx).

PR #3640, opened 2026-08-03
Restore "Key moves" label in AI Review

cMohammadT:
Restores the "Key moves" label above the worst-moves buttons in the AI Review panel.

The CSS for `.key-moves` was still present; only the markup had been removed. New users found the unlabeled buttons confusing (#3601).

Also sets the worst-move list container to column flex so the label sits above the moves.


anoek:
Thanks!

Issue #3547, opened 2026-05-05
Email preference screen has the toggles slightly off the screen on Mobile


Johnniedarkoo:
**Describe the bug**
Email preference screen has the toggles slightly off the screen on Mobile.
The toggles are accessible by horizontal scroll, but isn't not a good experience. 

Proposed fix: reduce width of the container so that it fits without horizontal scroll. 

This is on iOS iPhone 14 pro. 

**Screenshots**
<img width="1179" height="1873" alt="Image" src="https://github.com/user-attachments/assets/58dcd047-5263-4eec-ac2b-29847307262c" />
_The cutoff isn't a screenshot artefact, that is how far the screen goes_

ogsai:
Fixed by #3641 (merged): the preference-toggle-name element no longer uses a fixed 20rem width, so it shrinks instead of pushing the toggle off-screen on narrow phones. Verified on main (src/views/Settings/Settings.css).

PR #3641, opened 2026-08-03
Fix email preference toggles overflowing on mobile

cMohammadT:
The email notification preference name used a fixed 20rem width, which pushed the toggle off-screen on narrow phones (e.g. iPhone 14 Pro).

Allow the label to shrink and keep the toggle visible without horizontal scroll.

Addresses #3547.


anoek:
Nice, thanks!

PR #3644, opened 2026-08-07
UI bug: fix friend and notifications overlap

djma:
The bug happened because each dropdown had separate state:

- Notifications used `NavBar` state.
- Friends used its own local state.
- Both could be open at the same time, so CSS stacking made one appear over the other.

The fix uses the existing `MenuContext` as one shared menu state:

```ts
activeMenu = "notifications" | "friends" | "right-nav" | "settings" | null
```

Clicking notifications sets `activeMenu` to `"notifications"`. Clicking friends changes it to `"friends"`, which causes the notification component to unmount automatically. This removes the overlapping dropdown instead of trying to correct it with z-index values.

image

image


anoek:
Thanks!

Issue #3545, opened 2026-05-05
SGF Library page doesn't have a header text and the header row is too wide for the page


Johnniedarkoo:
**Describe the bug**
Three bugs:

1. SGF Library page doesn't have a header text, instead it shows a "/". 
2. The header row is also too wide for the page, requiring horizontal scroll to reach the Create Collection button. 
3. The "Verzameling Aanmaken" button in the top right overflows. Possibly in other languages too. 

iOS on iPhone 14 pro

**Screenshots**
![image](https://github.com/user-attachments/assets/92d97f8b-9bc9-468e-a635-9e521788d342)
![image](https://github.com/user-attachments/assets/a698bf33-1c92-4f35-95af-091cec22e7bd)

ogsai:
Fixed by #3654 (merged): the root breadcrumb now shows "SGF Library" instead of a bare "/", and the header row/create-collection button now wrap on narrow screens instead of requiring horizontal scroll. Verified on main (src/views/LibraryPlayer/LibraryPlayer.tsx, LibraryPlayer.css).

PR #3654, opened 2026-08-11
Fix SGF Library header title and mobile overflow (#3545)

cMohammadT:
Show "SGF Library" for the root breadcrumb (previously empty name rendered as "/" only). Allow the header row and create-collection controls to wrap on narrow screens so the button stays reachable without horizontal scroll.


anoek:
Thanks!

PR #3647, opened 2026-08-08
Show game name in preferred settings dropdown (#3592)

prisect:
The preferred settings dropdown in the challenge modal now shows the game name first when one is set (e.g. "My Game | Unranked, 19x19..."). Unnamed settings are unchanged. Named settings sort alphabetically before unnamed ones. Solves #3592.

Why: Saved settings are hard to distinguish when they all read "Unranked, 19x19, Japanese rules..." as shown by screenshot of #3592.

Also added unit and component tests (428 total pass). Type-check, lint prettier and build all clean

PR #3655, opened 2026-08-11
Preserve raw game name when saving preferred settings

prisect:
As I was looking back through my previous pr commit (#3647) I noticed the unnamed-game handling wasn't actually reachable. 

`getChallenge()` stamps a localized "Friendly Match" name on every game before saving so settings never had an empty name. Default-named games ended up sorting among the named ones instead of after them, which is what the original issue asked for.

This change keeps the raw typed name when saving so unnamed games stay
unnamed and sort last, also added tests for the save path plus the `Replication` mock they need.

You're welcome!


anoek:
Thanks for following back and getting that fixed!

PR #3656, opened 2026-08-11
Fix tournament list horizontal clipping on mobile (#3544)

cMohammadT:
On narrow viewports the tournament tables use nowrap cells, so content is wider than the screen. PaginatedTable constrained the table itself to 100dvw, which can clip the left side on iOS instead of allowing the wrapper to scroll.

Scope the fix to TournamentList: allow the table to size to its content and keep overflow-x scrolling on the wrapper (including schedule table).


anoek:
Thanks!

Issue #3642, opened 2026-08-04
Hover/redraw flicker in game review links


spookyboygo:
**Describe the bug**
I’m finding it difficult to follow a review because the screen flickers when I hover over the variation links. See video.

**To Reproduce**
Put mouse over a variation link in a review. See attached video.

**Expected behavior**
Text should not move when mouse is over it.

**Screenshots**
Video attached.

https://github.com/user-attachments/assets/940dd431-7220-4abf-bcab-45450df2ae2b

**Desktop (please complete the following information):**

Firefox on a Mac.

prisect:
browser issue?

anoek:
No, it's a space reservation issue, hover and that text jumps causing the mouse to not hover, which jumps back, which causes it to hover, etc..

anoek:
I couldn't reproduce this with the latest ui stuff, closing as fixed but please re-open if it's still an issue in some cases

PR #3657, opened 2026-08-11
Prevent variation link hover from reflowing chat text (#3642)

cMohammadT:
Underline was applied only on hover, which shifted chat line layout and made review variation/position links flicker under the cursor.

Reserve underline space always and only change decoration color on hover.

### DIFF
--- a/src/views/Game/GameChat.css
+++ b/src/views/Game/GameChat.css
@@ -96,6 +96,7 @@
         display: flex;
         align-items: flex-end;
         justify-content: center;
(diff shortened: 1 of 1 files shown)


anoek (changes requested):
Hmm, did you verify that you were able to reproduce the error before, and then this patched fixed the error?

I'm skeptical because the underline text-decoration doesn't change the dimensions of the rendered font, so wouldn't cause a reflow.

Issue #3011, opened 2025-03-22
Nav bar scrolls out of view on Chat page


shinuito:
**Describe the bug**
For some reason on the chat page, on mobile at least, the nav bar page seems to be scrollable and frequently scrolls out of view.

**To Reproduce**

You can scroll this element out of view and sometimes it happens by itself.

<img width="414" alt="Image" src="https://github.com/user-attachments/assets/3bc04264-9089-4117-a6d6-c0e0ff6925ac" />

<img width="414" alt="Image" src="https://github.com/user-attachments/assets/7208d4bf-77e0-484c-939b-429e4e76c877" />

**Expected behavior**
It should be the same as every other page and not scroll out of view.

**Smartphone (please complete the following information):**
 - iPhone11
 - safari

shinuito:
@gsouf might be a good person to ask. I see a lot of style changes to the nav bar for accessibility.

Maybe we can hide the bullet points as well

<img width="225" alt="Image" src="https://github.com/user-attachments/assets/51754136-08a3-4290-920f-ad93cc3f70e1" />

shinuito:
Yes I think that's a better way of phrasing it for the nav bar :)

Yes it's a funny one the bullet points, I'm not seeing it on desktop safari but only on mobile safari for some reason.

Thanks for having a look!

PR #3658, opened 2026-08-11
Keep site navbar fixed while Chat page is open (#3011)

cMohammadT:
ChatView is position:fixed under the navbar, but mobile browsers can still scroll the document and push the navbar off-screen. Lock body overflow while the chat page is mounted and disable overscroll chaining.


anoek:
This seems to make the perceived problem worse. Please test your patches before submitting PRs.

Issue #3461, opened 2026-03-27
On mobile view, active games (numbers) overlapping with rating charts details


zeemai:
**Describe the bug**
On mobile view, active games (numbers) overlaps with rating charts details

**To Reproduce**
1. Go to the User page
2. View on phones
3. Scroll down to Rating Charts
4. See the overlapping

**Expected behavior**
Not overlapping

**Screenshots**
<img width="308" height="330" alt="Image" src="https://github.com/user-attachments/assets/dae8f47c-264b-422a-a339-0bda451f5ab0" />

**Smartphone (please complete the following information):**
 - Device: [iPhone16 Pro]
 - OS: [iOS18.5]
 - Browser [Google App]
 - Version [413.00]

**Additional context**
It started few weeks ago out of nowhere.

viczommers:
@GreenAsJade @anoek seems like a simple css ts model fix - happy to write a pr if worthwile

GreenAsJade:
Yeah please go for it, just of course take care with the whole responsive layout effects :)

PR #3659, opened 2026-08-11
Prevent ratings chart controls overlapping details on mobile (#3461)

cMohammadT:
On narrow screens the absolute graph-type toggle can cover chart axis labels and game-count details. Flow the toggle normally under 768px and allow the chart container to scroll horizontally if needed.

### DIFF
--- a/src/views/User/User.css
+++ b/src/views/User/User.css
@@ -484,11 +484,22 @@
         .ratings-chart {
             display: flex;
             flex-direction: column;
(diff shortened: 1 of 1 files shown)


anoek (changes requested):
New version seems to introduce a scroll bar, I'm also not seeing what this is fixing, do you have a screenshot as to what this is fixing and that it is fixed?

Issue #2619, opened 2024-03-11
Game sorting shows a rengo game where it isnt my turn at the top


kobakitty:
I have this rengo game at the top of my list, altought its not my turn to play https://online-go.com/game/51256536
It looks like there was an undo just recently, i think that might have messed something up with the sorting?
Pic: 
![image](https://github.com/online-go/online-go.com/assets/29932807/5638b031-4fa7-4758-8395-84f8a336d294)

anoek:
Possibly related to https://github.com/online-go/online-go.com/issues/2619

PR #3660, opened 2026-08-11
Fix/issue 2452 rengo clock sort

### LINKED ISSUE #2452: Rengo clocks are probably sorted strangely
#2443 started sorting by the remaining time on clocks that weren't running. It assumed that `this.props.player.id` was always equal to exactly one of `clock.black_player_id` or `clock.white_player_id`. I was just looking at the code and developer console (to file #2451) and realized that this assumption is false for rengo games, since sometimes neither white nor black is "this" player.

I think what we want is:
- If it's `this.props.player.id`s turn, sort by the current clock (existing code is probably correct)
- Else, sort by whoever's turn it is, by checking `current_player_id` against `clock.black_player_id`

cMohammadT:
Players on a rengo team may not match black_player_id / white_player_id.
Detect side via team membership so clock and opponent sorts use the correct color for the listed player.

### Changes
- Add isPlayerBlack() helper (clock ids → rengo teams → game.black fallback)
- Use it for clock sort, opponent-clock sort, and opponent name sort

Fixes #2452
Related: #2619


anoek:
Hmm, I think 2619 might be the issue to try to fix, I think this patch isn't actually fixing something we want though as I *think* this patch would make it so if it is anyones turn to move on a rengo game it would sort as though it is *your* move, which isn't something we want. Please correct me if I'm wrong though and feel free to re-open.

Issue #3564, opened 2026-05-15
resign/info buttons are too close


scottdraves:
**Describe the bug**
When i move the mouse towards the resign button, instead the game info panel opens.

**To Reproduce**
1. Play a game
2. Move the mouse to the resign button
3. Info panel opens, covering the button

**Expected behavior**
Click to resign

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: latest chrome

here is how the buttons are positioned during the game, when i decide to resign:
<img width="530" height="449" alt="Image" src="https://github.com/user-attachments/assets/c6592369-4b5e-4763-805f-60cc7d8efccf" />

as i move the mouse towards the button, the info panel opens and hides the button:
<img width="473" height="482" alt="Image" src="https://github.com/user-attachments/assets/45ea3f7e-3ed5-4a00-b86c-e4775388bbcf" />

teauxfu:
This is at 

https://github.com/online-go/online-go.com/blob/1c52f403bfd6b28ef1109aea8621836fdcfd9fdc/src/views/Game/PlayButtons.tsx#L366

And those styles are here 

https://github.com/online-go/online-go.com/blob/1c52f403bfd6b28ef1109aea8621836fdcfd9fdc/src/views/Game/PlayControls.css#L83

this could probably be fixed easily with some padding or something, but it seems there's a lot going on in this style block, we'd want to be careful to test the result on small screens

PR #3662, opened 2026-08-12
Increase z-index and spacing for resign/cancel button

cMohammadT:
…ap with game info panel (#3564)

The resign button can be covered by the expanding Dock or info panel when the mouse approaches it. Give the button a higher stacking context and a small margin so it remains reachable.


anoek:
Not the way to fix this, also this will be obviated by #3575 so closing

PR #3665, opened 2026-08-14
Fix open redirect, JavaScript execution and tab-nabbing vulnerabilities

prisect:
A few things stacked together here.

The `?next=` param on `/sign-in` and `/register` was getting handed straight to `window.location.href` after login. No checks at all. Meaning someone could craft a `?next=javascript:....` link and once the victim logs in, it runs arbitrary code into OGS origin, with their brand new session attached. Thats basically full account takeover from one click.

Same param also let off-site URLs through, so people could get redirected clean off the site right after authenticating too.

Fix is `valid_next_url()`: it only lets same-origin relative paths or same-origin absolute URLs through, and it'll reject anything `javascript:`, `data:`, off-site, or protocol-relative. Wrote unit tests for both what it should accept and reject, including the sneaky ones with encoding and whitespace tricks people use to slip past filters

Another related issue: external links built from user input, in `chat_markup.tsx` and also `IncidentReportCard.tsx`, were using `target=_blank` **without** `rel=noopener`. That means the tab you just opened could reach back and control the original OGS tab through `window.opener`, classic reverse tab-nabbing, good for phishing. Added `rel="noopener noreferrer"` on those and updated the chat markup tests so they match now.

All tests pass

### DIFF
--- a/src/lib/url_validation.ts
+++ b/src/lib/url_validation.ts
@@ -21,3 +21,46 @@ export function is_valid_url(url: string): boolean {
     }
     return /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)\/?$/.test(url.toLowerCase());
 }
+
+/**
+ * Validates a `?next=` redirect target coming from the URL query string before
+ * it is handed to `window.location` after sign-in or registration.
+ *
+ * Only same-origin targets are accepted. Anything else - off-site URLs,
+ * `javascript:`, `data:`, and other non-http schemes - is rejected, since
+ * assigning it to `location.href` would either redirect the user off the site
+ * after authenticating, or, in the case of a `javascript:` URL, execute
+ * attacker-controlled code in the OGS origin with the freshly-established
+ * session.
+ *
+ * The value is always run through the URL parser rather than inspected with
+ * string prefix checks. Prefix checks are unsafe here: browsers strip embedded
+ * ASCII tab/newline characters while parsing a URL, so a value like `/%09/evil.com`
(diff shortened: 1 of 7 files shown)


prisect:
@anoek Also quick unrelated note - might be worth going through and closing out some issues that are already fixed and merged. Looks like a bunch of them are just sitting there done just need someone to actually close them out.

prisect:
Looks good, ready for merge.

src/lib/url_validation.ts
```diff
@@ -21,3 +21,46 @@ export function is_valid_url(url: string): boolean {
...
+ * string prefix checks. Prefix checks are unsafe here: browsers strip embedded
+ * ASCII tab/newline characters while parsing a URL, so a value like `/%09/evil.com`
+ * passes a `startsWith("/")` check but normalizes to `//evil.com` (an off-site
+ * redirect) once assigned to `location.href`. The URL parser performs the same
+ * normalization as the browser, so the origin comparison below sees exactly what
+ * the browser would navigate to.
+ *
+ * Returns the trimmed safe target, or null if it is not safe to navigate to.
+ */
+export function valid_next_url(next: string | null): string | null {
```
anoek, line 46: I think we have at least one external site that uses `?next` to go back to their site after logging in, let me check with the author

anoek: They are not using next so all is good

anoek:
Thanks!

src/lib/url_validation.ts
prisect: Very nice, thanks for that.

PR #3666, opened 2026-08-15
Fix open redirect via ?return= and harden remaining tab-nabbing links

prisect:
1. Open redirect via `?return=` on the game page: `is_valid_url()` accepts any host so `/game/123?return=https://evil.com` redirects off-site. Now uses the same-origin-only `valid_next_url()` used by the `?next=` fix.
2. Reverse tabnabbing: 17 `target="_blank"` links across 16 files still had no `rel` (including user-controlled group/site links). All now use `rel="noopener noreferrer"`

All tests pass

### DIFF
--- a/src/components/ACLModal/ACLModal.tsx
+++ b/src/components/ACLModal/ACLModal.tsx
@@ -152,6 +152,7 @@ export class ACLModal extends Modal<Events, ACLModalProperties, any> {
                                         target="_blank"
                                         href={`/group/${obj.group_id}`}
                                         className="group"
+                                        rel="noopener noreferrer"
                                     >
                                         {obj.group_name}
                                     </a>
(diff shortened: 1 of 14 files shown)


prisect:
Good for merge.

anoek:
We actually want to support `?return=` going back to other sites, I think it's the leagues that use that feature.

And my understanding is that `target="_blank"` implies `noopener` for awhile now, is that correct @claude?

prisect:
Let me do it for you.

anoek:
So since we want `return`, and `noopener` being a no-op these days, that leaves `noreferrer`. Most of what's changed in this patch are self links, which would mean it'd just be cutting off our own self reference analytics, and the couple of links elsewhere I don't really see the harm these days now that strict-origin-when-cross-origin is the default referrer-policy.

prisect:
Makes sense, agreed.

PR #3667, opened 2026-08-15
Restrict WebSocket host override to official routes for security

prisect:
`src/lib/sockets.ts` accepts anyhost from `localStorage["ogs.websocket_host"]` as the WebSocket endpoint and `src/main.tsx` sends the signed session JWT to that host on every connect (including refreshed user/jwt tokens)

the override now only accepts the three official gateway routes (`wss://online-go.com`, `wss://wsp.online-go.com`, `wss://wss.online-go.com`). Everything else is ignored with a warning. Dev mode is unaffected

All tests pass.

### DIFF
--- a/src/lib/sockets.ts
+++ b/src/lib/sockets.ts
@@ -26,6 +26,17 @@ const ROUTE_CLOUDFLARE = "wss://online-go.com";
 const ROUTE_GOOGLE_PREMIUM = "wss://wsp.online-go.com";
 const ROUTE_PUBLIC = "wss://wss.online-go.com";
 
+/* The only WebSocket gateway routes the client may connect to. The session
+ * JWT is transmitted to the chosen host on connect (see the socket
+ * `authenticate` call in main.tsx), so an arbitrary host here would silently
+ * exfiltrate the live session token. The `ogs.websocket_host` localStorage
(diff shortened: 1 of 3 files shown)


prisect:
Good for merge.

anoek:
This would break development and beta. Is there an actual risk vector we're addressing here?

prisect:
Short answer: socket.authenticate() sends the session JWT to whatever host ogs.websocket_host points to, and anything that can write localStorage can redirect it to an attacker host and steal the token. So that's the vector.

To be honest though, `config.user_jwt` alreayd lives in the same localStorage, so anyone who can write the override can usually just read the JWT anyway, this is defense-in-depth on top of #3668, not a new critical vuln.

And dev should be fine too. NODE_ENV=development resets the host after this block and `window.websocket_host` is untouched. If you have a workflow that legitimately sets a different host in dev, beta or else you can always add it locally. I just think having an allowlist is the better reflex

Let me know what you think about that.

anoek:
I think I'm not going to merge this because it's one of those things that I have a feeling I'm going to stub my toe on in the future, and since as you say, if an attacker already has access to localStorage then they can read the JWT directly anyways.

prisect:
Agreed, thanks for the review.

PR #3668, opened 2026-08-15
[CRITICAL] Fix/seekgraph dom xss via unescaped socket data

prisect:
DOM XSS in the seek graph via unescaped socket data. 

`src/components/SeekGraph/SeekGraph.tsx` builds the challenge detail line as an HTML string and assigns it to innerHTML line 894. Only the challenge name is escaped; every other field interpolated from the serverbroadcast challenge object (`challenger_color`, komi, width, height, handicap, time control system) **is concatenated raw**. 
A challenge with a malformed value in any of those fields executes as HTML in every visitor's browser, the seek graph is shown on the home page and to anonymous visitors. This is the same sink class as in my recently fixed pr the private-chat stored XSS one, with incomplete escaping.

Added a shared, unit-tested escapeHtml helper (src/lib/html.ts) and applied it to every server-supplied value before it is concatenated into the innerHTML string. Legitimate values render identically.

### DIFF
--- a/src/components/SeekGraph/SeekGraph.tsx
+++ b/src/components/SeekGraph/SeekGraph.tsx
@@ -820,20 +820,20 @@ export class SeekGraph extends TypedEventEmitter<Events> {
                 </React.StrictMode>,
             );
 
-            let details_html =
+            let details =
                 ", " +
                 (C.rengo ? _("Rengo") + ", " : "") +
                 (C.ranked ? _("Ranked") : _("Unranked")) +
                 ", " +
-                C.width +
+                String(C.width) +
                 "x" +
-                C.height +
+                String(C.height) +
                 (C.handicap === 0
                     ? ", " + _("no handicap")
-                    : C.handicap < 0
-                      ? ""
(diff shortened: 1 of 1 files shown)


prisect:
Rebase to remove the previous' PR changes.

prisect:
Good for merge.

anoek:
Hmm, not critical because it's not exploitable as those fields are controlled by the server, but I'm not against hardening.

However, let's not add a new escape function, instead we should probably just switch to using `textContent` instead of innerHtml, and just set the `details` class for the node instead of relying on building that with innerHtml. That should simplify things and keep from having yet another escape function we have to remember to use.

prisect:
Good point, I see what you mean here yeah most of these are unreachable from the user. Lemme just refactor all that real quick

prisect:
Details are now built as a plain string and assigned via textContent and the min/max rank "cause" node is now a DOM node with its class set directly. Removed the escapeHtml helper and the tests.

anoek:
Thanks!

Issue #3603, opened 2026-06-25
suggestion: hide "ranked" option in `ChallengeModal` vs bots


teauxfu:
Now that [ranked games vs bots are disabled](https://online-go.com/whats-new/4), we should probably remove this option when doing a rematch vs a bot

<img width="798" height="508" alt="Image" src="https://github.com/user-attachments/assets/89e01ed5-36e8-4034-b9d8-e17569286248" />

The game creation screen from play > play computer already handles this correctly 

<img width="1068" height="892" alt="Image" src="https://github.com/user-attachments/assets/e7eece39-f2f2-4e44-a6bc-c4d3c5e0fa9b" />

It looks like the place to affect this would be inside the `ChallengeModal`, using some piece of data that lets us detect if the user we're challenging is a bot. If they are, we can hide this checkbox 

https://github.com/online-go/online-go.com/blob/1c52f403bfd6b28ef1109aea8621836fdcfd9fdc/src/components/ChallengeModal/ChallengeModal.tsx#L1318-L1343

PR #3669, opened 2026-08-15
Fix rematch against bots showing ranked checkbox (#3603)

gueejla:
As referenced in #3603 the rematch page incorrectly shows the ranked checkbox for bot matches (which are only unranked now). I believe this PR should fix the issue by passing a new parameter is_bot_game to the goban engine, allowing the rematch logic to render the ranked checkbox properly.

I could be wrong about how the goban engine gets its values however, in this PR I did it by modifying the challenge sent out from the ChallengeModal. Please correct me if the goban engine gets its values differently from how I set them here.

I am also looking into writing a test case for this fix, but I figured getting input on if I passed the data properly should come first. Thank you!

### DIFF
--- a/src/components/ChallengeModal/ChallengeModal.api.tsx
+++ b/src/components/ChallengeModal/ChallengeModal.api.tsx
@@ -77,6 +77,8 @@ export function challengeRematch(
     //config.syncBoardSize();
     //config.syncTimeControl();
 
+    const is_bot_game = conf.is_bot_game ? true : false;
+
     const config: ChallengeModalConfig = {
         conf: {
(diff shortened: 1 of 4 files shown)


gueejla:
I can see what the github actions bot is saying that mode="computer" does more than just the checkbox. I made a more targeted fix for just the ranked selector appearing on bot rematches, not sure if it should be more general than that. This is my first PR to this repo so chances I missed something are high..

gueejla:
Think this is too complicated of a solution, closing this for now and will open a new PR if I find a cleaner way to make the fix

GreenAsJade:
Let us know if it needs backend support.  (I didn't absorb what made it complicated, just offering in case its needed :) )

Issue #3534, opened 2026-04-30
ChatChannelProxy event handlers re-emit args wrapped in an array


GreenAsJade:
## Summary

`ChatChannelProxy._onChat`, `_onChatRemoved`, `_onTopic`, `_onJoin`, and `_onPart` in `src/lib/chat_manager.ts` (lines 841-861) all re-emit channel events with the args collected into an array as the single payload, rather than spreading them. Listeners receive an array where the documented `Events` interface payload would be a single object.

## The bug

```ts
_onChat = (...args: any[]) => {
    this.emit.apply(this, ["chat", args]);   // emits ("chat", args) where args is the array
};
```

`emit.apply(this, ["chat", args])` is equivalent to `this.emit("chat", args)` — it passes \`args\` (the rest-collected array) as a single positional argument to \`emit\`. So when the underlying channel emits \`("chat", obj)\`, the proxy wraps that as \`args = [obj]\` and re-emits \`("chat", [obj])\`. The listener gets the array, not the original \`obj\`.

The same shape is in \`_onChatRemoved\` (line 859), \`_onTopic\` (844), \`_onJoin\` (847), and \`_onPart\` (850). All five have it.

## Why it has gone unnoticed

Every existing consumer of \`ChatChannelProxy\` events ignores the payload and re-derives state from \`proxy.channel.chat_log\` / \`users_by_join\` directly:

- \`src/components/Chat/ChatLog.tsx:428\` — \`function onChatMessage() { syncStateSoon(); }\` — no parameter.
- \`src/views/Kibitz/KibitzSharedStreamPanel.tsx\` — handler is \`refresh((value) => value + 1)\`, no parameter.
- \`src/views/Kibitz/KibitzController.ts\` (current) — payload-agnostic by design after #issue-context-below.

Because nothing reads the payload, the bug is silent.

## Why it surfaced now

The first version of \`KibitzController.onChatMessage\` (commit \`5e99eafe4\`, since reverted in \`d2a74779d\`) read the payload directly to keep chat-event handling O(1). It crashed inside \`mapChatToStreamItem\` reading \`.message.m\` on what it thought was a \`ChatMessage\` but was actually \`[ChatMessage]\`. The user-visible symptom was that pressing Enter in the kibitz room composer did nothing and chat history failed to render. Workaround landed in \`d2a74779d\`: the controller ignores the payload and re-derives from \`chat_log\`, matching every other consumer.

[...]

PR #3673, opened 2026-08-23
fix(chat_manager): stop double-wrapping ChatChannelProxy event payloads

Chaero0707:
Summary
ChatChannelProxy's re-emit handlers (_onChat, _onTopic, _onJoin, _onPart, _onChatRemoved, _onUnreadChanged in src/lib/chat_manager.ts) captured the underlying channel's single-argument event as (...args: any[]) and re-emitted it with this.emit.apply(this, [event, args]).
Because args is already the collected rest-array (e.g. [payload]), emit.apply(this, [event, args]) is equivalent to this.emit(event, args) — it passes the array as the single positional argument to emit, so subscribers receive [payload] instead of payload.
This matches a real, currently-open upstream bug report: https://github.com/online-go/online-go.com/issues/3534.
Fix
Forward args[0] directly for each handler, matching how the channel actually emits (event, payload) and how _onUserMetadataUpdate already handled this correctly in the same class. Also updated a stale comment in KibitzController.ts that documented the old (buggy) behavior as an accepted workaround.

Files changed
src/lib/chat_manager.ts
src/views/Kibitz/KibitzController.ts (comment only, no logic change)

### DIFF
--- a/src/lib/chat_manager.ts
+++ b/src/lib/chat_manager.ts
@@ -823,17 +823,22 @@ export class ChatChannelProxy extends TypedEventEmitter<Events> {
         this._destroy();
     }
 
+    // Re-emit the underlying channel's single-payload events as-is. The
+    // channel always emits (event, payload) - a single argument - so we
+    // must forward that same single argument, not wrap the captured
+    // (...args) rest-array in another array (which used to produce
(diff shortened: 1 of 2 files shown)


anoek:
Thanks!

Chaero0707:
My ogs nickname is chaerogo

PR #3678, opened 2026-08-28
refactor: reduce the duplicate function isStandardBoardSize

ClemensSahs:
I fund this function duplication. And the demoboard has already a dep on ChallengeModal so we can use it, right?


anoek:
Thanks for the cleanup!

PR #3697, opened 2026-09-04
Add a kibitz room card to the chat page's user column

GreenAsJade:
## What

Surfaces the new channel-scoped kibitz room (online-go/ogs#2464) on the chat page: a "Kibitz" card pinned at the bottom of the user column, linking to the room scoped to the current channel (`global-english` → `english-chat-live`).

- New self-contained `ChatKibitzRoom` component (+ css, + tests) in `components/Chat/`. It fetches `kibitz/directory` on mount and every 60s, finds the room by the payload's `preset.preset_key`, and renders:
  - the room title;
  - the current matchup on one line — long player names truncate individually with ellipsis (7ch floor) while short names, ranks, and "vs" always render whole, full text in the tooltip — or the room description (two-line clamp) while the room is finding a game;
  - the viewer count, clustered with its icon.
- Renders nothing when the channel has no scoped room, kibitz is disabled (same `kibitz/nav-config` `show_in_nav` gate as the NavBar), or the directory can't be fetched.
- `ChatUsersList` becomes header / scrolling entries / pinned footer, so the card stays visible regardless of chatter count.
- The card resets the global `button` chrome (`height: 2em`, shadow, radius) that would otherwise crush it to a one-line pill.

## Testing

- 6 new jest tests (unmapped channel, nav-disabled, matchup render, description fallback, no matching room, click navigation); Chat + Kibitz suites 287 tests green.
- `yarn type-check`, `yarn lint`, prettier, `yarn build` all clean.
- Kibitz e2e family (3 specs) passed against the local stack running both branches.
- Visually verified on the live dev stack in light and dark themes, in all three content states (real matchup, long names via a deliberately long username on the selected game, no-game description fallback).

Manual mobile + desktop browser testing still to be done before merge (note: on mobile the card is only visible when the users panel is toggled open).

 
Screenshot 2026-09-04 at 6 39 19 pm

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QXek6AzjoNpoo9ytnDrePS


anoek:
We should be using the push system instead of polling no?

GreenAsJade:
(Haven't merged yet because even though it is behind the feature toggle, it seemed prudent not to churn UI at the moment)

anoek:
@GreenAsJade can we switch to using the push system instead of constantly polling every 60 seconds for everyone connected to chat?

ogsai:
Push is in place (28f8eaed3) and the payloads match the backend. Since this ships together with online-go/ogs#2469, two follow-ups in `ChatKibitzRoom.tsx`:

### 1. Drop the per-room channel subscription

With ogs#2469 merged, every board change (the preset selector tick and the change-board endpoint) is also pushed as `room-updated` on `kibitz-rooms`, using the same `KibitzRoomDetailSerializer` payload as `board-changed`. The card doesn't need to subscribe to `room.channel` or listen for `board-changed` any more. Dropping it saves one subscription per connected chat user and stops each board change arriving twice.

- Remove `push_manager.subscribe(room_channel)` / `unsubscribe(room_channel)` and the `board-changed` handler.
- Once they're gone, `room_id` / `room_channel` are only used for matching, so the effect no longer has to re-run when the room changes (see #2).

### 2. Subscribe to the directory up front, and handle `room-created` / `rooms-refresh`

The card only subscribes after the initial fetch finds a matching room. So if the preset room is missing when the card loads, or gets removed and recreated, the card stays hidden until the socket reconnects. Preset rooms are seeded by migrations, so this is rare in practice, but the fix is cheap and makes the component simpler: merge the fetch effect and the push effect into one effect keyed only on `[preset_key, enabled]`, and match pushes by preset key instead of by the room we already know about. Every directory push that carries a room uses `KibitzRoomDetailSerializer`, which always includes `preset`, so matching on `preset.preset_key` is reliable.

```tsx
useEffect(() => {
    if (!preset_key || !enabled) {
        return;
    }
    let cancelled = false;

    const refresh = () => {
        get("kibitz/directory")
            .then((rooms: DirectoryRoom[]) => {
                if (!cancelled) {
                    setRoom((rooms ?? []).find((r) => r.preset?.preset_key === preset_key) ?? null);
                }
            })
            .catch(() => {
                if (!cancelled) {
                    setRoom(null);
                }
            });
    };

    // room-created and room-updated both carry the full detail payload;
    // either one (re)adopts our preset room.
    const onRoomPayload = (payload: DirectoryRoom) => {
        if (payload?.preset?.preset_key !== preset_key) {
            return;
        }
        setRoom((prev) => (prev?.id === payload.id ? { ...prev, ...payload } : payload));
    };

    const onViewerCountChanged = (payload: { channel?: string; viewer_count?: number }) => {
        if (typeof payload?.viewer_count !== "number") {
            return;
        }
        const viewer_count = payload.viewer_count;
        setRoom((prev) => (prev && prev.channel === payload.channel ? { ...prev, viewer_count } : prev));
    };

    const onRoomRemoved = (payload: { id?: string } | undefined) => {
        setRoom((prev) => (prev && prev.id === payload?.id ? null : prev));
    };

    const handlers = [
        push_manager.on("room-created", onRoomPayload),
        push_manager.on("room-updated", onRoomPayload),
        push_manager.on("viewer-count-changed", onViewerCountChanged),
        push_manager.on("room-removed", onRoomRemoved),
        push_manager.on("rooms-refresh", refresh),
    ];
    // Subscribe before the initial fetch so nothing published in between is missed.
    push_manager.subscribe(KIBITZ_DIRECTORY_CHANNEL);
    refresh();
    socket.on("connect", refresh);

    return () => {
        cancelled = true;
        for (const handler of handlers) {
            push_manager.off(handler);
        }
        push_manager.unsubscribe(KIBITZ_DIRECTORY_CHANNEL);
        socket.off("connect", refresh);
    };
}, [preset_key, enabled]);
```

This replaces both the current directory-fetch effect and the push effect, and the `room_id` / `room_channel` derived values go away. Also:
- Update the class docblock: state comes from the `kibitz-rooms` directory channel only.
- The backend never emits `rooms-refresh` today. Handling it is one line and matches `KibitzController`, so it's future-proof.
- `room-removed` only ever sends `{id, channel}`, so the `typeof payload === "string"` branch can go.

### Test changes (`ChatKibitzRoom.test.tsx`)
- "subscribes to the directory and room push channels": assert `subscribe` is called with `"kibitz-rooms"` only, and **not** with `kibitz-preset-english-chat-live`.
- "unsubscribes from both push channels on unmount": now only `kibitz-rooms`.
- "updates the matchup from a `board-changed` push" / "ignores `board-changed` pushes for other rooms": fire `room-updated` instead. For the "other rooms" case, use a payload whose `preset.preset_key` differs.
- New: directory has no matching room → fire `room-created` with the preset payload → card appears.
- New: `room-removed` for our id hides the card, and a later `room-updated`/`room-created` for the preset brings it back.
- New: subscribes to `kibitz-rooms` even when the initial directory fetch finds nothing.

anoek:
@claude please double check the suggestions in the most recent comment by ogsai

Issue #12, opened 2017-02-25
"Your Move" emails are no longer being sent out


OneWeirdDude:
### Summary

Hi, I no longer get any emails saying when it's my turn. I have all the boxes checked.

### Environment

 * Operating system: Windows 10
 * Browser: Google Chrome 
 * Browser Version: 56.0.2924.87 (64-bit)

### Steps To Reproduce:

I make moves and don't get emails when it's my turn until time starts running out.

cliffordwolf:
Same here. (I'm using chromium on 64 bit ubuntu, but obviously this is some kind of server issue.)

In case this is a bug that somehow effects only some users: I'm [cliffordvienna](https://online-go.com/user/view/337285) on OGS.

o0dano0o:
Same issue here.

Sniveler:
Just copying cliffordvienna's post format if it can help.
Same here, Mozilla on64bit Ubuntu.
Sniveler on OGS.

anoek:
Implemented on the backend, will be live in today's patch

o0dano0o:
THANK YOU!

nickeubank:
@anoek Just found this, but still having this issue. Was this resolved for most people?

anoek:
It's generally working, any remaining issues are probably related to spam filtering and that sort of stuff, as far as I know

nickeubank:
OK -- I'm pretty tech savvy, and I'm pretty darn sure it's not a spam issue (I've checked my filters repeatedly. Anything else I can try? Sorry, I know this kind of bug is a pain, but not sure what else to do... my account is JSeaTurtle if that helps.

mlopezviedma:
Are you sure you are not logged in on another device?

nickeubank:
@mlopezviedma Hmmm, interesting. I have logged in from ios devices. It's certainly not up and running anywhere, but I suppose maybe cookie status got confused? There a way to log off all devices?

PR #3704, opened 2026-09-04
Change answer for 4.24.12 from "Black" -> "seki"

bcspragu:
The puzzle shown under Beginner Level 1 > Seki > #12 shows seki, but only accepts black as an answer.

image

I'm a Go beginner, so there's always the chance I'm just wrong here, but I'm pretty confident that's seki.

Note: I didn't test this out locally (i.e. to make sure the casing of `"seki"` is right), but that matches other answers in this file.


anoek:
It's black because white cannot play to capture those black triangle stones, but black can play to capture the stones to the right of the black triangle stones, capture those, then capture the white triangle stones

bcspragu:
Ah thanks for the clarification, and sorry for the noise!

anoek:
No problem and no worries!

PR #3732, opened 2026-09-12
Reduce e2e runtime and stabilize parallel tests

ogsai:
The e2e suite used one worker, unconditional waits, and shared report state. This change runs 72 automated journeys in parallel and moves suitable coverage to normal tests. The standard `yarn test:e2e` command builds automatically and selects workers from RAM capacity. The full selection includes `@Slow`, uses zero retries, and fails on backend errors.

**Backend companion:** https://github.com/online-go/ogs/pull/2478. Land the companion and update `init_e2e` fixtures with this change.

There are **14 removed or merged automated journeys: 86 becomes 72**. The previously reported count of 92 also included four utility scripts and two visual checks. Those six remain in the repository but are excluded from the automated selection. Smoke tests keep their separate Docker command; `@Slow` remains included in the full run.

## Removed or merged tests

Paths below are under `e2e-tests/` unless stated otherwise.

| Removed browser test | Replacement and reason |
| --- | --- |
| `challenge-modal/ch-basic-creation.ts` | `ChallengeModal.test.tsx` checks the submitted request payload. The browser test intercepted the request, so a real browser/backend added little coverage. |
| `challenge-modal/ch-preferred-rank.ts` | Component tests check rank-specific preferred settings and request values. |
| `challenge-modal/ch-handicap-prefs.ts` | Component tests check handicap preferences and persistence. |
| `challenge-modal/ch-private-invite.ts` | Component tests check private/invite flags. The real invite-only Home journey remains. |
| `challenge-modal/ch-disable-analysis.ts` | Component tests check the analysis flag; the retained Home journey checks its displayed status. |
| `challenge-modal/ch-rengo.ts` | Component tests check rengo options and validation. |
| `cm/cm-ack-ack.ts` | Fold acknowledgement delivery and dismissal into `cm-vote-warn-not-ai.ts`; component tests cover immediate acknowledgement behavior. Avoids a second three-voter setup for the same outcome. |
| `cm/cm-escaping-one-at-a-time.ts` | `report_manager.test.ts` drives real report updates to check queue ordering, release after resolution, independent accused users, and own-report counts. Avoids creating multiple games to test client queue logic. |
| `moderation/mod-auto-warn-first-turn-disconnector.ts`, `mod-auto-warn-first-turn-escaper-black.ts`, `mod-dont-auto-warn-first-turn-blitz.ts` | Companion backend `first_turn_escaper_test.py` covers both colours, timeout/disconnect reasons, move counts, speed, and bot/missing-player exclusions. The white first-turn browser test still checks real timer-based warning delivery. |
| `moderation/warning-dialog-during-games.ts` | `AccountWarning.test.tsx` checks live/blitz suppression, display after a game, and correspondence display with controlled state and timers. |
| `profile/profile-sticky-rank-dist-graph-toggle.ts` | Merge into `profile-toggle-rank-dist.ts`, which now checks both toggle states across reloads. |
| `moderation/ai-detection-player-filter.ts` | Remove an obsolete result-link expectation that could pass without testing anything when results were empty. `player-check-ai-button.ts` retains current filter navigation and now requires the exact player ID and populated name. |

## Altered tests and fixtures

| Area | Change and reason |
| --- | --- |
| Multi-user setup | Create/login fixture accounts through the API, with unique device/IP data and initial rank/configuration. Registration/login forms retain smoke coverage; feature actions still use the UI. Removes repeated setup journeys and shared auth-state files. |
| CM and AI votes | Wait for the exact vote response to finish before navigation/context closure. Remove temporary disabled-button checks and fixed acknowledgement delays. Keep all report, warning, suspension and acknowledgement outcomes; fail HTTP errors without retrying writes. |
| Report counts and cleanup | Count the reporter's own reports and select full report IDs. Malicious-report filers are worker-specific; capture IDs from submission responses and cancel only reports created by that test. Both cancellation journeys require their unresolved row and a completed cancellation response; remove the three-second silent cleanup skip. Removes global queue/count/ID assumptions and cleanup that could affect another worker. |
| Escape-rate, last-warning, and sandbagging scenarios | Wait for persisted game completion instead of a 30-second pause per game; acknowledge queued warnings through the UI and wait for each response. Give non-clock scenarios longer game clocks so load cannot cause unintended timeouts. Predictive-boundary cleanup targets its own report directly. |
| AI vote, appeal, suspension, and annul fixtures | Use unranked games where ranking is not under test, avoiding external AI analysis and rating tasks. Browser-ID suspension still runs through a real game; backend tests add policy and persistence coverage. |
| Suspension ModLog | Capture the reporter's baseline before filing, then verify its count returns after cancellation. Replaces a baseline taken from the moderator after report creation. Finish an unranked fixture game after eight moves by resignation instead of duplicating empty-board scoring. Complete suspension, appeal submission, restoration and AI-report writes before the next step; wait for full appeal navigation. Retain SUSPENSION ModLog visibility assertions. |
| Basic scoring | Scope report counts to the reporter. Use an unranked game with a five-minute clock; remove the fixed startup wait and duplicate turn assertion. Wait for both clients to finish auto-scoring, observe the first acceptance on the opponent's page, and wait for actual game-log entries. |
| Board helpers and game clocks | Acceptance waits for both players to reach the returned game ID and load their boards; all 46 callers pass the challenger page. Every move reaches both displays before the next action. Resignation requires the Resign label so the helper cannot accidentally cancel a game still in its opening window. Non-clock fixtures use five-minute main time and five thirty-second periods, including early-report tests that inherited two-second clocks. Check final clock fields before submission. The real first-turn timeout remains. Board clicks wait for actionability before measuring their relative coordinates; centre clicks use the locator. A controlled resize made the old helper click D6 instead of H2. Three normal regressions preserve the intended intersection after resize. AI-review readiness is opt-in. |
| Two-move undo, both requester colours | Wait for the played moves on both boards before requesting. Poll each client’s request state and visible markers; after acceptance, require the expected display/engine move number and cleared markers on both boards. The old “Your move” wait could pass before the undo because two moves preserve the turn. Share the checks and extend the white-requester case through acceptance. Use the typed game-board reference. |
| Ladder vacation limits | Use six seeded players per worker and a fresh group/ladder. Load vacation controls before inspecting state; normalize P1/P5 before setup and verify cleanup, even after an interrupted run. Cleanup errors fail an otherwise green test. The reported failure reused accounts left on vacation by a premature negative visibility check. Wait for group/ladder joins and challenges to finish; verify deletion by API. The earlier cached-row product fix retains its three component regressions. |
| Kibitz room creation/editing | Use current More actions menus and picker controls; wait for the selected game's players before checking game information. Preserve owner/non-owner permission checks. |
| Mobile puzzle editor | Wait for setup and move-placement modes before placing stones. Measure and hit-test the board in one browser call, polling through layout changes while still rejecting a persistent overlay. |
| Tournament joins and start | All three tournament tests share completed join/membership checks and server-confirmed chat presence. Keep the director and entrants connected; live starts remove absent players. Require all player names/counts before starting, then wait for the start response and live results. Remove the director reload and fixed picker delay; retain McMahon, round-robin notifications and disable-vacation outcomes. |
| First-turn warning | Retain an explicit 90-second wait for the real game timer and warning delivery despite shorter default action timeouts. |
| `FragAIReview.test.tsx` | Fix the test user during each case and restore mocks afterward so background config updates cannot change moderator permissions mid-assertion. |
| Finished-game moderation fixtures | Seven journeys finish their existing moves by resignation: CM suspension, AI suspend/annul, AI warn/annul, AI first-warning/annul, AI ticket cancellation, appeal-template selection, and rejecting escaping reports during a live game. They test moderation, so they do not need repeated auto-scoring. Escaping-report fixtures have the other player resign, preserving the rule that a resigner cannot be reported for escaping. All moderation assertions remain. The five AI/appeal fixtures also use `TestInfo` and pass the handicap in its correct helper argument. |
| Shared scoring | Predictive escape-rate, informal warning/annul, and smoke scoring now use the synchronized scoring helper; existing informal-warning and escape-rate-history callers retain it. Smoke clocks use five-minute main time and five thirty-second periods. Short escape-rate games retain their moves and scored outcomes; predictive fixtures wait for the saved result directly instead of loading profile history. Both informal-warning tests keep the normal three-minute budget. |
| SGF permissions | Wait for loaded game state and use retrying class/button assertions for anonymous users, players, and spectators. Prevents reading the initial disabled state after reload; keeps all permission cases. |
| Generated usernames | Use a cryptographically random ten-digit suffix instead of random letters, avoiding words/acronyms rejected by registration. Validate the sixteen-character role limit. Five normal cases cover numeric boundaries and username length. |
| Warning acknowledgement and sandbagging warning | Remove three immediate disabled-OK snapshots after checking the box; the real countdown can expire before a slow worker observes them. `AccountWarning.test.tsx` owns exact timer/checkbox boundaries. Keep warning delivery, play restrictions, eventual enabled state, UI acknowledgement and dismissal. |
| Saved game results | Scoring, resignation, and cancellation helpers poll the read-only game API for saved `ended` state after the live outcome. The game server broadcasts completion before the background worker saves results and simultaneous-game metadata. A controlled worker pause reproduces the missing simul indicator; the same test passes after this wait. No action is retried. |
| Development image middleware | Three asset checks use Playwright’s HTTP request fixture and preserve exact fresh-byte hashing, missing-file 404/content type, and submodule fallback. Only the fresh-byte case creates a file and removes it in `finally`. Removes unnecessary application startup and four parallel `beforeAll` probe writes; cached-config coverage still uses a browser. |
| Development CDN configuration | Include navigation in the initial config-response budget, require a completed successful response, and wait for a seeded marker to disappear when the application applies that response. The old immediate snapshot could run before refresh and could time out while development modules were still loading. |
| Browser resources and duplicate setup | Close finished-game fixture players in four sandbagging journeys and the CM suspension journey; close P2–P4 after ladder joins. Create McMahon entrants only when ready to join, retaining their live connections through tournament start. Reuse the first suspension voter’s existing moderator page after escalation. Twenty initial CM setup sites across twelve journeys open their report directly and verify its ID/content after login and warning cleanup. Two reporters open directly at their game; the suspension subject stays on its connected game page. Removes redundant Home loads while retaining every actor, vote and outcome. |
| Shared UI writes | Add `actAndWaitForResponse`: arm the listener before one action, match method/path, require success and a completed response stream. Seven normal tests cover ordering, delayed headers/body, endpoint matching, HTTP errors, stream errors and timeouts. Direct/invite creation, invite deletion, warning acknowledgements and friend request/decline use it; card/message assertions still check the UI. Custom-game setup first waits for either the collapsed or expanded controls before choosing its action. |
| Account, suspension and appeal flows | Wait for suspension/restoration responses instead of modal closure or sleeps. Account-name tests use random names and wait for the initial fetch, successful write and config refresh; normal saves require a real reload, suspended saves reload settings explicitly to prove the update was ignored. Suspension fixtures disconnect the subject page before moderation to avoid racing the reload signal; CM suspension has no reload signal, so that test loads the persisted ban after the completed vote instead of sleeping three seconds; all seven appeal-message writes across five journeys wait for their completed POST, and appeal links wait for full navigation. The ModLog trace showed the moderator fetching the list before the appeal write completed; an immediately disabled Submit button only proved that the input cleared. System PM checks delivery on the connected recipient before navigation. |
| Browser-ID suspension | Replace the global seeded browser ID and cleanup login with a new account and historical/current ID pair per execution. Move the old account to another current ID, then suspend it. Keep UI registration with the historical ID, game-triggered suspension, and the Recently Blocked matched-account dropdown. Finish by opponent resignation after eight moves, with the subject page closed so its automatic suspension reload cannot race a scoring assertion. Failed/concurrent runs can no longer block registration. |
| Report submission, access and negative outcomes | Remove retries of whole report submissions; only opening the popover may retry. Claim/close waits for each response. Post-escalation counts require populated options. Unauthorized assessment/malicious-report views require HTTP 403, queue absence waits for initial synchronization, seeded CM login drains persisted warning/acknowledgement queues, and no-warning outcomes inspect the persisted queue after voting. Closing the malicious-report dialog verifies that no submission was sent. Report-count diagnostics no longer run after a failed journey and mask its original error; a normal regression covers error preservation. |

## Runner and product fixes

- The backend companion makes development heap-allocation tracking opt-in through `NODE_DEBUG_FLAGS`. A debugger caught the repeated scoring failure inside V8’s allocation tracker on Node 22.13.1; the old launcher enabled it unconditionally. Preserve snapshots/GC and explicit profiling, document the defaults, and restart existing Node service containers to apply them. Game rules and the Node version remain unchanged.
- Fix a shared `useData` race found by the suspended-account deletion journey under CPU load: the backend returns the banned ID, but a mounting reader writes an older snapshot back and erases it. Use `useSyncExternalStore`; only the explicit setter writes to the store. Functional setters read current store state; defaults are read fallbacks until explicitly set. Eight normal tests cover mount/commit ordering, concurrent consumers, setter composition, removal, defaults, key changes and StrictMode. Six fail before the fix, all eight pass afterward. Document the contract in `docs/data-hooks.md`.

- Make `test:e2e` and `test:e2e:quick` check the moderator password, build once, and run built assets with RAM-based worker selection. `test:e2e:built` reuses an existing build. Explicit `test:e2e:dev`, UI, and debug commands use the development frontend with one worker by default; development-server checks retain their own project. Listing, help, and CI smoke selection bypass the build and password requirement.
- Share one worker selector between the built runner and self-hosted CI: **2** up to 16 GiB, **6** above 16 and below 32 GiB, **8** from 32 to below 48 GiB, and **16** at 48 GiB or more. Round host capacity up to whole GiB for OS overhead and respect lower container limits without rounding. Use total capacity, not momentary free RAM. Preserve `E2E_WORKERS` and Playwright `--workers` overrides. Add 31 normal cases for tier boundaries, OS overhead, container limits, invalid/valid overrides, and built-runner integration.
- Add `make e2e` to run the same full command in `ogs_ui_1` with the local seed password. An exported `E2E_MODERATOR_PASSWORD` overrides that default, so a host shell can run the validated Docker path with one command. `E2E_WORKERS=4 make e2e` passes a worker override into the container.
- Report WebSocket proxy `ECONNRESET` events as concise warnings in both Vite development and built-preview servers. Emit the first warning immediately, summarize additional resets after five seconds, count duplicate proxy/socket callbacks once, and flush pending counts on shutdown. Preserve HTTP failures, other error codes, configured log levels, and Playwright exit status. Add 15 normal logger/config regression cases and extend launcher exit tests to verify flushing. Use an explicit TypeScript config import with `allowImportingTsExtensions` to avoid Vite's loader warning.
- Add opt-in `E2E_API_DELAY_MS` for real browser API latency checks; default zero. Document that Playwright routing also disables HTTP cache, so this is a stress mode rather than a runtime benchmark. Document synchronization contracts and recovery from dirty fixtures.
- Retain a JSON report at `test-results/e2e-results.json` with actual workers, assertions, traces, durations, Node/CPU/RAM/affinity metadata and injected delay. Do not dump credentials or environment variables. Error-boundary teardown checks run after successful journeys; failures retain their original assertion and still close owned contexts.
- Disable retries, bound actions/navigation, remove the five-second per-test delay and process-wide setup/teardown cleanup, and close only owned contexts in `finally`. Propagate launcher failures. New launcher tests cover arguments, proxy selection, shutdown, exit codes, automatic builds, stopping after build failure, early missing-password errors, and listing/help/smoke behavior.
- Fix account warning countdowns: refreshed copies of the same message do not restart the delay; a new message resets the checkbox and timer; the initial render respects the reading delay. Add component regression tests.
- Fix Kibitz default-room navigation races and preserve shared player records when cloning snapshots while still removing cycles. Add snapshot regression tests.
- Address the review findings: warning cleanup checks the queue after the tenth dismissal and only fails when another message remains. Kibitz room selection, creation, and deletion share a navigation guard that clears on any committed URL, including superseded navigation. Add 13 normal regression cases for both fixes; browser selection remains 72 journeys.
- Expose the rendered Home challenge ID so acceptance targets the current challenge and waits for its response instead of submitting a stale cached ID.
- Forward `/locale/` to the development server and remove current React-refresh and checker scripts from the built template. This fixes translation 404s and stale development imports; launcher tests cover both template forms and preserve application configuration.
- Document coverage, fixture ownership, measured results, and exact Docker commands in `docs/e2e-testing.md`; update the test-writing guide.

## Validation and limits

Author check: review account settings and preference controls in mobile and desktop browsers for the earlier shared data-hook fix. This portability follow-up changes tests, runner diagnostics and documentation only.

- **Three final full batches, 72/72 each**, zero retries/skips and no database reset: ordinary `make e2e` with automatic eight workers **293.4s including the fresh build**; sixteen workers with frontend/browsers restricted to four CPU cores **506.1s**; unrestricted sixteen-worker repeat **259.9s**. The host has 32 logical CPUs and 32 GiB installed RAM. Services/dependencies were ready. The fresh build reports **50.87s**, its browser batch **240.6s**. The ordinary full command meets five minutes here; the CPU-stress run does not.
- **638 normal tests pass**. TypeScript, lint, modified-file formatting and production build pass. Graphify is unavailable locally.
- **Controlled before/after evidence:** the original click helper lands on D6 when H2 is requested during a board resize; the corrected helper selects H2. Delayed background result processing reproduces the missing simultaneous-game indicator; waiting for saved `ended` state passes with processing held twenty seconds after live completion. Three normal regressions cover resize ordering and centre/scoped clicks.
- **CPU baseline: 70 pass / two fail in 522.1s**, after the earlier backend profiling fix. CM suspension and sandbagging warning reach their unchanged 180-second deadlines during continued progress. Both pass in the final complete four-core batch after redundant fixture work is removed. Backend processes remain alive. The CPU restriction affects the runner, built preview and browsers; backend services and the existing development server retain host CPU access; RAM alone does not establish CPU throughput.
- **5/5 focused checks pass with 500 ms API delay**: CM suspension, active-game escaping-report rejection, scoring, black-requester undo and mobile puzzle editing. Separate CI gameplay/scoring smoke also passes. Earlier follow-up checks retain 4/4 dirty-vacation recovery executions, 10/10 delayed appeals and 8/8 scoring/stalled-game checks after disabling unconditional allocation profiling.
- The previous native scoring failure enters V8 allocation tracking on Node 22.13.1. Companion `NODE_DEBUG_FLAGS` makes it opt-in; it does not patch V8. **Restart existing Node service containers after pulling the backend launcher change.** These final batches use the new defaults and stable service processes.
- Companion validation remains **337 Python passes, five existing skips**, default/override checks for nine launch targets and running-argument checks for eleven Node containers. Backend `make lint` passes Node checks then stops at existing baduk.com Node-type errors. No backend source changes in this follow-up.
- The remote machines supplied test names but no assertions, CPU counts or confirmed branch/service state. The reproduced causes are proven locally; they do not establish the cause of every remote entry. The JSON report now retains actual workers, machine metadata, assertions and trace paths for that comparison.

With local services, updated seeds, dependencies and Chromium ready, run `make e2e`. It builds and runs the complete suite in `ogs_ui_1` with the local seed password. Export `E2E_MODERATOR_PASSWORD` to override it. Reuse a current build with:

```sh
docker exec -e E2E_MODERATOR_PASSWORD ogs_ui_1 yarn test:e2e:built
```

The password must be exported for Docker's `-e` form. Fixture contracts, exact stress commands and measurement limits are in `docs/e2e-testing.md`.


GreenAsJade:
Superceded by #3738  (new PR for clarity about what went in)

PR #3735, opened 2026-09-13
fix(ui): Enforce client display preferences over URL parameter overrides

Chaero0707:
Summary
A malicious challenge/match creator could craft a /game/:id?... link carrying display-affecting query parameters — e.g. theme=book, zen=true, layout=compact, coords=kanji — that, if consumed by the client, would force a custom board theme, forced zen/layout mode, or hide critical UI (player names, match clock, chat, bottom settings toolbar) onto the opponent who clicks the link. This overrides the viewer's own client-side preferences and creates an unfair, non-consensual match experience.

This PR closes that vector by construction, before any such wiring can land, by introducing an explicit allowlist of the only query parameters the live Game room is permitted to read from the URL, and it hardens the pieces of the room UI most relevant to the report:

Sanitize incoming URL parameters — src/lib/url_display_params.ts defines ALLOWED_GAME_URL_PARAMS (currently just the already-validated return redirect target) and sanitizeGameSearchParams(), which strips everything else. Game.tsx and PlayControls.tsx now route useSearchParams() output through this allowlist instead of reading it raw, so theme=, zen=, layout=, coords=, etc. are silently dropped rather than ever reaching component state.
Client preferences always win — audited zen_mode end-to-end (GobanController.setZenMode → useZenMode hook → Game.tsx/PlayControls.tsx/PlayerCards.tsx) and confirmed it is only ever driven by the viewer's own start-in-zen-mode preference or their own manual toggle — never by game/challenge server data or the URL. Board theme is likewise only ever read from the viewer's own preferences.get("goban-theme-*") (see goban_theme_json.ts), never from remote/room config. The new allowlist guarantees this stays true for any URL-based path.
UI protection — confirmed the bottom tab bar / toolbar (TabBar.tsx, GobanView.tsx) has no code path that disables or hides it based on remote/URL-sourced state (only local toggleVisibility/activeTakeover state); the sanitizer also removes the only place a future hide_toolbar/hide_chat-style param could have been read from.
Changes
Added src/lib/url_display_params.ts — ALLOWED_GAME_URL_PARAMS allowlist + sanitizeGameSearchParams() helper, documented with the exact attack scenario it prevents.
Added src/lib/url_display_params.test.ts — unit tests asserting theme=book, zen=true, and combined layout=/coords=/hide_chat=/hide_toolbar= override attempts are stripped, while the legitimate return= param is preserved.
Modified src/views/Game/Game.tsx — searchParams is now the sanitized set, not the raw router value.
Modified src/views/Game/PlayControls.tsx — same sanitization applied where it independently reads useSearchParams().
Changelog
fix(security): live game room URLs can no longer inject display/theme/layout overrides (theme=, zen=, layout=, coords=, ...) — only a documented allowlist (return) is read from the URL.
chore: add regression tests locking the allowlist behavior in place.
Testing
Added/verified unit tests in src/lib/url_display_params.test.ts covering the exact vectors described in the report (theme=book, zen=true, combined layout/coords/hide-chat/hide-toolbar attempts) and confirming return= still passes through.
Manually traced zen_mode and goban-theme data flow (GobanController, preferences.ts, goban_theme_json.ts, TabBar.tsx) to confirm no other code path currently derives these from remote/URL data.
Fixes the reported vulnerability against https://online-go.com/game/86420438-style crafted links.


anoek:
Don't open AI slop PRs.

Issue #3538, opened 2026-05-01
Two-move undo doesn't notify the targetted player and infobox gives conflicting information


Johnniedarkoo:
**Describe the bug**
When requesting a two-move undo, the player to grant the undo request doesn't get a notification, and the information on screen is conflicting with the actual situation.  

This is not about one-move undo, which I also checked and works correctly. 

**To Reproduce**
We take the case of: black plays move 5, white plays move 6, then black presses "request undo". 

This is recreated here: https://beta.online-go.com/game/20472

What is expected is that:
1. white gets a notification (a counter increase)
2. white gets two buttons to grant or deny the undo request
3. white gets on-screen text explaining the situation along the lines of "the opponent requested an undo, it is up to you to grant or deny this"
4. black gets a button to cancel the undo request
5. black gets on-screen text explaining the situation

What happens is:
1. white doesn't get a notification **[incorrect]**
2. white gets two buttons to grant or deny the undo request **[correct]**
3. white gets on-screen text explaining the situation:
* A blue pop-up box: which says **Your opponent** may now decide to accept the undo request or not. Please note they are not obliged to do so. [incorrect: text should be **you** may now decide....]. Screenshot below. 
* It also says, for both players: "black to move". This is technically correct but mostly misleading because both players have possible actions, and neither is a "move". Either player first has to resolve the Undo request with cancel/grant/deny. 
4. black gets a button to cancel the undo request **[correct]**
5. black gets on-screen text explaining the situation, but it says "Your turn, [your name] wants to undo a move". Misleading, text should be something like:  "Your turn, cancel your undo request or wait for opponent to grant or deny it" **[incorrect]**

**Screenshots**
<img width="2877" height="1547" alt="Image" src="https://github.com/user-attachments/assets/74a0aac7-4ac3-4cef-ab39-1f5f6b48b5d3" />
_screenshot showing the player who has the buttons to accept or reject the undo request getting a conflicting (erroneous) message that it's the opponents choice to accept or reject the undo request. It was the opponent who created the undo request._

**Additional context**
Possibly related to https://github.com/online-go/online-go.com/pull/3458

PR #3739, opened 2026-09-14
Pick the undo help flow by who can answer the request, not by whose turn it is

Chirag6722:
Partially addresses #3538 (the wrong-player help text; see the last paragraph for what is left).

**What was wrong.** `GameStateHeader` chose between the two first-undo help flows by whose turn it is:

```ts
if (this_users_turn) triggerFlow("undo-request-received-intro");
else                 triggerFlow("undo-requested-intro");
```

For a one-move undo that works, because the player to move is always the one who answers. For a two-move undo the requester *is* the player to move, so both seats get the other seat's text: the requester is told "Your opponent may now decide to accept the undo request or not", and the opponent, who holds the Accept button, is told it is up to the requester. That is the screenshot in the issue.

**Change.** Use `useCanAnswerUndoRequest`, the same hook that decides whether the Accept Undo button is shown, so the text and the buttons cannot disagree. It also carries the fallback for older games that did not record `undo_requested_by`. Net change is one hook swap in `GameStateHeader.tsx`; `usePlayerToMove` is no longer used there.

**Tests.** New `GameStateHeader.test.tsx` renders the header with a `DynamicHelp.Api` double and asserts which flow fires:

- two-move undo, I am the requester and to move: `undo-requested-intro`
- two-move undo, I am the other seat and not to move: `undo-request-received-intro`
- one-move undo, opponent requests on my turn: `undo-request-received-intro` (guards the case that already worked)

The first two fail on `main` and pass with this change; the third passes on both. `yarn type-check`, `yarn lint` on the changed files, `yarn prettier:file`, and the `GameHooks`, `PlayButtons`, `PlayControls` suites pass locally. I have not run this against a live two-player game on beta; the tests drive the engine state directly.

**Not covered here.** The issue also reports that the answering player gets no notification-counter increase for a two-move undo. From what I can see that count comes from the server, so it is out of scope for this repository. The "Your turn" wording the issue proposes for the requester is a copy decision I did not want to make unasked.

AI help was taken for this change.


ogsai:
I reproduced a **P2 regression** at [GameStateHeader.tsx:77–84](https://github.com/online-go/online-go.com/blob/c0499d8cfac951595eb11cbe0ef81ac16e08105b/src/views/Game/GameStateHeader.tsx#L77-L84).

If the header mounts with an opponent’s undo request already pending, `useShowUndoRequested` starts as `true`, but `useCanAnswerUndoRequest` starts as `false`. The header first triggers `undo-requested-intro`. After the hook’s effect updates its state, the header also triggers `undo-request-received-intro`. The help controller keeps both flows active, so the recipient gets contradictory tutorials.

The reproduction uses six moves, black as the logged-in user, and white as the requester. This is a one-move undo. Set `undo_requested_by` and `undo_requested` **before** `renderHeader`:

```tsx
const controller = sixMoveGame({ black: ME, white: OPPONENT });
controller.goban.engine.undo_requested_by = OPPONENT.id;
controller.goban.engine.undo_requested = controller.goban.engine.cur_move.move_number;
const triggerFlow = jest.fn();
renderHeader(controller, triggerFlow);
expect(triggerFlow.mock.calls.map(([flow]) => flow)).toEqual([
    "undo-request-received-intro",
]);
```

On this PR, the actual calls are `undo-requested-intro` followed by `undo-request-received-intro`. The existing tests set the request after mount, so they miss this case.

**Proposed fix:** Extract the permission calculation in `useCanAnswerUndoRequest` into a memoized `readCanAnswer` callback. Use it both for the initial state and for event updates:

```tsx
const [can_answer, set_can_answer] = React.useState(readCanAnswer);
// In the subscription effect:
const syncCanAnswer = () => set_can_answer(readCanAnswer());
```

Keep the current participant, pushed-analysis, requester, and legacy fallback rules. Give the callback dependencies `[goban, in_pushed_analysis, user_id]` and the subscription effect dependencies `[goban, readCanAnswer]`.

I repeated the reproduction at commit `c0499d8cf`. A second test with the real `HelpController` and both undo flows also rendered both tutorial messages. Both regression tests passed with the base header. With the proposed initialization fix, all 46 tests across `GameStateHeader`, `GameHooks`, `PlayButtons`, and `PlayControls` passed.

These were Jest/jsdom tests. Manual browser tests remain necessary on mobile and desktop.

Chirag6722:
Good catch, thank you. Reproduced it with your test: on c0499d8 the calls were `["undo-requested-intro", "undo-request-received-intro"]`, because `can_answer` started as `false` and only caught up in the effect, while the old turn check was computed synchronously.

Pushed 26a09d0 with the fix you outlined: the permission read is a memoized `readCanAnswer` (deps `[goban, in_pushed_analysis, user_id]`), `useState` is initialised from it, and the subscription effect just calls `set_can_answer(readCanAnswer())` (deps `[goban, readCanAnswer]`). The participant, pushed-analysis, requester and legacy fallback rules are unchanged. Your mount-time test is added as the fourth case in `GameStateHeader.test.tsx`; it fails on c0499d8 and passes now. All 45 tests across the four Game suites pass, type-check and eslint are clean. Still jsdom only on my side; I have not run a live two-player game.

anoek:
Thanks!

